PR #3, opened 2021-02-23
Testbranch jay

JayMan91:
I created FirstUsage.md, as a guide to use python package, documentation and testing for first time users like me.


FirstUsage.md
```diff
@@ -36,11 +36,11 @@ setup(
     author='Tias Guns',
     author_email='',
     license='MIT',
-    # packages=find_packages(exclude=['sample_app.test']),
+    # packages=find_packages(exclude=['mccpy.test']),
```
tias, line 39: should be new name: cpmpy

FirstUsage.md
```diff
@@ -24,18 +24,70 @@ To use markdown the documentation says we have to `pip install recommonmark` and
...
 extensions = ['recommonmark']
 ```
-# Testing using tox https://tox.readthedocs.io/en/latest/
+# Building package Locally
+First create a `setup.py` file
+```python
+from setuptools import setup, find_packages
+
+setup(
+    name='mcppy',
```
tias, line 33: new name

FirstUsage.md
```diff
@@ -24,18 +24,70 @@ To use markdown the documentation says we have to `pip install recommonmark` and
...
+numpy>=1.20.0
+```
+Create a folder named `CPModules` (or whatever) and write all the class and definition inside it within `model.py` (or whatever).
+Now to install in test mode, first create a virtual environment. and then run after going to directory containing the `setup.py`
+```
+pip install -e .
+```
+To check whether the installation is successful, open python
+```python
+form CPModules.model import *
```
tias, line 59: should be 'from cpmpy import *'?

tias (changes requested):
Can you rebase on current master with new name?

FirstUsage.md
JayMan91: It's the name of the folder which in this case `cpmodules`, I think a good practice is to give it a same name as the package, but I d don't know

PR #7, opened 2021-02-23
Repo restructuring & Packaging & Workflows

antemooo:


setup.py
```diff
@@ -0,0 +1,32 @@
+from setuptools import find_packages, setup
+
+with open("README.md", "r") as readme_file:
+    long_description = readme_file.read()
+
+setup(
+    name='CP_Test_PY',
```
sourdough-bread, line 7: CpMPy

setup.py
```diff
@@ -0,0 +1,32 @@
...
+    description='A numpy-based light-weight Python library for conveniently modeling constraint problems in Python',
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/tias/cppy",
+    packages=find_packages(),
+    include_package_data=True,
+    zip_safe=False,
+    install_requires=[
+        'ortools>=7.2',
+        'numpy>=1.9',
```
sourdough-bread, line 20: Is this correct ? Was this not numpy >= 1.5 and ortools >=5.0?

sourdough-bread (commented):
Small comments on the pull requests:
setup.py

setup.py
antemooo: oh yeah, this will change ;)

setup.py
antemooo: I honestly need to check that first on PyPi, locally it does not locate the correct packages when I set numpy >= 1.5 and ortools >=5.0. Eventually py3.6 may already does not support numpy1.5 nor ortools5.0.

PR #47, opened 2021-09-29
Bug pysat.__add__ 

sourdough-bread:
iterates over all initial constraint instead of subconstraints of to_cnf(constraint)


tias:
I cherry-picked the 'Bug pysat.__add__' commit. Your pull request also contains your work on the example, which should be in a separate pull request...

PR #48, opened 2021-09-30
OCUS explanations generated for the frietkot problem

sourdough-bread:


examples/advanced/ocus_explanations.py
```diff
@@ -0,0 +1,246 @@
...
+        hittingset_solver.solve()
+        hs = [hs_var for hs_var in hs_vars if hs_var.value()]
+
+        if not sat.solve(assumptions=[hs_vars_Iend[hs_var] for hs_var in hs]):
+            if verbose:
+                print("OCUS found !!!!")
+                print("hs=", [hs_vars_Iend[hs_var] for hs_var in hs])
+
+            # deleting the hitting set solver
+            del hittingset_solver
```
tias, line 128: not sure you need to do this explciitly in Python?

examples/advanced/ocus_explanations.py
```diff
@@ -0,0 +1,246 @@
...
+    Guy = ~ketchup | curry | andalouse
+    Daan = ~ketchup | ~curry | andalouse
+    Celine = ~samurai
+    Anton = mayo | ~curry | ~andalouse
+    Danny = ~mayo | ketchup | andalouse | samurai
+    Luc = ~mayo | samurai
+
+    allwishes   = [Nora, Leander, Benjamin, Behrouz, Guy, Daan, Celine, Anton, Danny, Luc]
+
+    assum_Nora = boolvar(name="Nora")
```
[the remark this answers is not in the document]
sourdough-bread: I did that, to have the debug printing appear nicely.

examples/advanced/ocus_explanations.py
```diff
@@ -0,0 +1,302 @@
...
+    sat_model = set(v == v.value() for v in user_vars)
+
+    if verbose > 1:
+        print("\nOptimal Propagate")
+        print("-" * len("Optimal Propagate"))
+        print("\nInitial sat model:", sat_model)
+
+    while(True):
+        # negate the values of the model
+        blocking_clause = ~all(v == v.value() for v in sat_model)
```
tias, line 265: line 276 says that sat_model is itself a set of (v == v.value()) constraints, then v in sat_model then be a variable?

examples/advanced/ocus_explanations.py
```diff
@@ -0,0 +1,302 @@
...
+        set_to_hit = set(Iend_to_hs_vars[Iend_var] for Iend_var in F - S)
+
+        # Add complement as a new set to hit: sum x[j] * hij >= 1
+        hittingset_solver += (sum(set_to_hit) >= 1)
+
+        if verbose > 0:
+            print("\t Complement =", C)
+            print("\t set-to-hit =", set_to_hit)
+
+def optimal_propagate(hard, soft, solver="ortools", verbose=False):
```
tias, line 232: You are always solving with assump=soft, so it does not actually need the 'soft' formulation.

The function should also be called differently in case of integer variables...
what about 'solution_intersection' or smth? Note how you mention CNF in the comments, not a restriction here.

`def solution_intersection(model, solver="ortools", verbose=False)` should be doable. In that case, code can also be simpler I think, something like model2 = Model(model.constraints); while model2.solve(): model2 += blocking_clause; intersect = ...'. There might also be a way to do the intersections better... basically, user_vars is an array, so you can get an array of values of the same size... and then, some indicator on top, not sure. Smth is a bit weird with your current sets.

examples/advanced/ocus_explanations.py
```diff
@@ -0,0 +1,302 @@
...
+    for i,bv in enumerate(ind):
+        reified_soft += [bv.implies(soft[i])]
+
+    # to map indicator variable back to soft_constraints
+    indmap = dict((v,i) for (i,v) in enumerate(ind))
+
+    # adding the assumption variables
+    I = set(bv for bv in ind)
+
+    Iend = optimal_propagate(hard + reified_soft, I, solver, verbose)
```
tias, line 78: I think this should just be `solution_itnersection(Model(soft))` or smth similar, it does not need the reified formulation?

Iend is a partial assignment right, not necessarily a list/set of Booleans?

hmm.. maybe that is why the set(v == v.value()) worked below, CPMpy for Booleans converts that to BoolVarImpl or NegBoolView, so single objects, but this is not true for integer variables...

examples/advanced/ocus_explanations.py
```diff
@@ -0,0 +1,302 @@
...
+
+    while(I != Iend):
+        # explain 1 step using ocus
+        ocus_expl = explain_one_step_ocus(hard+reified_soft, cost, Iend, I, solver, verbose)
+
+        # project on known facts and constraints
+        soft_used = I & ocus_expl
+
+        # propagate and find information hold in all models
+        derived = optimal_propagate(hard + reified_soft, soft_used, solver, verbose) - I
```
tias, line 98: for the sake of this example and cleanliness of your code, I think you can skip this; go straight to the explanation. We might generate multiple steps for the same set of constraints, but no harm is done, it is just potentially a bit less efficient.

examples/advanced/ocus_explanations.py
```diff
@@ -0,0 +1,302 @@
...
+        unsatisfiable subsets of constraints. Journal of Automated Reasoning, 40(1), 1-33.
+
+        [3] Reiter, R. (1987). A theory of diagnosis from first principles.
+        Artificial intelligence, 32(1), 57-95.
+    """
+
+    sat = SolverLookup.lookup(solver)(Model(hard))
+
+    ## OPT Model: Variables
+    hs_vars = boolvar(shape=len(Iend))
```
tias, line 177: somehow, it would be great to make it clearer that 'sat' here is over a set of variables (Iend I guess), and that hs_model will be over another; and then you maintain the mapping...

but a clearer separation would be nice.

perhaps Iend -> sat_vars; sat -> sat_model
then hs_vars and hs_model as now. And then, your maps can be called varmap_sat_hs and varmap_hs_sat or so?

PR #53, opened 2021-10-01
Visual Sudoku example

CryoCardiogram:
Here is the visual sudoku advanced example, both in a .py script and a notebook. 
It's not ready/clean yet, there is a lot of room for improvement.


examples/advanced/visual_sudoku.py
```diff
@@ -0,0 +1,281 @@
...
+
+@torch.no_grad() #TODO: What does this do ?
+def predict_proba_sudoku(model, vizsudoku):
+    #TODO: Change shape of output to work for differnet kinds of sudoku shapes (4x4, 9x9, ...)
+    # reshape from 9x9x1x28x28 to 81x1x28x28
+    pred = model(vizsudoku.flatten(0,1))
+    # our NN returns 81 probabilistic vector: an 81x10 matrix
+    return pred.reshape(9,9,10).detach() # reshape as 9x9x10 tensor for easier visualisation
+
+if __name__ == '__main__':
```
tias, line 200: create a function `def visual_sudoku()` that you call here (rest is blank).

Put that `def visual_sudoku()` at the top: make your script readable from top to bottom

examples/advanced/visual_sudoku.py
```diff
@@ -0,0 +1,281 @@
...
+        [2, e, e,  e, e, e,  4, e, 9],
+        [e, e, e,  e, 7, e,  e, e, e],
+        [6, e, 9,  e, e, e,  e, e, 1],
+
+        [e, 8, e,  4, e, e,  1, e, e],
+        [e, 6, 3,  e, e, e,  e, 8, e],
+        [e, e, e,  6, e, 8,  e, e, e]
+    ])
+
+    # Emilio: Missing a bit of context here ?
```
tias, line 219: idd, also watch your variable names, you call it 'sample_sudoku_grid' which you then feed into a 'sample_...' function which means somethign else.

Also higher up, avoid 'model' for the NN as 'model' in CP/CPMpy is smth different...

examples/advanced/visual_sudoku.py
```diff
@@ -0,0 +1,281 @@
...
+    vs = sample_visual_sudoku(sample_sudoku_grid)
+
+    # TODO: (Minor) Does not work from outside of folder + what is clf ? (classifier?)
+    model = load_clf(LeNet,'lenet_mnist_e15.pt')
```
tias, line 223: do not use 'model', but something descriptive like 'nn'

tias: or clf, given the load_clf function name...

examples/advanced/visual_sudoku.py
```diff
@@ -0,0 +1,281 @@
...
+
+    sudoku_model += [dvar[sample_sudoku_grid > 0] == sample_sudoku_grid[sample_sudoku_grid > 0]]
+
+    # TODO: why do you need to create a new model ?
+    # old: model = Model(cons)
+
+    model.solve()
+    solution = dvar.value()
+
+    # TODO: pretty printing ?
```
tias, line 245: work on the flow...

You first want to compute the argmax solution, so have only the code for that, and print that result.
Some visu/print/annotation of how it makes a mistake would be good.... (a notebook could be better then? also forces you to be more linear...)

Then, write that you will now make a max likelihood formulation.

Ideally, the reader can see the difference very close to each other, e.g.

m = base_sudoku(grid)
m += (grid[given] == argmax[given])
m.solve()

m2 = base_sudoku(grid)
m2.optimize(np.sum((grid==i)*logprobs[i] for i in range()) # or some fancy one-liner to write this ; )
m2.solve()


THAT would be a great example...

then, advanced edition, can build on m2 perhaps?
while not_unique(grid.value()):
  m2 += # blocking clause
  m2.solve()


I'm not sure if it is doable, but if your 'def visu_sudoku():` would contain this, it would be a very pedagogical example...

CryoCardiogram:
I have updated a notebook and deleted the .py script. 

I have reworked the flow, added visualization and verbosity / markdown cells. 
Ready for a final review and/or merge.

PR #56, opened 2021-10-04
Converted visual examples to CPMPy + added planning problem

Nide44:


tias (commented):
in all examples, please separate out:

def model_probname(data1, data2, data3):
   m = Model()
   m += ...
   return m

def run():
    # data definitions
    data1 = 
    data2 =
 
    # build model
    (model, vars) = model_probname(data1, data2, data3)

    if model.solve():
         visu_prob(vars, data1, data2)

def visu_prob():
     ...lots of ugliness..

if __name__ == '__main__':
   run()


This makes it easier to digests and separtes the 'modeling' from the solving and the visualisation things...

examples/visual_examples/car_sequencing/run_car_seq.py
```diff
@@ -14,44 +14,87 @@
...
+    [0, 0, 0, 1, 0],
+    [0, 1, 0, 0, 1],
+    [0, 1, 0, 1, 0],
+    [1, 0, 1, 0, 0],
+    [1, 1, 0, 0, 0]]
+
+    (model, vars) = model_sequence(nCarConfigs, nSlots, nOptions, demand, per_slots, at_most, requires)
+
+    if model.solve():
+        for v in vars:
```
tias, line 37: for (name, var) in vars.items():
   print(f"{name}:\n{var.value()}")

would be a nice alternative?

examples/visual_examples/car_sequencing/run_car_seq.py
```diff
@@ -14,44 +14,87 @@
...
+            print(v + ":\n" + str(vars[v].value()))
+        visualize_sequence(vars, nSlots, nOptions, nCarConfigs, requires, demand, at_most, nCars, per_slots)
+
+def model_sequence(nCarConfigs, nSlots, nOptions, demand, per_slots, at_most, requires):
+    # Decision variables
```
tias, line 42: Please add more of the problem description in the comments..
what does 'line' represent (the sequence of cars?)
what does 'setup' represent...

examples/visual_examples/car_sequencing/run_car_seq.py
```diff
@@ -14,44 +14,87 @@
...
+    line = intvar(1, nCarConfigs, shape = nSlots)
+    setup = boolvar(shape = (nSlots, nOptions))
+
+    model = Model(
+        [sum((x == (i+1)) for x in line) == demand[i] for i in range(nCarConfigs)] # The amount of each type of car in the timetable has to be equal to the demand for that type
```
tias, line 47: model = Model()

# The amount of each...
m += []
is cleaner

what does `sum((x == (i+1)) for x in line)` intend to do? same right hand side for all x's?
that happens automatically (called broadcasting) when you write sum(x == (i+1)) == demand[i]

why the +1 btw? MiniZinc might be offset 1 but CPMpy is not, so if
line = intvar(0, nCarConfigs, shape=...) then you could simply do sum(x == i) == demand[i]?
or even:
m += [sum(x == i) == d for (i,d) in enumerate(demand)]
which at least to me reads nice : )

examples/visual_examples/car_sequencing/run_car_seq.py
```diff
@@ -14,44 +14,87 @@
...
+    )
+
+    for o in range(nOptions):
+        model += [sum(setup[s2][o] == True for s2 in range(s, s + per_slots[o])) <= at_most[o] for s in range(nSlots - per_slots[o])] # Check that no more than "at most" car properties are used per "per_slots" slots
```
tias, line 51: not readable enough, put more for-loops outside

for o in ..
  for s in ..
    selector = range(s, s + per_slots[o]))
    // an appropriate comment
    model += [sum(setup[selector,o]) <= atmost[o]]

something like the above? if 'selector' has meaning, use a more meaningful variable name

examples/visual_examples/car_sequencing/run_car_seq.py
```diff
@@ -14,44 +14,87 @@
...
+
+    requires = cpm_array(requires) # So it can be indexed by a variable
+    model += [setup[s,o] == requires[line[s]-1,o] for s in range(nSlots) for o in range(nOptions)] # Make sure that the properties in the timetable correspond to those per car type
```
tias, line 54: comment before the code line
model ++ [setup[s,:] == requires[line[s]-1,:] for s in range(nSlots)] ?
careful with the '-1' here if you make it offset 0 (which looks like again you should ; )

examples/visual_examples/car_sequencing/run_car_seq.py
```diff
@@ -14,44 +14,87 @@
...
+
+    return (model, {"line": line, "setup": setup})
+
+def visualize_sequence(vars, nSlots, nOptions, nCarConfigs, requires, demand, at_most, nCars, per_slots):
```
tias, line 58: Add a comment that all the rest below is only about making the nice visualistion

examples/visual_examples/packing/run_rect_packing.py
```diff
@@ -12,54 +12,69 @@
...
-    print(rect_width.value())
-    print(rect_area.value())
-    print(x.value())
-    print(y.value())
+def run():
+    # 2D dimensions of the items to be packed
+    widths  = [5, 6, 4, 3, 2, 4, 3, 1, 2, 1, 7, 3]
+    heights = [1, 2, 3, 2, 1, 2, 4, 6, 5, 1, 1, 2]
+
+    # Number of different items
```
tias, line 20: compute all the deterministaclly computed ones in 'model_rect_packing', so you don't have to pass them as an argument...

wrap the two arrays in cpm_array perhaps again.

examples/visual_examples/packing/run_rect_packing.py
```diff
@@ -12,54 +12,69 @@
...
+            print(v + ":\t" + str(vars[v].value()))
+        visualize_rect_packing(vars, widths, heights, n)
+
+def model_rect_packing(widths, heights, n, max_width_rect, 
+                        max_height_rect, min_width_rect, min_height_rect,
+                        min_area, max_area):
+    # Decision variables
+    rect_height = intvar(min_height_rect, max_height_rect)
+    rect_width = intvar(min_width_rect, max_width_rect)
+    rect_area = intvar(min_area, max_area)
```
tias, line 50: do we really need this one?
I mean, are its lb/ub smaller then all areas possible? If not, you can cut this and the constraint on line 56, and
write

model.minimize(rect_height * rect_width)

examples/visual_examples/packing/run_rect_packing.py
```diff
@@ -12,54 +12,69 @@
...
+    x = intvar(0, max_width_rect, shape = n)
+    y = intvar(0, max_height_rect, shape = n)
+
+    model = Model(
+        ### Necessary constraints
+        (rect_area == rect_height * rect_width), # Definition of the area of the needed space
+        [np.add(x, widths) <= rect_width], # Every item has to be within the width of the overall area
```
tias, line 57: with the new update, perhaps x + widths <= rect_width would actually work? note that widths is best a cpm_array in that case

examples/visual_examples/scheduling/run_scheduling.py
```diff
@@ -12,36 +12,49 @@
...
+def model_scheduling(jobs, lastT, nTasks, dur, taskToMach):
+
+    # Decision variables
+    start = intvar(0, lastT, shape=(len(jobs), nTasks))
+    end = intvar(0, lastT, shape=(len(jobs), nTasks))
+    makespan = intvar(0, lastT)
+
+    model = Model(
+        # Necessary constraints
+        [end[i][j] == start[i][j] + dur[i][j] for i in range(len(jobs)) for j in range(nTasks)], # The end of every task is the sum of its start and duration
```
tias, line 40: use the vectorized operations: end == start + dur

examples/wolf_goat_cabbage.py
```diff
@@ -8,14 +8,24 @@
...
-
-while stages < 20:
-
-    wolf_pos = intvar(0, 1, stages)
-    cabbage_pos = intvar(0, 1, stages)
-    goat_pos = intvar(0, 1, stages)
-    boat_pos = intvar(0, 1, stages)
+def run():
+    stage = 3
+    while stage < 20:
```
tias, line 13: while True?

examples/wolf_goat_cabbage.py
```diff
@@ -8,14 +8,24 @@
...
+            print("Found a solution for " + str(stage) + " stage!")
+            for v in vars:
+                print(v + ":\t" + str(vars[v].value()))
+            break
+        else:
+            print("No solution for " + str(stage) + " stage")
+            stage += 1
+
+def model_wgc(stage):
+    wolf_pos = intvar(0, 1, stage)
```
tias, line 25: boolvar?

examples/wolf_goat_cabbage.py
```diff
@@ -25,31 +35,25 @@
...
         # Boat keeps moving between shores
-        [boat_pos[i] != boat_pos[i-1] for i in range(1,stages)],   
+        [boat_pos[i] != boat_pos[i-1] for i in range(1,stage)],   
 
         # Final situation
-        (boat_pos[stages-1] == 1),
-        (wolf_pos[stages-1] == 1),
-        (goat_pos[stages-1] == 1),
-        (cabbage_pos[stages-1] == 1),
+        (boat_pos[stage-1] == 1),
```
tias, line 41: you can just write boat_pos[-1] == 1

examples/visual_examples/car_sequencing/run_car_seq.py
Nide44: Adjusted

examples/visual_examples/car_sequencing/run_car_seq.py
Nide44: Same comment as before: just to aesthetically talk about car type 1 to 6 instead of 1 to 5. :)

examples/visual_examples/car_sequencing/run_car_seq.py
Nide44: Done.

PR #57, opened 2021-10-06
Predict+opti example

JayMan91:
Please look into the CPMPY_Pred+Opt Notebook

### DIFF
--- a/examples/CPMPY_Pred+Opt.ipynb
+++ b/examples/CPMPY_Pred+Opt.ipynb
@@ -1,0 +1,906 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## TIas: do you really need all those packages??"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 19,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import torch\n",
+    "from torch import nn, optim\n",
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "\n",
+    "from sklearn.preprocessing import StandardScaler\n",
+    "from sklearn.metrics import mean_squared_error as mse\n",
+    "from sklearn.model_selection import cross_validate\n",
+    "from sklearn.metrics import make_scorer\n",
+    "import sklearn\n",
+    "import torch.utils.data as data_utils\n",
+    "import seaborn as sns\n",
+    "from cpmpy import *\n",
+    "# sns.set_theme()\n",
+    "sns.set_context(\"notebook\", font_scale=1.5, rc={\"lines.linewidth\": 2.5})"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### First we  load the data. We have training and test data."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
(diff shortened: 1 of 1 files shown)


examples/CPMPY_Pred+Opt.ipynb
```diff
@@ -0,0 +1,906 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## TIas: do you really need all those packages??"
```
tias, line 7: You probably want to remove this comment?

examples/CPMPY_Pred+Opt.ipynb
```diff
@@ -0,0 +1,906 @@
...
+    "from sklearn.metrics import make_scorer\n",
+    "import sklearn\n",
+    "import torch.utils.data as data_utils\n",
+    "import seaborn as sns\n",
+    "from cpmpy import *\n",
+    "# sns.set_theme()\n",
+    "sns.set_context(\"notebook\", font_scale=1.5, rc={\"lines.linewidth\": 2.5})"
+   ]
+  },
+  {
```
tias, line 33: I'd like you to put the CPMpy model already here, at the very very top

examples/CPMPY_Pred+Opt.ipynb
```diff
@@ -0,0 +1,906 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "import requests\n",
+    "import io\n",
+    "\n",
+    "response = requests.get('https://github.com/JayMan91/aaai_predit_then_optimize/raw/master/Data.npz')\n",
+    "response.raise_for_status()\n",
+    "data = np.load(io.BytesIO(response.content))\n",
```
tias, line 59: do you need io.BytesIO if you use response.text instead?

examples/CPMPY_Pred+Opt.ipynb
```diff
@@ -0,0 +1,906 @@
...
+   "metadata": {},
+   "source": [
+    "#### As we see, we have 552 training instances and 237 test instances"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Next we wil randomize the training and test data"
```
tias, line 168: this can be part of an earlies cell perhaps, with a comment; I thnk it could be good to have a bit fewer boxes...

examples/CPMPY_Pred+Opt.ipynb
```diff
@@ -0,0 +1,906 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Now we create an wrapper object, which repeatedly solve the knapsack problem where the cost are fed as pytorch tensor"
+   ]
+  },
+  {
+   "cell_type": "code",
```
tias, line 214: what about combining this cell and the previous into something like:

```
class problem_solver():
    def __init__(self, weights, capacity):
        self.x = boolvar(shape=w.shape)
        self.m = CPM_ortools(Model( sum(x*w <= c) ))

    def solve_cost(cost):
        # convert to int for ortools
        cost = (1000*cost).astype(int)
        self.model.maximize((self.x*cost).sum())
        self.solve()
        return self.x.value()

   def solve_from_pytorch(tcost):
         # pytorch in pytorch out
         cost = tcost..detach().numpy()
         ret = solve_cost(cost)
         return torch.from_numpy(ret).float()

```

like this?

tias (changes requested):
Thanks, looks very solid.

Some minor comments...

How is the runtime? If long, consider using fewer data,
and/or add a caching_problem_solver that inherits problem_solver but has a caching strategy...?

PR #67, opened 2021-10-13
Counterfactual example

IgnaceBleukx:
Reimplemented counterfactual explanations based on Korikov, A., & Beck, J. C.
Using 0-1 knapsack as an example.
Should we also implement another problem such as scheduling? Might be less useful as it will require a lot of code to generate the problems itself.

Notes:
Current print statements are quite conversational. Should we keep it like this or change it to a more descriptive style?


examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/python3
+"""
+Counterfactual explanation of a user query.
+
+Based on application on the knapsack problem of Korikov, A., & Beck, J. C. Counterfactual Explanations via Inverse Constraint Programming.
```
tias, line 5: mming, CP2021.

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+
+Usecase:
+ 1) Some optimal solution x* is provided to the user by a constraint optimization solver.
+ 2) User enters a query to change solution variables x_i to a new value.
+    This extra set of constraints is used to generate a new optimal solution x_d
+ 3) Using inverse optimization, we calculate the values of constraints the objective function which result in x_d being the optimal solution.
```
tias, line 11: 'val of constr of obj' sentence broken

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+    Extra constraint: only values corresponding to foil items are allowed to change.
```
tias, line 12: explain 'foil' before first use

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+ 4) The user is presented with the result.
```
tias, line 13: result here is? alternative set of objective coefficients (of the foil items)?

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/python3
```
tias, line 1: Fantastic!!!

chose another name though, NCE = Noise Contrastive Estimation for us...

counterfactual_explain or explain_obj_counterfactual? Something...

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+The algorithm consists of 2 alternating problems: the master problem (MP), and the sub problem (SP)
+
+Algorithm:
+    1 S = {}
+    2 Solve MP to obtain optimal d*
+    3 Solve SP with d* as c to obtain x_0
+    4 if objective(x_d) > objective(x_0):
+        add x_0 to S
```
tias, line 30: something more about the intuition....

you iterate optimal solutions to ||c-d|| until you find one for which min(f == x_d), something like that?

tias: maybe expl master prob first? not sure...

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+INFINITY = np.iinfo(np.int32).max
+verbose = False
+np.random.seed(0)
+
+def main():
+
+    n = 10  #Number of items in the knapsack
+    m = 5   #Number of items to change in the knapsack
+
+    values, weights, capacity, x = generate_knapsack_model(n)
```
tias, line 63: I prefer that you split out the generate from the solve, so x is computed on next line

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+
+############################################################################
+# All functions below this line are purely used for pretty printing results#
+############################################################################
+def print_knapsack_model(values, weights, capacity, x):
+    """
+        Pretty prints a knapsack model.
+    """
+    print("Solution to the following knapsack problem")
+    print(f"Capacity: {capacity}, used: {sum(x*weights)}")
```
tias, line 212: When running it, I did not know what 'used' was, it is unusual to already use the solution before you present it...

move used to below 'is:' and obj value

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+def main():
+
+    n = 10  #Number of items in the knapsack
+    m = 5   #Number of items to change in the knapsack
+
+    values, weights, capacity, x = generate_knapsack_model(n)
+    #TODO: verbose statement?
+    print_knapsack_model(values, weights, capacity, x)
+
+    x_user, foil_items = generate_foil_knapsack(values, weights, capacity, x, m)
```
tias, line 67: a comment like: # we emulate a foil query by a hypothetical user, our generation process ensures that...

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+    
+    # Pretty print user query
+    #TODO: verbose statement?
+    pp_uquery(x, foil_items)
+
+    #Find the values vector such that x_foil is optimal
+    d_star = master_problem(values, weights, capacity, x_user, foil_items)
```
tias, line 74: master_problem is a problem, not a solution method,
rename to smth like solve_branch_cut or whathever appropriate name for the technique?

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+
+        # Extension of the knapsack problem to the master problem
+        # Ensure the newly found values vector results in a better solution than any known solutions
+        constraints += [sum(d * x) >= sum(d * known) for known in known_solutions]
+        # Ensure values are only modified at foil indices 
+        constraints += [d[i] == values[i] for i in range(len(values)) if i not in foil_idx]
+        # Ensure the foil values assigned by the user remain the same
+        constraints += [x[i] == bool(x_d[i]) for i in foil_idx]
+       
+        master_model = Model(
```
tias, line 165: you don't need to rebuild the 'master_model' from scratch inside the loop...

CPMpy supprots incremental solving,
so you can make the model once before the loop,

and then do master_model += [sum(d*x) >= sum(d*new_solution)]
which is a lot more elegant : )

examples/advanced/nce_example.py
```diff
@@ -0,0 +1,237 @@
...
+
+        Based on the Numberjack model of Hakan Kjellerstrand
+    """
+    x = boolvar(len(values), name="x")
+    model = Model(
+                [sum(x * weights) <= capacity],
+                maximize = sum(x * values)
+            )
+    if model.solve() is not False:
+        return x.value()
```
tias, line 201: else? raise Excp is always a good option...

tias (changes requested):
Absolutely fantastic and elegant.

Esp make the master problem incremental though...

note that you could also make the subproblem incremental...
kn = make_knapsack_model(values, weights, capa)
then kn.solve() all good...

then later, you can do
kn.maximize(sum(x*new_values))
kn.solve()
now it has solved with the new objective...


we can then swap out the solver for a MIP solver like gurobi and it should automatically have the incrementality benefits...

examples/advanced/nce_example.py
IgnaceBleukx: Added some intuition

examples/advanced/nce_example.py
IgnaceBleukx: Made the master problem and sub problem iterative.
Thereby shortening the method massively.

PR #76, opened 2021-10-29
Cardinality Constraints (<,>,>=,<=,==)

sourdough-bread:
Basic support of unweighted Cardinality constraints in PySAT


tias:
There is one cardinality case missing: != with sum(x) != c equals to sum(x) < c | sum(x) > c;
so I hope it would be fairly easy to add this case too? then this is complete...

PR #79, opened 2021-11-05
Pysat int2bool

sourdough-bread:
Improved PySAT `make_cnf()`: 
- Decoupling of cnf encoding of the constraints allowing it to be reused by `__add__()` function.
- COrrected bug where variables of newly added constraints are not added to user_vars
- Some initial documentation of the process int->bool model translation

TODO:
- Handle linear constraints by mapping with [PySAT's Pseudo-Boolean library](https://pysathq.github.io/docs/html/api/pb.html)


cpmpy/transformations/to_bool.py
```diff
@@ -0,0 +1,317 @@
...
+    elif isinstance(constraint, (list, NDVarArray)):
+        for con in constraint:
+            bool_constraints += to_bool_constraint(con, ivarmap)
+
+    # CASE 4: base comparison constraints + ensure only handling what it can
+    elif isinstance(constraint, Comparison) and all(is_int(arg) or isinstance(arg, (_IntVarImpl, _BoolVarImpl)) for arg in constraint.args) :
+        bool_constraints += to_unit_comparison(constraint, ivarmap)
+
+    # CASE 5: global constraints
+    elif isinstance(constraint, (AllDifferent, AllEqual, Circuit, Table)):
```
tias, line 167: The way I see it is that, we want to replace in every constraint the intvar with its linear expression over Booleans.

Only, we also want to catch the special cases where it is a direct Comparison operator, and in that case replace the comparison by a direct expression on the Bools?

So, in any case, we should only 'do' something with constraints that contain an intvar, all others just pass through...
and the default should be to replace the intvar by a linexpr/wsum.
So then the question is, is this before or after flattening... I would say before!

tias (commented):
Looks good, see comments for open points

cpmpy/model.py
```diff
@@ -28,11 +28,18 @@
         Model
 """
 import numpy as np
+from cpmpy.expressions.variables import _BoolVarImpl
+from cpmpy.transformations.get_variables import get_variables_model
+from cpmpy.transformations.flatten_model import flatten_model
+
+from cpmpy.transformations.to_bool import intvar_to_boolvar, to_bool_constraint
```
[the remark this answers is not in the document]
sourdough-bread: Removed

cpmpy/model.py
```diff
@@ -27,7 +27,6 @@
 
         Model
 """
-import numpy as np
```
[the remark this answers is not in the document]
tias: so all changes to this file should go

cpmpy/transformations/get_variables.py
```diff
@@ -6,7 +6,7 @@
 import warnings # for deprecation warning
 from ..expressions.core import Expression
 from ..expressions.variables import _NumVarImpl,NegBoolView
-from ..expressions.utils import is_any_list
+from ..expressions.utils import is_any_list, is_int
```
tias, line 9: no cosmetic

tests/test_int2bool_pysat.py
```diff
@@ -0,0 +1,190 @@
...
+            iv1 > 6
+        )
+
+        s = CPM_pysat(m)
+        s.solve()
+
+        s += iv2 < 4
+        s.solve()
+    
+    # abs() not implemented yet
```
tias, line 34: its very good to add test cases too; lots of things here in comments though, remove or activate?

tests/test_model.py
```diff
@@ -3,10 +3,7 @@
 import os
 from os.path import join
 
-from numpy import logaddexp
```
tias, line 6: good catch, please commit these changes on master directly

sourdough-bread:
Ready for reviewing and merging.

PR #85, opened 2021-12-08
unified solver interfaces, and template

tias:
work in progress branch


cpmpy/solvers/TEMPLATE.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # initialise the native solver object
+        self.tpl_model = TEMPLATEpy.Model("cpmpy")
+
+        # initialise everything else and post the constraints/objective
+        # it is sufficient to implement __add__() and minimize/maximize() below
+        super().__init__(cpm_model, solver)
+
+        # TODO: these 4 in super()
+        self.cpm_status = SolverStatus(self.name)
```
IgnaceBleukx, line 83: Yes, would be best to do in solver indeed.
At the moment, Pysat contains a bug where solvername is "dummy" in status until solver is run.
Pushing the construction of cpm_status into the superclass will fix this.

IgnaceBleukx (commented):
Looks very nice. 
Also for myself, the layout of the solver interface is a lot clearer than starting from the current implemented solvers.

What tests would you have in mind to add to the file?
Just adding some variables and solving/optimizing it?

cpmpy/solvers/TEMPLATE.py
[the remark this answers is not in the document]
tias: Ah yes, maybe reworking the constructor of the superclass is a good starting point in aligning the interfaces?

tias:
About the tests, we'll have to see when we add a new solver interface, anything that can help an incremental development of that, e.g.

step 1: impl constructor
test: s = CPM_tpl(); s.status()
step 2: impl (empty) solve
s.solve(); s.status()
step 3: single var creation/solve
bv = BoolVar()
s = CPM_tpl(Model(bv))
s.solve(); bv.value()
etc...?

I'm not sure; just, stuff to check intermediate implementations of the interface

cpmpy/solvers/TEMPLATE.py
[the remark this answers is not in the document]
IgnaceBleukx: Edit: self.name cannot be called from superclass as it has no attribute name.
So best to still keep it in the subclass I think?

cpmpy/solvers/TEMPLATE.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # initialise variable handling
+        self.user_vars = set() # variables in the original (non-transformed) model
```
IgnaceBleukx, line 86: Right now, user_vars is a list. Altough I see no reason why it couldn't be a set?

cpmpy/solvers/TEMPLATE.py
```diff
@@ -0,0 +1,283 @@
...
+        Arguments:
+        - cpm_model: a CPMpy Model()
+        """
+        if not self.supported():
+            raise Exception("CPM_TEMPLATE: Install the python package 'TEMPLATEpy'")
+
+        self.name = "TEMPLATE" # solver's name
+
+        # initialise the native solver object
+        self.tpl_model = TEMPLATEpy.Model("cpmpy")
```
IgnaceBleukx, line 76: Do we make a solver object or the solver-specific model object here?
Maybe it would be good to add both?
Could be confusing for some solvers as OR-tools has both a solver and model object, but PySat has only the solver object...

cpmpy/solvers/TEMPLATE.py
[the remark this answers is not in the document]
tias: if we given name as an argument to the constructor then it becomes simpler overall?

cpmpy/solvers/TEMPLATE.py
tias: I think in MiniZinc it was a list so that we could output the variables 'in order' of use.

For all other solvers it should be a set, as that makes updating the sets on incremental addition more efficient. So, do switch to set and we'll see in the MiniZinc interface if that breaks something.

cpmpy/solvers/TEMPLATE.py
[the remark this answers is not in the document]
IgnaceBleukx: Yes, that's true, could also make more sense for users if they have more than 1 solver object at the same time

IgnaceBleukx:
> 

Hmm, some python-related issues with this.
Take pysat for example: the import `import pysat.formula` will not work as the file itself is also called pysat. We can go two ways I think: renaming solver interfaces to "solvername_interface", or we should add a testsuite with specific tests for the incremental development. The first one is the quick and dirty way imho ;) So best to go for the second imho.

cpmpy/solvers/TEMPLATE.py
```diff
@@ -0,0 +1,283 @@
...
+        if is_num(cpm_var):
+            return cpm_var
+
+        # special case, negative-bool-view
+        # work directly on var inside the view
+        if isinstance(cpm_var, NegBoolView):
+            return TEMPLATEpy.negate(self.solver_var(cpm_var._bv))
+
+        # create if it does not exit
+        if not cpm_var in self.varmap:
```
IgnaceBleukx, line 169: The create if does not exist part:
should we be able to do something like `self += [cpm_var]`?
I.e. posting a constraint with only one variable? It would certainly make the code very simple and could even mean this function can be put in the superclass. At the moment, I don't think we support it. But also not sure if we want something like this...

cpmpy/solvers/pysat.py
```diff
@@ -89,96 +92,50 @@ def __init__(self, cpm_model=None, solver=None):
...
-        else:
-            # store original vars
-            self.user_vars = get_variables_model(cpm_model)
-
-            # create constraint model (list of clauses)
-            cnf = self.make_cnf(cpm_model)
-
-        # create the solver instance
+        cnf = pysat.formula.CNF()
         self.pysat_solver = Solver(bootstrap_with=cnf.clauses, use_timer=True, name=solvername)
```
tias, line 110: no need to create an empty CNF to then bootstrap with it, I think?

cpmpy/solvers/pysat.py
```diff
@@ -243,100 +190,124 @@ def _after_solve(self, pysat_status):
...
-        assum_idx = frozenset(self.pysat_solver.get_core()) # to speed up lookup
+        # base case, just var or ~var
+        if isinstance(cpm_expr, _BoolVarImpl):
+            self.pysat_solver.add_clause([self.solver_var(cpm_expr)])
+        elif isinstance(cpm_expr, Operator):
+            if cpm_expr.name == 'or':
+                self.pysat_solver.add_clause([self.solver_var(var) for var in cpm_expr.args])
+            else:
+                raise NotImplementedError(
+                    "Only 'or' operator supported by CPM_pysat for now (more possible with aiger, contact us on github")
```
tias, line 226: the reason of 'notimplemented' error is no longer true.
If there is some non-or constraint here then the 'to_cnf' does not support it...

so smth like f"Operator {cpm_expr} not automatically translated to CNF, please report on github"

cpmpy/solvers/pysat.py
```diff
@@ -243,100 +190,124 @@ def _after_solve(self, pysat_status):
...
-                        atmost =  [cl + [-is_atmost] for cl in CardEnc.atmost(lits=lits, bound=bound-1, vpool=self.pysat_vpool).clauses]
-                        cnf.extend(atmost)
-                        
-                        ## add is_atleast or is_atmost
-                        cnf.append([is_atleast, is_atmost])
-                    else:
-                        raise NotImplementedError(f"Non-operator constraint {con} not supported by CPM_pysat")
-                else:
-                    raise NotImplementedError(f"Non-operator constraint {con} not supported by CPM_pysat")
+            CPMpy will return only those variables that are False (in the UNSAT core)
```
tias, line 304: not related to the unification, but 'those assumption variables that are False' would be more explicit...

Also, the next line, the 'though this interf...' part is now in CPMpy so can be cut.

cpmpy/solvers/minizinc.py
```diff
@@ -223,90 +272,19 @@ def objective_value(self):
...
-    def maximize(self, expr):
-        """
-            Maximize the given objective function
-
-            `maximize()` can be called multiple times, only the last one is stored
-        """
-        # do not add it to the model, support only one 'solve' entry
-        self.mzn_txt_solve = "solve maximize {};\n".format(self.convert_expression(expr))
 
     def clean_varname(self, varname):
```
tias, line 284: We had a private discussion on solver_var for minizinc.

I think, to keep everything unified, solver_var() should maintain which variables are already created, and if not yet, to add them to the mzn file on the fly.
Then, you can move this 'clean_varname' into solver_var: that will be the only place where you will need it.
And then, solver_var(cpm_var) actually returns the solver_var: the minizinc-compatible string name of the variable?

then the 'add variables' part can go out of __add__ just like it is not in the other interfaces
and the 'fill variables' will also become similar: for all vars in user_vars, call solver_var() to get their true handle, and get the corresponding value?


So if you ca

cpmpy/solvers/minizinc.py
```diff
@@ -110,22 +112,68 @@ def __init__(self, cpm_model=None, solver=None):
...
+            self.user_vars.add(var)
+
+        if not is_any_list(cons):
+            cons = [cons]
+
+        # we can't unpack lists in convert_expression, so must do it upfront
+        # and can't make assumptions on '.flat' existing either...
+        # this is dirty code that should not be reused, keeping it hidden in this function...
+        # TODO: can't we put it in convert_expression now?
+        def flatlist(lst):
```
tias, line 146: expressions.utils now had a flatlist function, that one can probably be used instead so this can go

cpmpy/solvers/minizinc.py
```diff
@@ -110,22 +112,68 @@ def __init__(self, cpm_model=None, solver=None):
...
+
+    def __add__(self, cons):
+        """
+            Add an additional (list of) constraints to the model
+        """
+        # Add variables in cons
+        for var in get_variables(cons):
+            if var in self.user_vars:  # Might be possible to do this prettier?
+                continue
+            if isinstance(var, _BoolVarImpl):
```
tias, line 133: this should go in solver_var() as for the other interfaces, see another comment below

cpmpy/solvers/ortools.py
```diff
@@ -64,24 +66,11 @@ def __init__(self, cpm_model=None, solver=None):
...
-            self.varmap = dict() # cppy var -> solver var
-        else:
-            # store original vars and objective (before flattening)
-            self.user_vars = get_variables_model(cpm_model)
-
-            # create model (includes conversion to flat normal form)
-            self.ort_model = self.make_model(cpm_model)
-
-        # create the solver instance
-        # (so its params can still be changed before calling solve)
```
tias, line 82: please keep this comment about why we also create ort_solver here, and not just internally in solve() which would also have been an option, except that you can then only changes its parameters/access it inside solve so already creating it here is more general.

cpmpy/solvers/ortools.py
```diff
@@ -97,17 +86,17 @@ def __add__(self, cons):
         """
         # store new user vars
         new_user_vars = get_variables(cons)
-        for v in frozenset(new_user_vars)-frozenset(self.user_vars):
-            self.user_vars.append(v)
+        for v in frozenset(new_user_vars) - frozenset(self.user_vars):
+            self.user_vars.add(v)
```
tias, line 90: now that user_vars is a set you can do user_vars.update(get_vars(cons)) or simlar, as you do in the pysat interface?

cpmpy/solvers/ortools.py
```diff
@@ -97,17 +86,17 @@ def __add__(self, cons):
...
 
-        flat_cons = flatten_constraint(cons)
+        flat_cons = only_bv_implies(flatten_constraint(cons))
         # add new (auxiliary) variables
```
tias, line 93: solver_var now creates new variables on the fly so this little code-block no longer needed

cpmpy/solvers/ortools.py
```diff
@@ -269,7 +227,7 @@ def _after_solve(self, ort_status):
         if self.ort_status == ort.FEASIBLE or self.ort_status == ort.OPTIMAL:
             # fill in variables
             for var in self.user_vars:
-                var._value = self.ort_solver.Value(self.varmap[var])
+                var._value = self.ort_solver.Value(self._varmap[var])
```
tias, line 230: _varmap should not be accessed directly, should call solver_var() instead

tias (commented):
nice, almost there

cpmpy/solvers/pysat.py
IgnaceBleukx: True, this is a leftover artefact of the make_model strategy we had.

cpmpy/solvers/pysat.py
IgnaceBleukx: Ok, added comment to refer to examples folder
In the future: refer to "tools-like" folder?

cpmpy/solvers/minizinc.py
IgnaceBleukx: Yes I agree that everything should go through solver_var(). Indeed, if we just return the minizinc-compatible string, a user can do the following once he has a solution: `mzn_var = cpm_var; sol.mzn_var`, just like in other solvers.
Still a little bit weird as we don't "post" a new variable to the minizinc interface, but this way at least the interface is unified.

On another note: maybe it would be good to create some sample scenarios how to use CPMpy and just swap out different solvers, so to make sure that all interfaces are unified?

cpmpy/solvers/ortools.py
IgnaceBleukx: True, also fixed for minizinc.

cpmpy/solvers/TEMPLATE.py
```diff
@@ -134,26 +149,22 @@ def solver_var(self, cpm_var):
         """
         # TODO: add `solver_vars(self, cpm_vars)` to SolverInterface class
 
-        if is_num(cpm_var):
```
IgnaceBleukx: Won't the removal of this check mean that we must do a lot more checks in the _post_constraint? Or we assume that all of the burden there has been overcome by the transformations?

cpmpy/solvers/TEMPLATE.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        return self.objective_value_
+
+
+    def solver_var(self, cpm_var):
+        """
+            Creates solver variable for cpmpy variable
+            or returns from cache if previously created
+        """
+        # TODO: add `solver_vars(self, cpm_vars)` to SolverInterface class
```
IgnaceBleukx, line 150: Todo can be removed, is done.

PR #93, opened 2022-01-04
Gurobi

IgnaceBleukx:
Pull request to merge solver inteface implemeting gurobi to master


cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+                "CPM_gurobi: Install the python package 'gurobipy' and make sure your licence is activated!")
+        import gurobipy as gp
+
+        # initialise the native solver object
+        self.env = gp.Env()
+        self.env.setParam("LogToConsole", 0)
+        self.env.setParam("OutputFlag", 0)
+        self.env.start()
+
+        self.gbi_model = gp.Model(env=self.env)
```
tias, line 86: it seems grb_model rather than gbi would be a more common abbreviation ; )

In this review I'll focus on the gurobi.py file, the linearizations can also be improved but lets do that in a second phase.

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+        self._objective_value = None
```
tias, line 87: I think we should make _objective_value a property of the superclass, as it is used in ORT and MZN also (and for PySAT it just always has to stay None, so also there); then objective_value() can be a superclass function too, and the solvers just have to populate self._objective_value whenever a solution is computed (more in that lower)

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+
+        # initialise everything else and post the constraints/objective
+        # it is sufficient to implement __add__() and minimize/maximize() below
+        super().__init__(name="gurobi", cpm_model=cpm_model)
+
+    def __add__(self, cpm_con):
```
tias, line 93: in the template __add__ is much lower, could you follow the same order?

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+
+        Note that we don't store the constraints in a cpm_model,
+        we first transform the constraints into primitive constraints,
+        then post those primitive constraints directly to the native solver
+
+        :param cpm_con CPMpy constraint, or list thereof
+        :type cpm_con (list of) Expression(s)
+        """
+        # add new user vars to the set
+        self.user_vars.update(set(get_variables(cpm_con)))
```
tias, line 105: I don't think the set() is needed, it is not in the template in any case

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+            Call the gurobi solver
+
+            Arguments:
+            - time_limit:  maximum solve time in seconds (float, optional)
+            - kwargs:      any keyword argument, sets parameters of solver object
+
+            Arguments that correspond to solver parameters:
+            <Please document key solver arguments that the user might wish to change
+             for example: log_output=True, var_ordering=3, num_cores=8, ...>
+            <Add link to documentation of all solver parameters>
```
tias, line 129: https://www.gurobi.com/documentation/9.5/refman/mip_models.html for the accessible version, maybe take some of them and list them in the line above as suggested

https://www.gurobi.com/documentation/9.5/refman/parameters.html#sec:Parameters for the full list

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+        _ = self.gbi_model.optimize(callback=solution_callback)
+
+        my_status = self.gbi_model.Status
+
+        # new status, translate runtime
+        self.cpm_status = SolverStatus(self.name)
+        self.cpm_status.runtime = self.gbi_model.runtime
+
+        # translate exit status
+        if my_status == GRB.OPTIMAL and self._objective_value is None:
```
tias, line 149: self._objective_value is a value that you should fill in yourself. You should check with self.gbi_model somehow if it was an optimisation problem or not... perhaps check the gurobi sudoku model to see what they do there

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+        # translate solution values (of user vars only)
+        if has_sol:
+            # fill in variable values
+            for cpm_var in self.user_vars:
+                solver_val = self.solver_var(cpm_var).X
+                if cpm_var.is_bool():
+                    cpm_var._value = bool(solver_val)
+                else:
+                    cpm_var._value = int(solver_val)
+
```
tias, line 171: here, within the 'if has_sol', you should get the objective_value from gbi_model and put it in self._objective_value; see ortools interface

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+            self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+        else:  # another?
+            raise NotImplementedError(
+                f"Translation of gurobi status {my_status} to CPMpy status not implemented")  # a new status type was introduced, please report on github
+
+        # True/False depending on self.cpm_status
+        has_sol = self._solve_return(self.cpm_status)
+
+        # translate solution values (of user vars only)
+        if has_sol:
```
tias, line 163: probably you should set self._objective_value = None here... to make sure it is overwritten when calling solve again (perhaps this hould be in the template too)

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+            # fill in variable values
+            for cpm_var in self.user_vars:
+                solver_val = self.solver_var(cpm_var).X
+                if cpm_var.is_bool():
+                    cpm_var._value = bool(solver_val)
```
tias, line 168: do you know what this function does? e.g. in terms of rounding?

I have had super tricky to catch bugs in the past, where gurobi would give a boolean a value of 0.9998 or so, so if you check bvar == 1 it will say no and your code thinks it had value 0.... maybe put a 'round()' around it first or so? pls check

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+        """
+        from gurobipy import GRB
+
+        if is_num(cpm_var):
+            return cpm_var
+
+        # special case, negative-bool-view
+        # work directly on var inside the view
+        if isinstance(cpm_var, NegBoolView):
+            return 1-self.solver_var(cpm_var._bv)
```
tias, line 187: ohhh, this worries me

solver_var() should return a solver-native object, not an expression...

This can break a lot of things, e.g. or(~bv0, ~bv1, ~bv2)?

Is there a gurobi-native way to do negation?

If not, maybe we should add a transformation that replaces ~bv by 1-bv in all expressions, and only continue from that?

cpmpy/solvers/gurobi.py
tias: more on rounding (for bool, and int below): https://support.gurobi.com/hc/en-us/community/posts/360055376572-Preferred-way-to-check-if-variable-value-is-non-zero

tias: while looking how to negate a Binary variable, some more namely a way to tell the MIP solver to be more precise about binaries: https://or.stackexchange.com/questions/5163/no-gurobi-i-really-do-want-this-variable-to-be-binary

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+        if not cpm_var in self._varmap:
+            if isinstance(cpm_var, _BoolVarImpl):
+                revar = self.gbi_model.addVar(vtype=GRB.BINARY, name=cpm_var.name)
+            elif isinstance(cpm_var, _IntVarImpl):
+                revar = self.gbi_model.addVar(cpm_var.lb, cpm_var.ub, vtype=GRB.INTEGER, name=str(cpm_var))
+            else:
+                raise NotImplementedError("Not a known var {}".format(cpm_var))
+            self._varmap[cpm_var] = revar
+            # Update model to make new vars visible for constraints
+            self.gbi_model.update()
```
tias, line 199: do we really need this? I think it will cause a performance penalty for large models...

we don't introspect the gbi_model, I don't think we need to tell it to update its internal datastructures now. We can just post constraints with that variable, and it iwll 'update' the model when we ask it to solve it?

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+            `minimize()` can be called multiple times, only the last one is used
+
+            (technical side note: any constraints created during conversion of the objective
+            are premanently posted to the solver)
+        """
+        from gurobipy import GRB
+
+        # make objective function non-nested
+        (flat_obj, flat_cons) = (flatten_objective(expr))
+        self += linearize_constraint(flat_cons)  # add potentially created constraints
```
tias, line 218: This linearizie not needed, this is done in __add__ anyway.

Given how for all solvers the minimize/maximize is so similar, should we perhaps make them superclass functions, and have one objective(self, expr, minimize=True) function instead to implement in the interfaces?

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+
+            What 'primitive' means depends on the solver capabilities,
+            more specifically on the transformations applied in `__add__()`
+
+            Solvers do not need to support all constraints.
+        """
+
+        # Base case: Boolean variable
+        if isinstance(cpm_expr, _BoolVarImpl):
+            return self.gbi_model.addLConstr(self.solver_var(cpm_expr), ">", 1)
```
tias, line 282: I don't understand that you write ">" for sense (2nd argument), sense should be:
   sense: Sense for the new constraint (GRB.LESS_EQUAL, GRB.EQUAL, or GRB.GREATER_EQUAL).
so ">" is confusing to me

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+        # Operators: base (bool), lhs=numexpr, lhs|rhs=boolexpr (reified ->)
+        elif isinstance(cpm_expr, Operator):
```
tias, line 286: I'm not sure you should keep this in just to throw an error... you can assume that the input to this function have been through your transformations and hence only of the expressions you want (and if not throw a general NotImplemented or Exception("unexpected...") at the end of this function...)

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+            # 'and'/n, 'or'/n, 'xor'/n, '->'/2
+            if cpm_expr.name in ["and", "or", "xor", "->"]:
+                raise Exception(f"{cpm_expr} should have been linearized, see /transformations/linearize.py")
+            else:
+                raise NotImplementedError("Not a know supported ORTools Operator '{}' {}".format(
+                    cpm_expr.name, cpm_expr))
+
+
+        # Comparisons: only numeric ones as 'only_bv_implies()' has removed the '==' reification for Boolean expressions
```
tias, line 295: comment not relevant as you want bv == and(...) to go to here

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+            if isinstance(lhs, _NumVarImpl):
+                lhs = self.solver_var(lhs)
+            elif isinstance(lhs, Operator) and (lhs.name == 'sum' or lhs.name == 'wsum'):
+                # a BoundedLinearExpression LHS, special case, like in objective
+                lhs = self._make_numexpr(lhs)
+                # assumes that gurobi accepts sum(x) >= y without further simplification
+
+            # post the comparison
+            if cpm_expr.name == '<=':
+                return self.gbi_model.addLConstr(lhs, "<", rvar)
```
tias, line 320: the "<" is again very confusing...

cpmpy/solvers/gurobi.py
tias: maybe this should be done in linearize actually? just like you translate an or?

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,376 @@
...
+                    if not isinstance(lhs.args[1], _NumVarImpl):
+                        a, b = self.solver_vars(lhs.args)
+                        return self.gbi_model.addLConstr(a / b, "=", rvar)
+                    raise Exception("Gurobi only supports division by constants")
+
+                elif lhs.name == 'pow':
+                    x, a = self.solver_vars(lhs.args)
+                    assert not isinstance(a, _NumVarImpl) or a.lb >= 0, f"Gurobi only supports power expressions with positive exponents."
+                    return self.gbi_model.addGenConstrPow(x, rvar, a)
+
```
tias, line 366: You are not using the single-reified wsum?
https://www.gurobi.com/documentation/9.5/refman/py_model_agc_indicator.html

that is a very powerful one that we will be able to use a lot I think... much better then bigM I would expect

tias (commented):
nice start, some comments

cpmpy/solvers/gurobi.py
IgnaceBleukx: Yes, I agree. It should indeed become a property of the superclass. I'll add it to the superclass on this branch while leaving the other solvers as is for now. The purpose of the branch might otherwise be compromised ;)

cpmpy/solvers/gurobi.py
IgnaceBleukx: Gurobi models have a default objective value "Linexp(0)" so I could check if the objective equals that expression?

cpmpy/solvers/gurobi.py
IgnaceBleukx: Yes, indeed. I'll add it in the template as well.

cpmpy/solvers/gurobi.py
IgnaceBleukx: Hmm something like "cpm_var.value = solver_varl >= 0.5" seems the way it is done in the sudoku example of gurobi

cpmpy/solvers/gurobi.py
IgnaceBleukx: This is a leftover artifact of a previous version I had, will remove it.

cpmpy/solvers/gurobi.py
IgnaceBleukx: Would indeed make more sense to put in the linearization lib. I will add it there

cpmpy/solvers/gurobi.py
IgnaceBleukx: That would make sense indeed. Difference between minimize and maximize is at most one line in all solvers so...

cpmpy/solvers/gurobi.py
IgnaceBleukx: Yes, it seems that I have missed that one.
The implies constraint got translated automatically by get_or_make_var. I will fix it in the next commit

IgnaceBleukx: On second tought, I don't think the indicator constraint of gurobi is directly applicable to our case. We cannot have something like (a -> b) == c as it gets flattened to (~a | b) == c

cpmpy/solvers/gurobi.py
IgnaceBleukx: After doing some more research, I don't think there is native gurobi way of doing this. I'll add a transformation for it then

cpmpy/solvers/gurobi.py
tias: I'm not sure you interpreted the docs right, it supports the following:
(BV == True) -> w * IVs >= 5
or
(BV = False) -> w * IVs = 1
etc

so BV -> lincons
for which I think you sometimes do a manual big-M in linearize.

Note that 'only_bv_implies' translates BV == lincons into two BV -> lincons
so the 'only_bv_implies' with a transformation that linearizes expressions to lincons and the above constraint is quite powerful for reified expressions...

IgnaceBleukx: Ok, yes I indeed misinterpreted the docs.
We do lose the power of the "generalconstraints" if we apply the 'only_bv_implies' transformation tho.
At the moment we can post `a & b == c` to gurobi as `grb_model.addGenConstrAnd(c, [a,b])` But this `==` will of course be translated to implications. I am not sure which is the more optimal way to go here.

tias: Yes, we would lose that and don't want to lose that.

Perhaps we should add an 'eq_exept' argument that leaves constraints with that expr.name untouched in only_bv_implies 'BE == BV' part? This might be something we will need for other transformations too at some point, a list of 'only these' or 'all execpt these' names...

To be discussed. because only_bv_implies also transforms BoolExpr -> BV to ~BV -> ~BoolExpr, e.g. the Boolean literal at the left-hand side; I don't think you now support BoolExpr -> BV, so this part of only_bv_implies would also be usefl?

tias (commented):
from meeting

cpmpy/transformations/linearize.py
```diff
@@ -0,0 +1,326 @@
...
+        if lhs.name == "sum":
+            new_var, cons = get_or_make_var(-rhs)
+            return cons + [lhs + new_var <= -1]
+        rhs_minus_1, cons = get_or_make_var(rhs - 1)
+        return cons + [lhs <= rhs_minus_1]
+
+    if cpm_expr.name == ">":
+        lhs, rhs = cpm_expr.args
+        if lhs.name == "sum":
+            new_var, cons = get_or_make_var(-rhs)
```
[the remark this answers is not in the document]
IgnaceBleukx: I suppose this should not become wsum -1 is a constant?

cpmpy/solvers/ortools.py
```diff
@@ -296,328 +210,321 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, **kwargs)
...
+            Requires or-tools >= 8.2!!!
         """
-        if is_num(cpm_expr):
-            return cpm_expr
+        assert (self.ort_status == ort.INFEASIBLE), "get_core(): solver must return UNSAT"
 
-        # decision variables, check in varmap
-        if isinstance(cpm_expr, _NumVarImpl): # _BoolVarImpl is subclass of _NumVarImpl
-            return self.ort_var(cpm_expr)
+        # use our own dict because of VarIndexToVarProto(0) bug in ort 8.2
```
tias, line 511: stray comment.

Your change broke it though: get_core() expects a list of CPMpy variables, but you return the ortools variables.

I don't think we have a way of going from the ort_var to the cpm_var. So perhaps the old 'own list' approach is still the best? (there never has been another implementation)

tias (changes requested):
broken get_core()

cpmpy/solvers/ortools.py
IgnaceBleukx: Indeed, this should return CPMpy vars...
Also present in the solvers branch probably as I branched off that one...
Can't we retrieve the cpm_vars from the _varmap? They should be mapped there, although in the other direction...

IgnaceBleukx: smth along the lines of:
```
ort_assum_vars = [self.ort_model.VarIndexToVarProto(i).name for i in assum_idx]
return [cpm_var for cpm_var in self._varmap if cpm_var.name in ort_assum_vars]
```

cpmpy/transformations/linearize.py
```diff
@@ -0,0 +1,411 @@
...
+        if lhs.name == "wsum":
+            return [Operator("wsum", [lhs.args[0] + [-1], lhs.args[1] + [1]]) >= rhs]
+        else:
+            cons = lhs - 1 >= rhs
+        cons = flatten_constraint(cons)
+        cons = linearize_constraint(cons)
+        return cons
+
+    if cpm_expr.name == "!=":
+        lhs, rhs = cpm_expr.args
```
tias, line 131: add smth like
-- TODO: this can be optimized, example cases:
  - if rhs is constant and lhs is var, then make a Boolean for every value of lhs, add sum(bool_lhs) = 1, also sum(bool_lhs*idx_lhs)=lhs potentially, and then that bool_lhs[rhs] <= 0.  While probably not very understandable, this is how it seems to be implemented in numberjacks mipwrapper: https://github.com/eomahony/Numberjack/blob/53fa9e994a36f881ffd320d8d04158097190aad8/Numberjack/solvers/MipWrapper/MipWrapper.cpp#L95
  - other case of rhs is constant but lhs is expression, make `lhs == aux` for an appropriate auxiliary variable then do the above

(I'm not suggesting to implement this necessarily, just to add the above for future reference...)

tias (commented):
Very nice and solid, also the additional tests are a fantastic addition

tias (approved):

cpmpy/transformations/flatten_model.py
```diff
@@ -590,7 +589,7 @@ def normalized_numexpr(expr):
 
     elif isinstance(expr, Operator) and expr.name == 'wsum': # unary
         weights, sub_exprs  = expr.args
-        flatvars, flatcons = zip(*[get_or_make_var(arg) for arg in sub_exprs]) # also bool, reified...
+        flatvars, flatcons = map(list, zip(*[get_or_make_var(arg) for arg in sub_exprs])) # also bool, reified...
```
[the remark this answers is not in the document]
IgnaceBleukx: Yes, the original flatvars and flatcons were tuples, will commit this directly to master

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,435 @@
...
+                # functional globals only exist for equality in gurobi
+                # example: min(x) > 10 :: min(x) == aux, aux > 10
+                # create the equality and overwrite lhs with auxiliary (will handle appropriate bounds)
+                (lhs, cons) = get_or_make_var(lhs)
+                self += cons
+
+            # all but '==' now only have as lhs: const|ivar|sum|wsum
+            # translate ivar|sum|wsum so they can be posted directly below
+            if isinstance(lhs, _NumVarImpl):
+                lhs = self.solver_var(lhs) # Case can be omitted -> handled in _make_num_expr
```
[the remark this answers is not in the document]
IgnaceBleukx: Depends if we want to support multiplication in the objective as well?

cpmpy/transformations/linearize.py
```diff
@@ -0,0 +1,411 @@
...
+        constraints += [sum(np.arange(n) * sigma) == idx]
+        constraints += [Comparison(cpm_expr.name, np.dot(arr,sigma), cpm_expr.args[1])]
+
+        return linearize_constraint(flatten_constraint(constraints))
+
+
+
+
+
+    if isinstance(cpm_expr, GlobalConstraint):
```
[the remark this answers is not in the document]
IgnaceBleukx: Yes, maybe we could give the decompose() a non-default argument "transform" which applies the transformation(s) given in a list to the decomposition directly?
Then this case can be handled in _post_constraint and can be made specific to any solver?

cpmpy/transformations/linearize.py
```diff
@@ -0,0 +1,411 @@
...
+        return linearize_constraint(only_bv_implies(flatten_constraint(cpm_expr.decompose())))
+
+    return [cpm_expr]
+
+
+def is_lin(cpm_expr):
```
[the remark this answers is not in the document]
IgnaceBleukx: Was a helper function I was planning to use in only_positive_bv but ended up not using, can be removed.

cpmpy/expressions/core.py
```diff
@@ -438,7 +442,11 @@ def value(self):
         elif self.name == "pow": return arg_vals[0] ** arg_vals[1]
         elif self.name == "-":   return -arg_vals[0]
         elif self.name == "abs": return -arg_vals[0] if arg_vals[0] < 0 else arg_vals[0]
-        
+        elif self.name == "and": return all(arg_vals)
+        elif self.name == "or" : return any(arg_vals)
+        elif self.name == "xor": return sum(arg_vals) % 2 == 1
+        elif self.name == "->": return (not arg_vals[0]) or arg_vals[1]
+
```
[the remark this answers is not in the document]
IgnaceBleukx: Cherry picked commits and put in master

cpmpy/transformations/flatten_model.py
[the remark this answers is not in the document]
IgnaceBleukx: Cherry picked commit and pushed to master

tests/test_examples.py
```diff
@@ -24,7 +24,21 @@ class TestExamples(unittest.TestCase):
     loader = importlib.machinery.SourceFileLoader("example", example)
     mod = types.ModuleType(loader.name)
     loader.exec_module(mod)
-    
+
+    # run again with gurobi
+    if "npuzzle" in example or "tst_likevrp" in example:
```
IgnaceBleukx, line 29: We should make this better/modular
i.e. a list of solver names to run the tests on at the top of the file

cpmpy/solvers/gurobi.py
```diff
@@ -0,0 +1,435 @@
...
+    Module details
+    ==============
+"""
+
+from .solver_interface import SolverInterface, SolverStatus, ExitStatus
+from ..expressions.core import *
+from ..expressions.variables import _BoolVarImpl, NegBoolView, _IntVarImpl, _NumVarImpl, intvar
+from ..transformations.flatten_model import flatten_constraint, flatten_objective, get_or_make_var
+from ..transformations.get_variables import get_variables
+from ..transformations.linearize import linearize_constraint, no_negation
```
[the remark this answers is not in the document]
IgnaceBleukx: Change to actual name of function `only_positive_bv`

cpmpy/transformations/linearize.py
```diff
@@ -0,0 +1,411 @@
...
+    if isinstance(cpm_expr, Comparison):
+        lhs, rhs = cpm_expr.args
+        return isinstance(lhs, _NumVarImpl) or lhs.name == "sum" or lhs.name == "wsum"
+
+    if isinstance(cpm_expr, Operator) and cpm_expr.name == "->":
+        cond, subsexpr = cpm_expr.args
+        return isinstance(cond, _BoolVarImpl) and is_lin(subsexpr) and is_num(subsexpr.args[1])
+
+
+def only_positive_bv(cpm_expr):
```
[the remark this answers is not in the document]
IgnaceBleukx: This is the new version of the "no_negation" transformation, somehow this was not called from gurobi.py, it is fixed now.

cpmpy/solvers/gurobi.py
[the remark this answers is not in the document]
IgnaceBleukx: Yes we do, will implement

cpmpy/transformations/linearize.py
```diff
@@ -0,0 +1,411 @@
...
+            var, cons = get_or_make_var(var * arg)
+            constraints += cons
+
+        return constraints + [Comparison(cpm_expr.name, var, rhs)]
+
+    gen_constr = ["min","max","abs","pow"] #General constraints supported by gurobi, TODO: make composable for other solvers
+    if cpm_expr.name == '==':
+        lhs, rhs = cpm_expr.args
+        if lhs.is_bool() and not isinstance(lhs, _BoolVarImpl) and isinstance(rhs, _BoolVarImpl):
+            # BE == BV :: ~BV -> ~BE, BV -> BE
```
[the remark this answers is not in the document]
IgnaceBleukx: Will impelement transform() method for solverinterfaces
Quadratic objective functions for later

cpmpy/transformations/linearize.py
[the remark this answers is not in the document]
IgnaceBleukx: Remove

PR #94, opened 2022-01-14
Add set_solution_hint to Model

jrobben:
Add a method to set a solution hint in Model which will be offered to the solver when solve() is called


tias:
Not all solvers support solution hinting, its peculiar to SAT-based methods. But if we add it to Model, we should also add it to the other solver interfaces....

If you know you are using ortools, you could also do instead of
```o = Model()'''
replace by
``` o = SolverLookup.get("ortools")'''
and everything (including solution hints) will continue to work.

Would that suit your use as well?

PR #102, opened 2022-03-09
Copy method for model and constraints

IgnaceBleukx:


cpmpy/expressions/core.py
```diff
@@ -131,6 +134,23 @@ def is_bool(self):
     def value(self):
         return None # default
 
+    def _copy_args(self, memodict={}):
```
tias, line 137: add a one line comment what it does, and also why it exists (versus just being implemented in copy())

cpmpy/expressions/core.py
```diff
@@ -131,6 +134,23 @@ def is_bool(self):
...
+                if isinstance(arg, Expression):
+                    memodict[arg] = arg.copy(memodict)
+                elif is_num(arg) or isinstance(arg, bool):
+                    memodict[arg] = arg
+                else:
+                    raise ValueError(f"Not a supported argument to copy: {arg}")
+            copied.append(memodict[arg])
+        return copied
+
+    def copy(self, memodict = {}):
```
tias, line 150: Add docstring

cpmpy/expressions/core.py
```diff
@@ -315,6 +339,11 @@ def value(self):
         elif self.name == ">=": return (arg_vals[0] >= arg_vals[1])
         return None # default
 
+    def copy(self, memodict={}):
```
tias, line 342: Add docstring

cpmpy/expressions/globalconstraints.py
```diff
@@ -138,6 +138,10 @@ def decompose(self):
         """
         return None
 
+    def copy(self, memodict={}):
```
tias, line 141: Add docstring... same for all the rest : )

cpmpy/model.py
```diff
@@ -231,4 +231,18 @@ def from_file(fname):
         """
         with open(fname, "rb") as f:
             return pickle.load(f)
-    
+
+    def copy(self, memodict={}):
+        """
+            Copies a the model to a new instance.
```
tias, line 237: Returns a copy of the model?

something about the 'memodict' argument too.

and something about it being a shallow copy or deep copy?

IgnaceBleukx: I was thinking about shallow/deep copy distinctions as well. But I can see no reason why someone would need a shallow copy of a model.
As variables are still shared between models then, solving one model will affect the variables in the other one.
Might still be usefull to clarify that it is a deep copy anyway indeed

IgnaceBleukx:
Updated the pullrequest with results of our earlier discussion.
Renamed the newly implemented copy methods to `deepcopy` and added a shallow copy method for models.

Issue #96, opened 2022-01-21
lb/ub computation of expressions


tias:
This is currenlty done in cpmpy/transformations/flatten.py, function 'get_or_make_var()'

Should we take it out into a separate 'transformation' that only computes the bounds?

Or should we add a function get_lb()/get_ub() to the Expression objects directly?

Undecided, opinions welcome.

IgnaceBleukx:
I think bounds calculation can become a separate transformation.
This would allow other transformations to calculate bounds without having to deal with the side effects of get_or_make_var, such as constraint generation.
Adding them in the core would add a lot of code, although it would allow us to cache the result of get_lb()/get_ub() so future lookups can be more efficient. Of course assuming that uses do not change an expression manually...

As the code is structured around the transformations, my preference goes to implementing a separate transformation for this

IgnaceBleukx:
Related to this is the following bug:
```python
arr = boolvar(shape=3, name="arr")
idx = intvar(lb=0, ub=2, name="idx")

cons  = arr[idx]

get_or_make_var(cons)
>> (IV0, [arr[0],arr[1],arr[2]][idx] == IV0)
```
Instead of creating an integer variable, we should create a new boolean variable.

This messes things up when assigning cons to a new variable and using only_bv_implies. IV0 then occurs as base constraint which is not allowed as it is numeric...
For example:

```python
a = boolvar(name="a")
b = boolvar(name="b")

cons = a == (arr[idx] & b)
only_bv_implies(flatten_constraint(cons))
>> [ (~a) -> AND([IV0 == 0, b]), (a) -> AND([IV0,b]), [arr[0],arr[1],arr[2]][idx] == IV0]
 


```

tias:
We had shortly dsicussed the latter in person.

The real issue is that an Element of a an array that is all Boolean, should be a Boolean expression (so its .is_bool() should be true. (This is btw a fantastic reason why it was good that we decided that 'is_bool()' is a property, and not a subclassing into BoolExpr and NumExpr)

Anyway, if we make Element detect whether it is Bool or not, interesting things happen in unexpected ways... see #128
Now an Element() can be a top-level constraint, so ortools tries to decompose it...

What now? Should we special-case in the sovlers that a top-level element should an Element(arr, idx) == True solver constraint? Or should this be in a transformation rule? 

Do we even want this? Should the user not write `cons = a == ((arr[idx] == 1) & b)`? Like in who killed agatha...

[...]

PR #103, opened 2022-03-15
Bare minimum z3 interface

tias:
Not implemented yet:
    - constraints : )
    - solving with assumptions
    - optimisation/objective


cpmpy/solvers/z3.py
```diff
@@ -0,0 +1,303 @@
...
+        # we store them directly in `solver_var()` itself.
+        #self.user_vars.update(get_variables(cpm_con))
+
+        # only complication is that a list is implicitly an 'And' for us,
+        # as well as being a list of arguments, so top-level lists first
+        if is_any_list(cpm_con):
+            # recursively
+            for con in cpm_con:
+                self.__add__(con)
+        else:
```
IgnaceBleukx, line 228: This isn't really compatible with our ideas for putting __add__ in the superclass and supporting by supporting a `transform()` function. 
I think the logic for posting the constraint to the z3 solver should still be contained in the `_post_constraint` method as is done in other solvers.

IgnaceBleukx:
To implement optimization: we need the `Optimize` object instead of the `Solver` object of Z3. Not sure what the downsides are for using `Optimize` for SAT checking

IgnaceBleukx:
Z3 doesn't directly support min/max consrtaints but it is quite easy to implement using upper/lower bounds on the arguments.
E.g.: `max(x,y,z) = m <=>  m >= x.ub & m >= y.ub & m >= z.ub`
So one reason more to fix #96 :)

IgnaceBleukx:
> To implement optimization: we need the `Optimize` object instead of the `Solver` object of Z3. Not sure what the downsides are for using `Optimize` for SAT checking

to answer my own doubts:
https://stackoverflow.com/a/71754240/18784299

cpmpy/solvers/utils.py
```diff
@@ -120,13 +123,12 @@ def lookup(name=None):
             solvername,_ = solvername.split(':',maxsplit=1)
 
         # find CPM_slv
```
tias, line 125: everything below this is a bugfix, should be in a separate commit

cpmpy/solvers/z3.py
```diff
@@ -0,0 +1,424 @@
...
+            sol = self.z3_solver.model() # the solution (called model in z3)
+            # fill in variable values
+            for cpm_var in self.user_vars:
+                sol_var = self.solver_var(cpm_var)
+                if isinstance(cpm_var, _BoolVarImpl):
+                    cpm_var._value = bool(sol[sol_var])
+                elif isinstance(cpm_var, _NumVarImpl):
+                    cpm_var._value = sol[sol_var].as_long()
+
+            # TODO
```
tias, line 163: the TODO seems resolved?

cpmpy/solvers/z3.py
tias: I see, I agree. So yes, no transforms here and so this content would become post_constraint, go ahead

examples/advanced/VRP by learning from historical data.ipynb
```diff
@@ -526,7 +526,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.4"
+   "version": "3.9.2"
```
tias, line 529: remove from commit

tias:
Looks great, I left a few minor comments in the code (incl answer to your post_constraint question).

It looks feature complete, is it? If so, do merge after fixing the comments!

IgnaceBleukx:
The reification of decomposable constraints is not fully working yet (such as min, max, abs). But this requires some more refactoring of the code I believe.

tias:
I would be fine if you would throw 'NotImplementedErrors' for those cases and open a bugreport for it. We do need need feature completeness for it to be useful (e.g. no wsum in pysat yet, but still useful as is).
Then it can already be merged and used and tested, e.g. in our incremental work

cpmpy/solvers/z3.py
IgnaceBleukx: For future reference: the transform() function of z3 should then just make sure it returns a list of constraints

PR #104, opened 2022-03-15
Use ortools as a propagator

IgnaceBleukx:
Using the presolve procedure of ortools as propagator to tighten bounds of cpm_vars
This is an example of low level solver access


tias (commented):
Yes, but pls add as examples/advanced/ortools_native_propagate.py or similar

and start your file with a docstring explaining what this example does, and what it shows


also if I understand the 'bounds' correctly, then it is a list of 'ranges', so it can represent domains with 'holes' in it; this could be good to make more explicit, e.g. y != 3 prolly becomes [0,2] [4,5]?

PR #107, opened 2022-03-17
Pysat interrupt

sourdough-bread:
Solve pysat instance with a time limit (auto-interrupt in case runtime exceeds)


tests/test_pysat_interrupt.py
```diff
@@ -0,0 +1,88 @@
...
+            for lit in clause:
+                lit_var = lit_cpmvar[abs(lit)]
+                new_clause.append(~lit_var if lit < 0 else lit_var)
+            m +=any(c for c in new_clause)
+        s = CPM_pysat(m)
+        assumption = [lit_cpmvar[1]]
+        # offset for additional stuff done by cpmpy after solving
+        time_limit, time_offset = 1, 0.5
+        tstart_solving = time.time()
+        s.solve(assumptions=assumption, time_limit=1)
```
IgnaceBleukx, line 83: should be change to `time_limit=time_limit` in function call

sourdough-bread: Done, thanks for the suggestion.

PR #108, opened 2022-03-18
Pysat wsum

sourdough-bread:
Weighted sum

- Added the case where an expression is of the form a * x - b * y where a and b : sum should transform to weighted sum
- Added support with Pysat: weighted sum interpreted as pseudo boolean constraint

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -214,14 +214,18 @@ class Expression(object):
     def __add__(self, other):
         if is_num(other) and other == 0:
             return self
+        print(f"add: {self=} {other=}")
         return Operator("sum", [self, other])
     def __radd__(self, other):
         if is_num(other) and other == 0:
             return self
+        print(f"radd: {self=} {other=}")
         return Operator("sum", [other, self])
 
     # substraction
     def __sub__(self, other):
+        print("self", self)
+        print("other:", other)
         # if is_num(other) and other == 0:
         #     return self
         # return Operator("sub", [self, other])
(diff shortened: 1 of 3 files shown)


cpmpy/expressions/core.py
```diff
@@ -454,6 +455,8 @@ def value(self):
 def _wsum_should(arg):
     """ Internal helper: should the arg be in a wsum instead of sum """
     # Undecided: -x + y, -x + -y?
+    if isinstance(arg, Operator) and arg.name == "-":
```
tias, line 458: I am not sure we should always convert a sum with a - into a wsum...
Ideally we do very few transformatiosn to the expression prior to calling 'flatten()', so that the debugprint of the model looks recognizable for the user.

So, are you sure we need this? Is the special case not handlable somewhere else?

cpmpy/expressions/core.py
```diff
@@ -465,6 +468,10 @@ def _wsum_make(arg):
         return arg.args
     elif arg.name == 'mul':
         return [arg.args[0]], [arg.args[1]]
+    elif arg.name == "-" and isinstance(arg.args[0], Operator):
+        # - (3 * y)
+        w, x = _wsum_make(arg.args[0])
```
tias, line 473: that looks very broad catching... so for any operator, we forably turn it into a wsum and make the 'w' negative?

is there not a more specific case you want to catch?

maybe at least a wsum_shuld on args0?


or is it the recursive case you want to catch; e.g. a wsum in a wsum? I'm again not sure we should do this her ein 'expression', but perhaps rather in flatten...?

cpmpy/solvers/pysat.py
```diff
@@ -224,6 +235,9 @@ def _post_constraint(self, cpm_expr):
             Post a primitive CPMpy constraint to the native solver API
         """
         from pysat.card import CardEnc
+        if not CPM_pysat.pb_supported():
```
tias, line 238: is this not only in case 'PB' is used?

i would not want users not to be able to solve plain SAT models if they want to, because they don't have a library installed they dont want to use.


Also, it looks like you are completely replacing CardEnc by PBenc, would it not make sense to have both, and only use PBenc in case of a wsum?

cpmpy/expressions/core.py
sourdough-bread: No we don't. I added the rewrite rules to the flattening.

cpmpy/expressions/core.py
sourdough-bread: Changed the rewriting rules for - in mul, wsum, .... to flattening.

cpmpy/solvers/pysat.py
sourdough-bread: Adapted the PBenc to Cardenc (regular sum) and PBEnc (wsum)

PR #114, opened 2022-04-01
Pysat wsum

sourdough-bread:
# Weighted sum

- Added case where a boolean variable-based expression is of the form a * x - b * y where a and b: sum should transform to weighted sum.
    - Moved implementation into the flattening to stay true to the user's model.
- Added support with PySAT: 
    - weighted sum interpreted as pseudo boolean constraint
    - sum encoded as a cardinality constraint


cpmpy/expressions/core.py
```diff
@@ -238,6 +238,7 @@ def __mul__(self, other):
             return self
         return Operator("mul", [self, other])
     def __rmul__(self, other):
+
```
tias, line 241: remove all style-only changes, your commit does not actually change core.py

if you want to do style changes, do it as a separate commit

cpmpy/transformations/flatten_model.py
```diff
@@ -185,6 +185,34 @@ def flatten_constraint(expr):
     - Numeric inequality (>=,>,<,<=,): Numexpr >=< Var     (CPMpy class 'Comparison')
     - Reification (double implication): Boolexpr == Var    (CPMpy class 'Comparison')
         """
+        left, right = expr.args[0], expr.args[1]
+
+        if isinstance(left, Operator) and left.name == "wsum" and any(_should_wsum(xi) for xi in left.args[1]):
```
tias, line 190: add a comment about what/why you are doing this here

cpmpy/transformations/flatten_model.py
```diff
@@ -185,6 +185,34 @@ def flatten_constraint(expr):
...
+                if _should_wsum(xi):
+                    wni, xni = _wsum_make(xi)
+                    wni = [wnij * wi for wnij in wni]
+                    w_new += wni
+                    x_new += xni
+                else:
+                    w_new.append(wi)
+                    x_new.append(xi)
+
+            return flatten_constraint(
```
tias, line 203: its a bit unusual to want to flatten something that you just created. Can't you ensure the arguments are flattened, and then just create your expression?

cpmpy/transformations/flatten_model.py
```diff
@@ -449,7 +477,7 @@ def normalized_boolexpr(expr):
     """
     assert(not __is_flat_var(expr))
     assert(expr.is_bool()) 
-
+    #print(f"\nnormalized_boolexpr: {expr=}" )
```
tias, line 480: no style changes (e.g. only adding spaces) nor adding stray prints

cpmpy/transformations/flatten_model.py
```diff
@@ -675,3 +703,23 @@ def negated_normal(expr):
         #raise NotImplementedError("negate_normal {}".format(expr))
         return expr == 0 # can't do better than this...
 
+def _should_wsum(sub_expr):
```
tias, line 706: expressions.core already has a _wsum_should and _wsum_make. They are almost identical?

Do you really need this here? If for the '-', then perhaps make a _wsum_should_flatten() that calls _wsum_should as much as possible... same for _wsum_make_flatten (if a new one is really needed...)

tias (commented):
feedback

cpmpy/transformations/flatten_model.py
sourdough-bread: Removed prints

cpmpy/transformations/flatten_model.py
sourdough-bread: New one allows for flattening of complex expressions that can be simplified into 1 wsum.

sourdough-bread:
New version can be reviewed.

tias:
I wanted to include it in the release I am about to make, so I made some quick cleanups (space changes, solver_vars instead of list of solver_var, a few one-line docs).

But then I noticed:
1) there are two test failing, including in test_flatten, which need to be fixed
2) current master is not merge into it yet, so there might be more (or less) tests failing after that.

If you could fix that, then it can be merged for the next release, it looks ready otherwise (but so there might be some subtle cases still in the new flatten code)

sourdough-bread:
I merged master branch into pysat_wsum and fixed the tests.
If expression contains abs it is not transformed into wsum:

```python3
IV = cp.intvar(0,9, shape=3)
cp.Model(abs(l[0]-l[1])- abs(l[2]-l[1]) < 0)
```
Previously transformed to, but not handled by ortools in post_constraint:
wsum([1, -1], [abs(l[0]-l[1]), abs(l[2]-l[1])])

cpmpy/transformations/flatten_model.py
```diff
@@ -185,8 +187,20 @@ def flatten_constraint(expr):
     - Numeric inequality (>=,>,<,<=,): Numexpr >=< Var     (CPMpy class 'Comparison')
     - Reification (double implication): Boolexpr == Var    (CPMpy class 'Comparison')
         """
+        # <<<<<<< HEAD
```
tias, line 190: something went wrong in your merge here... see another inline merge below.

sourdough-bread:
Ready to be merged all tests passing.

sourdough-bread:
Only tests for int-based constraints are non-passing. 
All tests passing.

cpmpy/solvers/pysat.py
```diff
@@ -285,9 +296,51 @@ def _post_constraint(self, cpm_expr):
...
+                if left.name == "mul":
+                    # single weight,value pair, in list
+                    weights = [left.args[0]]
+                    lits = [self.solver_var(left.args[1])]
+                else: # wsum
+                    weights = left.args[0]
+                    lits = self.solver_vars(left.args[1])
+
+                clauses = []
+                if cpm_expr.name == "<" or (cpm_expr.name == "!=" and bound >= len(lits)):
```
IgnaceBleukx, line 316: You are missing the weights in the condition.
Now `10 * a + 5 * b != 5` is converted to `10 * a + 5 * b <= 4`

cpmpy/solvers/pysat.py
```diff
@@ -285,9 +296,51 @@ def _post_constraint(self, cpm_expr):
...
+                    clauses += PBEnc.leq(lits=lits, weights=weights, bound=bound-1, vpool=self.pysat_vpool).clauses
+                elif cpm_expr.name == "<=":
+                    clauses += PBEnc.leq(lits=lits, weights=weights, bound=bound,vpool=self.pysat_vpool).clauses
+                elif cpm_expr.name == ">" or (cpm_expr.name == "!=" and bound <= 0):
```
IgnaceBleukx, line 320: What about negative weights? I think you need the weights in the condition here as well

cpmpy/transformations/flatten_model.py
```diff
@@ -621,3 +633,46 @@ def negated_normal(expr):
             return negated_normal(Operator('and', expr.decompose()))
         else:
             raise NotImplementedError("negate_normal {}".format(expr))
+
+
+def _wsum_should_flatten(arg):
```
IgnaceBleukx, line 638: Check these additions with pull request #182 and with @tias  :)

tests/test_constraints.py
```diff
@@ -15,7 +15,7 @@
 # Exclude certain operators for solvers.
 # Not all solvers support all operators in CPMpy
 EXCLUDE_OPERATORS = {"gurobi": {"mod"},
-                     "pysat": {"sum", "wsum", "sub", "mod", "div", "pow", "abs", "mul","-"}}
```
IgnaceBleukx, line 18: I wil make some adjustments here. Pysat does not have to support all `sum's` and `wsums` right? Only sums and wsums where args are bool.

IgnaceBleukx (changes requested):
Some edge cases leftover in post_constraint and more automatic testing can be done (I'll fix the test_constraints file for this)

cpmpy/solvers/pysat.py
```diff
@@ -243,40 +253,41 @@ def _post_constraint(self, cpm_expr):
...
+
             # only handle cardinality encodings (for now)
-            if isinstance(cpm_expr.args[0], Operator) and cpm_expr.args[0].name == "sum" and all(
-                    isinstance(v, _BoolVarImpl) for v in cpm_expr.args[0].args):
-                lits = self.solver_vars(cpm_expr.args[0].args)
-                bound = cpm_expr.args[1]
+            if isinstance(left, Operator) and left.name == "sum" and all(isinstance(v, _BoolVarImpl) for v in left.args):
+                lits = self.solver_vars(left.args)
                 if not is_int(bound):
                     raise NotImplementedError(f"PySAT sum: rhs `{bound}` type {type(bound)} not supported")
```
IgnaceBleukx, line 263: Ideally, this would be something like `NotSupportedError`, as we cannot implement it :)

cpmpy/transformations/flatten_model.py
tias: that commit was about being conservative at expression creation time about when to create a wsum, e.g. the expression should be close to what the user wrote.

Here we are in flatten, and it makes sense to do some additional optimisations to avoid unnecessary auxiliary variable creation (which is what this is).

Could it make sense/be possible to separate the addition of these optimisations from the wsum branch? Yes, less expressions will be supported then in this branch (because any auxiliary intvar created is not supported), but then I can look at the optimisation at a different pace and with e.g. a list of example expressions we want to optimize...

tias (changes requested):
see open comments

also, we merged the `transform()` refactor so pysat interface changed a bit, which needs conflict/merging.

but importantly, I want the changes to 'flatten_model.py' as a separate pull request. These are optimisations that improve the flattening, but imho can be treated independently of pysat. I think it also means that the pysat_wsum tests should become generic flatten tests?

If yes, we can merge this branch before checking the optimisations in more detail. (the preferred option)

If no, we should first get the optimisations on that separate branch right, and then merge this

tias:
@IgnaceBleukx you probably best do the resolve of test_constraints yourself

tias:
@IgnaceBleukx the tests indeed test bvar*bvar, which pysat does not support (and throwns an exception on now); so I had to remove 'mul' from the tests to not trigger this test.

However, it also tests that for z3, which does not support it either.
So can you instead remove the bv*bv test from the automatic constraint generation and not test that? I don't understand where it is generated or how to stop it being generated...


Note that this code is rebased on the unmerged #316, you wait with a further code review till that one is merged.

cpmpy/solvers/pysat.py
```diff
@@ -285,9 +296,51 @@ def _post_constraint(self, cpm_expr):
...
+            # WEIGHTED !
+            elif isinstance(left, Operator) and (left.name in ["wsum", "mul"]) and is_int(bound):
+
+                if not CPM_pysat.pb_supported():
+                    raise ImportError("Please install PyPBLib: pip install pypblib")
+                from pysat.pb import PBEnc
+
+                if left.name == "mul":
+                    # single weight,value pair, in list
+                    weights = [left.args[0]]
```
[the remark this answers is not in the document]
tias: the current _pysat_pb throws a not-implemented
(for multiplications of boolvars it could add an auxiliary literal, but Ok to just refuse in general)

cpmpy/solvers/pysat.py
tias: current version generates '[(BV0) -> (sum([10, 5] * [a, b]) <= 4), (~BV0) -> (sum([10, 5] * [a, b]) >= 6)]'
which I think is correct? @IgnaceBleukx can you resolve this thread if you agree?

tias: OK, I was confused here... it is the 'linearize' transfo that I added today for testing that did that rewriting.
You mean that it should be bound >= sum(weights) I think... but then indeed what about negatives (your comment lower). We could do bound >= sum(max(0,weights)) but what are we doing here then...

Either the bound is that max, and you can just post it as <= bound-1, or the bound is higher then the max and then the constraint is always true...

Maybe we should just drop this 'optimisation'?
I'm hoping that the PB encoding itself will check the 'always true' edge cases and so cover this appropriately?

cpmpy/solvers/pysat.py
tias: removed special case in to-be commited version; resolve if OK (ignace)

cpmpy/solvers/pysat.py
tias: actually OK to throw NotImplementedError...

e.g. if rhs is a sum, then the entirety is a linear constraint and if some transformation (see below, eg does linearize do this?) write all linear equalities in some normal form then it would be supported...

tias:
OK, so first, we need to check for 'out of bounds' bounds in both card/pb, as the encoding libraries simply refuse to encode a trivially true/false constraint.

I might also accidentally have removed the bv < 1 support case (or it used to be in simplify_bool and now it aint no more?)...

Then, all pb/cardinality expressions are expected to be of the following normalized form:
wsum <= const  (potentially reified)

From a CPMpy design that is fine: a solver interface only implements the constraint formulations it 'natively' supports.

So then the question is, do we have a transformation that rewrites e.g. wsum <= var or wsum >= wsum into such 'normalized' linear constraints?
I would expect 'linearize'? @IgnaceBleukx ?

But this interface is more sensitive to the creation of auxliary integer variables (e.g. for bv[0] > bv[1]+bv[2]). So how far do we want to go with this?

hbierlee:
I'm working on this PR, everything seemed to work quite well already with the only failing tests were for trivially sat/unsat constraints for which linearize now returns, `BoolVal(True)/BoolVal(False)` respectively.

The last failing test was that `only_positive_bv` for linear solvers receives a `NegBoolView` (e.g. `~p`), which I've handled by `p<=0`. I'm not sure if there is a better way either within CPMpy or for linear solvers to fix boolean variables?

I haven't carefully read all the code yet, I will do this now to see if there are other improvement, but I think we can re-review the PR already to see if it's mergeable.

hbierlee:
After offline discussion, the question remains whether outputting a Boolean value from linearize is still optimal (since it's not a linear; if we go this route we should update linearize's spec). An alternative solution is to return actual linears `[] = 0` and `[] = 1` for true/false which does (technically) adhere to the spec?

tests/test_pysat_cardinality.py
```diff
@@ -96,6 +87,40 @@ def test_pysat_atleast_equals(self):
...
+            self.bvs[0] > (self.bvs[1] | self.bvs[2]),
+        ]
+
+        ## check all types of linear constraints are handled
+        for expression in expressions:
+            cp.Model(expression).solve("pysat")
+
+    def test_pysat_oob(self):
+
+        def test_encode_pb_oob(self):
```
IgnaceBleukx, line 105: Why this nested def? The inner one will not be called by pytest

tests/test_solvers.py
```diff
@@ -310,6 +310,18 @@ def test_pysat(self):
...
+    @pytest.mark.skipif(not CPM_pysat.supported(),
+                        reason="PySAT not installed")
+    def test_pysat_card(self):
+        b = cp.boolvar()
+        x = cp.boolvar(shape=5)
+
+        cons = [sum(x) > 3, sum(x) <= 2, sum(x) == 4, (sum(x) <= 1) & (sum(x) != 2),
+                b.implies(sum(x) > 3), b == (sum(x) != 2), (sum(x) >= 3).implies(b)]
+        for c in cons:
+            cp.Model(c).solve("pysat")
```
IgnaceBleukx, line 322: Also assert the solve call here

cpmpy/solvers/pysat.py
```diff
@@ -258,10 +269,14 @@ def transform(self, cpm_expr):
         cpm_cons = toplevel_list(cpm_expr)
         cpm_cons = decompose_in_tree(cpm_cons)
```
IgnaceBleukx, line 270: `alldifferent` has a specialized decomposition in linearize, so you can also add it to the `supported` set here.

cpmpy/transformations/linearize.py
```diff
@@ -69,16 +69,19 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
 
         # boolvar
         if isinstance(cpm_expr, _BoolVarImpl):
-            newlist.append(sum([cpm_expr]) >= 1)
+            if "bv" in supported:
```
IgnaceBleukx, line 72: Hmm we typically don't do this as the supported set contains names of constraints, but I can see why it's done here. 
Personally, I would prefer it to stay a sum but no strong objection to this change.
In any case, update the description of the `linearize` normal form above.

IgnaceBleukx (commented):
Nice! Seems like this acient PR is getting there : )
I fully agree with putting pysat directly after linearize, with the `has_subexpr` optimization that was merged recently, the overhead should be ok (maybe we do have to test that tho... see #512 ).

I'm not sure we need to full transformation stack however, see the comments in the files themselves.
You can also remove the old code instead of commenting it, that makes the review process (imho) a little easier

hbierlee:
@IgnaceBleukx I believe it's mergeable; tests seem stable and I think the code/docs are decent.

hbierlee:
One second..

hbierlee:
Approximately one minute later..

Now I'm a lot happier with the changes to linearize, because it is simplified by never returning Boolean literals (instead turning them into trivial constraint, linear or clausal, depending on the solver). An even better approach would be to simply fix them to their value (so they might propagate further), but I believe this is not usually done so should be left for another issue.

cpmpy/solvers/pysat.py
```diff
@@ -381,48 +413,77 @@ def get_core(self):
...
+        if cpm_expr.name == "<=":
+            return CardEnc.atmost(**pysat_args).clauses
+        elif cpm_expr.name == ">=":
+            return CardEnc.atleast(**pysat_args).clauses
+        elif cpm_expr.name == "==":
+            return CardEnc.equals(**pysat_args).clauses
+
+    def _pysat_pseudoboolean(self, cpm_expr):
+        """ Convert CPMpy comparison of `wsum` (over Boolean variables) into PySAT list of clauses """
+        assert_normalized_bool_lin(cpm_expr)
```
IgnaceBleukx, line 443: This assertion might slow down the posting process quite a bit due to the isinstance checks. We'll do that in `solver_var` either way, so I don't think it's really necessary to add this assertion here too.

cpmpy/solvers/pysat.py
```diff
@@ -381,48 +413,77 @@ def get_core(self):
...
+            )
+
+        lits = self.solver_vars(lhs.args)
+        pysat_args = { "lits": lits, "bound": rhs, "vpool": self.pysat_vpool }
+
+        if cpm_expr.name == "<=":
+            return CardEnc.atmost(**pysat_args).clauses
+        elif cpm_expr.name == ">=":
+            return CardEnc.atleast(**pysat_args).clauses
+        elif cpm_expr.name == "==":
```
IgnaceBleukx, line 438: Terminate this elif-chain with an else throwing a ValueError : )

cpmpy/solvers/pysat.py
```diff
@@ -381,48 +413,77 @@ def get_core(self):
...
+        # unpack and transform to PySAT arguments
+        lhs, rhs = cpm_expr.args
+        lits = self.solver_vars(lhs.args[1])
+        pysat_args = {"weights": lhs.args[0], "lits": lits, "bound": rhs, "vpool":self.pysat_vpool }
+
+        if cpm_expr.name == "<=":
+            return PBEnc.atmost(**pysat_args).clauses
+        elif cpm_expr.name == ">=":
+            return PBEnc.atleast(**pysat_args).clauses
+        elif cpm_expr.name == "==":
```
IgnaceBleukx, line 463: same here: also raise ValueError in else-case

cpmpy/transformations/linearize.py
```diff
@@ -52,33 +52,41 @@
...
     Any other unsupported global constraint should be decomposed using `cpmpy.transformations.decompose_global.decompose_global()`
+    - `reified`: whether the constraint is fully reified
 
     """
 
     newlist = []
     for cpm_expr in lst_of_expr:
-
-        # boolvar
+        # Boolean literal are handled as trivial linears or unit clauses depending on `supported`
```
IgnaceBleukx, line 71: literals

cpmpy/transformations/linearize.py
```diff
@@ -100,11 +108,23 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
                 # BV -> LinExpr
                 elif isinstance(cond, _BoolVarImpl):
                     lin_sub = linearize_constraint([sub_expr], supported=supported, reified=True)
-                    newlist += [cond.implies(lin) for lin in lin_sub]
+                    for lin in lin_sub:
+                        # linearize might return True/False for trivial constraints
+                        if is_true_cst(lin):
```
IgnaceBleukx, line 113: Can you add a TODO here to check performance of this check? The happy-path is now the longest one which might make the transformation slower.

cpmpy/transformations/linearize.py
```diff
@@ -179,6 +199,18 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
             [cpm_expr] = canonical_comparison([cpm_expr])  # just transforms the constraint, not introducing new ones
             lhs, rhs = cpm_expr.args
 
+            # check trivially true/false (not allowed by PySAT Card/PB)
+            if cpm_expr.name in ('<', '<=', '>', '>=') and is_num(rhs):
```
IgnaceBleukx, line 203: What if i post a constraint `x + y + z == 10`, this is trivially UNSAT but not catched here?

IgnaceBleukx: In the future, we can actually move these checks into the `get_bounds` function on Comparisons (and Boolean operators)

IgnaceBleukx (commented):
Nice, I think the testing-infrastructure is getting pretty robust, but left some comments on the changed to linearize and the solver-interface itself

cpmpy/transformations/linearize.py
hbierlee: I've added a test for this, it seems to be ok.

I've realized that instead of guessing when PySAT will raise an exception for trivial unsatisfiable constraints, we should just catch this error. However, there are two issues from preventing this approach at the moment, for which I created a ticket for it https://github.com/pysathq/pysat/issues/194). Essentially, PySAT has a case where it is incorrect about trivial unsat, and it throws a generic ValueError (which is dangerous to catch). It also seems they do support negative weights. So the integration can be much simplified.

The question is whether we fix the current clunky approach or whether we wait for a PySAT release (potentially with our own PR).

hbierlee:
The current PR is mergeable, but could be improved in the following ways:

- The main issue as discussed above: the integration with PySAT could be simpler if we fix https://github.com/pysathq/pysat/issues/194, since we are now unnecessarily normalizing negative weights, and guessing which trivially unsat constraints need to be removed, while we could just offload this to pysat once. The question is whether to merge now or wait until this is fixed upstream (something I could make a PR for).
- decomposition of `among` seems to output wsum over integer variables. I removed its test, I'm not sure if I need to do other things to say, pysat does not support this? Or should we further investigate (e.g. git bisect) why `among` regressed?

cpmpy/transformations/linearize.py
tias: If its correct now, you can merge it now.

If it becomes easier later, we can make it easier later : )

hbierlee:
Hi @IgnaceBleukx if you can find the time, could you have a final look? Probably we can merge after that.

hbierlee:
Hi @IgnaceBleukx if you can find the time, could you have a final look? Probably we can merge after that.

IgnaceBleukx (commented):
Two more small things:
- The (seamingly) unused assertion-helper function (I might have missed  it in the code)
- (minor) try to cache the supportedness instead of doing the check for each constraint we post

cpmpy/solvers/pysat.py
```diff
@@ -386,48 +423,69 @@ def get_core(self):
...
+            return PBEnc.equals(**pysat_args).clauses
+
+"""
+  An assert to test whether cpm_expr is a normalized Boolean linear constraint of the form `LinExpr <=/==/>= Constant`.
+"""
+def assert_normalized_bool_lin(cpm_expr):
```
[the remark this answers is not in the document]
hbierlee: No because it's too "heavy", but for me it is very useful to have heavy debug asserts which are only run during development/tests. I'm not sure how to guarantee that users do not run this. We have the [`-O` flag](https://docs.python.org/3/tutorial/modules.html#compiled-python-files), but are we sure this is always enabled in production for all users? And does it impact the assertions in the dependencies (possibly breaking them?)? Unsure (so unsafe).

So, my suggestion is to remove it for this PR, then hardcode our own DEBUG flag which is checked-in to be `False`, and enabled for testing and you can manually change it during development. Then we can add "heavy" pre- and post-condition checks (based on the docs) on many of the transformations (I think you said this was complicated, but it can't be as complicated as writing the transformation, right? Also they don't have to be fully comprehensive)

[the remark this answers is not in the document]
IgnaceBleukx: Ok so I agree on removing this function from the PR, not sure how to proceed with Debugging/Assertion related stuff in the future...
I suppose there is some pythonic way of doing things like that -- feel free to open an issue where we can discuss (and potentially implement such debug-flags in a follow-up PR).

After this is removed and my other comment is resolved, ready to merge imo.

cpmpy/solvers/pysat.py
```diff
@@ -386,48 +423,69 @@ def get_core(self):
...
+            raise ValueError(f"PySAT: Expected Comparison to be either <=, ==, or >=, but was {cpm_expr.name}")
+
+    def _pysat_pseudoboolean(self, cpm_expr):
+        """ Convert CPMpy comparison of `wsum` (over Boolean variables) into PySAT list of clauses """
+        if cpm_expr.args[0].name != "wsum":
+            raise NotSupportedError(
+                f"PySAT: Expect {cpm_expr} to be a 'wsum'"
+            )
+
+        if not CPM_pysat.pb_supported():
```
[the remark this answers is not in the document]
tias: like cpo's `get_docp()` I guess you mean? which is instead of a staticmethod, a class method that creates cls._docp (here cls._pb_supported) and returns that once it exists?

I agree that doing imports and a try/catch on each constraint will have a high overhead, and should be cached.

cpmpy/solvers/pysat.py
tias: +1

cpmpy/solvers/pysat.py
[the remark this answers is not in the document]
tias: yes, not appropriate here. Adding asserts on 'else' cases above is a lightweight way to catch some of what this would catch.

The other way is just a very different type of test; e.g. a test where you do PySAT.transform(cons) and then test the things below... so not as inline test, but as outside test.

cpmpy/solvers/pysat.py
```diff
@@ -255,18 +266,25 @@ def transform(self, cpm_expr):
...
         cpm_cons = toplevel_list(cpm_expr)
-        cpm_cons = decompose_in_tree(cpm_cons)
+        cpm_cons = decompose_in_tree(cpm_cons, supported=frozenset({"alldifferent"}))
         cpm_cons = simplify_boolean(cpm_cons)
-        cpm_cons = flatten_constraint(cpm_cons)
+        cpm_cons = flatten_constraint(cpm_cons)  # flat normal form
         cpm_cons = only_bv_reifies(cpm_cons)
         cpm_cons = only_implies(cpm_cons)
-        cpm_cons = canonical_comparison(cpm_cons)
+        cpm_cons = linearize_constraint(cpm_cons, supported=frozenset({"sum","wsum", "and", "or", "bv"}))  # the core of the MIP-linearization
```
tias, line 287: right, so I understand you added it for and/or, but I dont think "bv" should be explicitly listed?

cpmpy/transformations/linearize.py
```diff
@@ -277,6 +302,17 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
...
+                lb,ub = lhs.get_bounds()
+                t_lb = eval_comparison(cpm_expr.name, lb, rhs)
+                t_ub = eval_comparison(cpm_expr.name, ub, rhs)
+                if t_lb and t_ub:
+                    continue
+                elif not t_lb and not t_ub:
+                    newlist += linearize_constraint([BoolVal(False)], supported=supported) # post the linear version of False
```
tias, line 313: that seems like a very indirect way... you are in this function; just do the 'or' check here, then return Or(false), otherwise sum(...)?

tests/test_globalconstraints.py
```diff
@@ -1302,7 +1302,8 @@ def test_among(self):
         iv = cp.intvar(0,10, shape=3, name="x")
 
         for name, cls in cp.SolverLookup.base_solvers():
-
+            # The decomposition of this global introduces (as of yet) unsupported integer variables for PySAT
+            if name in ("pysat",): continue
```
tias, line 1306: no need for a membership of a 1 sized list...? just name == "pysat"...

tias (changes requested):
yes, impressive nr of extra tests too.

Some minor things to change

cpmpy/transformations/linearize.py
hbierlee: If I move that code here, then linearize can't handle the expression of BoolVal(False) (it's a trivial expression, but it still solver-dependent). So this seems like the right way to me.

tias: I meant just copy the 2-line if here? I guess this is such a rare case that it doesn't actually matter to save 1 function call, so you can also leave it as is I realize now

cpmpy/solvers/pysat.py
hbierlee: Ah, indeed should be removed (no longer uesd in linearize since boolvars are now properly linearized for each solver)

cpmpy/solvers/pysat.py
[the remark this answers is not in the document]
hbierlee: I resolved this similar to CPO (although I also avoid the import if PBLib is not installed, maybe a slight improvement). I heard that an import statement once caused huge overhead, although I'm a bit surprised by this as I believe imports are cached (so compared to the function call it should make no difference, as this SO answer claims https://stackoverflow.com/questions/57346840/does-python-cache-imported-files and not entirely confirmed here in https://docs.python.org/3/reference/import.html)

cpmpy/transformations/linearize.py
hbierlee: Ah duplicate it; it is only 2 lines, I agree, but this can change in the future, and I rather stick to the approach where we just use linearize to linearize something not linear

tias (commented):
minor things

cpmpy/transformations/linearize.py
```diff
@@ -3,6 +3,9 @@
 
 Linearized constraints have one of the following forms:
 
+Unsatisfiable:
```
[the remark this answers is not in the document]
hbierlee: Ah I'm surprised too. This is outdated, since BoolVal(False) is not a linear constraint.

hbierlee:
Good comments, I've resolved them.

One change now is that I just now see `pblib` as a "recommended" optional dependency by automatically installing it when we run `pip install cpmpy[pysat]` (so it will just do `pip install python-sat[pblib]`). I think this is the best approach as most users want to solve (weighted) sums, and advanced user or (or when pblib breaks) can work around this by still doing `pip install python-sat`.

IgnaceBleukx (approved):
Looks good to me! Fine to merge this.

PR #115, opened 2022-04-04
Xor global constraint

IgnaceBleukx:
Implemented the xor-constraint as a global constraint with it's decomposition.
Edited the post_constraint code for all implemented solvers to work with the new global constraint.

Issues (help wanted):
- `xor([BV1,BV2,BV3]) -> BV4` is not working in pysat because of negated_normal procedure. Not sure how to fix this... Maybe we should stay in XOR-space then and negate the last argument in the XOR?
- Should `XOR` also inherit from `Operator` or only from `GlobalConstraint`?


cpmpy/expressions/__init__.py
```diff
@@ -20,6 +20,6 @@
 # others need to be imported by the developer explicitely
 from .variables import boolvar, intvar, cpm_array
 from .variables import BoolVar, IntVar, cparray # Old, to be deprecated
-from .globalconstraints import AllDifferent, AllEqual, Circuit, Table, Minimum, Maximum, Element
+from .globalconstraints import AllDifferent, AllEqual, Circuit, Table, Minimum, Maximum, Element, XOR
```
tias, line 23: We used single capital alwasy for now, e.g. Xor()... any strong reason to want to full-cap here?

cpmpy/expressions/globalconstraints.py
```diff
@@ -102,12 +102,14 @@ def my_circuit_decomp(self):
...
 import warnings # for deprecation warning
-from .core import Expression
+from .core import Expression, Operator
 from .variables import boolvar, intvar, cpm_array
-from .utils import flatlist, all_pairs, argval
+from .utils import flatlist, all_pairs, argval, is_num
 
 
 # Base class GlobalConstraint
+from ..transformations.flatten_model import get_or_make_var
```
tias, line 111: this should be on line 108 instead of below the # Base class... comment I think

cpmpy/expressions/globalconstraints.py
```diff
@@ -285,3 +287,38 @@ def value(self):
...
+        return cons + [prev_var]
+
+    def value(self):
+        return sum(argval(a) for a in self.args) % 2 == 1
+
+    def __repr__(self):
+        if len(self.args) == 2:
+            return "{} xor {}".format(*self.args)
+        return "xor({})".format(self.args)
+
```
tias, line 324: does it require a change for the deepcopy? this is in master now

cpmpy/expressions/globalconstraints.py
```diff
@@ -285,3 +287,38 @@ def value(self):
...
+        The 'xor' constraint for more then 2 arguments.
+        Acts like cascaded xor operators with two inputs
+    """
+
+    def __init__(self, arg_list):
+        if len(arg_list) == 2 and is_num(arg_list[1]):
+            arg_list[0], arg_list[1] = arg_list[1], arg_list[0]
+        i = 0  # length can change
+        while i < len(arg_list):
+            if isinstance(arg_list[i], GlobalConstraint) and arg_list[i].name == "xor":
```
tias, line 301: you could isinstance this on XOR/Xor directly...

cpmpy/expressions/globalconstraints.py
```diff
@@ -285,3 +287,38 @@ def value(self):
...
         return "{}[{}]".format(self.args[0], self.args[1])
 
+class XOR(GlobalConstraint):
+    """
+        The 'xor' constraint for more then 2 arguments.
+        Acts like cascaded xor operators with two inputs
+    """
+
+    def __init__(self, arg_list):
+        if len(arg_list) == 2 and is_num(arg_list[1]):
```
tias, line 297: this is to make it more canonical, e.g. num first? not sure that is really needed, but if you prefer then do add a line-comment perhaps

tests/test_constraints.py
```diff
@@ -3,38 +3,38 @@
 from cpmpy import boolvar, intvar, Model, cpm_array
 from cpmpy.expressions.core import Comparison, Operator
 from cpmpy.expressions.globalconstraints import *
-from cpmpy.solvers import CPM_gurobi, CPM_ortools, CPM_minizinc
+from cpmpy.solvers import CPM_gurobi, CPM_ortools, CPM_minizinc, CPM_pysat
```
tias, line 6: I'm not sure this file/changes is intended in this pull request?
should be separate I think

tias (commented):
nice, few minor changes perhaps

cpmpy/expressions/__init__.py
IgnaceBleukx: No special reason. I'll change it to make it coherent with other global constraints

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: You mean swapping `len(arg_list) == 2` and `is_num(arg_list[1])`?
Might result in anoying errors when length of arg_list is 1...

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Yes, does not follow Globalconstraint constructor syntax, so needs its own `_deepcopy` method

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Fixed with extra comment, similar to the one in the Operator constructor

tests/test_constraints.py
IgnaceBleukx: I used this test set to test the new XOR constraint in all solvers, so that's why it has changed for Pysat as well...

Issue #74, opened 2021-10-27
How to use ortool's global constraints which are not integrated by CPMpy?


hakank:
The ortools CP-SAT solver has quite a few constraints that's not (yet) integrated in CPMpy, for example;  AddAutomaton, AddCumulative,  AddForbiddenAssignments, AddInverse, AddNoOverlap, AddNoOverlap2D, AddReservoirConstraint. 

It would be great to be able to use these constraints even if they are now defined in globalconstraints.py and/or ortools.py.

There are two methods in principles. Here I use `AddAutomaton` as an example (since I want to use it in my Nonogram solver :-)).

First, the approach that was suggested by Tias in a mail exchange a while ago. For a simple model it works by adding the constraint to the solver object, e.g. 
```
import sys
import numpy as np
from cpmpy import *
from cpmpy.solvers import *

def get_different_solution(m, x):
  m += [any([t.value() != t for t in x])]

# Callback function for printing the solution
def print_solution(x):
    for a in x:
        print([val for val in a.value()])

def automaton_test(n=7,num_sols=0):
    # The DFA for AddAutomaton
    transitions = [
        (0,0,0), # 0*
        (0,1,1), # 1*
        (1,1,1), # 
        (1,0,2), # 0*
        (2,0,2)  #
        ]

    initial_state = 0 
    # all states are accepting states
    accepting_states = [0, 1, 2]

    # declare variables
    x = boolvar(shape=n,name="x")
    z = intvar(0,n,name="z")
    
    # constraints
    
    # Currently one have to include x in the model.
    model = Model(z == sum(x) )
    
    ss = CPM_ortools(model)
    # This (the next two lines) was suggested by Tias and it works:
    ort_vars = [ss.varmap[v] for v in x]
    ss.ort_model.AddAutomaton(ort_vars, initial_state,accepting_states,transitions)

    num_solutions = 0
    while ss.solve():
        num_solutions += 1
        print("x:",x.value(), "z:", z.value())
        get_different_solution(ss,x)
    print("num_solutions:",num_solutions)

n=7
automaton_test(n)
```

This works and is fast. Note that is use the `CPM_ortools` object and one have to translate the CPMpy variables to ORtools variables (the `ss.varmap` step).

However, for larger models this can be quite messy.  It would be really nice if one could use the constraint as any other constraint, i.e. using via `model += [AddAutomaton(...)]` and without the step to translate the CPMpy variables.

[...]

PR #121, opened 2022-05-26
NativeConstraint prototype

tias:
See #74 

A proposal to make native constraints (API calls) more accessible, like we use global constraints.

```python
from cpmpy import *
from cpmpy.expressions.globalconstraints import NativeConstraint
x = intvar(0,10, shape=3)

# CPMpy standard
s = SolverLookup.get("ortools")
s += Table(x, [[0,1,2],[2,1,0]])
print(s.solveAll())

# CPMpy with currently recommended 'direct solver access' use
s = SolverLookup.get("ortools")
s.ort_model.AddAllowedAssignments(s.solver_vars(x), [[0,1,2],[2,1,0]])
print(s.solveAll())

# Proposal in branch
s = SolverLookup.get("ortools")
s += NativeConstraint("AddAllowedAssignments", (x, [[0,1,2],[2,1,0]]))
print(s.solveAll())

# Proposal also allows to indicate some args are const/novar/should be passed as is
# the top two variants (now) also do not scan the data part
s = SolverLookup.get("ortools")
s += NativeConstraint("AddAllowedAssignments", (x, [[0,1,2],[2,1,0]]), arg_novar=[1])
print(s.solveAll())
```

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -387,3 +387,31 @@ class Xor(GlobalConstraint):
        """
         copied_args = self._deepcopy_args(memodict)
         return Xor(copied_args)
+
+class NativeConstraint(Expression):
+    """
+        A constraint whose name corresponds to a native solver API function,
+        and whose arguments match those of the solver function.
+
+        A solver interface will map the arguments to native solver variables
+        where possible, and will call the native API function.
+    """
+    # is_bool: whether this is normal constraint (True or False)
+    def __init__(self, name, arg_list, arg_novar=None):
+        """
+            arg_list: list of arguments to give to the function with name 'name'
+            arg_novar: list of arguments that contain no variables,
+                       that can be passed 'as is' without scanning for variables
+        """
+        super().__init__(name, arg_list)
+        self.arg_novar = arg_novar
+
+    def is_bool(self):
+        """ is it a Boolean (return type) Operator?
+        """
+        return True
+
+    def deepcopy(self, memodict={}):
+        copied_args = self._deepcopy_args(memodict)
+        return type(self)(self.name, copied_args)
+
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -454,9 +454,9 @@ class CPM_ortools(SolverInterface):
         elif cpm_expr.name == 'alldifferent':
             return self.ort_model.AddAllDifferent(self.solver_vars(cpm_expr.args))
         elif cpm_expr.name == 'table':
-            assert (len(cpm_expr.args) == 2)  # args = [array, table]
-            array, table = self.solver_vars(cpm_expr.args)
-            return self.ort_model.AddAllowedAssignments(array, table)
+            assert (len(cpm_expr.args) == 2)
+            array, table = cpm_expr.args
+            return self.ort_model.AddAllowedAssignments(self.solver_vars(array), table)
         else:
             # TODO: NOT YET MAPPED: Automaton, Circuit, Cumulative,
             #    ForbiddenAssignments, Inverse?, NoOverlap, NoOverlap2D,
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -27,6 +27,7 @@ import sys  # for stdout checking
 
 from .solver_interface import SolverInterface, SolverStatus, ExitStatus
 from ..expressions.core import Expression, Comparison, Operator
+from ..expressions.globalconstraints import NativeConstraint
 from ..expressions.variables import _NumVarImpl, _IntVarImpl, _BoolVarImpl, NegBoolView
 from ..expressions.utils import is_num, is_any_list
 from ..transformations.get_variables import get_variables_model, get_variables
@@ -438,6 +439,14 @@ class CPM_ortools(SolverInterface):
             raise NotImplementedError(
                         "Not a know supported ORTools left-hand-side '{}' {}".format(lhs.name, cpm_expr))
 
+        # Proposal: dynamic mapping of cpm_expr.name to API call? see #74
+        elif isinstance(cpm_expr, NativeConstraint):
+            # get the native function, will raise an AttributeError if it does not exist
+            native_function = getattr(self.ort_model, cpm_expr.name)
+            native_args = [a if cpm_expr.arg_novar is not None and i in cpm_expr.arg_novar
+                             else self.solver_vars(a) for i,a in enumerate(cpm_expr.args)]
+            # len(native_args) should match nr of arguments of `native_function`
+            return native_function(*native_args)
 
         # rest: base (Boolean) global constraints
         elif cpm_expr.name == 'xor':
@@ -455,7 +464,6 @@ class CPM_ortools(SolverInterface):
             
             # global constraint not known, try posting generic decomposition
             self += cpm_expr.decompose() # assumes a decomposition exists...
-            # TODO: dynamic mapping of cpm_expr.name to API call? see #74
             return None # will throw error if used in reification
         
         raise NotImplementedError(cpm_expr)  # if you reach this... please report on github


IgnaceBleukx:
Yes, I like this idea a lot.
I am wondering however if we want to support **kwargs in the constructor as well? I think it's more natural to post constraints using just the "native" names of its parameters. `arg_novar` can then be a list of names instead of indices?

cpmpy/expressions/globalconstraints.py
```diff
@@ -387,3 +387,31 @@ def deepcopy(self, memodict={}):
        """
         copied_args = self._deepcopy_args(memodict)
         return Xor(copied_args)
+
+class NativeConstraint(Expression):
```
IgnaceBleukx, line 391: Don't we want to subclass `GlobalConstraint` for this? Then current transformations will be able to work as is

maxasauruswall:
Hello All, 

I employ the ortools interval constraints quite a bit in my code.

I'd love to use this NativeConstraint to get all my problem's constraints into the cpmpy model.constraints list (we make use of this downstream, and having some constraints only in the ortools solver makes the code messy). 

Unfortunately, the ortools interval constraints mix the concept of constraint and variable by requiring that the constraint itself be named ([see docs](https://developers.google.com/optimization/reference/python/sat/python/cp_model#newintervalvar)).

Since string values can't be part of the `arg_list`, even if included in the `arg_novar` list (because they cannot be reduced to variable expressions, unlike numbers?), I've created a little work-around. I'm wondering if this is an approach folks would consider, or if there is a better way to do this? 

Below are screenshots of an example test (I put it as a new case in the `test_native.py::TestNativeORTools` class), along with the changes to the `NativeConstraint` class and the ORTools solver method for posting native constraints.

Thanks again for the wonderful package! 
Cheers, 
Max



image

image

image

maxasauruswall:
Apologies, there is a typo in the above code. The `arg_novar` param should have been `[1]`, not "[size]".

tias:
Yes, the name and more examples.

1. the name:
I'm currently most in favor of "DirectConstraint" as in, a constraint that is directly posted to the solver through an API call with the same name.
Or indeed "NativeConstraint" a constraint that has a correspondingly named native API call that will be called.

1. examples:

I think the IntervalVar of Max is really the missing piece. But the provided example is not enough: it mixes variable and constraint indeed, with as purpose to be used in other constraints (e.g. AddCumulative). So, what do we need to do to support that?

Perhaps we need to add to our ORT object a
```
def IntervalVar(name=..., ...):
  return s.ort_solver.addIntervalVar(name=....)
```
meaning that if you do `s = SolverLookup("ortools")` you can do `v = s.IntervalVar(...)`
after which, the NativeCosntraint/DirectConstraint could allow:
```
s.NativeConstraint("AddCumulative", ([v], 2, 4), arg_novar=[1,2])
```
or so?

If we can think of a clean way to have both NativeConstraint and these intervalvar/optionalvars, then we greatly increase the things we can model in CPMpy (for this and future solvers)

IgnaceBleukx:
1. DirectConstraint or NamedConstraint are my favorites at the moment. Although DirectConstraint might be more google'able friendly...
2. Adding the `AddIntervalVar` to the ORT-object would not really help Max in this case as he needs the list of constraints donwstream. To do this, the constraint has to be posted to the model instead of the solver... I think we could make a `DirectVar` as well which subclasses our Variable class...

tias:
`DirectVar` is a really neat idea. Because then the model is still separated from the solver object...

Could you work out an example to see if that would work?

IgnaceBleukx:
I implemented the DirectVar idea for OR-Tools. It is surprisingly elegant. I only had to make minor changes to `get_variables` and `solver_var`.
Added a test for the NoOverlap constraint of OR-tools and it seems to work as intended.
We should add some more documentation to the class DirectVar nevertheless.

cpmpy/expressions/globalconstraints.py
tias: No, because then the `decompose()` function should be supported, while the idea is that this can only be used directly.

transformations should continue to work though... it should handle any expression (e.g. leave these as is)

cpmpy/expressions/__init__.py
```diff
@@ -18,8 +18,8 @@
 
 # we only import methods/classes that are used for modelling
 # others need to be imported by the developer explicitely
-from .variables import boolvar, intvar, cpm_array
+from .variables import boolvar, intvar, cpm_array, DirectVar
```
tias, line 21: for user-facing functions we are using lowercase (e.g. boolvar, intvar) I think I will lowercase this too

tias: or is this the _impl and should we allow shape=... too

also, there is a difference between the name of a variable and the name of the internal API call...

will think some more about it

tias:
I've did the reverse exercise: how would we want to model with these new capabilities? See 'examples/jobshop_ortools.py' added, a slight modification from jobshop.py, where I use the current 'direct solver access' capabilities of CPMpy to create custom variables and constraints.

Current, variables:
```python
# make intervalvars... ensures start + dur = end
intervals = np.zeros(shape=(machines_count,jobs_count), dtype=object)
for m in all_machines:
    for j in all_jobs:
        intervals[m,j] = model.ort_model.NewIntervalVar(
            model.solver_var(start_time[m,j]), jobs_data[j,m], model.solver_var(end_time[m,j]), f"interval[{m},{j}]")
```

where we should discuss what we want, for example maximal convenience could be with:
```python
intervals = directvar("NewIntervalVar", [start_time, jobs_data.T, end_time], shape=start_time.shape, name="interval")
```
where `directvar()` is a helper function with a shape, that creates an nparray like `intvar()` does (e.g. of some custom `_DirectVariable` class)

side-note: in ortools creating these variables ensures that their constraints are ensured. It is a mix of variable and constraint... so we should probably allow to
```python
m += intervals
```
because people might create intervalvars without using them in future constraints while still wanting the relation between start/size/end to be enforced...


Then, the overlap constraint is with current direct access:
```python
for j in all_jobs:
    model.ort_model.AddNoOverlap(intervals[:,j])
```

and could become:
```python
for j in all_jobs:
    model += DirectConstraint("AddNoOverlap", intervals[:,j])
```

Please comment @IgnaceBleukx and others watching this.

IgnaceBleukx:
> where directvar() is a helper function with a shape, that creates an nparray like intvar() does (e.g. of some custom _DirectVariable class)

That would absolutely be the preferred way of doing things indeed. I hadn't thought of  adding a `shape` arg but it makes perfect sense!

> side-note: in ortools creating these variables ensures that their constraints are ensured. It is a mix of variable and constraint... so we should probably allow to

This is very easy to do right? Very similar to something like `model += boolvar()`. `_post_constraint` will call `solver_var`, thereby creating the interval variables on the OR-Tools native solver object (and thus corresponding constraints). We simply need a check at the start of `_post_constraint`. If and when we decide to go trough with this idea (and we should), this must also become part of the TEMPLATE.py I think.

Now most of the Fuzz-test bugs are fixed, I think this pull request and the bounds-computation issues are highest on the wishlist!

tias:
Yes, this is very high on the wishlist as it expands the capabilities a lot. Also now it will be possible to use the ortools optionalintervalvars etc with the same easy, and without us having to declare it at CPMpy level for all solvers...

I've implemented the necessary components and hooked them into ortools. Works nicely.

We should now add more examples (e.g. ortools vrp with its shiny new multiplecircuit? the PCTSP of maxime? an RCPSP or similar with optional intervals?) it would also be great if we could do, and demonstrate, floatvars in gurobi with this?

We should also rework the docs accordingly. Currently it is described in 'solvers' under 'native solver access', but I think that should now become part of the modeling docs? Maybe all of the 'solvers' doc page should become part of the modeling page actually... I noticed many other packages do not mind having long individual pages where users get an overview on everything without having to click to different subsections...

IgnaceBleukx:
I think the DirectConstrait/Var should definetely be in the modelling part of the documentation indeed.
And yes, we need a lot of examples for this... Many of the arguments/features are non-trivial such as the `insert_name_at_index`.

For me what is still missing from this is a way to get the value of a `DirectVar` in the solution of a problem. Especially for float variables in Gurobi this is very usefull.

tias:
We now have two very generic building blocks that makes it easier for users to use the underlying solver, 'declarative style'.

Anything further than that, like getting a value, sounds very solver specific. If users want to do that they can use the standard direct solver access... we can show that in the examples too...

Not sure about gurobi, but I guess it would be something like (well, in this case you perhaps dont even need the solver object, but just the native variable object?, untested!):
```
v = directvar("addVar", [0, 2.5])
s = SolverLookup.get("gurobi")
s += v > 0.5
s.solve()
print("val: ", s.solver_var(v).X)
```

to be tried when making examples...


also if the 'insert_name_at_index' proves to messy, we can switch to named arguments and 'name_as_argument' option or so... all really to be seen as we use it...

tias:
We will split this branch into two:

- DirectConstraint, this is clear, and we want it in the next release like this (for all solvers). It will get a separate pull request
- DirectVar, we need to play with it a bit more, it should live in this branch a bit longer

I will initiate this now

tias:
Converted to (lesser priority) draft. Please say something here if a 'DirectVariable' is what you need, we need a bit more use cases/testing before merging this.

IgnaceBleukx:
I have worked some more on this feature, also adding tests for all solvers and some documentation in `modelling.md` on how to use the `directvar`.

Should be ready for review

tias:
There are two clear use cases:

1. Float variables; we will not support float variables, it requires a complete overhaul of the language and DirectVars would just be an ugly hack that doesn't even go halfway.
2. Interval variables and Optional interval variables. See #606, instead of generic DirectVariables, we will instead add OptionalCumulative and OptionalNoOverlap; where solver interfaces can create interval variables behind the scenes like we do for Cumulative and NoOverlap.

Issue #105, opened 2022-03-17
Implication of globals in ortools


IgnaceBleukx:
```python

x,y = intvar(0,3,shape=2)
p = boolvar()

s1 = SolverLookup.lookup("ortools")()

s1 += p.implies(max([x,y]) != 1)
s1.solve()
>> True

s2 = SolverLookup.lookup("ortools")()

s2 += p.implies(max([x,y]) == 1)
s2.solve()
>>  Exception: ('OR-Tools says: model invalid:', 'Enforcement literal not supported in constraint: enforcement_literal: 0 int_max { target: 4 vars: 1 vars: 2 vars: 3 }')
```

While posting the constraint in the first case, it gets split up in the following:
` [max([x,y]) == IV0, p -> IV0 != 1]`

In the second case, this does not happen as `MaxEquality` is natively supported by OR-tools.

So to answer this todo: yes :)
https://github.com/CPMpy/cpmpy/blob/b821f441f9b38613ed4221d43b59a9344837f5dc/cpmpy/solvers/ortools.py#L355

IgnaceBleukx:
Ok, so one way to solve this is to always replace the global constraint with a new integer/boolean variable on the rhs of the implication.
Question remains: do we want a new transformation for this or can this simply be done in `_post_constraint`? Imo it would be better to put it in a new transformation, together with this code snipped that is now present in both ortools and gurobi:
https://github.com/CPMpy/cpmpy/blob/f85c8e64f0d7397887ff11d6529053d7202de620/cpmpy/solvers/ortools.py#L378-L385
@tias what do you think?

tias:
OK, so this includes quite a number of things.

Cleanest would probably be a new generic transformation, (not in flatten itself?), that rewrites reified constraints as appropriate, e.g.:

# Case 1, Operator == BV with Operator.bool() = true
#   assume Boolean Operators 'and' and 'or' and '->' are supported by solver...
#   (-> could be rewriten to or, should be in separate transf? can discuss in Github)
# Case 2, Comparison(OP, lhs, rhs) == BV
#   assume supported by solver if lhs and rhs are const/var
#   have list of supported lhs's? such as sum and wsum...
#   in any other case (e.g. any NumExpr lhs), decompose to [aux = get_or_make_var(lhs), Comparison(OP, aux, rhs) == BV]
#   warning, this 'get_or_make_var' may only be valid for total expressions (that always have a valid result)
[...]

PR #127, opened 2022-06-09
transf: add reify_rewrite, generic way of handling reification if onl…

tias:
…y few constraints support it by the solver

Implementation of the discussion in #105 
used in ortools and gurobi

only limitly tested, e.g. 
```
ivs = intvar(1,3, shape=5, name="ivs")
rv = boolvar(name="rv")

e = []
e += [(rv == AllDifferent(ivs))]
e += [(Element(ivs, intvar(1,9)) <= 2).implies(rv)]
e += [rv.implies(max(ivs) < 5)]
sat = Model(e).solve("ortools")
print(sat, ":", rv.value(), ivs.value())
```

@IgnaceBleukx can you test it more extensively with your test suite? Quite possible there are edge cases we are not covering here...

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -373,7 +373,7 @@ class Xor(GlobalConstraint):
 
     def decompose(self):
         if len(self.args) == 2:
-            return (self.args[0] + self.args[1]) == 1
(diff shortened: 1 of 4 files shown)


IgnaceBleukx:
Looks very nice!
Except for the XOR case, all tests work just fine

PR #128, opened 2022-06-10
globals: Element, detect whether arr will return a Bool

tias:
to support the (side-tracked) discussion in #96


tias:
Closing, will not implement, use Element(arr, idx) == 1 (or ==0) if you want a logical expression.

The main issue is that 'Element(arr,idx)' would then suddenly be a valid stand-alone constraint, but solvers can only and the equality version, and so do we as a language. Changing this requires changes at multiple places (language, transformations and solvers)

PR #132, opened 2022-06-27
Decaprio

IgnaceBleukx:
Implemented the Hamming distance surrogate from our hyperparameter tuning paper.

### DIFF
--- a/cpmpy/tools/tune_solver.py
+++ b/cpmpy/tools/tune_solver.py
@@ -1,14 +1,18 @@
 """
     This file implements parameter tuning for constraint solvers based on SMBO and using adaptive capping.
-    Based on the paper "Model-Based Algorithm Configuration with Adaptive Capping and Prior Distributions by I. Bleukx, S. Berden et al.
+    Based on the following paper:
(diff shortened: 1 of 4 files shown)


tias (approved):
excellently executed.

The TODO support time_limit, this is a manual time.time() check I guess?
could be given to first solve call, and then max(best_r, t.time()-t0) kind of thing?

IgnaceBleukx:
Yes indeed. Idea is to be able to give a time budget to the tuner to try configurations. You suggestion might be a good way to implement it indeed

cpmpy/tools/tune_solver.py
```diff
@@ -0,0 +1,83 @@
+import numpy as np
+
+from ..solvers.utils import SolverLookup, param_combinations
+from ..solvers.solver_interface import ExitStatus
+
+class ParameterTuner:
+
+    def __init__(self, solvername, model, all_params=None, defaults=None):
+        """
+            Parameter tuner based on DeCaprio method [ref_to_decaprio]
```
tias, line 10: you really have to insert publication details before we merge.

Ideally at top of file, something like
"""
   Parameter tuning with adaptive capping and...

   Based on the paper "..." by ... at ...
   - DOI: url with DOI (ppl can find the bibtex there)
   - paper: link to pdf
   - code used in the paper: link

   The code currently implements ...

PR #134, opened 2022-07-28
Pysat rc2

sourdough-bread:
Basic implementation of MaxSAT PySAT RC2 solver.

### DIFF
--- a/cpmpy/solvers/pysat_rc2.py
+++ b/cpmpy/solvers/pysat_rc2.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+##
+## pysat.py
+##
+"""
+    Interface to PySAT's API
+
+    PySAT is a Python (2.7, 3.4+) toolkit, which aims at providing a simple and unified
+    interface to a number of state-of-art Boolean satisfiability (SAT) solvers as well as
+    to a variety of cardinality and pseudo-Boolean encodings.
+    https://pysathq.github.io/
+
+    This solver can be used if the model only has Boolean variables,
+    and only logical constraints (and,or,xor,implies,==,!=) or cardinality constraints.
+
+    Documentation of the solver's own Python API:
+    https://pysathq.github.io/docs/html/api/solvers.html
+
+    WARNING: CPMpy uses 'model' to refer to a constraint specification,
+    the PySAT docs use 'model' to refer to a solution.
+
+    ===============
+    List of classes
+    ===============
+
+    .. autosummary::
+        :nosignatures:
+
+        CPM_rc2
+"""
+from ..expressions.core import Operator
+from ..expressions.variables import _BoolVarImpl
+from ..transformations.to_cnf import to_cnf
+from ..transformations.get_variables import get_variables
+from .pysat import CPM_pysat
+from .solver_interface import SolverInterface, SolverStatus, ExitStatus
+
+class CPM_RC2(CPM_pysat):
+    """
+    Interface to PySAT's API
+
+    Requires that the 'python-sat' python package is installed:
+    $ pip install python-sat
+
+    See detailed installation instructions at:
+    https://pysathq.github.io/installation.html
+
+    Creates the following attributes (see parent constructor for more):
+    pysat_vpool: a pysat.formula.IDPool for the variable mapping
+    pysat_solver: a pysat.solver.Solver() (default: glucose4)
+    """
+
+    @staticmethod
+    def supported():
+        # try to import the package
+        try:
+            import pysat
+            # there is actually a non-related 'pysat' package
+            # while we need the 'python-sat' package, some more checks:
+            from pysat.formula import IDPool
+            from pysat.solvers import Solver
+            return True
+        except ImportError as e:
+            return False
+
+    @staticmethod
+    def solvernames():
+        """
+            Returns solvers supported by PySAT on your system
+        """
+        from pysat.solvers import SolverNames
+        names = []
+        for name, attr in vars(SolverNames).items():
+            if not name.startswith('__') and isinstance(attr, tuple):
+                if name not in attr:
+                    name = attr[-1]
+                names.append(name)
+        return names
+
+    def __init__(self, cpm_model=None, subsolver="glucose3"):
+        """
+        Constructor of the native solver object
+
+        Requires a CPMpy model as input, and will create the corresponding
+        PySAT clauses and solver object
+
+        Only supports satisfaction problems (no objective)
+
+        Arguments:
+        - cpm_model: Model(), a CPMpy Model(), optional
+        """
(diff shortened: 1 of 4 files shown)


cpmpy/solvers/pysat_rc2.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+##
+## pysat.py
```
IgnaceBleukx, line 4: Update documentation

cpmpy/solvers/pysat_rc2.py
```diff
@@ -0,0 +1,172 @@
...
+        elif isinstance(expr, Operator) and expr.name == "wsum":
+            weights, vars = expr.args
+            pysat_assum_vars = self.solver_vars(vars)
+            for weight, pysat_assum_var in zip(weights, pysat_assum_vars):
+                self.pysat_solver.add_clause([pysat_assum_var], weight=weight)
+        else:
+            raise NotImplementedError(f"Expression {expr} not handled")
+
+    def get_core(self):
+        raise NotImplementedError("Does not work.")
```
IgnaceBleukx, line 172: I think the error message should be a little more elaborate :) Something like "RC2 does not support unsat core extraction, check out the PySat solver for this functionality..."

cpmpy/solvers/utils.py
```diff
@@ -72,6 +73,7 @@ def base_solvers():
                 ("pysat", CPM_pysat),
                 ("pysdd", CPM_pysdd),
                 ("minizinc", CPM_minizinc),
+                ("rc2", CPM_RC2)
```
IgnaceBleukx, line 76: Just a thought: maybe we can group the list of solvers for each paradigm, or sort them alphabetically?

cpmpy/solvers/pysat_rc2.py
```diff
@@ -0,0 +1,172 @@
...
+        assert subsolver in CPM_RC2.solvernames(), f"Wrong solver ({subsolver}) selected from available: ({CPM_RC2.solvernames()})"
+
+        self.pysat_solver = RC2(formula=WCNF(), solver=subsolver)
+        self.pysat_vpool = self.pysat_solver.pool
+
+        SolverInterface.__init__(self, name="rc2", cpm_model=cpm_model)
+
+    def solve(self, assumptions=None):
+        """
+            Call the PySAT solver
```
IgnaceBleukx, line 107: Call the RC2 solver

cpmpy/solvers/pysat_rc2.py
```diff
@@ -0,0 +1,172 @@
...
+            - time_limit:  maximum solve time in seconds (float, optional). Auto-interrups in case the
+                           runtime exceeds given time_limit.
+                           Warning: the time_limit is not very accurate at subsecond level
+            - assumptions: list of CPMpy Boolean variables that are assumed to be true.
+                           For use with s.get_core(): if the model is UNSAT, get_core() returns a small subset of assumption variables that are unsat together.
+                           Note: the PySAT interface is statefull, so you can incrementally call solve() with assumptions and it will reuse learned clauses
+        """
+        ## TODO: adding assumptions to the solving !
+        if assumptions is not None:
+            print("[Warning] Assumption variables interpreted as hard clauses!")
```
IgnaceBleukx, line 119: I would remove this comment. Add it to the documentation. When repeated solving the terminal will be full of warnings...

IgnaceBleukx: Also add an example on how to use the maxsat solver (for example the frietkot example, but add it to the examples list so you can refer to it from the documentation here).

IgnaceBleukx (commented):
Looks good.
I would add an example on how to use a maxsat solver in CPMpy. (I.e., add implication variables to every clause and implement weights on them in the obj function).
If you can find a somewhat larger example, maybe we can compare performance of RC2 vs OR-tools in the example?
Just a thought :)

cpmpy/solvers/pysat_rc2.py
sourdough-bread: Added example

cpmpy/solvers/pysat_rc2.py
```diff
@@ -0,0 +1,178 @@
...
+            Call the MaxSAT RC2 solver
+
+            Arguments:
+            - assumptions: list of CPMpy Boolean variables that are assumed to be true.
+                           MaxSAT solver does not support incremental solving and assumptions.
+                           CPM_rc2 adds the assumptions as hard clauses to the MaxSAT solver.
+        """
+        if assumptions is not None:
+            print("[Warning] Assumption variables interpreted as hard clauses!")
+            pysat_assum_vars = self.solver_vars(assumptions)
```
IgnaceBleukx, line 116: Can you extract the clauses from the pysat model? If so, I would re-initialize the RC2 solver to remove assumptions after solving. This way the semantics are consistent with other solvers. As the constraints do not have to pass by our transformations again, this should be quite efficient

sourdough-bread: Absolutely, I will keep track of the clauses and save them into a WCNF formula that can be reinitialized at every solve call with no assumptions.

tias:
Looks great.

Only the `
    def _post_clauses(self, clauses):
        self.pysat_solver.append_formula(clauses)`
funciton seems overkill, only used in 1 (with inheritance 2) places, for a oneliner....

Did you run it with the full test-suite? If that is relevant/possible, check with Ignace, he documented it better recently.

Should that be fine (or not applicable) then do merge!

IgnaceBleukx (changes requested):
Left some comments in the code. 
Main things are changes in documentation and semantics of solveAll parameters.

cpmpy/solvers/pysat.py
```diff
@@ -222,23 +222,31 @@ def __add__(self, cpm_con):
...
         for con in cpm_cons:
-            self._post_constraint(con)
+            clauses += self._enocde_constraints(con)
+
+        self._post_clauses(clauses)
 
         return self
 
-    def _post_constraint(self, cpm_expr):
+    def _post_clauses(self, clauses):
```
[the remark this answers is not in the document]
sourdough-bread: No reason, I rolled back to __post_constraint

cpmpy/solvers/pysat.py
```diff
@@ -222,23 +222,31 @@ def __add__(self, cpm_con):
...
+        self.pysat_solver.append_formula(clauses)
+
+    def _enocde_constraints(self, cpm_expr):
```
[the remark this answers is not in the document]
sourdough-bread: rolled back to _post_constraint. Transformation should be kept in separate method.

sourdough-bread:
Needs further testing for constraints and also for solution printing.

PR #147, opened 2022-11-05
Example for maximal propagate set of constraints

IgnaceBleukx:
Calculating maximal consequence in CPMpy.


tests/test_examples.py
```diff
@@ -7,7 +7,11 @@
 import pytest
 from cpmpy import *
 
-EXAMPLES = glob(join("..", "examples", "*.py")) + glob(join(".", "examples", "*.py")) + glob(join(".", "examples/advanced", "*.py"))
+EXAMPLES = glob(join("..", "examples", "*.py")) + \
+           glob(join(".", "examples", "*.py")) + \
+           glob(join("..", "examples/advanced", "*.py")) + \
+            glob(join("..", "examples/csplib", "*.py"))
+
```
sourdough-bread, line 14: It would be better to put this into a separate pull request or merge directly with master.

examples/advanced/maximal_propagate.py
```diff
@@ -0,0 +1,39 @@
...
+def maximal_propagate(constraints, solvername="ortools"):
+    # For faster propagation, use incremental solver such as pysat, z3 or gurobi.
+    vars = get_variables(constraints)
+    visisted_domain = {var: set() for var in vars}
+
+    solver = SolverLookup.get(solvername)
+    solver += constraints
+
+    while solver.solve():
+        cons = False
```
sourdough-bread, line 20: Ideally, we should be able to solve with an assumption variable that allows for temporarily turning on/off the added constraints:

`solver.solve(assumptions=[bv])`

examples/advanced/maximal_propagate.py
```diff
@@ -0,0 +1,39 @@
...
+Example showing repeated solving for deriving the maximal consequence of set of constraints.
+
+Iteratively finds a new solutions by forcing at least one variable to have a value not seen in any other solution.
+"""
+
+from cpmpy import *
+from cpmpy.transformations.get_variables import get_variables
+
+
+def maximal_propagate(constraints, solvername="ortools"):
```
sourdough-bread, line 11: Proposed improvements:
- Users can choose to use their own solver or start a new one.
- Users can choose which variables to propagate (i.e. explanations use-case).

```python
def maximal_propagate(constraints, vars=None, solver=None, solvername="ortools"):
```

examples/advanced/maximal_propagate.py
```diff
@@ -0,0 +1,39 @@
...
+    # For faster propagation, use incremental solver such as pysat, z3 or gurobi.
+    vars = get_variables(constraints)
+    visisted_domain = {var: set() for var in vars}
+
+    solver = SolverLookup.get(solvername)
+    solver += constraints
+
```
sourdough-bread, line 18:
```python
    if solver is None:
        solver = SolverLookup.get(solvername)
        solver += constraints

    bv = boolvar()
    solver += any([bv, ~bv])
```

examples/advanced/maximal_propagate.py
```diff
@@ -0,0 +1,39 @@
...
+Iteratively finds a new solutions by forcing at least one variable to have a value not seen in any other solution.
+"""
+
+from cpmpy import *
+from cpmpy.transformations.get_variables import get_variables
+
+
+def maximal_propagate(constraints, solvername="ortools"):
+    # For faster propagation, use incremental solver such as pysat, z3 or gurobi.
+    vars = get_variables(constraints)
```
sourdough-bread, line 13: ## Ensure variables are in `cpm_array` format.
```python
    if vars is None:
        vars = get_variables(constraints)
    vars = cpm_array(vars)
```

examples/advanced/maximal_propagate.py
```diff
@@ -0,0 +1,39 @@
...
+    visisted_domain = {var: set() for var in vars}
+
+    solver = SolverLookup.get(solvername)
+    solver += constraints
+
+    while solver.solve():
+        cons = False
+        for var, values in visisted_domain.items():
+            values.add(var.value())
+            cons |= all(var != val for val in values)
```
sourdough-bread, line 23: Remove from loop (Outdent the `cons |= ....`)
```python
solver += bv.implies(all(vars != vars.value()))
```

examples/advanced/maximal_propagate.py
```diff
@@ -0,0 +1,39 @@
...
+        solver += cons
+    # exhausted all possible domain values
```
sourdough-bread, line 25: Disable assumption var:
```python
solver += ~bv
```

examples/advanced/maximal_propagate.py
```diff
@@ -0,0 +1,39 @@
...
+    return visisted_domain
```
sourdough-bread, line 26: I'm not sure if it is better to return assignments or variables and their corresponding domains.

sourdough-bread (commented):
Some ideas for improvement.

examples/advanced/maximal_propagate.py
IgnaceBleukx: What would be the usecase for providing your own solver? Add some extra hard constraints or?
Providing the vars might be a good idea indeed, this will also limit the number of iterations needed

examples/advanced/maximal_propagate.py
IgnaceBleukx: Well as it is a propagator, you would want the domains allowed by the constraints you provide, right?

IgnaceBleukx:
I think adding assumption variables with an instantiated solver is a bit overkill for this example, I like to keep it quite simple and readable, yet functional

examples/advanced/maximal_propagate.py
JoD: Yeah, the restricted domains seems the most flexible choice (and matches nicely with the algorithm).

JoD:
The algorithm does not give the intersection of all solutions, but instead it gives the union of all possible values for all variables. E.g., for `x \in {0,1,2,3}` and constraint `x>1` the algorithm will reply `x \in {2,3}`, while the intersection of solutions would be `x \not \in {0,1}`. The difference is trivial, but both interpretations lead to a different algorithm. The intersection one would add one disjunction at each iteration, while the current adds a disjunction of conjunctions, so arguably the intersection one is more efficient. But it requires finite domains. We can discuss this later (hard to explain in a comment), the algorithm is fine for now.

examples/advanced/maximal_propagate.py
JoD: Providing your own solver can improve efficiency (in case of incremental solving) but then you also have to have assumption variables to not make the solver contain inconsistent constraints. It's ok to keep things simple and go for the advanced use case when the need arises.

IgnaceBleukx:
> The algorithm does not give the intersection of all solutions, but instead it gives the union of all possible values for all variables. E.g., for `x \in {0,1,2,3}` and constraint `x>1` the algorithm will reply `x \in {2,3}`, while the intersection of solutions would be `x \not \in {0,1}`. The difference is trivial, but both interpretations lead to a different algorithm. The intersection one would add one disjunction at each iteration, while the current adds a disjunction of conjunctions, so arguably the intersection one is more efficient. But it requires finite domains. We can discuss this later (hard to explain in a comment), the algorithm is fine for now.

Ah yes I see, I added the alternative to the file as well. This will indeed be more efficient probably. Should we keep both versions in the file or just pick one?

IgnaceBleukx:
Surprisingly, the disjuctive (propagate_alt) version seems to be slower compared to the original method. I think it has partly to do with the size of the disjunction in early iterations of the algorithm. With large domains it is huge, while in the original version, the disjunction starts out small and gets bigger over time. So I would suggest to just keep the first version if that's ok with you?

JoD:
> So I would suggest to just keep the first version if that's ok with you?

That's ok with me.

tests/test_examples.py
tias: I agree, I'll put it in master already so I can test them for release

tias:
One last thing is the signature, which is currently

```def maximal_propagate(vars=None, constraints=[], solvername="ortools"):```

I propose:

```def maximal_propagate(constraints, vars=None, solver="ortools"):```

Because required arguments should go before optional ones (and not have default values); and because `.solve()` has 'solver=str' as argument. So for consistency I would propose it here as well. I did the same when creating the 'mus' tool, there too it is is 'solver=str' (I did hesitate about it)


**Finally,**

should this be part of our 'tools' too, instead of an advanced example?

We use it in our explanation work, it will be needed for a stepwise-explain advanced example or tool, so maybe just make it a tool straight away?

(and would it then have both implementations? It could...

JoD:
Algorithmically and interface-wise, the PR is in good shape. I cannot see any test of the functionality. Should this still be added or am I missing something?

tias:
Looks great, I guess some advanced examples could now be changed to use this?

Issue #167, opened 2022-11-23
Can't deepcopy Cumulative global constraint


HelgeS:
Version: 0.9.10

The new `Cumulative` global constraint can't be deepcopied.
Instead, an `TypeError` exception is thrown.

Reproduction example:
```
from cpmpy import *
import numpy as np

m = Model()
start = intvar(0, 10, 4, "start")
duration = np.array([1, 2, 2, 1])
end = start + duration
demand = np.array([1, 1, 1, 1])
capacity = 2
m += AllDifferent(start)
m += Cumulative(start, duration, end, demand, capacity)

print(m)

m2 = m.deepcopy()  # should not throw an exception

print(m2)
print(repr(m) == repr(m2))  # should be True
```

Actual output:
```
Constraints:
    alldifferent(start[0],start[1],start[2],start[3])
    cumulative([start[0],start[1],start[2],start[3]],[1,2,2,1],[1 + (start[0]),2 + (start[1]),2 + (start[2]),1 + (start[3])],[1,1,1,1],2)
Objective: None
Traceback (most recent call last):
  File "/home/helge/tmp/repro.py", line 15, in <module>
    m2 = m.deepcopy()  # should not throw an exception
  File "/home/helge/tmp/venv_cpmpy/lib/python3.10/site-packages/cpmpy/model.py", line 240, in deepcopy
    copied_cons = [cpm_cons.deepcopy(memodict) for cpm_cons in self.constraints]
  File "/home/helge/tmp/venv_cpmpy/lib/python3.10/site-packages/cpmpy/model.py", line 240, in <listcomp>
    copied_cons = [cpm_cons.deepcopy(memodict) for cpm_cons in self.constraints]
  File "/home/helge/tmp/venv_cpmpy/lib/python3.10/site-packages/cpmpy/expressions/globalconstraints.py", line 141, in deepcopy
    copied_args = self._deepcopy_args(memodict)
  File "/home/helge/tmp/venv_cpmpy/lib/python3.10/site-packages/cpmpy/expressions/core.py", line 145, in _deepcopy_args
    if arg not in memodict:
TypeError: unhashable type: 'NDVarArray'
```

I debugged it and found the following reasons:
a) Its parameters are a list of lists and this cannot be processed by `Expression._deepcopy_args`, which expects a single flat list of arguments. [Link](https://github.com/CPMpy/cpmpy/blob/master/cpmpy/expressions/core.py#L144)
[...]

PR #169, opened 2022-11-23
Enable deep copy of Cumulative global constraint

HelgeS:
This addresses issue #167 

I added a test to reproduce the error (and that it's fixed).
The changes did not break any other test.

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -151,7 +151,7 @@ class Expression(object):
         copied = []
         for arg in arglist:
             if is_any_list(arg):
-                copied.append(self._deepcopy_arg_list(arg, memodict))
+                copied += [self._deepcopy_arg_list(arg, memodict)]
                 continue
 
             if arg not in memodict:
@@ -161,7 +161,7 @@ class Expression(object):
                     memodict[arg] = arg
                 else:
                     raise ValueError(f"Not a supported argument to copy: {arg}")
-            copied.append(memodict[arg])
+            copied += [memodict[arg]]
         return copied
 
     def deepcopy(self, memodict = {}):
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -140,8 +140,20 @@ class Expression(object):
             Used in copy() methods of expressions to ensure there are no shared variables between the original expression and its copy.
             :param: memodict: dictionary with already copied objects, similar to copy.deepcopy()
         """
+        return self._deepcopy_arg_list(self.args)
+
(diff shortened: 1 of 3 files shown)


tias:
Hi Helge,

Great, thanks a lot for catching and fixing this!

To me the fix looks fine, and idd no errors.

@IgnaceBleukx you wrote the original code (both the deepcopy and the cumulative), can you acknowledge and merge this change?

cpmpy/expressions/globalconstraints.py
```diff
@@ -418,10 +418,10 @@ class Cumulative(GlobalConstraint):
...
     """
     def __init__(self, start, duration, end, demand, capacity):
-        super(Cumulative, self).__init__("cumulative",[start,
-                                                       duration,
-                                                       end,
-                                                       demand,
+        super(Cumulative, self).__init__("cumulative",[flatlist(start),
+                                                       flatlist(duration),
+                                                       flatlist(end),
+                                                       flatlist(demand),
```
IgnaceBleukx, line 424: You should check first if the demand parameter is a list or not. Our implementation of the Cumulative constraint supports using a constant demand, but calling flatlist on an integer (or variable) will result in an error.

cpmpy/expressions/core.py
```diff
@@ -140,8 +140,20 @@ def _deepcopy_args(self, memodict={}):
...
+        """
+            Create and return a deep copy of the arguments in `arglist`.
+            Recursively deep copy nested lists.
+            :param: memodict: dictionary with already copied objects, similar to copy.deepcopy()
+        """
         copied = []
-        for arg in self.args:
+        for arg in arglist:
+            if is_any_list(arg):
+                copied.append(self._deepcopy_arg_list(arg, memodict))
```
IgnaceBleukx, line 154: In most places in the code we use `lst += [new_element]` instead of `lst.append(new_element)`. But this is fine too :)

IgnaceBleukx (commented):
Thank you for adding this, I will make a secion in the docs on how to add a new global constraint and what not to forget when doing so (as I clearly forgot to test for this functionality :) ).
Left some minor comments in the code, fine to merge for me when these are addressed.a

cpmpy/expressions/globalconstraints.py
HelgeS: Well spotted! I added a check before `flatlist` is called and a corresponding test case.

cpmpy/expressions/core.py
HelgeS: Okay, changed it, it's better to be consistent.

PR #171, opened 2022-11-24
Bug170 flatten model

sourdough-bread:
Bug with normalization of bool expressions: `normalized_boolexpr((b) != (k == 1))` was transformed to 
```
((BV0) == (b)) == (BV1)
(k == 1) == (BV0)
```
should be 
```
((BV0) == (~b)) == (BV1)
(k == 1) == (BV0)
```


sourdough-bread:
Bug fix pushed to master by @tias.

PR #173, opened 2022-11-28
166 assertionerror when using a value of 00 with type numpyfloat32 as weight in a weighted sum objective

IgnaceBleukx:
Fixed by testing if `int(arg) == arg` and allowing for empty decimal parts of the float.

### DIFF
--- a/cpmpy/expressions/utils.py
+++ b/cpmpy/expressions/utils.py
@@ -29,7 +29,8 @@ from itertools import chain, combinations
 def is_int(arg):
     """ can it be interpreted as an integer? (incl bool and numpy variants)
     """
-    return isinstance(arg, (bool, np.bool_, int, np.integer))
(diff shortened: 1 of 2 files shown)


tias:
OK, there are multiple things at play here...

- a float is not an int, so I don't really agree with your fix, it can come back to us in different ways (because it does not convert it to an int, nor should it)
- we might want to support transforming int-equivalent floats to int in some expressions, e.g. sums and wsums. If so, we should do that check and conversion there
- this bug is because instead of creating a single wsum, it is split up into two expressions and our later wsum check does not merge it into a wsum either so it treats it as a nested expression...

Our design principle is that expressions stated by the user should be similar to that. We do have some simplications for and/or and sum with neutral elements. For multiplication/wsum, we should probably better not simplify 0*x to 0 at expression creation; leave that to a later clean up stage.

see #182 for follow-up

PR #174, opened 2022-11-28
158 importing a file and adding constrains can turn problems wrongly unsatisfiable

tias:


tests/test_model.py
```diff
@@ -41,6 +41,23 @@ def test_io_nempty(self):
...
+        bv = cp.boolvar()
+        m = cp.Model( iv > 3, ~bv )
+        m += (iv[0] == 5)
+        m.to_file(fname)
+
+        self.assertEqual(_BoolVarImpl.counter, 1)
+        self.assertEqual(_IntVarImpl.counter, 3)
+        _BoolVarImpl.counter = 0  # don't try this at home
+        _IntVarImpl.counter = 0  # don't try this at home
+        loaded = cp.Model.from_file(fname)
```
IgnaceBleukx, line 56: But now the user cannot make any variables before loading the model, we should handle this case as well I believe.

tias: This is the testcase.

The actual code takes the max of the current and loaded counter, for that reason.

rubenkindt:
Would the following situation also be handled? 

user creates a int variable (it gets autonamed to "IV0")
then imports a model already containing that variable name
user appends the first variable to model in some constraint
model.solve()

I think that the counter would be correct but the situation of sending multiple variables with the same name to the solver would still occur.

tias:
I could add a warning if the current bv_counter is non-zero...
  Or we could go all the way in renaming the variables of the imported model above the counter; but I'm not sure we should?

Your statement is also true for named variables, but that is something a user should be careful with I think, I don't think we should rename variables for these edge cases... I mean, worst case we could add an argument to load that says rename=True/False, which would do renaming, but I'm not sure there is a real need for that?

tias:
Added the warning.

A 'rename' argument is still an option too, but as long as nobody needs I'm not implementing it

Issue #153, opened 2022-11-14
Gurobi sometimes does not set amount of solutions and exit status correctly


rubenkindt:
solver: gurobi
CPMpy: version v0.9.9 commit e79b3afedc934a9437c2ddb3a9f54d7e2d7bd3ee

Gurobi does not set the amount of solutions and exit status correctly. It does find at least one solution, but says NOT_RUN and sets its found solutions to strange numbers, (16 for the "send more ...", while OR-Tools finds 1 and 100 for the second problem while OR-Tools finds 9)

[send_most_money16650521758194163.zip](https://github.com/CPMpy/cpmpy/files/10002243/send_most_money16650521758194163.zip)
[secondExample.zip](https://github.com/CPMpy/cpmpy/files/10002251/secondExample.zip)

```python 
from cpmpy import *

file = "send_most_money16650521758194163" #don't forget to unzip
m = Model().from_file(file)
sol = m.solveAll(solver="gurobi", solution_limit=100)
print(sol)
print(m.status())

m2 = Model().from_file(file)
sol = m2.solveAll(solver="ortools", solution_limit=100)
print(sol)
print(m2.status())
```

expected behavior: 
Same number of solution like other solvers ("ortools",...) with it exitstatus not set to NOT_RUN or no solution with exitstatus to NOT_RUN

bug found while working on my master thesis.

IgnaceBleukx:
This issue touches upon several misspecifications actually:
1) OR-tools does not support finding all optimal solutions
2) Gurobi does support finding all optimal solutions but we did not implement it correctly. The current setting will search for `all` solutions, in decreasing quality according to the objective. 
```python
x = intvar(0,2,shape=3, name="x")
m = Model(sum(x) >= 1, minimize=sum(x))
m.solveAll(solver="gurobi",solution_limit=100, display=lambda : print(x.value(), sum(x.value())))
>>[0 0 1] 1
[1 0 0] 1
[0 1 0] 1
[2 0 0] 2
...
[2 2 1] 5
[2 2 2] 6

m.solveAll(solver="ortools",solution_limit=100, display=lambda : print(x.value(), sum(x.value())))
>> [0 0 1] 1
``` 
I'll have to look into the second model more closely as it contains many constraints...

PR #176, opened 2022-11-28
Solve all semantics

IgnaceBleukx:
Revisisted solveAll semantics of all solvers.
Changes:
- Raise exception for OR-tools when optimization problem
- Stop gurobi pool search at sub-optimal solutions
- Changed z3 objective_value_ type to float instead of var
- Generic solver interface: stop search at sub-optimal solutions

Still open is the following issue:
When searching for all solutions with a solver using the generic `solveAll` implementation, the solver will be in an invalid state at the end (as all solutions are exhausted) and will therefore always answer `False` to subsequent `solve` calls.

Ways to fix this are:
1) raise a warning to the user this is happening, but this shouldn't happen when solveAll is called from the`Model` class (as the solver will be reinitialized for subsequent calls anyway
2) make a copy of the solver interface (my least favorite method)
3) add an assumption variable to every blocking constraint and disable assumption variables afterwards (but then the solver has to support assumption variables...)


cpmpy/solvers/ortools.py
```diff
@@ -207,7 +207,9 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, **kwargs)
 
             Returns: number of solutions found
         """
-        # XXX: check that no objective function??
+        if self.ort_model.HasObjective():
+            raise Exception("OR-tools does not support finding all optimal solutions.")
```
tias, line 211: I think these two lines can go into the template... by default solvers will not support finding all optimal solutions... also the generic one does not (and should not)

I am not convinced of your solver_interface generic implementation either, you should add a constraint that the subsequetn solutions ahve to be better than the optimal one after finding it. But I don't think we should go there... we can just stick to 'solveAll is for satisfaction problems'... and let the user model it manually otherwise
(though I am fine with leaving it in for gurobi now that you alrady implemetned it, we just raise for all other solvers)

IgnaceBleukx: Why should we add a constraint forcing the objective to be better? We just walk over the different optimal solutions by adding no-goods as usual?

IgnaceBleukx:
I added the exception for minizinc as well as the python interface does not support finding all optimal solutions.

cpmpy/solvers/ortools.py
tias: That is what you do now, and then you break when a sub-optimal solution is found.

But in this case we know that the solution needs at least/most a certain value, so it does not need to search for that lower/upper bound each time again...

I don't have use cases, but I would assume that it would be faster to give the solver more information.
At the very least, in the very last call (that would be suboptimal), it can stop as soon as its lb/ub is below the optimal value, without having to prove that whatever value it then has is the now optimal one...

cpmpy/solvers/solver_interface.py
```diff
@@ -199,9 +199,17 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, **kwargs)
 
             Returns: number of solutions found
         """
-        # XXX: check that no objective function??
+        # TODO raise warning solver obj will be invalid? Or add assumption variables?
```
tias, line 202: Yes, Wout introduced a warnings.warn recently, we can do that here too, that would be good.

cpmpy/solvers/ortools.py
IgnaceBleukx: Yes, but if we implement it as we did in `solver_interface` the solver will here also become invalid... But as OR-tools is not incremental i guess that doesn't really matter here. I'll implement it and raise a warning as in `solver_interface`

IgnaceBleukx: Aha, but then we need to store the objective function in the solver object as well. I don't see any harm in doing so, but seems silly to keep it in there just for this (obscure) feature?

sourdough-bread:
> Still open is the following issue:
> 
> When searching for all solutions with a solver using the generic solveAll implementation, the solver will be in an invalid state at the end (as all solutions are exhausted) and will therefore always answer False to subsequent solve calls.
> 
> Ways to fix this are:
> 
> - raise a warning to the user this is happening, but this shouldn't happen when solveAll is called from the `Model` class (as the solver will be reinitialized for subsequent calls anyway

Add a `from_model` flag when `solveAll` is called from the `Model` class ? If 
```Python
if from_model:
     warnings.Warn(....)
```
> - make a copy of the solver interface (my least favorite method)

Same for me, this is also my least favorite option. 
If this is the only way, then we should keep track of the transformed models for the solvers. 
For instance, we hash the model and keep the transformed/reified/... model in the cache and if it's not we reload the transformed model into a new solver without the transformations.
It would still need to map the CPMpy variables to the solver variables.

> - add an assumption variable to every blocking constraint and disable assumption variables afterward (but then the solver has to support assumption variables...)

We can at least implement this use-case for the solvers that we know. And for those we don't, we could Try adding a reified `cpm_expr` of this style  `bv.implies(cpm_expr)` and see if it raises an error. 
If it doesn't, we call a subroutine like `_solveAll_assum`.

cpmpy/solvers/TEMPLATE.py
```diff
@@ -257,3 +257,51 @@ def _post_constraint(self, cpm_con):
...
+        if self.TEMPLATE_solver.HasObjective():
+            raise Exception("TEMPLATE does not support finding all optimal solutions")
+
+        # A. Example code if solver supports callbacks
+        if is_any_list(display):
+            callback = lambda : print([var.value() for var in display])
+        else:
+            callback = display
+
+        self.solve(time_limit, callback=callback, enumerate_all_solutions=True, **kwargs)
```
sourdough-bread, line 287: If this is the template implementation for `solveAll`, I don't see an advantage to calling `solve` instead of manually blocking the found solutions. See [comment](https://github.com/CPMpy/cpmpy/pull/176#issuecomment-1368747103) for ideas on `solveAll`.

cpmpy/solvers/gurobi.py
```diff
@@ -398,21 +399,32 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, **kwargs)
...
             for cpm_var in self.user_vars:
                 solver_val = self.solver_var(cpm_var).Xn
                 if cpm_var.is_bool():
                     cpm_var._value = solver_val >= 0.5
                 else:
                     cpm_var._value = int(solver_val)
+
             # Translate objective
             if self.grb_model.getObjective().size() != 0:                # TODO: check if better way to do this...
-                self.objective_value_ = self.grb_model.getObjective().getValue()
```
sourdough-bread, line 415: Nothing to see in the docs either ...

cpmpy/solvers/TEMPLATE.py
sourdough-bread:
```Python
        if time_limit is not None:
            raise NotImplementedError("TEMPLATE: TODO, implement time_limit")
```

I would use a warning for this and have a default implementation in `solverInterface` with `time_limit`.

sourdough-bread: `216:        raise NotImplementedError("TEMPLATE: Not a know supported numexpr {}".format(cpm_expr))`

KNOW -> KNOWN

sourdough-bread: **Line 234:**
I would at least write the transformations that can be done in the order that they should be applied to something similar to `ortools.py`:
```

cpm_cons = only_bv_implies(cpm_cons) # everything that can create
```

cpmpy/solvers/solver_interface.py
```diff
@@ -199,9 +199,17 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, **kwargs)
 
             Returns: number of solutions found
         """
```
sourdough-bread, line 201: `hasObjective(..)` could be implemented here as default ?

cpmpy/solvers/z3.py
```diff
@@ -164,7 +164,7 @@ def solve(self, time_limit=None, assumptions=[], **kwargs):
             if isinstance(self.z3_solver, z3.Optimize) and \
```
sourdough-bread, line 164: See comment in SolverInterface to move objective check to solver interface ?

sourdough-bread (changes requested):
Looks good, added some ideas for improvement.
@IgnaceBleukx tag me for proofreading before merging.

cpmpy/solvers/TEMPLATE.py
IgnaceBleukx: We should do the changes you proposed in TEMPLATE.py directly in master.
Regarding calling `solve` in `solveAll`, this is actually the way both OR-tools and Gurobi handle the solve all call

cpmpy/solvers/ortools.py
IgnaceBleukx: Ok will leave it for now and raise error when calling solveAll with objective.

cpmpy/solvers/solver_interface.py
IgnaceBleukx: Done :)

IgnaceBleukx:
Implemented requested changes. @sourdough-bread or @tias can you do a final review so this can be merged? Will be able to close issue #153

tias:
Reviewed, good to merge for me

tias:
@sourdough-bread ready for you to proofread and merge as you requested.

sourdough-bread (approved):
Looks good, nothing to report on my side.

PR #178, opened 2022-11-28
177 cumulative can not assume args are numpy array

IgnaceBleukx:
Fixed wrong assumption arguments are numpy arrays

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -102,6 +102,8 @@
 
 """
 import warnings # for deprecation warning
+import numpy as np
+
 from .core import Expression, Operator, Comparison
 from .variables import boolvar, intvar, cpm_array
 from .utils import flatlist, all_pairs, argval, is_num, eval_comparison, is_any_list
@@ -455,8 +457,6 @@ class Cumulative(GlobalConstraint):
         return cons
 
     def value(self):
-        import numpy as np
-
         argvals = [np.array([argval(a) for a in arg]) if is_any_list(arg)
                    else argval(arg) for arg in self.args]
 
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -455,7 +455,15 @@ class Cumulative(GlobalConstraint):
         return cons
 
     def value(self):
-        start, dur, end, demand, cap = [argval(a) for a in self.args]
+        import numpy as np
+
+        argvals = [np.array([argval(a) for a in arg]) if is_any_list(arg)
+                   else argval(arg) for arg in self.args]
+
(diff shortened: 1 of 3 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -455,7 +455,15 @@ def decompose(self):
         return cons
 
     def value(self):
-        start, dur, end, demand, cap = [argval(a) for a in self.args]
+        import numpy as np
```
tias, line 458: why here and not at top level?

cpmpy/expressions/globalconstraints.py
```diff
@@ -455,7 +455,15 @@ def decompose(self):
...
+
+        argvals = [np.array([argval(a) for a in arg]) if is_any_list(arg)
+                   else argval(arg) for arg in self.args]
```
tias, line 461: this still does not guarantee that it is a np.array (the latter case), which you need for the .all() below to work?

if it does, please add a comment that we assume the args are a numpy array here...

cpmpy/expressions/globalconstraints.py
```diff
@@ -455,7 +455,15 @@ def decompose(self):
...
+
+        if any(a is None for a in argvals):
+            return None
+
+        start, dur, end, demand, cap = argvals
         # start and end seperated by duration
         if not (start + dur == end).all():
```
tias, line 468: an alternative fix could be not to assume that it is an nparray...

e.g. `if all(s + d == e for s,d,e in zip(start, dur, end))` or similar?

tests/test_constraints.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
 
-SOLVERNAME = None
+SOLVERNAME = "ortools"
```
tias, line 6: I've done that in a separate PR with additional comments, so need not be here

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Can be on top level, will fix

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: start, end and duration will always be lists (as fixed by Helge's pull request), so non-numpied args here are only the capacity and potentially the demand if fixed for all tasks.

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Also possible, but non-numpiefied decomposition of the capacity limit is much harder imho

IgnaceBleukx:
Handled the initialization with some more care

IgnaceBleukx:
Added a testcase. Oke to merge @tias ?

Issue #172, opened 2022-11-25
Memory crash when one side of mod is a cpm_array


Dimosts:
This issue is related to issue #96. 

The bug appears when trying to flatten the mod expression, when either the left or the right side (or both of course) includes a reference to a cpm_array, with a variable being the index. This can be demonstrated in the following simple case.

```python
    m = Model()

    F = intvar(1, 10, shape=10, name="vars")

    array = np.array( [ 5, 6, 7, 8, 9, 10, 12, 14, 16, 18, 29 ])
    array = cpm_array(array)

    m += array[F[2]] % 5 > 1

    m.solve()
```

When a cpm_array is used with a variable as the index, the bounds are assumed to be [-2147483648, 2147483647] which are the bounds for an integer variable. Then, in order to find the bounds for the mod expression, it tries to check all possibilities, and then find the minimum and maximum resulting value. This leads to memory crashes, as it requires (a minimum of) 32GB to do it.

The problem is in the function "get_or_make_var(expr)", in the following part:
```python
            elif flatexpr.name == 'mod': # binary
                l = np.arange(lbs[0], ubs[0]+1)
                r = np.arange(lbs[1], ubs[1]+1)
                # check all possibilities
                remainders = np.mod(l[:,None],r)
                lb, ub = np.min(remainders), np.max(remainders)
                ivar = _IntVarImpl(lb,ub)
```

Such a case can be resolved in two ways. Either by being able to calculate the bounds of the cpm_array (which shouldn't be difficult), or by just continuing with  [-2147483648, 2147483647] being the bounds of the mod expression, without any further calculation.

IgnaceBleukx:
Yes, this has the same root cause as issues #164 and #96. We should handle this issue soon!

Dimosts:
It has the same root but can be solved independently for now, just by avoiding to really calculate the bounds of mod when we do not have real bounds in the left or right side of the mod. In that case, we could just return the intvar bounds
`ivar = _IntVarImpl(-2147483648, 2147483647)`

PR #179, opened 2022-11-28
issue #172 fix - memory crash on mod - update flatten_model.py

Dimosts:
This concerns issue #172.

The bug appeared when trying to flatten the mod expression, when either the left or the right side (or both of course) includes a reference to a cpm_array, with a variable being the index. When a cpm_array is used with a variable as the index, the bounds are assumed to be [-2147483648, 2147483647] , as actual bounds are not calculated (so, it is a more generic bug that could happen in other cases too).

The memory crash was the result of the following:

```
 elif flatexpr.name == 'mod': # binary
                l = np.arange(lbs[0], ubs[0]+1)
                r = np.arange(lbs[1], ubs[1]+1)
                # check all possibilities
                remainders = np.mod(l[:,None],r)
                lb, ub = np.min(remainders), np.max(remainders)
                ivar = _IntVarImpl(lb,ub)
```

When trying to check all possibilities for the results of the mod, it was needed to create arrays that were too large (best case in the above example: 32GB).

This pull request just avoids to really calculate the bounds of 'mod' when we do not have real bounds in the left or right side of the mod. In that case, we can just return the intvar bounds:

``ivar = _IntVarImpl(-2147483648, 2147483647) ``

So, the bounds calculation for the 'mod' expression is just changed to the following:

```
elif flatexpr.name == 'mod': # binary
                if (ubs[0]+1) - lbs[0] > 1000000 or (ubs[1]+1) - lbs[1] > 1000000:
                    ivar = _IntVarImpl(-2147483648, 2147483647)
                else:
                    l = np.arange(lbs[0], ubs[0]+1)
                    r = np.arange(lbs[1], ubs[1]+1)
                    # check all possibilities
                    remainders = np.mod(l[:,None],r)
                    lb, ub = np.min(remainders), np.max(remainders)
                    ivar = _IntVarImpl(lb,ub)
```

### DIFF
--- a/cpmpy/transformations/flatten_model.py
+++ b/cpmpy/transformations/flatten_model.py
@@ -319,6 +319,7 @@ def get_or_make_var(expr):
             elif flatexpr.name == 'mod': # binary
 
                 if (ubs[0]+1) - lbs[0] > 1000000 or (ubs[1]+1) - lbs[1] > 1000000:
+                    # special check: if the bounds are too loose we can not check all possibilities below
                     ivar = _IntVarImpl(-2147483648, 2147483647)
                 else:
                     l = np.arange(lbs[0], ubs[0]+1)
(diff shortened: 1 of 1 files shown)


tias:
OK. The real issue is that the bounds of an array are not computed... TODO #96

Issue #157, opened 2022-11-14
Creating a circuit of 1 element crashes "ortools" and "gurobi"


rubenkindt:
solver: "ortools" and "gurobi"
CPMpy: version v0.9.9 commit e79b3afedc934a9437c2ddb3a9f54d7e2d7bd3ee
minizinc-python release version 0.7.0 commit {https://github.com/MiniZinc/minizinc-python/commit/a195cf63fcfbc98665d70ab64efb5424db25bd7e)

Creating a circuit of 1 element crashes both "ortools" and "gurobi".

```python 
from cpmpy import *
x = intvar(0, 1, name="x")
m=Model()
m += Circuit([x])
m.solve(solver="ortools") # same with "gurobi"
```

expected behavior: 
No crash like other solvers ("minizinc:ortools" ...).

bug found while working on my master thesis.
edit: wrong solvers in line 1

Dimosts:
The bug appears when the "decompose" method of the expression is called, in _post_constraint(). The decompose method is called only in ortools and gurobi and that is why this happens there.

In the decompose method of Circuit, it is assumed that it contains more than one variables.
So, "order" is assumed to be an array of variables
`        order = intvar(0,n-1, shape=n) ` 
and then an index is used to define that the last one has no succesor
`order[n-1] == 0`

However, when only 1 variable is given, then n = 1, resulting in "order" not being subscriptable.

This seems an easy fix, just returning nothing when it contains only 1 (or even 0?) variable, skipping the decomposition.

Dimosts:
Has been solved.

PR #181, opened 2022-11-28
Circuit with 1 variable fix - issue #157 fix

Dimosts:
This pull request is fixing issue #157 

This issue appears when a circuit of 1 element is created and it crashes both "ortools" and "gurobi".

The bug appears when the "decompose" method of the expression is called, in _post_constraint(). The decompose method is called only in ortools and gurobi and that is why this happens there.

In the decompose method of Circuit, it is assumed that it contains more than one variables.
So, "order" is assumed to be an array of variables
order = intvar(0,n-1, shape=n)
and then an index is used to define that the last one has no succesor
order[n-1] == 0

However, when only 1 variable is given, then n = 1, resulting in "order" not being subscriptable.

It was an easy fix, just returning nothing when it Circuit constraint contains less than 2 variables, skipping the decomposition.

It also needed a check in ortools.py, to check if None is returned before adding the decomposed expression to the model of the solver.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -215,6 +215,10 @@ class Circuit(GlobalConstraint):
         """
         succ = cpm_array(self.args)
         n = len(succ)
+
+        if n <= 1:
+            return None
+
         order = intvar(0,n-1, shape=n)
         return [
             # different successors
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -433,7 +433,9 @@ class CPM_ortools(SolverInterface):
             #    ReservoirConstraint, ReservoirConstraintWithActive
             
             # global constraint not known, try posting generic decomposition
-            self += cpm_expr.decompose()  # assumes a decomposition exists...
+            c = cpm_expr.decompose()
+            if c is not None:
+                self += c  # if a decomposition exists...
             # TODO: DirectConstraint/NativeConstraint from cpm_expr.name to API call? see #74
             return None # will throw error if used in reification
         


tias:
I think an empty list [] is a valid 'set' of constraints, at least all methods should be robust to handling an empty list.

So if you return that no special checks are needed (otherwise we have to carefully check all places where decompose is called and verify in all those places that it does not return None)

Dimosts:
Returning just an empy list [] throws a "IndexError: list index out of range" at other parts of the code.

IgnaceBleukx:
Why not just return `True` ? This is a valid translation of the empty constraint I think.

Dimosts:
Yes returning [True] is the best solution, will change it to that.

Another issue arising though is the following:

 If Circuit([x]) is the only constraint involving the variable x, then the when the cpmpy model tries to get the value of the variable, it crashes, as the solver doesn't have the variable (no constraint involving x as Circuit([x]) is replaced with True).

This happens in both ortools and gurobi. E.g. in ortools.py it starts at lines 186-187:

```
                cpm_var._value = self.ort_solver.Value(self.solver_var(cpm_var))
```

as we have stashed the variable in solver vars (i.e. self.user_vars), but then we never added a constraint involving this var in the solver.

```
        self.user_vars.update(get_variables(cpm_con))   <- here we add the variable

        cpm_cons = only_bv_implies(cpm_cons) 

        for con in cpm_cons:
            self._post_constraint(con)   <- here we do not post the constraint
```

Thus, the command `self.ort_solver.Value(self.solver_var(cpm_var))` crashes (a few function calls later). 

I think this should be solved in this issue too, as I think we do not have (yet) any other case that this may happen, so no need to open as a seperate issue. 

I am closing the pull request, and will check on how to solve this.

PR #182, opened 2022-11-29
Wsum care

tias:


cpmpy/expressions/core.py
```diff
@@ -326,6 +328,11 @@ def __rpow__(self, other, modulo=None):
 
     # unary mathematical operators
     def __neg__(self):
+        # special case, -(w*x) -> -w*x
+        if self.name == 'mul' and is_num(self.args[0]):
+            return Operator(self.name, [-self.args[0], self.args[1]])
+        elif self.name == 'wsum':
+            return Operator(self.name, [-self.args[0], self.args[1]])
```
IgnaceBleukx, line 335: Negating  args[0] will fail as this is a list. Added a test for this as well which fails atm.
should be `[-w for w in self.args[0]]`

IgnaceBleukx (commented):
Added one comment as negation of wsum is not implemented correctly.

On another note: should we add a new transformation "simplify" which removes the 1's and 0's from multiplications, sums and wsums?

sourdough-bread:
All tests passing, ready to be merged. @Wout4 @tias

PR #186, opened 2022-12-02
151 prob001 car sequencepy sliding window view only numpy 1200+

IgnaceBleukx:


examples/csplib/prob001_car_sequence.py
```diff
@@ -18,11 +18,12 @@
...
 from numpy.lib.stride_tricks import sliding_window_view
 
+import cpmpy
 from cpmpy import *
 
-def car_sequence(n_cars, n_options, n_classes, n_cars_p_class, options, capacity=None, block_size=None, **kwargs):
+def car_sequence(n_cars, n_options, n_classes, n_cars_p_class, options, capacity=None, blocks=None, **kwargs):
     # build model
-    model = Model()
+    model = cpmpy.SolverLookup.get()
```
tias, line 26: if you don't care what solver it is, just use Model(), only use SolverLookup for a specific solver (will change back in additional commit)

PR #187, opened 2022-12-05
Added new file for custom cpmpy exceptions

Wout4:
Added the cpmpy_expressions.py file where we can add our own exception classes.
They would all have to inherit from the CPMpyException
Added one exception class already as an example.

Does it make sense to have this file in the expressions folder? Or is it better to have a new folder

I will be using cpmpy exceptions in other bug fixes, that's why I am posting this pull request first.

### DIFF
--- a/cpmpy/exceptions.py
+++ b/cpmpy/exceptions.py
@@ -10,3 +10,4 @@ class CPMpyException(Exception):
 
 class MinizincPathException(CPMpyException):
     pass
+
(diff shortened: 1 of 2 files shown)


tias:
I think it should be in the root folder, e.g. where model.py is, to have an exceptions.py; it is not an expression, nor a transformation, nor a solver, nor an add-on tool; instead it is shared by all.

Wout4:
I realise i called the branch cpmpy_expressions in stead of exceptions that is a typo..

IgnaceBleukx (approved):

Issue #146, opened 2022-11-03
Boolean comparison in z3


IgnaceBleukx:
In CPMpy, a user can write something like this:
```python
a,b = boolvar(shape=2)
(a & b) <= 1 
```
But this is not supported by the z3 API. Currently, the interface raises the following error: `TypeError: '<=' not supported between instances of 'BoolRef' and 'int'`. We should catch these cases and translate to the proper boolean expression.

tias:
It goes a bit further even, including (a & b)  <= (a | b) as well as (a & 1) or (a & 0) etc, it is simply much stricter on the Booleanness...

So, to fix in next release... (it can be avoided by writing the constraints more carefully as a user, for now)

Wout4:
We have a fix for this in pull request #189

Wout4:
(a & 1), a | 0, etc will also crash linearize.py when using other solvers as an integer is not an expression. Will open a separate issue for that

Dimosts:
I tried to replicate this issue, and I saw that this is resolved, as pull request #189 was merged. I am closing this.

PR #189, opened 2022-12-07
185 z3 translation bug

Wout4:
fix for bug 185
Added test cases

### DIFF
--- a/cpmpy/solvers/z3.py
+++ b/cpmpy/solvers/z3.py
@@ -368,7 +368,11 @@ class CPM_z3(SolverInterface):
                     return z3.And(self._z3_expr(any(lhs == a for a in rhs.args)),
                                   self._z3_expr(all([lhs <= a for a in rhs.args])))
 
-                lhs, rhs = self._z3_expr(cpm_con.args)
+                st = str(rhs)
+                if str(rhs) == '0' and self.lhs_evaluates_to_bool(cpm_con):
+                    return z3.Not(self._z3_expr(lhs))
+                else:
+                    lhs, rhs = self._z3_expr(cpm_con.args)
                 return (lhs == rhs)
 
 
@@ -427,3 +431,21 @@ class CPM_z3(SolverInterface):
         assert (len(self.assumption_dict) > 0), "Assumptions must be set using s.solve(assumptions=[...])"
 
         return [self.assumption_dict[z3_var] for z3_var in self.z3_solver.unsat_core()]
+
+
+    '''
+    Checks if lhs of a comparison evaluates to a boolean expression or a numerical one
+    '''
+    def lhs_evaluates_to_bool(self, cpm_con):
+        assert isinstance(cpm_con, Comparison) and cpm_con.name == "==", "only use this function for == comparisons"
+        lhs, rhs = cpm_con.args
+        # base cases
+        if isinstance(lhs, _BoolVarImpl):
+            return True
+        elif isinstance(lhs, _NumVarImpl):
+            return False
+        elif isinstance(lhs, Operator):
+            return lhs.is_bool()
+        elif isinstance(lhs, Comparison):
+            return True
+
--- a/tests/test_solvers.py
+++ b/tests/test_solvers.py
@@ -329,6 +329,17 @@ class TestSolvers(unittest.TestCase):
         self.assertFalse(s.solve(assumptions=bv))
         self.assertTrue(len(s.get_core()) > 0)
 
+        m = cp.Model(~(iv[0] != iv[1]))
+        s = cp.SolverLookup.get("z3", m)
+        self.assertTrue(s.solve())
+
+        m = cp.Model((iv[0] == 0) & ((iv[0] != iv[1]) == 0))
+        s = cp.SolverLookup.get("z3", m)
+        self.assertTrue(s.solve())
+
+        m = cp.Model([~bv, ~((iv[0] + abs(iv[1])) == sum(iv))])
+        s = cp.SolverLookup.get("z3", m)
+        self.assertTrue(s.solve())
 
     def test_pow(self):
         iv1 = cp.intvar(2,9)


cpmpy/solvers/z3.py
```diff
@@ -368,7 +368,11 @@ def _z3_expr(self, cpm_con, reify=False):
                     return z3.And(self._z3_expr(any(lhs == a for a in rhs.args)),
                                   self._z3_expr(all([lhs <= a for a in rhs.args])))
 
-                lhs, rhs = self._z3_expr(cpm_con.args)
+                st = str(rhs)
+                if str(rhs) == '0' and self.lhs_evaluates_to_bool(cpm_con):
```
IgnaceBleukx, line 372: Also check for other nums instead of only '0' (See main comment in review)
Why not just use `lhs.is_bool()`?

tests/test_solvers.py
```diff
@@ -329,6 +329,17 @@ def test_z3(self):
...
 
+        m = cp.Model(~(iv[0] != iv[1]))
+        s = cp.SolverLookup.get("z3", m)
+        self.assertTrue(s.solve())
+
+        m = cp.Model((iv[0] == 0) & ((iv[0] != iv[1]) == 0))
+        s = cp.SolverLookup.get("z3", m)
+        self.assertTrue(s.solve())
+
+        m = cp.Model([~bv, ~((iv[0] + abs(iv[1])) == sum(iv))])
```
IgnaceBleukx, line 340: Ideally, add these cases to the `test_constains.py` suite

IgnaceBleukx (changes requested):
Some more cases should be handled.
I..e, `a | b >= 1`, `a > b < 10` etc...
So basically any `Comparion(BoolExpr, num)`

You can then translate the lhs to `~lhs` if the comparison on the rhs implies lhs == 0

cpmpy/solvers/z3.py
Wout4: lhs.is_bool() is not recursive so it would fail on for example
~(iv[0] + iv[1] == 0)
because is.bool(iv[0] + iv[1] == 0) returns false

Wout4: Not sure what the expected behavior is for Comparion(BoolExpr, num).
for example: Model((iv[1] + iv[0] == 0) == 0) or Model((iv[1] + iv[0] == 0) == 1) work fine
but Model((iv[1] + iv[0] == 0) == 2) also crashes other solvers
should any nonzero integer be considered true here? because then this is a bug when flattening

Wout4: Similarly: Model([(iv[1] + iv[0] == 0) == bv[0], bv[0] == 1]) works if bv is a boolvar but crashes if it is an intvar

IgnaceBleukx: Hmm, (iv[0] + iv[1]) ==. 0 should definitely return`True` on `is_bool()`
In the Z3 interface we do not apply any flattening to our expressions as their API supports nested expressions by default so we just recursively translate the CPMpy expressions to Z3 expressions.

For the other cases, it might be a bit more tricky indeed. Maybe for now only consider 0 and 1 and raise an error for other values?
I think your last case is related to issue #146, should we fix that bug in this pull request too? Would be most logical to do so I think.

Wout4:
Added boolean comparisons to test_constraints. 
They do fail when testing boolexpr > 1 because this is unsat
Also failing on cumulative() but this is solver independent I think
Otherwise all cases seem to be working now

IgnaceBleukx:
> They do fail when testing boolexpr > 1 because this is unsat 
Yes, maybe add a special case for this... and set the rhs to > 0 or smt.
A little bit ugly but definately usefull to keep them in the testsuite.

cpmpy/solvers/z3.py
```diff
@@ -368,7 +392,23 @@ def _z3_expr(self, cpm_con, reify=False):
...
+                #comparing a boolexpr with an integer or float:
+                if is_num(rhs) and not is_bool(rhs) and lhs.is_bool():
+                    iv = _IntVarImpl(0, 1)  # 0,1 is the valid domain, as this represents true or false
+                    mhs = self._z3_expr(iv)  # turn into z3 expression
+                    lhs, rhs = self._z3_expr(cpm_con.args)
+                    if is_any_list(lhs):
+                        olhs = lhs[0]
+                        orhs = lhs[1]
+                        #keep auxiliary assignments in the first element of the list
+                        return [z3.And(olhs, orhs == mhs), mhs == rhs]
```
tias, line 405: nonono, you can not sometimes return a list (tuple where first element is special wrt second/others) and sometimes a single element.
- python allows it but you should avoid doing it
- especially on a function that is called in so many places!
- if you must, you should always return a tuple with a first special element and then second element a potentially empty list (like in flatten)

but here you really shouldn't...

A design principle of CPMpy is that the user can create any type of expression, and a solver should only accept the expressions that it can straightforwardly map to its internal API.

The heavy lifting of going from any expression to those supported is put into transformations, such that other (future) solvers with similar needs can reuse the same transformations...

So here, you should create a new transformation file 'strict_bool' or whathever is more appropriate, and create these auxiliary channeling variables in there, at CPMpy expression level.

Then the z3 interface can just post the cpmpy expressions as before (with previously unsupported cases now transformed to expressions (with new vars) that it does support)

Wout4: I added it as a transformation in comparison.py
But actually @JoD thought it might not be needed after all, and he was correct.
Where needed boolrefs are now cast to int as follows: if boolref then 1 else 0

Don't know if the transformation will be useful in the future so maybe we should not keep it?

In some cases (comparing 2 boolrefs with >,<,<= or >=) it might be more efficient to use implies, can still implement this if it seems better to you. ( b >= c is equivalent to c.implies(b) ) 

I also added a function is_boolref to utils, to avoid having to check for the 'is_bool' attribute every time.
This won't be needed if we use fishhook in the future but I think it's useful for now.

Issue #162, opened 2022-11-18
naming a variable starting with a '+' crashes all MiniZinc's subsolvers


rubenkindt:
Solver: all Minizinc's subsolvers
CPMpy: version v0.9.9 commit e79b3afedc934a9437c2ddb3a9f54d7e2d7bd3ee
minizinc-python release version 0.7.0 commit {https://github.com/MiniZinc/minizinc-python/commit/a195cf63fcfbc98665d70ab64efb5424db25bd7e)

This one was a fun one to find, I made a typo when naming a variable and the MiniZinc subsolvers crashed. It works with other symbols as well, I tried -, #, %, * and more. It does not occur with 'ortools' or 'gurobi'.

```python 
from cpmpy import *

b = boolvar(name="+")
m = Model()
m += b==0
m.solve(solver="minizinc:ortools")
```

expected behavior: 
Treat the variable name as a name instead of code.

Bug found while working on my master thesis.

Wout4:
MiniZinc does not allow special characters in identifiers. they must be "sequences of lower and uppercase alphabetic characters, digits and the underscore _ character. They must start with an alphabetic character" 
(https://www.minizinc.org/doc-2.5.3/en/modelling.html)

Also a list of keywords is excluded to be used as identifiers: https://www.minizinc.org/doc-2.5.3/en/spec.html#spec-identifiers
Empty string also gives an error.

@IgnaceBleukx 
What would be the best place to add checks for this, when using any of the minizinc subsolvers?
Or do we want to enforce these rules for all solvers, for consistency reasons?

Giving invalid names in minizinc itself also throws an error, so i don't think they have a check we can barrow

IgnaceBleukx:
We should check for this in the `minizinc.py` solver interface in the `solver_var` function. Here all variables will pass trough before being posted to the minizinc solver.
Maybe best thing to do is to manually create a set of non-allowed variable names using the minizinc specification. Quite cumbersome but might be the only proper way of doing it I'm afraid...

Dimosts:
If the user gives an "invalid name", should we just return an "invalid variable name" error, or maybe "correct" the name of the variable so the solving process continues normally?

Wout4:
@tias Already had a bit of code in there that cleaned the variable name from some illegal characters (but only a handful of them).

Do you think it's best to clean all possible variable names, or to throw an exception saying which names are allowed?

Wout4:
[...]

PR #193, opened 2022-12-09
162 naming a variable starting with a + crashes all minizincs subsolvers

Wout4:
fix for issue #162


cpmpy/solvers/minizinc.py
```diff
@@ -284,6 +287,22 @@ def solver_var(self, cpm_var) -> str:
             varname = cpm_var.name
             mzn_var = varname.replace(',', '_').replace('.', '_').replace(' ', '_').replace('[', '_').replace(']', '')
 
+            #test if the name is a valid minizinc identifier
+            #name must be of this form
+            regex = '^[A-Za-z][A-Za-z0-9_]*$'
```
tias, line 292: `solver_var()` is a hot path, I mean that it will be called a lot.

So, it would be better to compile the regex in the constructor and then use that here everytime instead.
https://docs.python.org/3/howto/regex.html#compiling-regular-expressions

Wout4: I added the compiled regex in the constructor
Keeping the assignment of keywords in solver_var() is ok?

tias:
make keywords a frozendict class variable, ideally

Issue #188, opened 2022-12-07
z3 type casting float to int


Wout4:
When solving with z3 variables of type float are converted to integers when posting the constraints. This leads to incorrect solutions.

Example:
x = intvar(0,1)
m1 = Model((x >= 0.1) & (x != 1))
print(m1.solve("z3")) #should be false
print(x.value()) #z3 accepts x = 0 as a solution because 0.1 gets type casted to 0

The bug is located in solvers/z3.py at line 278
Can probably just add a case for floats, unless z3 does not support that?

PR #194, opened 2022-12-09
188 z3 type casting float to int

Wout4:
Funnily enough z3 had a similar bug, they have fixed it and when running with their updated solver I can confirm that the fix in this pull request is also needed to resolve issue #188

However the z3 fix is not yet in their current pip release so maybe we should wait on this merge until that is the case.
(otherwise the test that I added will fail)


tias:
For completeness, if the test suite gives you

```
        x = cp.intvar(0, 1)
        m = cp.Model((x >= 0.1) & (x != 1))
        s = cp.SolverLookup.get("z3", m)
>       self.assertFalse(s.solve())
E       AssertionError: True is not false
```
then you should
```
pip install --upgrade z3-solver
```

PR #196, opened 2022-12-20
191 mus tool crashes with only 1 soft constraint

sourdough-bread:
```
from cpmpy import *
from cpmpy.tools.mus import mus


bv = boolvar()
mus(soft=[bv], hard=[~bv]) # crashes
```

> Source of the error is assum.implies(candidates). When assum is a single boolvar and candidates is a list (of length 1), it fails.
> 
> We should check when invoking bv.implies(rhs) rhs is either a single expression or a list of length 1, and extract this value I think. Or change to all(candidates) in the mus tool, but this might have some side effect by itself... to check.

When `assum` is a single boolvar, it now converts the assumption variable to a `NDVarArray` of length 1.

### DIFF
--- a/cpmpy/tools/mus.py
+++ b/cpmpy/tools/mus.py
@@ -5,6 +5,7 @@ Loosely based on PySat's MUSX:
 https://github.com/pysathq/pysat/blob/master/examples/musx.py
 
 """
+import numpy as np
 from cpmpy import *
 from cpmpy.expressions.variables import NDVarArray
 from cpmpy.transformations.get_variables import get_variables
@@ -33,7 +34,7 @@ def mus(soft, hard=[], solver="ortools"):
 
     assump = boolvar(shape=len(soft), name="assump")
     if len(soft) == 1:
-        assump = NDVarArray([assump])
+        assump = NDVarArray(shape=1, dtype=object, buffer=np.array([assump]))
 
     m = Model(hard+[assump.implies(candidates)]) # each assumption variable implies a candidate
     s = SolverLookup.get(solver, m)
--- a/cpmpy/tools/mus.py
+++ b/cpmpy/tools/mus.py
@@ -6,6 +6,7 @@ https://github.com/pysathq/pysat/blob/master/examples/musx.py
 
 """
 from cpmpy import *
+from cpmpy.expressions.variables import NDVarArray
 from cpmpy.transformations.get_variables import get_variables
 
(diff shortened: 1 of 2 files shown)


cpmpy/tools/mus.py
```diff
@@ -29,8 +31,11 @@ def mus(soft, hard=[], solver="ortools"):
     """
     # order so that constraints with many variables are tried and removed first
     candidates = sorted(soft, key=lambda c: -len(get_variables(c)))
-    
+
     assump = boolvar(shape=len(soft), name="assump")
+    if len(soft) == 1:
```
IgnaceBleukx, line 36: I think this should actually be resolved in `core.py`.
What we want is equivalent behaviour to: `bv.implies(all([some_cons]))` which extracts the 1 element from the list on the rhs and returns a single expression.
So maybe we can check wether the rhs is a list when calling `.implies()` in an expression that is not a NDVarArray?
Anyway, if that's too much hassle this fix is fine for me :)

IgnaceBleukx: if you leave this, maybe put a comment an refer to this bug request?

tests/test_tools_mus.py
```diff
@@ -21,6 +21,33 @@ def test_circular(self):
...
+        When assum is a single boolvar and candidates is a list (of length 1), it fails.
+        """
+        bv = boolvar(name="x")
+        hard = [~bv]
+        soft = [bv]
+
+        mus_cons = mus(soft=soft, hard=hard) # crashes
+        self.assertEqual(set(mus_cons), set(soft))
+
+    def test_bug_191_many_soft(self):
```
IgnaceBleukx, line 35: also add for mus_naive?

cpmpy/tools/mus.py
sourdough-bread: That would be a small improvement in core.py especially for the case where the implied constraint is either a single boolvar or an empty list.
For instance
```python
from cpmpy import *; 
import cpmpy as cp; 
bv, bv2 = boolvar(shape=2);
some_cons = [~bv2];
```

```python
>>> bv.implies(some_cons2)
(BV0) -> [~BV1]
>>>bv.implies([])
(BV0) -> []
```

sourdough-bread:
Added test case for naive mus and solved issue with unit soft constraint.

IgnaceBleukx:
Will merge

PR #200, opened 2022-12-22
Make line-expr more linear

IgnaceBleukx:
Implemented two new transformations for linear expressions:
1) only allow variables on lhs of comparisons
2) only allow constants on rhs of comparisons

Some solvers (such as Exact) require these types of linear constraints a input.


Wout4:
Looks good to me

JoD:
Tnx!

PR #201, opened 2022-12-23
Added floordivision '//' on integer variables

Wout4:
I put a syntax warning on using / in stead of a deprecation warning, because this prints the warning message.

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -73,6 +73,7 @@
         Comparison
         Operator
 """
+import warnings
 from types import GeneratorType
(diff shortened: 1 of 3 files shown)


IgnaceBleukx:
I suppose this prints a warning every time you use "/"? Maybe it would be good to just print it once and check if we have printed it already?

Wout4:
It only prints once in my IDE, I guess python warnings do that automatically

tias:
excellently executed, including the choice of using the warnings library. I think there will be future modeling cases where we might want to warn the user about undesirable options

PR #202, opened 2022-12-29
Glasgow Constraint Solver

tias:
Matthew's GCS solver interface

Unfinished: multiple constraint tests failing


tias:
Turns out that the transformations were called with different arguments then what they should be.

I updated the docs to make it clearer what they should be.


All the expression tests now pass, so it seems this would be ready to merge...

I do have one request @mmcilree , can we call it 'gcs' instead of 'glassgowconstraintsolver'?
s.SolverLookup.get("glassgowconstraintsolver", m) is really a lot to type... "gcs" would be nicer... just like gcspy is your lib
(we don't support both a shorthand and longhand, just a single one)

tias:
Hi Matthew @mmcilree ,

We've refactored our solver interface and solver making guidelines a bit, based also on this effort, and to make it easier to debug.

The two changes are that there should be a `transform()` function that does the transformations and returns the list of transformed constraints (very useful for debugging). And then the `__add__()` function can directly loop over the result of transform and post the constraint one by one (so no more need for a `_post_constraint()`, that gets merged into the add).

Would you be up for making those changes and testing what the current state is? Would be happy to see it progres...

(and we have a summerschool 2nd week of july, could be nice for participants to also have gcs available)

mmcilree:
Hi, the gcspy package is very out of date with respect to the Glasgow constraint solver - lots of the solver API has changed since I wrote the python bindings. It's on my todo list to update it and make sure it all works correctly, I'll try to get to it possibly next week.

cpmpy/solvers/gcs.py
```diff
@@ -0,0 +1,413 @@
...
+            'circuit', 
+            'xor'}
+        cpm_cons = decompose_in_tree(cpm_cons, supported)
+        cpm_cons = flatten_constraint(cpm_cons)  # flat normal form
+
+        # NB: GCS supports full reification for linear equality and linear inequaltiy constraints
+        # but no reification for linear not equals and not half reification for linear equality. 
+        # Maybe a future transformation (or future work on the GCS solver).
+        cpm_cons = reify_rewrite(cpm_cons, supported=frozenset(['==']))
+        cpm_cons = only_numexpr_equality(cpm_cons, supported=frozenset(["sum", "wsum"]))  # supports >, <, !=
```
tias, line 228: If you want some of your reified sum/wsums to pass, you could not reify_rewrite them above and copy the only_numexpr_equality function (which checks for reification, the documentation is outdated) and handle the special cases there..

cpmpy/solvers/gcs.py
```diff
@@ -0,0 +1,413 @@
...
+        then post those primitive constraints directly to the native solver
+        :param cpm_con CPMpy constraint, or list thereof
+        :type cpm_con (list of) Expression(s)
+        """
+        # add new user vars to the set
+                # add new user vars to the set
+        get_variables(cpm_cons, collect=self.user_vars)
+
+        for con in self.transform(cpm_cons):
+            self._post_constraint(con)
```
tias, line 252: for tiny efficiency speedups, we recommend putting the body of _post_constraint here instead of doing a function call for every element in the list; that is what we now do in the other solvers too.

cpmpy/solvers/gcs.py
```diff
@@ -0,0 +1,413 @@
...
+        elif isinstance(cpm_expr, Operator):
+            # 'and'/n, 'or'/n, '->'/2
+            if cpm_expr.name == 'and':
+                return self.gcs.post_and(self.solver_vars(cpm_expr.args))
+            elif cpm_expr.name == 'or':
+                return self.gcs.post_or(self.solver_vars(cpm_expr.args))
+
+            # Reified constraint: BoolVar -> Boolexpr or BoolVar == Boolexpr
+            # LHS must be boolvar due to only_bv_reifies
+            elif cpm_expr.name == '->' or cpm_expr.name == '==':
```
tias, line 282: Operator can not have name ==

tias: (the outer elif checks for Operator)

cpmpy/solvers/gcs.py
```diff
@@ -0,0 +1,413 @@
...
+                # Reified Comparison
+                elif isinstance(bool_expr, Comparison):
+                    lhs = bool_expr.args[0]
+                    rhs = bool_expr.args[1]
+                    if bool_expr.name == '==':
+                        return self.gcs.post_equals_reif(*self.solver_vars([lhs, rhs]), reif_var, fully_reify)
+                    elif bool_expr.name == '<=':
+                        return self.gcs.post_less_equal_reif(*self.solver_vars([lhs, rhs]), reif_var, fully_reify)
+                    elif bool_expr.name == '<':
+                        return self.gcs.post_less_if(*self.solver_vars([lhs, rhs]), reif_var, fully_reify)
```
tias: the other functions are post_X_reif and this one is post_X_if?

also on line 324 you use post_or_if while on 292 you use post_or_reif

cpmpy/solvers/utils.py
```diff
@@ -24,8 +24,9 @@
 from .gurobi import CPM_gurobi
 from .ortools import CPM_ortools
 from .minizinc import CPM_minizinc
-from .pysat import CPM_pysat
 from .z3 import CPM_z3
+from .gcs import CPM_gcs
+from .pysat import CPM_pysat
```
tias, line 29: very small preference not to touch unneeded lines

tests/test_constraints.py
```diff
@@ -11,6 +11,7 @@
 #   make sure that `SolverLookup.get(solver)` works
 # also add exclusions to the 3 EXCLUDE_* below as needed
 SOLVERNAMES = [name for name, solver in SolverLookup.base_solvers() if solver.supported()]
+SOLVERNAMES = ['gcs']
```
tias, line 14: be sure to remove this before merge

nasty_nesting.py
```diff
@@ -0,0 +1,15 @@
+from cpmpy import *
```
tias, line 1: I think this is from the docs, right?

You could add it to examples/, then it will be run as a test for all solvers too.
(maybe add a comment on top saying that it is from the docs to demonstrate the free use of nested expressions?)

tias (commented):
code review, only minor comments

cpmpy/solvers/gcs.py
```diff
@@ -0,0 +1,413 @@
...
+            (if this argument is not supplied the proof files created by the solver are deleted).
+
+            Arguments that correspond to solver parameters:
+            #TODO document solver parameters (there are none at the moment)
+        """
+        # ensure all user vars are known to solver
+        self.solver_vars(list(self.user_vars))
+
+        if time_limit is not None:
+            raise NotImplementedError("Glasgow Constraint Solver does not currently support timeouts.")
```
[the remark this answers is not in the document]
mmcilree: Have implemented timeouts for gcs now.

cpmpy/solvers/gcs.py
```diff
@@ -0,0 +1,413 @@
...
+
+        # translate solution values (of user specified variables only)
+        self.objective_value_ = None
+        if has_sol:
+            # fill in variable values
+            for cpm_var in self.user_vars:
+                sol_var = self.solver_var(cpm_var)
+                if isinstance(cpm_var, _BoolVarImpl):
+                    # Convert back to bool
+                    cpm_var._value = bool(self.gcs.get_solution_value(sol_var))
```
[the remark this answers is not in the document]
mmcilree: I'm not sure about this one. What do you mean by well-type? GCS always returns integers for solution values, so I thought I'd need to convert 0-1 back to False/True for boolvars.

tias:
then indeed you need the bool() part for Boolean variables

cpmpy/solvers/gcs.py
```diff
@@ -0,0 +1,413 @@
...
+            or returns from cache if previously created
+        """
+        if is_num(cpm_var): # shortcut, eases posting constraints
+            return self.gcs.create_integer_constant(cpm_var)
+
+        # special case, negative-bool-view
+        # work directly on var inside the view
+        if isinstance(cpm_var, NegBoolView):
+            # gcs only works with integer variables, so not(x) = -x + 1
+            return self.gcs.add_constant(self.gcs.negate(self.solver_var(cpm_var._bv)), 1)
```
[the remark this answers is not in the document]
mmcilree: Also not sure about this?
`gcs.negate(x)` maps a new var id for the view `-x` (without creating a new underlying variable).
`gcs.add_constant(x, c)` similarly maps a new id for the view `x + c`.

So `~a|b` would be translated to `Or{view_of_a, b}` where `view_of_a` is a reference to `-a + 1` (we have a specific wrapper for viewed variables).
Since our `Or` constraint just asserts that at least one of it's arguments is non-zero this shouldn't cause any problems?

.github/workflows/python-test.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           pip install pycryptosat
           pip install z3-solver
           pip install exact
+          pip install gcspy
```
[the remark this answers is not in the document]
mmcilree: NB: I don't currently have permission to edit this file. 

```
 ! [remote rejected]   glasgowconstraintsolver -> glasgowconstraintsolver (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/python-test.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/CPMpy/cpmpy.git'
```

Wout4 (commented):
I'm having some trouble adding an objective to a model:
e.g.
```
a = boolvar()
m = Model(a == 1)
m.minimize(a).
m.solve('gcs')
```
This crashes..
Other than that test are successful.

cpmpy/solvers/gcs.py
```diff
@@ -0,0 +1,593 @@
...
+            elif cpm_expr.name == 'inverse':
+                self.gcs.post_inverse(self.solver_vars(cpm_expr.args[0]), self.solver_vars(cpm_expr.args[1]))
+            elif cpm_expr.name == 'alldifferent':
+                self.gcs.post_alldifferent(self.solver_vars(cpm_expr.args))
+            elif cpm_expr.name == 'table':
+                self.gcs.post_table(self.solver_vars(cpm_expr.args[0]), cpm_expr.args[1])
+            elif cpm_expr.name == 'negative_table':
+                self.gcs.post_negative_table(self.solver_vars(cpm_expr.args[0]), cpm_expr.args[1])
+            elif isinstance(cpm_expr, GlobalConstraint):
+                # GCS also has SmartTable, Regular Language Membership, Knapsack constraints
```
Wout4, line 583: cpmpy's ShortTable is SmartTable

Wout4:
Some test fail due to pycroptosat not being installed, will merge this anyway as that does not seem to happen on master

PR #204, opened 2023-01-01
140 refactoring of solverinterface

tias:
Implements the generic `transform()` function in the solverinterface, and implements it for the different solver interfaces.

Includes the necessary docs changes.

Can make debugging solver interfaces (esp ones with complex transformations) much easier and transparent...

It also allows calling transformations again on already transformed expressions (e.g. when decomposing a global constraint, which itself might consists of global constraints that need decomposition. I don't have the best example for that, but it exists or will exists. Maybe reified n-ary xor is/was already so). It could also be passed to functions/transformations that can not guarantee that their output satisfies earlier transformations in the chain... I again dont have a concrete example but it exists or will exist.

Please review.

Also, I think it exposes or creates the following bug. Should probably be solved in a next/future commit.
Our tests dont seem to test reification of globals, we could consider adding that (at some point it looked like the current ortools fails on the below too, but repo was in a messy state)
```
s = SolverLookup.get("pysat")
bv = boolvar(3)
s += (boolvar().implies(Xor(bv)))
s.solve()
```


tests/test_constraints.py
```diff
@@ -130,13 +131,9 @@ def global_constraints():
         -  AllDifferent, AllEqual, Circuit,  Minimum, Maximum, Element,
            Xor, Cumulative
     """
-    if SOLVERNAME is None:
+    if SOLVERNAME is None or SOLVERNAME in EXCLUDE_GLOBAL:
```
IgnaceBleukx, line 134: Why add the second part of the condition? Now global constraints are not tested for any solver?
I think the previous version was better.

cpmpy/solvers/solver_interface.py
```diff
@@ -170,16 +165,62 @@ def solver_vars(self, cpm_vars):
...
+            the `cpmpy/transformations/` directory.
+
+            See the 'Adding a new solver' docs on readthedocs for more information.
+
+        :param cpm_expr: CPMpy expression, or list thereof
+        :type cpm_expr: Expression or list of Expression
+
+        :return: list of Expression
+        """
+        return cpm_expr  # overwrite this function and call the transformations you need
```
IgnaceBleukx, line 207: this should return a list

cpmpy/solvers/solver_interface.py
```diff
@@ -170,16 +165,62 @@ def solver_vars(self, cpm_vars):
...
+
     def _post_constraint(self, cpm_expr):
         """
-            Post a primitive CPMpy constraint to the native solver API
+            Post a supported CPMpy constraint directly to the underlying solver's API
+
+            What 'supported' means depends on the solver capabilities, and in effect on what transformations
+            are applied in `__transform()__`.
```
IgnaceBleukx, line 214: the func definition is just `transform()` without double underscores

cpmpy/solvers/solver_interface.py
```diff
@@ -170,16 +165,62 @@ def solver_vars(self, cpm_vars):
...
-            What 'primitive' means depends on the solver capabilities,
-            more specifically on the transformations applied in `__add__()`
+            Solvers can raise 'NotImplementedError' for any constraint not supported after transformation
 
-            Solvers do not need to support all constraints.
+        :param cpm_expr: list of CPMpy expressions
+        :type cpm_expr: list of Expression
         """
-        return None
+        raise NotImplementedError()  # abstract function, overwrite
```
IgnaceBleukx, line 221: Maybe better to put the comment in the args of the error?

IgnaceBleukx (commented):
Yes, this is nice. Left some small comments but overall looks fine!
For the documentation, maybe good to also mention the 'transform()' function in the debugging guide?

Regarding your comment on reified global constraints: we DO test them in `test_constraints` but there was an if-check which ommitted all of them... (left a comment in the code)

The bug you mentioned is also present in master, and actually we excluded that one from the test_constraints for OR-tools for this reason probably. The XOR is tricky, as it is/was treated as an `Operator` in early stages of the repo, but is now a `GlobalConstraint` so it probably falls trough some checks in transformations.

Best to open a separate bug report for the XOR one

tests/test_constraints.py
tias: yes, I made a mistake here, has since been fixed in master and merged

tias:
I addressed your comments, and updated the debugging guide.

Reified globals are indeed tested (again), will test the XOR on master first.

Can you run test_constraints on gurobi still? Can't do that here. Also, if you know a way to run the test_constraints on all supported solvers...

IgnaceBleukx:
I ran the test_constraints for gurobi, all finished without issue.
I feel like the tests are running faster than before... Could the refactoring have a significant impact on runtime you think? I guess it can as we only do the `transform()` instead of the entire `__add__` when decomposing global constraints!

We can certainly test all supported solvers, but might be a little tricky to generate the correct constraints. I'll have a look at it but that should't prevent this pull request to be merged already. (Will do the testing either in master or a separate pull request)

IgnaceBleukx:
Tests now automatically run for all supported solvers. This uncovered some bugs regarding xors and cumulative constraints. Some small doc-changes as well, should be ready to merge now.

IgnaceBleukx:
We can also use `pytest.skip` instead of the dictionaries on top of the file. We could then skip any tests throwing a `NotImplementedError`. It would make the testfile a lot easier to read and you get a nice overview of tests skipped anyway, so still easy to check which _should_ have been tested/passed but were skipped due to NotImplementedErrors

tias:
That is a great idea.

It is also a good motivation for having CPMpy specific exceptions.

E.g. I added some 'skipifs' to the test suite in case a solver is not 
installed, but if we could have a skipif on an exception, we could 
simply have a generic catch on a 'CPMpySolverNotAvailable exception or so...


T

IgnaceBleukx:
Ok, doing the above with regards to skipping the tests in `test_constraints.py` uncovered a whole class of errors in `flatten_constraint` which should be fixed before this is merged! Will make a separate issue and pull request.

cpmpy/transformations/to_cnf.py
```diff
@@ -126,6 +126,9 @@ def flat2cnf(constraints):
                 cnf += [a0sub | a1 for a0sub in subcnf]
                 continue
 
+        elif expr.name == "xor":
```
tias, line 129: I find this a bit of a shortcut.

More proper would be, this is a proposal, to add a 'supported_globals=frozenset()' argument to the top-level function, and then here do smth like 'if expr.is_bool() and isinstnace(expr, globalCons) and expr.name not in supported_globals: to_cnf(decompose())


it is actually part of the question where globals should be decomposed, and what trnasform they should get, but at least making this one a bit more generic would be good?

rest of pull request looks ready

IgnaceBleukx: Yes I agree, this also goes hand in hand with issue #219... Here we also want to leave some constraints for the solver to encode to CNF... I'll introduce a "supported" argument for this one.

tias:
I've taken out the Xor decomposition change and the testsuite change as separate commits to master, then this can be squash-committed as a single change

PR #207, opened 2023-01-05
an ITE() helper function and count/gcc globals

tias:
part of a modeling exercise. Should be 'properly' integrated and tested, potentially


Wout4:
Only works on ortools for now

cpmpy/expressions/globalconstraints.py
```diff
@@ -242,8 +242,20 @@ def deepcopy(self, memdict={}):
...
 
-    # TODO: value()
-
+    def value(self):
+        from .python_builtins import all
+        pathlen = 0
+        idx = 0
+        visited = []
+        arr = [argval(a) for a in self.args]
+        while(idx not in visited):
```
JoD, line 251: should `visited` be a set to avoid this linear test? Or is this overkill?

cpmpy/expressions/globalconstraints.py
```diff
@@ -242,8 +242,20 @@ def deepcopy(self, memdict={}):
...
+            if idx not in range(len(arr)):
```
JoD, line 252: More efficient: `if 0 <= idx <= len(arr)`?

cpmpy/expressions/globalconstraints.py
```diff
@@ -242,8 +242,20 @@ def deepcopy(self, memdict={}):
...
+                break
+            visited.append(idx)
+            pathlen += 1
+            idx = arr[idx]
+
+        return all([(pathlen == len(self.args)), (AllDifferent(self.args).value()), (arr[-1] == 0)])
```
JoD, line 258: More efficient: `return pathlen == len(self.args) and arr[-1]==0`.

We can drop the alldiff because `pathlen==len(self.args)` is equivalent to it.

cpmpy/expressions/globalconstraints.py
```diff
@@ -292,6 +304,18 @@ def deepcopy(self, memodict={}):
...
         return Minimum(copied_args)
 
+    def decompose_comparison(self, cpm_op, cpm_rhs):
+        """
+        Decomposition if it's part of a comparison
+        """
+        from .python_builtins import any, all
+
+        arr = argval(self.args)
+        return [any(eval_comparison(cpm_op, j, cpm_rhs) & (all([j <= a for a in arr])) for j in arr)]
```
JoD, line 314: - why use bit-wise `&` instead of `and`?
- this is a quadratic blow-up. We can avoid this by introducing an auxiliary symbol, say `_min` where `any(x>=_min for x in arr) and all(x<=_min for x in arr)` and then just assert `eval_comparison(cpm_op,x,cpm_rhs)`.

cpmpy/expressions/globalconstraints.py
```diff
@@ -316,8 +340,20 @@ def deepcopy(self, memodict={}):
...
         return Maximum(copied_args)
 
+    def decompose_comparison(self, cpm_op, cpm_rhs):
+        """
+        Decomposition if it's part of a comparison
+        """
+        from .python_builtins import any, all
+
+        arr = argval(self.args)
+        return [any(eval_comparison(cpm_op, j, cpm_rhs) & (all([j >= a for a in arr])) for j in arr)]
```
JoD, line 350: Same comment as for `Minimum`.

cpmpy/expressions/globalconstraints.py
```diff
@@ -496,3 +532,114 @@ def deepcopy(self, memodict={}):
         return Cumulative(*copied_args)
 
 
+
+# from Hakan's global decompositions.
```
JoD, line 536: comment should be moved to `def count`

cpmpy/expressions/globalconstraints.py
```diff
@@ -496,3 +532,114 @@ def deepcopy(self, memodict={}):
...
+        """
+
+    def __init__(self, a, gcc):
+        super().__init__("gcc", [a,gcc])
+
+    def decompose(self):
+        a, gcc = self.args
+        ub = max([v.ub for v in a])
+        assert (len(gcc) == ub + 1), f"GCC: length of gcc variables {len(gcc)} must be ub+1 {ub + 1}"
+        return [Count(a, i) == v for i, v in enumerate(gcc)]
```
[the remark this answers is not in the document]
Wout4: Those functions should disappear, since we want our global constraints to be subclassing GlobalConstraint. Was just a draft of the decomposition

cpmpy/transformations/linearize.py
```diff
@@ -227,6 +227,12 @@ def linearize_constraint(cpm_expr):
 
         return linearize_constraint(flatten_constraint(constraints))
 
+    if cpm_expr.name in [">=", "<=", "=="] and cpm_expr.args[0].name == "count":
```
JoD, line 230: Can `<` and `>` still occur here? If not, add a comment.

cpmpy/expressions/globalconstraints.py
```diff
@@ -496,3 +532,114 @@ def deepcopy(self, memodict={}):
...
+    """
+
+    def __init__(self,arr,val):
+        super().__init__("count", [arr,val], is_bool=False)
+
+    def value(self):
+        arr, val = self.args
+        val = argval(val)
+        argvals = [argval(a) for a in arr]
+        return sum([argv == val for argv in argvals])
```
JoD, line 604: Previous line can be merged with this: `return sum([argval(a) == val for a in arr])`

cpmpy/expressions/globalconstraints.py
```diff
@@ -496,3 +532,114 @@ def deepcopy(self, memodict={}):
...
+        return Count(arr, val)
+
+    def decompose_comparison(self, cmp_op, cmp_rhs):
+        """
+        Count(arr,val) can only be decomposed if it's part of a comparison
+        """
+        from .python_builtins import any
+
+        arr, val = self.args
+        return [any((eval_comparison(cmp_op, j, cmp_rhs) & (Operator('sum',arr == val) == j) for j in range(len(arr) + 1)))]
```
JoD, line 617: Would `eval_comparison(cmp_op, Operator('sum',arr==val), cmp_rhs)` work? The `any` seems unnecessary.

JoD:
Overall, seems good work. I've added a couple of comments, some of which are probably due to my limited understanding of the existing codebase. There are a couple of decompositions that perhaps can be improved, and a couple of efficiency optimizations.

JoD:
The main thing I'm wondering is if we have tests, e.g., for the value functions. Is there a policy on what should be added or how these are tested?

cpmpy/expressions/globalconstraints.py
Wout4: I implemented this change, but for the auxiliary symbol we need to calculate bounds. So will only work properly after we added get_bounds to expressions (as I am working on in branch Bound_calculations)

Wout4:
closed because everything is split up in smaller pr's

PR #209, opened 2023-01-13
only check minizinc path on windows systems

Wout4:


tias:
better throw 'e' with a warning if minzinc is in the path after all

PR #218, opened 2023-01-25
Issue 157 fix circuit constraint

Dimosts:
This pull request is fixing issue #157 

This issue appears when a circuit of 1 element (or even no element) is created and it crashes both "ortools" and "gurobi".

The bug appears when the "decompose" method of the expression is called, in _post_constraint(). The decompose method is called only in ortools and gurobi and that is why this happens there.

In the decompose method of Circuit, it is assumed that it contains more than one variables. So, "order" is assumed to be an array of variables

`order = intvar(0,n-1, shape=n)`

and then an index is used to define that the last one has no succesor.

`order[n-1] == 0`

However, when only 1 variable is given, then n = 1, resulting in "order" not being subscriptable.

This could be handled by just returning True when only one variable is given (as there is no real ordering required in that case), but that creates a different problem:

If Circuit([x]) is the only constraint involving the variable x, then the when the cpmpy model tries to get the value of the variable, it crashes, as the solver doesn't have the variable (no constraint involving x as Circuit([x]) is replaced with True).

This happens in both ortools and gurobi. E.g. in ortools.py it starts at lines 186-187:

```
                cpm_var._value = self.ort_solver.Value(self.solver_var(cpm_var))
```

as we have stashed the variable in solver vars (i.e. self.user_vars), but then we never added a constraint involving this var in the solver.

```
        self.user_vars.update(get_variables(cpm_con))   <- here we add the variable

        cpm_cons = only_bv_implies(cpm_cons) 

        for con in cpm_cons:
            self._post_constraint(con)   <- here we do not post the constraint

```

Thus, the command self.ort_solver.Value(self.solver_var(cpm_var)) crashes (a few function calls later).

The issue is solved by returning a true constraint, that involves the variable given to the Circuit constraint, so it will be considered in the solver. That is, the constraint `var >= lower bound of var` is returned.

In case no variables are given to the Circuit constraint, an exception is raised.


IgnaceBleukx:
It feels a little hacky, but I wouldn't know of another way to fix this...
The only thing that can be done I think is to explicitely make the variables when adding the non-decomposed version of the constraint to the solver.
I.e., for OR-tools, on line 312, we can do `self.solver_vars(get_variables(cpm_cons))` but this will create some overhead when posting constraints. In fact, I believe this is the way we did it some time ago...

No strong opinion which fix to chose. Fine to merge for me.

Dimosts:
I resolved a conflict that existed with master.

I think it is ok to merge this way, to avoid adding an overhead. Just adding a fake constraint if only one variable is given resloves this, and I think we have done it in other cases.

Dimosts:
It seems that the post_constraint() has changed in ortools.py from when the issue/pull request has opened. Instead of decomposing, now Circuit constraint is directly posted to ortools.

In order to do that, First the arcs are are found and are then given to the Circuit constraint of ortools

```
        elif cpm_expr.name == "circuit":
            # ortools has a constraint over the arcs, so we need to create these
            # when using an objective over arcs, using these vars direclty is recommended
            # (see PCTSP-path model in the future)
            x = cpm_expr.args
            N = len(x)
            arcvars = boolvar(shape=(N,N), name="circuit_arcs")
            # post channeling constraints from int to bool
            self += [b == (x[i] == j) for (i,j),b in np.ndenumerate(arcvars)]
            # post the global constraint
            # when posting arcs on diagonal (i==j), it would do subcircuit
            ort_arcs = [(i,j,self.solver_var(b)) for (i,j),b in np.ndenumerate(arcvars) if i != j]
            return self.ort_model.AddCircuit(ort_arcs)
```

So, the previous problem has been removed. Now, in the case of 0 or 1 variables given in the Circuit constraint, the system prints the following error: "ValueError: AddCircuit expects a non-empty array of arcs". it is good that it is not crashing, but I think we would need a more informative message for the user, instead of posting the error from ortools.

I am changing the pull request. Will just add a CPMpy exception at that point, printing to the user an appropriate message.

tias:
sorry for that, I didn't realize it would intervene here.

But for gurobi (or z3, or... any other using the decomposition) it would still be an issue? So better add the raise exception in the decompose too? You can reopen this branch and do that here (e.g. it would be what is in this branch?).

Dimosts:
It is indeed better that way.

I just introduced an exception when the constraint is initialized, checking how many variables the constraint is given. This covers all constraints.

Issue #183, opened 2022-11-29
Revisit linearize transformation


IgnaceBleukx:
With wsum optimizations (#182) and other changes in place, we should revisit the linearize transformations, as well defining a better "linear normal form" so we know exactly what kind of expressions can come out of the transformations.

Furthermore, we should also define the "supported" parameter as many MIP-solvers support API calls such as `z = AND(x,y)`etc

IgnaceBleukx:
Proposal for a grammer for linear expressions:

```
Linearize returns any of :
1) LinExpr
2) Boolvar -> LinExpr 
3) Expr                       (Expr.name in supported)
-----------------------------------------------
LinExpr:
1) Expr <= Const              (Expr.name in {"sum","wsum"})
2) Expr >= Const              (Expr.name in {"sum","wsum"})
3) Expr == Const              (Expr.name in {"sum","wsum"})
```

This differs from the current specification as variabels can be part of the rhs of LinExpr (except in reified expressions where they have to be constants).
Note the current imlementation is very highly tailored to support as much as possible the API of Gurobi. With Exact, SCIP and potentially other linear solvers in the pipeline, linearization should be generalized somewhat more.

With the above changes in place, `only_var_lhs` and `only_const_rhs` can also go.

JoD:
- I think we also need to add
```
4) BoolVar <- LinExpr
```
- What is meant with `3) Expr   (Expr.name in supported)`? Is this how we allow Gurobi-supported operations (i.e., min/max/abs or multiplication with a Boolean variable)? If so, `max(x,y)` is an `Expr`, but not a constraint (no Boolean type), so I do not see it yet.

JoD:
This grammar is close to what Exact allows: https://gitlab.com/JoD/exact/-/blob/main/src/Exact.hpp. 

In the above notation, Exact also allows:
```
Boolvar <-> LinExpr
Const >= Expr >= Const
```
But it is more restrictive as it only allows nested `LinExpr` of the form
```
Expr >= Const
```
There exist simple conversions between all (splitting `==` or `<->` is a straightforward operation), so I can live with the above grammer.

JoD:
Ok, I'm going to propose two variations: a maximal one (easy for CPMpy) and a minimal one (easy for the solvers). I'm ignoring the "supported" rule for now.
```
**Maximal:**
1) LinExpr
2) Boolvar -> LinExpr
3) Boolvar <- LinExpr
4) Boolvar <-> LinExpr
[...]

Issue #140, opened 2022-10-18
Refactoring of solverinterface


IgnaceBleukx:
Discussed some time ago but not implemented yet:
It would be good to change put the `__add__` functionality into the superclass `SolverInterface` as the only difference between solvers is the transformations applied to the constraints before calling `_post_constraint`.
Therefore, it would make sense to have a separate `transform()` function for every solver interface to implement. This also allows users to get the final representation of the constraints before they are posted to the solver.
Especially relevant for debugging purposes.

PR #220, opened 2023-01-27
Exact backend

JoD:
Blocking issue: runnint `test_constraints.py` yields (variations on) the following error:
`FAILED tests/test_constraints.py::test_comparison_constraints[([x y z][l]) > (l)] - NotImplementedError: Exact: Unexpected lhs mul expr (x) * (BV140)`

This should not happen, as `mul` is an excluded operator for `exact`.

Nice-to-have issues:
- compilation of Exact in Ignace's workflow. It compiles on macOS, but not the way Ignace prefers.
- no exclusion of operators `min`, `max` and `abs`: these can quite straightforwardly be linearized, but this does not seem to happen fully yet.


tias:
I just merged #140 Refactoring of Sovlerinterfaces.

It splits `__add__` into a new `transform()` function that all solvers (should) implement and the `_post_cosntraiont` as before.

Could you change your interface accordingly? The transform function will make debugging wrong transformations much easier btw, as sketched in
https://cpmpy.readthedocs.io/en/latest/how_to_debug.html#debugging-an-unsatisfiable-model

also the testsuite is now such that if your solver is in SolverLookup.base_solvers() then test_constraints will automatically be run on it.

tias:
All doc fixes are great. Could you push (or pull request) them in a separate commit (doc fixes can be applied directly on master)

As to the pigeonhole example:
- is it just another example, independent of exact? Then please post it as a separate pull requestion without specifying a solver for it
- does it serve to make a point about exact? Then please make the example show that point, e.g. contrasting it with one of the other solvers (but for all other solvers than ortools you should add a check whether it is installed)

JoD:
The pigeonhole example was just me doing a sanity check of the backend. It may make a point about Exact, will have to see if OR-tools can be confused to run exponentially.

Will update to the new interface and split off the document changes.

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def getSolAndObj(self):
+        if not self.xct_solver.hasSolution():
+            self.objective_value_ = None
+            return False
+
+        # fill in variable values
+        exact_vals = self.xct_solver.getLastSolutionFor([self.solver_var(cpm_var) for cpm_var in self.user_vars])
+        i = 0
+        for cpm_var in self.user_vars:
```
tias, line 98: I think user_vars is a set, which is an unordered type.
I would not trust Python to iteration in this and the previous line in the same order...
maybe
```python
lst_vars = list(self.user_vars)
lst_vals = self.xct_solver.getLas...
for cpm_var, val in zip(lst_vars, lst_vals):
    cpm_var._value = val
    if....
```

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,488 @@
...
+        assert subsolver is None, "Exact does not allow subsolvers."
+
+        # initialise the native solver object
+        self.xct_solver = xct()
+
+        # for solving with assumption variables,
+        self.assumption_dict = None
+
+        # objective can only be set once, so keep track of this
+        self.has_objective = False
```
tias, line 84: there might be a function has_objective() since a recent commit, maybe use has_objective_ as var in that case

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,488 @@
...
+        if not self.has_objective:
+            self.has_objective=True
+            self.xct_solver.init([],[])
+
+        # set time limit?
+        if time_limit is not None:
+            self.xct_solver.setOption("timeout",str(time_limit))
+        # set additional keyword arguments
+        for (kw, val) in kwargs.items():
+            print(kw,val)
```
tias, line 135: these should go at some point : )

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,488 @@
...
+
+        return solsfound
+
+
+    def solver_var(self, cpm_var):
+        """
+            Creates solver variable for cpmpy variable
+            or returns from cache if previously created
+        """
+        assert not is_num(cpm_var)
```
tias, line 228: are you sure you want to do this?

the 'DirectConstraint' in a pull request is not bothered with calling solver_var on constants, it expects to get a value back if it is a value. Note that you could even do your maximum/str() trick here if you wish...

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,488 @@
...
+        :param cpm_expr: CPMpy expression, or list thereof
+        :type cpm_expr: Expression or list of Expression
+
+        :return: list of Expression
+        """
+        # apply transformations, then post internally
+        # expressions have to be linearized to fit in MIP model. See /transformations/linearize
+        cpm_cons = flatten_constraint(cpm_expr)  # flat normal form
+        cpm_cons = reify_rewrite(cpm_cons, supported=frozenset(['sum', 'wsum']))  # constraints that support reification
+        cpm_cons = only_bv_implies(cpm_cons)  # anything that can create full reif should go above...
```
tias, line 341: below it seems that you have _reif and _reif_left and _reif_right, then you don't need this 'only_bv_implies, which translates full reifs and reif_rights (in fact I think you undo some of these transformations below)

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,488 @@
...
+            assert isinstance(lhs, _NumVarImpl) or lhs.name == "sum" or lhs.name == "wsum",f"Unknown linear expression {lhs} on right side of indicator constraint: {cpm_expr}"
+
+            xct_coefs, xct_vars, xct_rhs = self._make_numexpr(lhs,rhs)
+
+            if isinstance(cond, NegBoolView):
+                cond, bool_val = self.solver_var(cond._bv), False
+            else:
+                cond, bool_val = self.solver_var(cond), True
+
+            if sub_expr.name == "==":
```
tias, line 440: The design idea of our solver interfaces so far is that the transformations should transform expressions into those (almost) directly supported by the solver.

I have the feeling that here you are doing a bit too much transformation-like work. That we would better put these into a new transformation?

tias (commented):
first further look

cpmpy/solvers/exact.py
JoD: Good question. I was not thinking about outside code calling on this, so I assumed that I could add this as a precondition. I am confused that values and variables are used interchangeably. I guess they function as _terminals_ in the expression tree.

The standard code 
```
if is_num(cpm_var): # shortcut, eases posting constraints
    return cpm_var
```
leads to some `cpm_var` in a direct constraint being replaced, but some `cpm_var` remain, but they actually represent values. So when posting the direct constraint, these `cpm_var` should still be replaced by solver-specific values? This may not be the best approach.

Maybe we should introduce some kind of `solver_val` that the solver can translate to? And adjust `solver_var` accordingly.

cpmpy/solvers/exact.py
JoD: Yeah, I agree. Also see https://github.com/CPMpy/cpmpy/issues/183.

cpmpy/solvers/exact.py
JoD: I can't seem to find (variations on) `has_objective()`. Renamed to `objective_is_set` to avoid name clashes.

cpmpy/solvers/exact.py
JoD: Relevant point, fixed.

cpmpy/solvers/exact.py
JoD: Yeah, fixed :)

cpmpy/solvers/exact.py
JoD: From a practical point of view, I can add the check, and write a wrapper that adds the assert, only called when translating the constraints (when no values should be passed).

cpmpy/solvers/exact.py
JoD: Ah, I did not realize that this is what `only_bv_implies` does. Yeah, should be removed.

Exact only accepts pure Boolean variables in the head of the reification, but CPMpy also allows negated variables (i.e., literals). Part of the work below addresses that: for negative literals, the direction of the implication has to be changed for Exact.

JoD: Alas, just removing it yields
```
AssertionError: Left hand side of implication (or([BV1035, BV1038, BV1041])) -> (p >= 1) should be boolvar
```
https://github.com/CPMpy/cpmpy/blob/master/cpmpy/transformations/linearize.py#L329

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+
+
+    def __init__(self, cpm_model=None, subsolver=None):
+        """
+        Constructor of the native solver object
+
+        Requires a CPMpy model as input, and will create the corresponding
+        Exact solver object xct_solver.
+
+        xct_solver can be modified externally before
```
tias, line 66: we have a uniform way in the constructor to lists the properties tht are stored in this class, e.g. I see `xct_solver`, assum_dict, solv_inited, ...
please list exhaustively here

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+
+        # objective can only be set once, so keep track of this
+        self.solver_is_initialized = False
+        self.has_objective = False
+        self.objective_minimize = True
+
+        # initialise everything else and post the constraints/objective
+        super().__init__(name="exact", cpm_model=cpm_model)
+
+    def getSolAndObj(self):
```
tias, line 95: this looks like a private function, not something you expect a user of the interface to use? If so you could name it `_getSolAndObj()` also it is not really a getter because it returns true/false, so maybe `_fillObjAndVars()`?

cpmpy/solvers/exact.py
tias: maybe also why you store it, e.g. solve_minimize because it only supports minimisation so you have to manually flip the sign or so?

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+
+        # translate objective
+        self.objective_value_ = self.xct_solver.getObjectiveBounds()[1] # last upper bound to the objective
+        if not self.objective_minimize:
+            self.objective_value_ = -self.objective_value_
+
+        return True
+
+    def solve(self, time_limit=None, assumptions=None, **kwargs):
+        # TODO: test this function
```
tias, line 114: done? : )

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+                           get_core() returns a small subset of assumption variables that are unsat together.
+
+            Additional keyword arguments:
+            The Exact solver parameters are defined by https://gitlab.com/JoD/exact/-/blob/main/src/Options.hpp#L207
+        """
+        from exact import Exact as xct
+
+        if not self.solver_is_initialized:
+            assert not self.has_objective
+            self.xct_solver.init([],[])
```
tias, line 131: why wait till here? why not in constructor?
your init does not need to know vars/constraints, so I think it can be done in the constructor

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+            self.xct_solver.setOption(kw,str(val))
+
+        # set assumptions
+        if assumptions is not None:
+            assert(all(is_bool(v) for v in assumptions))
+            assump_vals = [int(not isinstance(v, NegBoolView)) for v in assumptions]
+            assump_vars = [self.solver_var_no_num(v._bv if isinstance(v, NegBoolView) else v) for v in assumptions]
+            self.assumption_dict = {xct_var: (xct_val,cpm_assump) for (xct_var, xct_val, cpm_assump) in zip(assump_vars,assump_vals,assumptions)}
+            self.xct_solver.setAssumptions(assump_vars, assump_vals);
+            # NOTE: setAssumptions does not clear previous assumptions, is this ok?
```
tias, line 145: good point... I kind-of expect only the vars in the latest call should be True. At the very least the function definition should document what it does (e.g. only latest, or all)
in OrTools we first clear, and then add
(how can you actually clear them from the CPMpy interface? we can only give those that should be true... if we can't clear from the interface, I think we should clear by default)

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+            self.cpm_status.exitstatus = ExitStatus.UNKNOWN # TODO
+        elif my_status == 3: # found timeout
+            self.cpm_status.exitstatus = ExitStatus.UNKNOWN # TODO
+        else:
+            raise NotImplementedError(my_status)  # a new status type was introduced, please report on github
+
+        return self.getSolAndObj()
+
+    def solveAll(self, display=None, time_limit=None, solution_limit=None, **kwargs):
+        # TODO: test this function
```
tias, line 176: done? : )

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+                - time_limit: stop after this many seconds (default: None)
+                - solution_limit: stop after this many solutions (default: None)
+                - any other keyword argument
+
+            Returns: number of solutions found
+        """
+
+        assert not self.has_objective # TODO: is this true?
+
+        if not self.solver_is_initialized:
```
tias, line 195: if this is in the constructor, you avoid having two places where you do this

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+        # set additional keyword arguments
+        for (kw, val) in kwargs.items():
+            print(kw,val)
+            print(type(self.xct_solver))
+            self.xct_solver.setOption(kw,str(val))
+
+        solsfound = 0
+        while solution_limit == None or solsfound < solution_limit:
+            # call the solver, with parameters
+            my_status = self.xct_solver.runFull(self.has_objective,time_limit if time_limit is not None else 0)
```
tias, line 208: it looks like you could do one solve() and then a loop where you continue to call runFull, is that conceptually easier? if not, OK like this.
Maybe write that the clue is that you can call `invalidateLastSol()`

btw, since you have 'getObjandSOl()' you could also implement display?

best to raise NotImplementedError on anything not (yet) implemented (like the todo's below, and the display above)

tias: and the time limit and solution limit

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+                self.xct_solver.addVariable(revar,str(lb), str(ub), encoding)
+            else:
+                self.xct_solver.addVariable(revar,lb,ub, encoding)
+        else:
+            raise NotImplementedError("Not a known var {}".format(cpm_var))
+        self._varmap[cpm_var] = revar
+        return revar
+    
+
+    def solver_var_no_num(self, cpm_var, encoding="onehot"):
```
tias, line 258: can you document the purpose of this function?

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+        self += flat_cons  # add potentially created constraints
+        self.user_vars.update(get_variables(flat_obj))  # add objvars to vars
+
+        # make objective function or variable and post
+        xct_coefs,xct_vars,xct_rhs = self._make_numexpr(flat_obj,0)
+        if not self.objective_minimize:
+            xct_coefs = [-x for x in xct_coefs]
+
+        self.xct_solver.init(xct_coefs,xct_vars)
+        self.solver_is_initialized = True
```
tias, line 287: I don't think this is the right place.

Also, what if I add an objective before a constraint? or no, it will actually call add, so add will always init the solver (and set this param)

cpmpy/solvers/exact.py
tias: still an issue with the new linearize?

tias (commented):
part 1, I just need one more deeper look at the __add__ but it looks clean and direct

cpmpy/solvers/exact.py
tias: looking at the file directly instead of the diff, I see that you do explain the xct_solver in the class docstring; so please add the others there too.

I now also see that you wrote about the inits that the objective can only be initialized once, which explains why you need these other attribs, so repeating that in the class docstring too
(this resolves multiple of my comments)

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+            elif isinstance(cpm_expr, BoolVal):
+                self._add_xct_constr([], [], True, 0 if cpm_expr.args[0] else 1, False, 0)
+
+            # a direct constraint, pass to solver
+            elif isinstance(cpm_expr, DirectConstraint):
+                assert False, "Exact does not support direct constraints"
+
+            else:
+                raise NotImplementedError(cpm_expr)  # if you reach this... please report on github
+
```
tias, line 487: I think you need to return self, I think one test is failing for that (e.g. s += con; s becoming None)

tias (commented):
with the latest pip version of exact, I get 4 errors (one of which is the return self one I think):

FAILED tests/test_solveAll.py::TestSolveAll::test_solveall_no_obj - AssertionError: Items in the second set but not the first:
FAILED tests/test_solveAll.py::TestSolveAll::test_solveall_with_obj - AttributeError: 'NoneType' object has no attribute 'minimize'
FAILED tests/test_solvers.py::TestSolvers::test_false - ValueError: max() arg is an empty sequence
FAILED tests/test_solvers.py::TestSolvers::test_installed_solvers_solveAll - cppyy.gbl.std.invalid_argument: void Exact::setOption(const std::string& op...
====== 4 failed, 2988 passed, 1 skipped, 9 warnings in 215.40s (0:03:35) =======

tias (commented):
with the latest pip version of exact, I get 4 errors (one of which is the return self one I think):

FAILED tests/test_solveAll.py::TestSolveAll::test_solveall_no_obj - AssertionError: Items in the second set but not the first:
FAILED tests/test_solveAll.py::TestSolveAll::test_solveall_with_obj - AttributeError: 'NoneType' object has no attribute 'minimize'
FAILED tests/test_solvers.py::TestSolvers::test_false - ValueError: max() arg is an empty sequence
FAILED tests/test_solvers.py::TestSolvers::test_installed_solvers_solveAll - cppyy.gbl.std.invalid_argument: void Exact::setOption(const std::string& op...
====== 4 failed, 2988 passed, 1 skipped, 9 warnings in 215.40s (0:03:35) =======

tias (commented):
with the latest pip version of exact, I get 4 errors (one of which is the return self one I think):

FAILED tests/test_solveAll.py::TestSolveAll::test_solveall_no_obj - AssertionError: Items in the second set but not the first:
FAILED tests/test_solveAll.py::TestSolveAll::test_solveall_with_obj - AttributeError: 'NoneType' object has no attribute 'minimize'
FAILED tests/test_solvers.py::TestSolvers::test_false - ValueError: max() arg is an empty sequence
FAILED tests/test_solvers.py::TestSolvers::test_installed_solvers_solveAll - cppyy.gbl.std.invalid_argument: void Exact::setOption(const std::string& op...
====== 4 failed, 2988 passed, 1 skipped, 9 warnings in 215.40s (0:03:35) =======

cpmpy/solvers/exact.py
JoD: Done. DirectConstraint also works now.

examples/php.py
```diff
@@ -0,0 +1,64 @@
...
+Fit P pigeons into N holes, such that no two pigeons share the same hole.
+Relevant proof complexity problem (only has exponential resolution proofs, but has short cutting plane proofs).
+"""
+
+# load the libraries
+import numpy as np
+from cpmpy import *
+from cpmpy.solvers import CPM_ortools
+
+def php(P,H):
```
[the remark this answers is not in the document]
JoD: I've removed the example, it was a sanity check for the implementation.

tests/test_constraints.py
```diff
@@ -16,13 +15,15 @@
...
                   "pysdd": {"circuit", "element","min","max","allequal","alldifferent","cumulative"},
+                  "exact": {"circuit","max","min"},
                   }
 
 # Exclude certain operators for solvers.
 # Not all solvers support all operators in CPMpy
 EXCLUDE_OPERATORS = {"gurobi": {"mod"},
                      "pysat": {"sum", "wsum", "sub", "mod", "div", "pow", "abs", "mul","-"},
                      "pysdd": {"sum", "wsum", "sub", "mod", "div", "pow", "abs", "mul","-"},
+                     "exact": {"mod","pow","div","mul","min","max","abs"},
```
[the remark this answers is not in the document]
JoD: I've removed as many as possible.

Most work now, except for `abs`, which should become a global constraint with a simple decomposition (see #258), and for `min` and `max` in a reification, which will be fixed by the decompose globals PR.

cpmpy/solvers/exact.py
JoD: Done!

JoD: Done!

cpmpy/solvers/exact.py
JoD: - implemented NotImplementedError where appropriate
- display should work, will add a test still
- I think looping over runFull is easier. Let's keep it as is.
- time limit and solution limit should work.

cpmpy/solvers/exact.py
JoD: Done. The idea is to just have a private wrapper that checks whether the invariant holds that whatever we request is not a number. This should be the case, otherwise things will break internally. The external `solver_var` can handle numbers (e.g., for `DirectConstraint`).

If you're concerned about efficiency, I can optimize the wrapper a bit.

cpmpy/solvers/exact.py
JoD: No idea, testing it gave errors where the Exact wrapper does not yet handle bidirectional reification. 

Personally, I'm just trying to keep the wrapper as simple as possible - I'd rather have CPMpy do all the work :) So I'm fine with the current implementation.

cpmpy/solvers/exact.py
JoD: Yeah, cleared by default by `solve` and `solveAll`, seems best for now.

cpmpy/solvers/exact.py
JoD: Yes :)

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+                    raise NotImplementedError(
+                    "Unexpected condition constraint for Exact '{}' {}".format(lhs.name,cpm_expr))
+
+            # True or False
+            elif isinstance(cpm_expr, BoolVal):
+                self._add_xct_constr([], [], True, 0 if cpm_expr.args[0] else 1, False, 0)
+
+            # a direct constraint, pass to solver
+            elif isinstance(cpm_expr, DirectConstraint):
+                assert False, "Exact does not support direct constraints"
```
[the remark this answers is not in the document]
JoD: Done, with test.

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+
+            elif isinstance(cpm_expr, Operator) and cpm_expr.name == "->":
+                # Indicator constraints
+                # Take form bvar -> sum(x,y,z) >= rvar
+                cond, sub_expr = cpm_expr.args
+                assert isinstance(cond, _BoolVarImpl), f"Implication constraint {cpm_expr} must have BoolVar as lhs"
+                assert isinstance(sub_expr, Comparison), "Implication must have linear constraints on right hand side"
+
+                lhs, rhs = sub_expr.args
+                assert isinstance(lhs, _NumVarImpl) or lhs.name == "sum" or lhs.name == "wsum",f"Unknown linear expression {lhs} on right side of indicator constraint: {cpm_expr}"
```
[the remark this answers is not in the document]
JoD: Yeah, the code contains quite a bit of such asserts (guaranteed by some transformation). For me, this was a sanity check. Should I remove them again?

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+
+        # transform and post the constraints
+        for cpm_expr in self.transform(cpm_expr_orig):
+            # Comparisons: only numeric ones as 'only_bv_implies()' has removed the '==' reification for Boolean expressions
+            # numexpr `comp` bvar|const
+            if isinstance(cpm_expr, Comparison):
+                lhs, rhs = cpm_expr.args
+                xct_coefs, xct_vars, xct_rhs = self._make_numexpr(lhs,rhs)
+
+                # Thanks to `only_numexpr_equality()` only supported comparisons should remain
```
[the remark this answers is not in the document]
JoD: Fixed

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+        xrhs += rhs
+
+        if is_num(lhs):
+            xrhs -= lhs
+        elif isinstance(lhs, _NumVarImpl):
+            xcoefs += [1]
+            xvars += [self.solver_var_no_num(lhs)]
+        elif lhs.name == "sum":
+            xcoefs = [1]*len(lhs.args)
+            xvars = self.solver_vars(lhs.args)
```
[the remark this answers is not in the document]
JoD: At the points `solver_var_no_num` is called, the code cannot handle numeric values. So the assert checks it is not the case (which -if correct- is guaranteed by the previous transformations) and documents where we expect these numeric values to have been split from the variables.

`solver_var_no_num` is called in multiple places, so replacing it with an assert seems cumbersome. I can also just delete `solver_var_no_num` and not have the assert.

It seems a matter of code style more than anything, so I'm happy to change if need be.

cpmpy/solvers/exact.py
JoD: Done

cpmpy/solvers/exact.py
JoD: It's needed to initialize the objective, which may not be ideal, but I'm not going to refactor that part of Exact soon.

cpmpy/solvers/exact.py
JoD: Extended class docstring

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+            self.xct_solver.addConstraint([self.fix(x) for x in xct_coefs],xct_vars,uselb,self.fix(lb),useub,self.fix(ub))
+
+    def _add_xct_reif(self,head,xct_coefs,xct_vars,lb):
+        maximum = max([max([abs(x) for x in xct_coefs]),abs(lb)])
+        if maximum > 1e18:
+            self.xct_solver.addReification(head,[str(x) for x in xct_coefs],xct_vars,str(lb))
+        else:
+            self.xct_solver.addReification(head,[self.fix(x) for x in xct_coefs],xct_vars,self.fix(lb))
+
+    def _add_xct_reif_right(self,head, xct_coefs,xct_vars,xct_rhs):
```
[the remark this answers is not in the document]
JoD: Yeah, I'm happy to let CPMpy do the work, as the plain reification will immediately be split in too anyway by Exact.

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,492 @@
...
+                assert isinstance(lhs, _NumVarImpl) or lhs.name == "sum" or lhs.name == "wsum",f"Unknown linear expression {lhs} on right side of indicator constraint: {cpm_expr}"
+
+                xct_coefs, xct_vars, xct_rhs = self._make_numexpr(lhs,rhs)
+
+                if isinstance(cond, NegBoolView):
+                    cond, bool_val = self.solver_var_no_num(cond._bv), False
+                else:
+                    cond, bool_val = self.solver_var_no_num(cond), True
+
+                if sub_expr.name == "==":
```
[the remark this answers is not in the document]
JoD: Added the following comment:
```
        # NOTE: the transformations that are still done specifically for Exact are two-fold:
        # 1) transform '==' and '<=' to '>='
        # 2) transform implications with negative conditions to ones with positive consequences
        #
        # 1) seems quite general and is a candidate to function as an independent transformation.
        # 2) seems very solver-specific.
```
That seems enough to remember what's going on for now. We can make `1)` a full-fledged transformation when other systems seem to need it too.

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,540 @@
...
+        while solution_limit == None or solsfound < solution_limit:
+            # call the solver, with parameters
+            my_status = self.xct_solver.runFull(False,time_limit if time_limit is not None else 0)
+            assert my_status in [0,1,2,3], "Unexpected status code for Exact."
+            if my_status == 0: # found unsatisfiability
+                break
+            elif my_status == 1: # found solution, but not optimality proven
+                assert self.xct_solver.hasSolution()
+                solsfound += 1
+                self._fillObjAndVars()
```
tias, line 221: these two can go inside the 'if display is not None'

cpmpy/solvers/exact.py
tias: ok

cpmpy/solvers/exact.py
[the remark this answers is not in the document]
tias: explicit can be good, it just looked redundant

cpmpy/solvers/exact.py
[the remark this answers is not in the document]
tias: OK

tias (commented):
the blocking issue must be from the decomposition of 'Element', but I don't really see where it could come from just looking at the code.

JoD:
Ok, all tests pass now. @tias @IgnaceBleukx @Wout4 if one of you can have a last look, it may be ready.

cpmpy/solvers/exact.py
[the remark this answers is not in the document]
JoD: Ok, I guess the assert has outlived its usefulness (which was mainly during development anyway). I've removed it, it's now calling `_solver_var` like a normal interface would :)

cpmpy/solvers/exact.py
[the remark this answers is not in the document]
JoD: No prob, removed this one.

cpmpy/solvers/exact.py
JoD: Ah, indeed :) Moved `_fillObjAndVars()`, but kept `solsfound+=1`. I can imagine someone wanting to both run a callback and know the total number of solutions.

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+    - self.objective_given: whether an objective function is given to xct_solver
+    - self.objective_minimize: the direction of the optimization (if false then maximize)
+    as Exact can only minimize
+    """
+
+    @staticmethod
+    def supported():
+        # try to import the package
+        try:
+            import exact
```
IgnaceBleukx, line 60: Should we check for version here? 
People might have an older version of the Python API? (Which was the case for me ;) )

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+            self.xct_solver.setOption("verbosity","0")
+        self.xct_solver.clearAssumptions()
+
+        # set additional keyword arguments
+        for (kw, val) in kwargs.items():
+            self.xct_solver.setOption(kw,str(val))
+
+        # set assumptions
+        if assumptions is not None:
+            assert(all(v.is_bool() for v in assumptions))
```
IgnaceBleukx, line 146: This will fail if someone gives a non-CPMpy-expression as assumption, maybe add a clearer error message?

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+            else:
+                self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+        elif my_status == 1: # found solution, but not optimality proven
+            assert self.xct_solver.hasSolution()
+            self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+        elif my_status == 2: # found inconsistency over assumptions
+            self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+        elif my_status == 3: # found timeout
+            self.cpm_status.exitstatus = ExitStatus.UNKNOWN
+            return 0
```
IgnaceBleukx, line 175: Should return `False`

IgnaceBleukx: But ideally, use `self._solve_return(self.status)` which will ensure consistency between solvers as it is a function in the superclass `SolverInterface`

IgnaceBleukx: See: https://github.com/CPMpy/cpmpy/blob/d53809aa4eabea0808d2842975dfb876e1bdf14d/cpmpy/solvers/ortools.py#L187

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+        else:
+            raise NotImplementedError(my_status)  # a new status type was introduced, please report on github
+
+        return self._fillObjAndVars()
+
+    def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from_model=False, **kwargs):
+        """
+            Compute all solutions and optionally display the solutions.
+
+            This is the generic implementation, solvers can overwrite this with
```
IgnaceBleukx, line 185: But its no the generic implementation :)

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+        """
+            Returns whether the solver has an objective function or not.
+        """
+        return self.objective_given
+
+
+    def solver_var(self, cpm_var, encoding="onehot"):
+        """
+            Creates solver variable for cpmpy variable
+            or returns from cache if previously created
```
IgnaceBleukx, line 247: Maybe also document the non-default `encoding` parameter?

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+        # return it if it already exists
+        if cpm_var in self._varmap:
+            return self._varmap[cpm_var]
+
+        # create if it does not exist
+        revar = str(cpm_var)
+        if isinstance(cpm_var, _BoolVarImpl):
+            self.xct_solver.addVariable(revar,0,1)
+        elif isinstance(cpm_var, _IntVarImpl):
+            lb = cpm_var.lb
```
IgnaceBleukx, line 264: code suggestion: `lb, ub = cpm_var.get_bounds()`

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+    def solver_var(self, cpm_var, encoding="onehot"):
+        """
+            Creates solver variable for cpmpy variable
+            or returns from cache if previously created
+        """
+        if is_num(cpm_var):  # shortcut, eases posting constraints
+            return cpm_var
+
+        # variables to be translated should be positive
+        assert(not isinstance(cpm_var, NegBoolView))
```
IgnaceBleukx, line 253: Add error message after assert?

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+        # make objective function or variable and post
+        xct_coefs,xct_vars,xct_rhs = self._make_numexpr(flat_obj,0)
+        if not self.objective_minimize:
+            xct_coefs = [-x for x in xct_coefs]
+
+        self.xct_solver.init(xct_coefs,xct_vars)
+        self.solver_is_initialized = True
+        self.xct_solver.setOption("verbosity","0")
+
+    def _make_numexpr(self, lhs, rhs):
```
IgnaceBleukx, line 306: I would like this function to have same heading as other solver interfaces.
You can easily extract the lhs and rhs from the args of a comparison

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+        cpm_cons = only_bv_implies(cpm_cons)  # anything that can create full reif should go above...
+        cpm_cons = linearize_constraint(cpm_cons, supported=frozenset({"sum","wsum"}))  # the core of the MIP-linearization
+        cpm_cons = only_positive_bv(cpm_cons)  # after linearisation, rewrite ~bv into 1-bv
+        return cpm_cons
+
+        # NOTE: the transformations that are still done specifically for Exact are two-fold:
+        # 1) transform '==' and '<=' to '>='
+        # 2) transform implications with negative conditions to ones with positive consequences
+        #
+        # 1) seems quite general and is a candidate to function as an independent transformation.
```
IgnaceBleukx, line 365: I agree, not sure, But I think the `Scip` solver actually also only accepts `<=` and `>=`

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+        # 2) seems very solver-specific.
```
IgnaceBleukx, line 366: also agree : )

cpmpy/solvers/exact.py
IgnaceBleukx: Ah although the return type is also different anyway... Ok maybe its fine to leave it as is then.

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+
+    @staticmethod
+    def fix(o):
+        return o.item() if isinstance(o, np.generic) else o
+
+    def _add_xct_constr(self, xct_coefs,xct_vars,uselb,lb,useub,ub):
+        if any([not isinstance(x, numbers.Integral) for x in xct_coefs+[lb,ub]]):
```
IgnaceBleukx, line 373: you can drop the `[]` in the `any`,  it should with generators as well.

IgnaceBleukx: Same for `max` below

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+            Any CPMpy expression given is immediately transformed (through `transform()`)
+            and then posted to the solver in this function.
+
+            This can raise 'NotImplementedError' for any constraint not supported after transformation
+
+            The variables used in expressions given to add are stored as 'user variables'. Those are the only ones
+            the user knows and cares about (and will be populated with a value after solve). All other variables
+            are auxiliary variables created by transformations.
+
+        :param cpm_expr: CPMpy expression, or list thereof
```
IgnaceBleukx, line 418: parameter name is `cpm_expr_orig`

IgnaceBleukx (commented):
Nice! I too am wondering why the `init` of the solver does not happen in the constructor of the CPmpy interface? It seems cumbersome to do it at 3 different places as is the case now?

cpmpy/solvers/exact.py
JoD: ok

cpmpy/solvers/exact.py
JoD: done

cpmpy/solvers/exact.py
JoD: good to know, let's revisit that once we get to SCIP.

cpmpy/solvers/exact.py
JoD: Added a comment describing the remaining transformations. We can refactor those later based on information by other solvers.

cpmpy/solvers/exact.py
JoD: removed the wrapper in the end.

cpmpy/solvers/exact.py
```diff
@@ -0,0 +1,528 @@
...
+        self.solver_is_initialized = False
+        self.objective_given = False
+        self.objective_minimize = True
+
+        # initialise everything else and post the constraints/objective
+        super().__init__(name="exact", cpm_model=cpm_model)
+
+    def _fillObjAndVars(self):
+        if not self.xct_solver.hasSolution():
+            self.objective_value_ = None
```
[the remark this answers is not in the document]
JoD: done

cpmpy/solvers/exact.py
JoD: Added an error message, is this what you aimed for?

cpmpy/solvers/exact.py
JoD: done

cpmpy/solvers/exact.py
JoD: done

cpmpy/solvers/exact.py
JoD: done

cpmpy/solvers/exact.py
JoD: ok

cpmpy/solvers/exact.py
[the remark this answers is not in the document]
JoD: done

cpmpy/solvers/exact.py
JoD: I've moved the encoding to a class attribute for Exact, and made "log" default. And checked that this actually works :)

Later, I'd like to make this attribute changeable from the interface, but there was no obvious way now. Let's discuss this later.

JoD:
Running Exact on bigtest gave a couple of issues. The ones on Exact's side seem fixed with the latest commits. The ones that seem to have to do with the transformations are here:  #359, #362, #363

There are no known issues, I'm happy to merge this branch now.

PR #221, opened 2023-02-03
183 revisit linearize transformation

IgnaceBleukx:
First version of the new linearization transformation.

Work in progress and some tests are still failing! Mainly due to the `!=` linearization.
For this operator, we can either do the traditional big-M implemention or use indicator constraints.
For the latter one, we risk introducing unsupported constraints depending on lhs and rhs of original `!=` constraint...
To be discussed.


cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
-            if isinstance(rhs, _NumVarImpl):
-                # Vars on rhs of linexpr not supported in indicator constraints
-                if isinstance(lhs, _NumVarImpl):
-                    new_lhs = lhs + (-1) * rhs
-                elif lhs.name == "sum":
-                    new_lhs = Operator("wsum", [[1]*len(lhs.args) + [-1], lhs.args + [rhs]])
-                elif lhs.name == "wsum":
-                    new_lhs = lhs + (-1 * rhs)
+        if isinstance(cond, _BoolVarImpl) and isinstance(sub_expr, _BoolVarImpl):
+            # shortcut for BV -> BV exprs
```
tias, line 82: this can be more uniform/explicit, that BV1 -> BV2 equals ~BV1 | BV2 equals ((1-BV1) + BV2 >= 1) where in case of BV being a negboolview an additional NBV=(1-BV) is applied.
but then I guess you purposefully already group the consts? Can be a bit more expl in docs still : )

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
-        if len(lhs.args) == 2:
-            # multiplication of var and constant
-            if _wsum_should(lhs):
-                return [Comparison(cpm_expr.name, Operator("wsum", _wsum_make(lhs)), rhs)]
-            return [cpm_expr]
+        # BV -> LinExpr
+        if isinstance(cond, _BoolVarImpl):
+            lin_sub = linearize_constraint(sub_expr, supported=supported)
+            return [cond.implies(lin) for lin in lin_sub]
+        # # LinExpr -> BV Updated: unsupported by new grammar
```
tias, line 99: it will not appear in the input grammar?
or it can but then wont be supported? Because equiv to ~bv -> ~linexpr, which we could do somewhere

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
-            var, cons = get_or_make_var(var * arg)
-            constraints += cons
 
-        return constraints + [Comparison(cpm_expr.name, var, rhs)]
+        # linearize unsupported operators
+        if lhs.name not in supported: # TODO: add mul, (abs?), (mod?), (pow?)
+            if isinstance(lhs, _NumVarImpl) and is_num(rhs):
+                pass
+            elif isinstance(lhs, _NumVarImpl) and isinstance(rhs, _NumVarImpl):
+                lhs, rhs = lhs + -1 * rhs, 0  # bring numvar to lhs
```
tias, line 114: use 2 lines please : )

cpmpy/transformations/linearize.py
```diff
@@ -247,185 +219,52 @@ def only_positive_bv(cpm_expr):
...
-
-
-    if isinstance(cpm_expr, Operator) and cpm_expr.name == "->":
+            if isinstance(lhs,NegBoolView):
+                lhs, rhs = -lhs._bv, rhs - 1
+
+        elif lhs.name == "sum":
+            weights = [-1 if isinstance(a, NegBoolView) else 1 for a in lhs.args]
+            lhs = sum(w * a._bv if isinstance(a, NegBoolView) else w * a for w,a in zip(weights,lhs.args))
+            rhs -= sum(w < 0 for w in weights)
```
tias, line 233: a bit obscure... isn't a generic function possible that gets w_lhs, w_exp, rhs as input and outputs the same but without negboolviews?

cpmpy/solvers/gurobi.py
```diff
@@ -266,7 +266,7 @@ def transform(self, cpm_expr):
         cpm_cons = flatten_constraint(cpm_expr)  # flat normal form
         cpm_cons = reify_rewrite(cpm_cons, supported=frozenset(['sum', 'wsum']))  # constraints that support reification
         cpm_cons = only_bv_implies(cpm_cons)  # anything that can create full reif should go above...
-        cpm_cons = linearize_constraint(cpm_cons)  # the core of the MIP-linearization
+        cpm_cons = linearize_constraint(cpm_cons, supported=frozenset({"sum", "wsum", "sub", "min", "max","mul", "abs","pow","div"}))  # the core of the MIP-linearization
```
JoD, line 269: - I cannot see a situation where keeping "sub" is better than just translating it to a "wsum". I'd drop it.
- We will also need a separate transformation for "min", "max", "abs".
- Could "mod" be missing?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
-- Boolvar -> LinExpr == Constant
-- Boolvar -> LinExpr >= Constant
-- Boolvar -> LinExpr <= Constant
+Where BoolVar is a boolean variable or its negation.
 
+General comparisons or expressions
+-----------------------------------
+- GenExpr                               (GenExpr.name in supported, GenExpr.is_bool())
+- GenExpr == Var/Constant               (GenExpr.name in supported)
+- GenExpr <= Var/Constant               (GenExpr.name in supported)
```
JoD, line 35: `GenExpr.is_num()`

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
-        return [cpm_expr >= 1]
+            return [sum([cpm_expr._bv]) <= 0]
+        return [sum([cpm_expr]) >= 1]
 
+    # conjunction
     if cpm_expr.name == "and":
-        if all(arg.is_bool() for arg in cpm_expr.args):
-            return [sum(cpm_expr.args) >= len(cpm_expr.args)]
-        raise Exception("Numeric constants or numeric variables not allowed as base constraint")
+        return [sum(cpm_expr.args) >= len(cpm_expr.args)]
```
JoD, line 70: Ideally, no negated variables are left in these sums. E.g., the negation is removed for `# boolvar`. Here it is not. I'd vote to remove it at this point (it is typically only a matter of changing signs and rhs), as it would fit more closely with a true linearization.

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
-                    new_lhs = lhs + (-1) * rhs
-                elif lhs.name == "sum":
-                    new_lhs = Operator("wsum", [[1]*len(lhs.args) + [-1], lhs.args + [rhs]])
-                elif lhs.name == "wsum":
-                    new_lhs = lhs + (-1 * rhs)
+        if isinstance(cond, _BoolVarImpl) and isinstance(sub_expr, _BoolVarImpl):
+            # shortcut for BV -> BV exprs
+            if isinstance(cond, NegBoolView):
+                if isinstance(sub_expr, NegBoolView):
+                    return [cond + -1 *sub_expr >= 0] # ~BV -> ~BV
```
JoD, line 85: Will this yield a weighted sum or a sum with a multiplication?

cpmpy/transformations/linearize.py
JoD: Yeah, it can happen and it should be supported. In general, it requires new variables though.

E.g. `x==y -> p` as input should lead to `aux1 or aux2 or p`, `aux1 -> x>y`, `~aux1 -> x<=y`, `aux2 -> x<y`, `~aux2 -> x>=y`, which can all be translated with previous methods.

I am not sure what "new grammar" refers to here. The new grammar we discussed on https://github.com/CPMpy/cpmpy/issues/183#issuecomment-1412363735 is the grammar that linearize outputs, so it does not restrict the input.

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
 
-        return constraints + [Comparison(cpm_expr.name, var, rhs)]
+        # linearize unsupported operators
+        if lhs.name not in supported: # TODO: add mul, (abs?), (mod?), (pow?)
+            if isinstance(lhs, _NumVarImpl) and is_num(rhs):
+                pass
+            elif isinstance(lhs, _NumVarImpl) and isinstance(rhs, _NumVarImpl):
+                lhs, rhs = lhs + -1 * rhs, 0  # bring numvar to lhs
+
+            elif lhs.name == "sub":
```
JoD, line 116: Aha, here `sub` is linearized, so it should be erased from `supported`.

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
+                constraints = [sum(sigma) == 1]
+                constraints += [sum(np.arange(n) * sigma) == idx]
+                # translation with implication:
+                constraints += [s.implies(Comparison(cpm_expr.name, a, cpm_expr.args[1])) for s, a in zip(sigma, arr)]
+                return linearize_constraint(constraints, supported=supported)
+
+            # other global constraints
+            elif isinstance(lhs, GlobalConstraint) and hasattr(lhs, "decompose_comparison"):
+                decomp = lhs.decompose_comparison(cpm_expr.name, rhs)
+                return linearize_constraint(flatten_constraint(decomp),supported=supported)
```
JoD, line 137: Do we not assume everything unsupported is already flattened? Otherwise, could, e.g., a `sum` expression also have nested terms?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,183 @@
...
+
+        # now fix the comparisons themselves
+        if cpm_expr.name == "<":
+            new_rhs, cons = get_or_make_var(rhs - 1)
+            return [lhs <= new_rhs] + cons
+        if cpm_expr.name == ">":
+            new_rhs, cons = get_or_make_var(rhs + 1)
+            return [lhs >= new_rhs] + cons
+        if cpm_expr.name == "!=":
+            # Big-M implementation, ensure no new reifications are added
```
JoD, line 168: - a new (smart) reification *is* added, see `z=boolvar()`.
- we count on solvers to do their own Big-M by allowing the `BoolVar -> LinExpr` output. We do this because we hope they can do it more efficiently. The same assumption should hold here, so we should just translate to that type of constraint.
- so just add `z -> linearize(lhs > rhs)` and `~z -> linearize(lhs < rhs)`?

cpmpy/transformations/linearize.py
```diff
@@ -247,185 +219,52 @@ def only_positive_bv(cpm_expr):
...
-        if isinstance(lhs, Operator) and (lhs.name == "sum" or lhs.name == "wsum"):
-
-            nn_subsexpr = only_positive_bv(subexpr)
-            return linearize_constraint([cond.implies(nn_expr) for nn_expr in nn_subsexpr])
-        else:
-            raise NotImplementedError(f"Operator {lhs} is not supported on left right hand side of implication in {cpm_expr}")
+        if isinstance(cond, _BoolVarImpl): # BV -> Expr
+            subexpr = only_positive_bv(subexpr)
+            return[cond.implies(expr) for expr in subexpr]
+        # else: # Expr -> BV Update: Unsupported by new grammar
```
JoD, line 257: Here this is ok :)

cpmpy/transformations/linearize.py
JoD: I see that below, the `only_positive_bv` takes care of this. Good. But then we do not need to do it above either.

JoD:
Looks like it will be a nice improvement!

If we want to be really thorough, we could add a check that tests whether the output adheres to the predefined grammar, and run it during testing. We can use that same method to check the input of other transformations, e.g., `only_positive_bv`, and be more confident that the code does what it is supposed to.

cpmpy/transformations/linearize.py
JoD: Also, this linearization will not work when the `LinExpr` is nested, e.g., `p -> x!=y`. Here we do need both reifications.

cpmpy/solvers/gurobi.py
IgnaceBleukx: Gurobi supports `sub` in its API (see `_make_numexpr`) above so I thought it would be best to let Gurobi handle it themselves :)

JoD: This is  gurobi-specific. If `sub` is not passed in `supported`, it will be transformed anyway.

JoD: So the rationale behind `supported` is that anything that is natively supported by the solver. `mod` is not, so linearize will complain, which is ok.

cpmpy/transformations/linearize.py
JoD: Conclusion after discussion: current approach seems fine, except that we want to keep the newly introduced reification when not nested (as there is no risk of infinite recursive reifications).

IgnaceBleukx:
Updated some more, I think it is ready to put it into master.
Bigtest is failing on this branch as there is a model `football16650495281380026` that has a constraint `~bv >= False` which should not be possible to write after our discussion on using booleans as integers.

tias:
if you typecheck/throw a 'notallowed' error somewhere at expression level on x >= False, then the bigtest test will skip : )

Wout4:
i thought this is a case where we can just transform it into x >= 0?

tias:
hmm, you are right, we agreed that Booleans can be used as integer, so bv > False and iv > False should both be possible, its iv & bv that is not allowed. This will be for a new 'simplify_constants' or so transformation, it might become a bit tricky/ugly for Boolean constants used as integer... e.g. 1+BoolVal(False) >= boolvar...

IgnaceBleukx:
I'll start by implenting the new transformation removing BoolVals from integere experessions, than this issue is solved as well.

cpmpy/solvers/gurobi.py
```diff
@@ -265,12 +265,12 @@ def transform(self, cpm_expr):
...
         # expressions have to be linearized to fit in MIP model. See /transformations/linearize
         cpm_cons = flatten_constraint(cpm_expr)  # flat normal form
-        cpm_cons = decompose_global(cpm_cons, supported={"min","max", "element", "alldifferent"}) #element and alldiff have specialized MIP decomp in linearize
+        cpm_cons = decompose_global(cpm_cons, supported={"min","max","alldifferent"}) # alldiff has specialized MIP decomp in linearize
         cpm_cons = reify_rewrite(cpm_cons, supported=frozenset(['sum', 'wsum']))  # constraints that support reification
         cpm_cons = only_bv_implies(cpm_cons)  # anything that can create full reif should go above...
-        cpm_cons = linearize_constraint(cpm_cons)  # the core of the MIP-linearization
         cpm_cons = only_numexpr_equality(cpm_cons, supported=frozenset(["sum", "wsum", "sub"]))  # supports >, <, !=
-        cpm_cons = only_positive_bv(cpm_cons)  # after linearisation, rewrite ~bv into 1-bv
+        cpm_cons = linearize_constraint(cpm_cons, supported=frozenset({"sum", "wsum","sub","min","max","mul","abs","pow","div"}))  # the core of the MIP-linearization
```
JoD, line 272: So `wsum, sum, sub, min, max, mul, div, abs` etc. remain after linearize? Is there a transformation that turns all of these into `wsum`s?

cpmpy/solvers/gurobi.py
```diff
@@ -265,12 +265,12 @@ def transform(self, cpm_expr):
...
+        cpm_cons = only_positive_bv(cpm_cons)  # after linearization, rewrite ~bv into 1-bv
```
JoD, line 273: Can we have `5*~bv` as input here, so that it is transformed to `5*(1-bv)`? How do we avoid introducing nested expressions here?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
-        lin_comps = linearize_constraint(sub_expr)
-        lin_exprs = []
-        for l_expr in lin_comps:
-            lhs, rhs = l_expr.args
-            if lhs.name == "mul" and _wsum_should(lhs):
-                # multiplication with var and constant, convert to wsum
-                lhs = Operator("wsum", _wsum_make(lhs))
+        # xor
+        elif cpm_expr.name == "xor" and len(cpm_expr.args) == 2:
+            return [sum(cpm_expr.args) >= 1]
```
JoD, line 83: `return [sum(cpm_expr.args) == 1]` ?

JoD: Note: for `xor` of arbitrary length, this works too if we allow an auxiliary variable, e.g., `xor(p,q,r,s,t)` becomes `p+q+r+s+t == 1+2*auxiliary`.

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+        # reification
+        elif cpm_expr.name == "->":
+            # determine direction of implication
+            cond, sub_expr = cpm_expr.args
+            assert isinstance(cond, _BoolVarImpl), f"Linearization of {cpm_expr} is not supported, lhs of implication must be boolvar. Apply `only_bv_implies` before calling `linearize_constraint`"
 
-        return [cond.implies(l_expr) for l_expr in lin_exprs]
+            if isinstance(cond, _BoolVarImpl) and isinstance(sub_expr, _BoolVarImpl):
+                # shortcut for BV -> BV, convert to disjunction and apply linearize on it
+                return linearize_constraint(~cond | sub_expr, reified=reified)
```
JoD, line 93: What if `cond` is a negation, e.g., `~p`? Do we introduce double negation then? Can `only_positive_bv` fix this?

In general, I'd expect to not introduce more negations if not needed. Perhaps this could become `linearize_constraint(sub_expr >= cond)`?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
-            constraints += cons
 
-        return constraints + [Comparison(cpm_expr.name, var, rhs)]
+        if isinstance(lhs, _NumVarImpl) and isinstance(rhs, _NumVarImpl):
+            # bring numvar to lhs
+            lhs = 1 * lhs + -1 * rhs
+            rhs = 0
+
+        # linearize unsupported operators
+        elif isinstance(lhs, Operator) and lhs.name not in supported: # TODO: add mul, (abs?), (mod?), (pow?)
```
JoD, line 110: mul, div, mod, pow are not linearizable unless in special circumstances (e.g., div/mul by constant). `abs` will be fixed as a global constraint.

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+
+            if lhs.name == "sub":
+                # convert to wsum
+                lhs = sum([1 * lhs.args[0] + -1 * lhs.args[1]])
```
JoD, line 114: I do not understand why sometimes we translate to a `+` over `*`, sometimes a `sum` sometimes a `wsum`. Should it not mostly be a `wsum`, and exceptionally a `sum` (if all weights are exactly 1)?

In this case, the two terms probably need to be split, i.e., `lhs = sum([1 * lhs.args[0] , -1 * lhs.args[1]])` (`,` instead of `+`).

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+
+            elif lhs.name == "mul" and is_num(lhs.args[0]):
+                lhs = Operator("wsum",[[lhs.args[0]], [lhs.args[1]]])
+            else:
+                raise TransformationNotImplementedError(f"lhs of constraint {cpm_expr} cannot be linearized, should be any of {supported | set(['sub'])} but is {lhs}. Please report on github")
```
JoD, line 119: There may be more cases we can cover, e.g., `3*x*5` can be linearized.

cpmpy/solvers/gurobi.py
IgnaceBleukx: For gurobi these indeed remain in the expressions.
`sub, min, max` (and soon `abs`)  will be linearized if they are not part of the supported set.
`mul` gets linearized to a `wsum` of one argument if the expression consists of a constant and a variable, otherwise we raise an error as we cannot linearize `var * var`.
For the `div` case, I am not sure how to linearize floordivision? If we can do so, we can cover the same case as mul. I.e., a `var / const`

cpmpy/transformations/linearize.py
JoD: Here's what I would do with `x mod y` and `x div y`: introduce the equality `x = y*[x div y] + [x mod y]` for auxiliary symbols `[x div y]` and `[x mod y]` (with appropriate bounds). Then this will simplify to the `*` and `+` operators, which in linearize will just check that `y` is a constant, which is what we want.

cpmpy/transformations/linearize.py
JoD: And even better, make it a `wsum`: `lhs = wsum([1, -1], [lhs.args[0], lhs.args[1]])` ?

cpmpy/solvers/gurobi.py
IgnaceBleukx: Anything that comes out of linearize is either a wsum, sum or some supported expression/global constraint.
For sum and wsum, it is trivial to put all of the constants on the rhs while eliminating `~bv`.
For the other expressions, we loop over the arguments in the expression and replace `~bv` with a new variable `bvn` and add the constraint `bvn == 1-bv`.
Aha but we still have to linearize this new constraint... will at that!
For global constraints we do the same thing, although I haven't found a MIP solver yet supporting global constraints so it hasn't come up yet :)

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+                for i, (w, arg) in enumerate(zip(*lhs.args)):
+                    if is_num(arg):
+                        rhs -= w * arg
+                    else:
+                        new_weights.append(w)
+                        new_args.append(arg)
+                lhs = Operator("wsum",[new_weights, new_args])
+
+
+        if isinstance(lhs, Operator) and lhs.name == "mul" and is_num(lhs.args[0]):
```
JoD, line 153: what if there are more than 2 multiplication arguments? Should we throw here?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+
+            elif lhs.name == "mul" and is_num(lhs.args[0]):
+                lhs = Operator("wsum",[[lhs.args[0]], [lhs.args[1]]])
+            else:
+                raise TransformationNotImplementedError(f"lhs of constraint {cpm_expr} cannot be linearized, should be any of {supported | set(['sub'])} but is {lhs}. Please report on github")
+
+        elif isinstance(lhs, GlobalConstraint) and lhs.name not in supported:
+            raise ValueError("Linearization of `lhs` not supported, run `cpmpy.transformations.decompose_global.decompose_global() first")
+
+        if is_num(lhs) or (isinstance(lhs, Operator) and lhs.name in {"sum","wsum"}):
```
JoD, line 124: What if `lhs` is a numeric variable? Is this case covered by `is_num(lhs)`?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+            # bring all vars to lhs
+            if isinstance(rhs, _NumVarImpl):
+                if isinstance(lhs, Operator) and lhs.name == "sum":
```
JoD, line 127: what if `lhs.name == "wsum"`?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+            elif lhs.name == "wsum":
+                new_weights, new_args = [],[]
+                for i, (w, arg) in enumerate(zip(*lhs.args)):
+                    if is_num(arg):
+                        rhs -= w * arg
+                    else:
+                        new_weights.append(w)
+                        new_args.append(arg)
+                lhs = Operator("wsum",[new_weights, new_args])
+
```
JoD, line 151: How about the case where `is_num(lhs)` ?

cpmpy/transformations/linearize.py
IgnaceBleukx: If cond is a negated boolean variable, we'll negate it and it will become a positive variable (in the constructor).
If cond is a positive boolvar, it will become negative and be removed by only_positive_bv

cpmpy/transformations/linearize.py
IgnaceBleukx: yes

cpmpy/transformations/linearize.py
IgnaceBleukx: As mentioned offline: the `sum` constructor checks whether its arguments are multiplications of const + var and translates to `wsum` if possible

cpmpy/transformations/linearize.py
IgnaceBleukx: Indeed, but we should simplify this to `15*x` using either some simplifcation transformation of at construction time of the `mul` operator

cpmpy/transformations/linearize.py
IgnaceBleukx: Indeed, we should check for that!

cpmpy/transformations/linearize.py
JoD: First is good. However, it would be good to have the invariant that linearize does not introduce negations: we risk getting loops, because `only_positive_bv` calls linearize again on auxiliary constraints. I know the loop does not occur in practice, but it will definitely not occur of we have this invariant.

Also, we should avoid making many auxiliary variables (which is done by `only_positive_bv`).

So let's go with the above suggestion?

cpmpy/solvers/gurobi.py
JoD: Hmm, it is correct, but I don't quite like the `bvn == 1-bv`. That's introducing a new variable for no good reason... 

So we get `3*~x >= ~y` to become `[3*xn >= yn, xn + x == 1, yn + y == 1]`, which is 5 constraints in Exact instead of just one: `y-3x >= 2`. 

The natural solution would be to have `3*~x >= ~y` become `3*(1-x) >= 1-y`, apply distributivity, then normalize during linearization. This is tricky, because we introduce nesting where we want to keep things flattened.

This would actually be doable if everything was a weighted sum: `ws([3],[~x]) >= ws([1],[~y])` can easily be transformed to `ws([3,-3],[1,x]) >= ws([1,-1],[1,y])` after which normalization (moving all constants to the right and all non-constants to the left) fixes the problem.

So here's my proposal: turn every multiplication/regular sum to a weighted sum, run `only_positive_bv` then `linearize_constraint`.

Alternatively, make `only_positive_bv` generate weighted sums, split the normalization from linearization, linearize -> only_positive_bv -> normalize.

cpmpy/transformations/linearize.py
JoD: So, TODO for later? Ok if I make an issue?

cpmpy/transformations/linearize.py
JoD: ok!

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+            rhs = 0
+
+        # linearize unsupported operators
+        elif isinstance(lhs, Operator) and lhs.name not in supported: # TODO: add mul, (abs?), (mod?), (pow?)
+
+            if lhs.name == "sub":
+                # convert to wsum
+                lhs = sum([1 * lhs.args[0] + -1 * lhs.args[1]])
+
+            elif lhs.name == "mul" and is_num(lhs.args[0]):
```
JoD, line 116: Add `and len(lhs.args)==2` ?

Can we have multiplications with one argument?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+        if cpm_expr.name == ">":
+            new_rhs, cons = get_or_make_var(rhs + 1) # if rhs is constant, will return new constant
+            return [lhs >= new_rhs] + cons
+        if cpm_expr.name == "!=":
+            # Special case: BV != BV
+            if isinstance(lhs, _BoolVarImpl) and isinstance(rhs, _BoolVarImpl):
+                return [lhs + rhs == 1]
+
+            if reified or (isinstance(lhs, (Operator, GlobalConstraint)) and lhs.name not in {"sum","wsum"}):
+                # lhs is sum/wsum and rhs is contant OR
```
JoD, line 170: typo: `conStant`

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
-        elif lhs.is_bool() and not isinstance(lhs, _BoolVarImpl) and isinstance(rhs,_NumVarImpl):
-            #boolexpr == intvar, so introduce an intermediate boolvar bv, boolexpr == bv, bv == intvar
-            bv = boolvar()
-            c1 = Comparison('==', lhs, bv)
-            c2 = Comparison('==', bv, rhs)
-            return linearize_constraint(flatten_constraint([c1, c2]))
-
-    if cpm_expr.name == "xor" and len(cpm_expr.args) == 2:
-        return [sum(cpm_expr.args) == 1]
+                # introduce new indicator constraints
```
JoD, line 182: This code seems unreachable. Missing `else:` or one less indentation?

JoD: Also, if tests did not catch this, please add :D

cpmpy/transformations/linearize.py
```diff
@@ -203,40 +200,16 @@ def linearize_constraint(cpm_expr):
         # Linear decomposition of alldifferent using bipartite matching
         sigma = boolvar(shape=(len(cpm_expr.args), 1 + ub - lb))
 
-        constraints = [sum(row) == 1 for row in sigma]  # Exactly one value
-        constraints += [sum(col) <= 1 for col in sigma.T]  # All diff values
+        constraints = [sum(row) == 1 for row in sigma]  # Each var has exactly one value
+        constraints += [sum(col) <= 1 for col in sigma.T]  # Each value is assigned to at most 1 variable
```
JoD, line 205: In ManyWorlds, I have both: the at-most-one for non-Integer variables (string variables, e.g., for pigeonhole and sudoku) and the binary one for integer variables (I assume integer variables have a large domain. Use string variables if not (e.g., for sudoku I use the strings `"1", "2"` etc.)).

What I would do here is to check the size of the domain. If it is larger than quadratic the number of variables, go for the quadratic encoding. Otherwise, use the at-most-one?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+                # lhs is sum/wsum and rhs is contant OR
+                # lhs is GenExpr and rhs is constant or var
+                #  ... what requires less new variables?
+                # Big M implementation
+                z = boolvar()
+                # Calculate bounds of M = |lhs - rhs| + 1
+                _, M = (abs(lhs - rhs) + 1).get_bounds()
+
+                cons = [lhs - M*z <= rhs-1, lhs - M*z >= rhs-M+1]
+                return linearize_constraint(flatten_constraint(cons), supported=supported, reified=reified)
```
JoD, line 179: Why `flatten_constraint`?

cpmpy/transformations/linearize.py
```diff
@@ -5,190 +5,187 @@
...
+            if reified or (isinstance(lhs, (Operator, GlobalConstraint)) and lhs.name not in {"sum","wsum"}):
+                # lhs is sum/wsum and rhs is contant OR
+                # lhs is GenExpr and rhs is constant or var
+                #  ... what requires less new variables?
+                # Big M implementation
+                z = boolvar()
+                # Calculate bounds of M = |lhs - rhs| + 1
+                _, M = (abs(lhs - rhs) + 1).get_bounds()
+
+                cons = [lhs - M*z <= rhs-1, lhs - M*z >= rhs-M+1]
```
JoD, line 178: Took a while to understand, but I think it is correct. Please add the following:
```
M is chosen so that 
lhs - rhs + 1 =< M*z
rhs - lhs + 1 =< M*~z
holds.
```

tests/test_trans_linearize.py
```diff
@@ -121,11 +166,11 @@ def test_others(self):
...
 
         cons = cp.max([a,b,c,5]) <= rhs
-        cons = only_var_lhs(cons)[0]
+        cons = linearize_constraint(cons)[0]
         self.assertEqual("(max(a,b,c,5)) <= (r)", str(cons))
 
         cons = cp.AllDifferent([a, b, c])
-        cons = only_var_lhs(cons)[0]
+        cons = linearize_constraint(cons)[0]
         self.assertEqual("alldifferent(a,b,c)", str(cons))
```
JoD, line 176: Add a case `p.implies(x-2y+3z != ~w)`? I'm mainly interested in checking the nested `!=`...

cpmpy/transformations/linearize.py
JoD: Also, maybe rewrite as `wsum` to avoid introducing the subtractions? Or are these also rewritten as `wsum` already?

JoD (commented):
Ok, let me know what you think :)

cpmpy/solvers/gurobi.py
IgnaceBleukx: The distributivity trick is exactly what happens in the current pipeline:
`3 * ~x >= ~y` will be linearized to `[sum([3, -1] * [~x, ~y]) >= 0]`. After which `only_positive_bv` will transform it to `[sum((-3, 1) * (x, y)) >= -2]`.

The case I was talking about for when we take out the `~bv` is for example in `max(bv1, ~bv2) <= 1`. This constraint will be kept as is by linearize in case `max` is in the supported list.
Then `only_positive_bv` will transform it to `[max(bv1, aux) <= 1, bv2 + aux = 1]`

cpmpy/transformations/linearize.py
IgnaceBleukx: Aha I see, yes will adapt it to your suggestion.

cpmpy/transformations/linearize.py
IgnaceBleukx: This is covered above for now, but will adapt the code so its covered here. That way it is more uniform

cpmpy/transformations/linearize.py
IgnaceBleukx: That case is covered by the 'else' case below. I added an assert to make this more clear. 
For the 'sum' case, I explicitely had to make a 'wsum' with one negative weight. If we make our 'sum' constructor a little smarter to always to this, the if-then-else can go away here.

cpmpy/transformations/linearize.py
IgnaceBleukx: All variables are on the lhs at this point, so if `is_num(lhs)` is true here, we have a simple Python comparison which will automatically be evaluated. I'll add an assert here to make it more explicit

cpmpy/transformations/linearize.py
IgnaceBleukx: Nope, this is prohibited by its constructor

cpmpy/transformations/linearize.py
IgnaceBleukx: its certainly reachable and will be when the input expression is `wsum != const` for example (or != var but this var will be put to lhs by previous code)
Then we create the new constraints `aux -> wsum < const & ~aux -> wsum > const`.
Also checked using the coverage tool this case is actually triggered by the tests :)

cpmpy/transformations/linearize.py
IgnaceBleukx: This can indeed be added in a later iteration. Ok to leave it for now and create an issue

cpmpy/transformations/linearize.py
IgnaceBleukx: Yes, that is a good idea! I couldn't do that before as we did not have the `get_bounds()`

cpmpy/transformations/linearize.py
IgnaceBleukx: In theory, rhs can be a variable at this point if lhs is not a sum/wsum.
So for example when linearizing `max(x,y,z) != p` we cannot bring `p` to the lhs...
Notice however that in practice this does never occur as we rewrite `max(x,y,z) != p` as `[max(x,y,z) == iv, p != iv]` using `only_numexpr_equality`. But it is not a precondition for the `linearize` function...

cpmpy/transformations/linearize.py
JoD: Issue created.

cpmpy/transformations/linearize.py
IgnaceBleukx: Implemented suggestions and also added the optimization of using 2 different big M constants

PR #222, opened 2023-02-07
flatten: avoid unnecessary sum decompositions

tias:
Alternative and more extensive version of the optimisations Emilio implemented in the #114 branch.

 I think its fairly solid, please have a look


sourdough-bread (approved):
Looks good. For the test_flatten, I think we should give a specific name to the integer variables (self.iv) and boolvars (self.bv) which are stringified during the tests to ensure no errors if extra variables are created.

Issue #219, opened 2023-01-26
to_cnf: to CNF encoding of cardinality constraints


sourdough-bread:
```Python
b, x, y, z = boolvar(4)
c1 = b.implies(x+y+z <=1)
c2 = ~b | (x+y+z <=1)
```
The implication `c1` is translated to an `OR` constraint in `to_cnf(c1)`:
```Python
[(~BV0) or (sum([BV1, BV2, BV3]) <= 1)]
```
and `c2`is decomposed in `to_cnf(c2)` as:
```Python
[(~BV0) or (BV4),
 (sum([BV1, BV2, BV3]) > 1) or (BV4),
 (~BV4) or (sum([BV1, BV2, BV3]) <= 1)]
```
potentially introducing a loop.

JoD:
Where exactly do you see the potential loop?

sourdough-bread:
> Where exactly do you see the potential loop?

```Python
c2 = ~b | (x+y+z <=1)
```
is the same expression as

```Python
(~BV4) or (sum([BV1, BV2, BV3]) <= 1)]
```
Which means there is no decomposiiton of the `c2` constraint. If it's already in CNF, it should return the c2 constraint instead of creating extra constraint:
```Python
(~BV0) or (BV4),
 (sum([BV1, BV2, BV3]) > 1) or (BV4)
```

tias:
There might be something bigger at play.

```
bv0 = boolvar()
bvs = boolvar(3)
Model(bv0 | Xor(bvs)).solve("ortools")
```

does not work either. The 'Xor' global constraint is not flattened. For any solver.

Fixing that might also fix the `c2 = ~b | (x+y+z <=1)` case.

(note that the above was sidestepped in #204 by adding globalconstraint decomp in to_cnf, but so the problem is deeper)

tias:
what I reported was a different issue I now realize, which is also resolved in the mean time.

@sourdough-bread the derivation of c2 follows the CNF rules:

`~b | (x+y+z <=1)` is translated into `[~b | auxbv, auxbv == (x+y+z <=1)]` with the double reification split into `[~b | auxbv, auxbv -> (x+y+z <=1), auxbv <- (x+y+z <=1)]` with the latter one flipped to a right-pointing implication `[~b | auxbv, auxbv -> (x+y+z <=1), ~auxbv -> (x+y+z > 1)]` and then both implications converted to a disjunction.

[...]

PR #223, opened 2023-02-10
Transformations care, for to_cnf

tias:
Started from #219 which showed that `to_cnf` should not remove implications on non-clauses.

Then wondered whether `only_bv_implies` does not already do that, and fixed/added some optimisations to that and flatten along the way.

Indeed, after flatten and only_bv_implies; the only thing remaining to get a CNF (on the logical part) is to convert implications over a clause into a clause... easy peasy.

this makes to_cnf a kind of meta-transformation that combines a number of existing transformations...


tias:
Some constraint tests failing.

Interesting case:
```
bvs = boolvar(3)
bv = boolvar()
e = all(bvs) == bv
s = SolverLookup.get("pysat")
s.transform(e)
```
so `and(x1,x2,x3) == bv` is flat, but `bv -> and(x1,x2,x3)` would not be output in this branch rather `bv -> x1, bv -> x2...`

but only-bv-implies outputs `bv -> and(x1,x2,x3)`. So should only-bv-implies run flatten again?

tias:
I've reworked it to follow the new transform style.

I think some of these could make sense to move out into a 'normalize' that is run before, but I don't think that is the priority now...

Note that the origin of these changes was simplifying to_cnf, which succeeded, with fewer auxiliary variables created for other models (and tests) too.

It would be great to see this merged in rather soon...

sourdough-bread (commented):
Changed print, the rest is looking good.

sourdough-bread (approved):

PR #224, opened 2023-02-13
Bound calculations

Wout4:


cpmpy/expressions/core.py
```diff
@@ -531,9 +534,69 @@ def value(self):
...
+        elif self.name == 'sub':
+            return tuple(np.subtract(get_bounds(self.args[0]),get_bounds(self.args[1])))
+        elif self.name == 'div':
+            b1, b2 = get_bounds(self.args[0])
+            b3, b4 = get_bounds(self.args[1])
+            if b3 <= 0 <= b4:
+                raise ZeroDivisionError("division by domain containing 0 is not supported")
+            bounds = [b1 // b3, b1 // b4, b2 // b3, b2 // b4]
+            return min(bounds), max(bounds)
+        elif self.name == 'mod':
```
IgnaceBleukx, line 563: This one is really tricky... Are we sure this covers all cases?

IgnaceBleukx (commented):
This needs some explicit testing I think.
Afterwards ok to merge for me.
The rework of `linearize` also depends on this as we need some bounds calculation there as well.

cpmpy/expressions/core.py
```diff
@@ -531,9 +534,69 @@ def value(self):
...
+            b3,b4 = get_bounds(self.args[1])
+            bounds = [b1 * b3, b1 * b4, b2 * b3, b2 * b4]
+            return min(bounds), max(bounds)
+        elif self.name == 'sum':
+            return sum([get_bounds(x)[0] for x in self.args]), sum([get_bounds(x)[1] for x in self.args])
+        elif self.name == 'wsum':
+            return sum(self.args[0] * np.array([get_bounds(x)[0] for x in self.args[1]])), \
+                   sum(self.args[0] * np.array([get_bounds(x)[1] for x in self.args[1]]))
+        elif self.name == 'sub':
+            return tuple(np.subtract(get_bounds(self.args[0]),get_bounds(self.args[1])))
```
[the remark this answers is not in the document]
Wout4: python substract does not work on tuples that's why it's written like this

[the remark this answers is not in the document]
IgnaceBleukx: I would prefer to do something similar as in the other cases.
You can also group these cases somewhat by arity I think.

cpmpy/expressions/core.py
Wout4: pretty sure, also added tests now! wsum was wrong though

cpmpy/expressions/core.py
```diff
@@ -550,13 +550,22 @@ def get_bounds(self):
             lbs, ubs = zip(*[get_bounds(x) for x in self.args])
             return sum(lbs), sum(ubs)
         elif self.name == 'wsum':
-            lbs, ubs = zip(*[get_bounds(x) for x in self.args])
-            return sum(self.args[0] * np.array(lbs)), \
-                   sum(self.args[0] * np.array(ubs))
+
```
IgnaceBleukx, line 553: We had a pretty clean and fast one for this case: https://github.com/CPMpy/cpmpy/blob/b75f577006f598ec0220a0a81b55461a9336be39/cpmpy/transformations/flatten_model.py#L336
Maybe better to re-use that one?
Also, would it be usefull to have a helper function "get_bounds_args" that returns the bounds of an expressions arguments?

cpmpy/expressions/core.py
```diff
@@ -550,17 +550,10 @@ def get_bounds(self):
             lbs, ubs = zip(*[get_bounds(x) for x in self.args])
             return sum(lbs), sum(ubs)
         elif self.name == 'wsum':
+            weights, vars = self.args
+            bounds = np.array([x for x in zip(*[np.multiply(w, get_bounds(fvar)) for w, fvar in zip(weights, vars)])])
```
Dimosts, line 554: Be sure to check that the bounds are ints (or make them ints), as in cases like in issue #243, when we have weights with float numbers, the bounds become float too and it will crash

IgnaceBleukx: That is not really the task of the `get_bounds()` function I think, but rather when we construct the new variable.
@Wout4 You can write this even a little more efficient as follows:
```python
var_bounds = np.array(list(zip(*[get_bounds(arg) for arg in vars])))
bounds = var_bounds * weights
....
```

Wout4: oke so for now get_bounds can return floats, since that might be useful to get a more accurate calculation. I changed the creation of intvars to call math.floor and math.ceil

tias (commented):
I have (attempted to) rebase it on top of the CountGlobals branch

some tests are failing though...

code looks otherwise OK but I would have to review it again after rebased on master with countglobals

cpmpy/expressions/utils.py
```diff
@@ -109,3 +109,12 @@ def eval_comparison(str_op, lhs, rhs):
         raise Exception("Not a known comparison:", str_op)
 
 
+def get_bounds(expr):
+    # can return floats, use floor and ceil when creating an intvar!
+    if hasattr(expr,'get_bounds'):
```
tias, line 114: I am not convinced by this.

I think it would be clearer to do
```if isinstance(expr): return expr.get_bounds(); else return expr,expr```?

**Review by tias — APPROVED** [review]
reworked a little bit, also left 2 final comments.

Otherwise looks good, cleans up flatten nicely

File: cpmpy/solvers/z3.py [hunk 6]
```diff
@@ -353,6 +353,9 @@ def _z3_expr(self, cpm_con, reify=False):
             elif rhs_is_expr and rhs.name == "element":
                 arr, idx = rhs.args
                 return self._z3_expr(all([(idx == i).implies(Comparison(cpm_con.name, lhs, arr[i])) for i in range(len(arr))]))
+            if isinstance(lhs,GlobalConstraint) and lhs.name == "count":
+                # I think this case is covered in the next line
```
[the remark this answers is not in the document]
Wout4: true, I also removed the element case

cpmpy/expressions/utils.py
Wout4: This would lead to circular import for Expression, I put the import in the get_bounds function

cpmpy/solvers/z3.py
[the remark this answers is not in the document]
Wout4: (we still have to change the element decomposition to allow indexes out of range, think ignace has that in another pr)

cpmpy/expressions/core.py
```diff
@@ -554,6 +554,72 @@ def value(self):
...
+            lb1, ub1 = get_bounds(self.args[0])
+            lb2, ub2 = get_bounds(self.args[1])
+            if lb2 <= 0 <= ub2:
+                raise ZeroDivisionError("% by domain containing 0 is not supported")
+            if ub1 <= 0 and ub2 < 0:
+                return lb1, 0
+            elif ub1 <= 0 and lb2 > 0:
+                return 0, ub2
+            elif lb1 >=0 and lb2 > 0:
+                return 0, ub1
```
JoD, line 596: The above three cases need to be scrapped? It does not matter what `ub1`or `lb1` is, it only matters what `ub2` or `lb2` is. E.g., `-13%-2` is `-1`, so line 591 giving the bound `-13,0` seems overly permissive.

tests/test_expressions.py
```diff
@@ -193,5 +193,103 @@ def test_nullarg_mul(self):
...
+                self.assertTrue(val <= ub)
+            for lhs in range(*z.get_bounds()):
+                val = Operator(name, [lhs]).value()
+                self.assertTrue(val >= lb1)
+                self.assertTrue(val <= ub1)
+            for lhs in range(*y.get_bounds()):
+                val = Operator(name, [lhs]).value()
+                self.assertTrue(val >= lb2)
+                self.assertTrue(val <= ub2)
+
```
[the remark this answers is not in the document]
JoD: After reviewing, the bounds of `mod` seem indeed overly permissive.

cpmpy/expressions/utils.py
```diff
@@ -109,3 +109,13 @@ def eval_comparison(str_op, lhs, rhs):
...
 
+def get_bounds(expr):
+    # can return floats, use floor and ceil when creating an intvar!
+    from cpmpy.expressions.core import Expression
+    if isinstance(expr,Expression):
+        return expr.get_bounds()
+    else:
+        assert is_num(expr), f"All Expressions should have a get_bounds function, `{expr}`"
+        if is_bool(expr):
+            return 0, 1
```
JoD, line 120: This is a known bool, right? Then we can return `0, 0` or `1, 1`, depending on whether the bool is true or false.

JoD (commented):
Looks like a nice improvement!

JoD (commented):
Pushed fixes. @IgnaceBleukx : can you do a quick sanity check on 5b3361d609f0c3191c26a602c2065a893a47a61a ? Then we can commit, and global decompose can grab the changes.

cpmpy/expressions/core.py
```diff
@@ -554,38 +553,31 @@ def get_bounds(self):
...
+                return 0, ub2 - 1
         elif self.name == 'pow':
             lb1, ub1 = get_bounds(self.args[0])
             lb2, ub2 = get_bounds(self.args[1])
             if lb2 < 0:
                 raise NotImplementedError("Power operator: For integer values, exponent must be non-negative")
             bounds = [lb1**lb2, lb1**ub2, ub1**lb2, ub1**ub2]
-            if ub2 > 0:  # even/uneven behave differently when base is negative
-                bounds += [lb1 ** (ub2 - 1), ub1 ** (ub2 - 1)]
+            if lb1 < 0 and 0 < ub2:  # even/uneven behave differently when base is negative
```
IgnaceBleukx, line 566: Just making sure I get this case: so we add a special case as our get_bounds() functions is not guaranteed to get the tightest bounds? So the actual bounds can be odd while calculated bounds are even and therefore we need this case?

JoD: No, the problem is that, e.g., (-2)^2 is positive, but (-2)^1 is negative. So the lower bound depends on either the largest or the second largest exponent value when the base term can be negative. So we need to add it.

PR #225, opened 2023-02-13
Not operator

Wout4:


JoD:
It is hard to review when the diff also contains other MRs.

Thinking about how to resolve this...

Maybe:
If this MR needs the others, we need to get those in the master first.
If not, this MR needs to based on the master.

cpmpy/expressions/core.py
```diff
@@ -52,7 +53,7 @@
     - ~x            x == 0
```
tias, line 53: will need doc update

cpmpy/expressions/core.py
```diff
@@ -350,7 +351,7 @@ def __abs__(self):
         return Operator("abs", [self])
     # 'not' for now, no unary constraint for it
```
tias, line 352: needs doc update

cpmpy/expressions/core.py
```diff
@@ -350,7 +351,7 @@ def __abs__(self):
...
     def __invert__(self):
-        return (self == 0)
+        return Operator("not", [self])
```
tias, line 354: can do a typecheck on self.is_bool() here

also in Operator (or here?) you can overload it and check that it is not 'name == "not"' otherwise you can skip to remove the double negation and return self.args[0]

tias (commented):
also in line 517 of core.py, there is something weird...

we check that if 'x == 1' and x is_bool, then it is just 'x'. And because we overwrite 'eq' for this, we also need to overwrite hash, which is a bit ugly.

But we (I) implemented it rather inconsistently... we do not check for != 0, or that !=1 on bool would be == 0... or (after this commit) that ==0 should be ~x. I wonder though that may be we should leave it the way the user wrote it  here, and leave it for 'normalize' to normalize these? (removing can be here, normalizing would be in a separate commit ; )

IgnaceBleukx:
Will the new transformation `push_down_negation` also be part of this pull request?

tias:
Getting rid of all cases is through a normalize transformation, e.g. to also capture boolvar() >= 1 etc.

Our design principle is that expressions as posted should be as close as possible to what the user wrote (e.g. when they print it immediately after). So we can remove trivial things like a + 0 :: a, and I guess b == 1 to b, but nothing too clever/surprising.

cpmpy/expressions/globalconstraints.py
```diff
@@ -605,5 +630,7 @@ def get_bounds(self):
         """
         Returns the bounds of the (numerical) global constraint
         """
-        arr, val = self.args
-        return 0, len(arr)
+        return [0, len(self.args[0])]
```
tias, line 633: should this be a tuple?
I liked the previous explicit readable
```
arr, val = self.args
return 0, len(arr)
```

tias: this and the repr below are best commited directly on master I think

cpmpy/expressions/variables.py
```diff
@@ -511,6 +511,8 @@ def __rpow__(self, other, modulo=None):
     
     # VECTORIZED Bool operators
     # __invert__ not needed because translated to == 0 and that is handled properly
+    def __invert__(self):
```
tias, line 514: also update the commetn above

cpmpy/solvers/z3.py
```diff
@@ -299,6 +299,11 @@ def _z3_expr(self, cpm_con, reify=False):
                 return z3.Or(self._z3_expr(cpm_con.args))
             elif cpm_con.name == '->':
                 return z3.Implies(*self._z3_expr(cpm_con.args, reify=True))
+            elif cpm_con.name == 'not':
+                if is_boolexpr(cpm_con.args[0]):
+                    return z3.Not(self._z3_expr(cpm_con.args[0]))
+                else:
+                    return self._z3_expr(cpm_con.args[0] == 0)
```
tias, line 306: we, cpmpy, should never allow a not expression around a non-boolexpr, so that this case can never trigger

cpmpy/expressions/globalconstraints.py
Wout4: Ah this is a merge slip-up, in master it is correct (we don't need the repr, since it is implemented like the default anyway)

Wout4:
Ok so i think we will leave it to a separate transformation to do these cleanups, also the push down negation will be in another pr

cpmpy/expressions/core.py
```diff
@@ -322,9 +318,10 @@ def __pos__(self):
         return self
     def __abs__(self):
         return Operator("abs", [self])
-    # 'not' for now, no unary constraint for it
     def __invert__(self):
-        return (self == 0)
+        if not (is_boolexpr(self)):
+            raise TypeError("Not operator is only allowed on boolean expressions")
```
tias, line 323: for bonus points, also print the expression given as argument

cpmpy/transformations/flatten_model.py
```diff
@@ -578,18 +587,18 @@ def negated_normal(expr):
...
-        # stay in xor space
-        # only negate last element
-        from ..expressions.globalconstraints import Xor  # avoid circular import
-        return Xor(expr.args[:-1] + [negated_normal(expr.args[-1])])
-
     else: # circular if I import GlobalConstraint here...
+        if hasattr(expr, "decompose_negation"):
+            # for global constraints where the negation of the decomposition is not equivalent
+            # to the negated global constraint (due to auxiliary variables, i.e. Circuit)
+            return Operator('and', expr.decompose_negation())
```
tias, line 601: we are not very consistent with the use of our builtin-overwritten `any()` and `Operator('and',...)`.

one benefit of `all()` is that it removes constants from the args, one benefit of Operator is that we dont have to import our all, and that we save the effort of a function call and going over the loop...

we were already not consistent, so not for you to fix in this PR, but to keep in mind that we might want to make it consistent at some point...

tias (approved):
Great.

One bonus point comment, and one future thought.

Great extensive tests as well! Do merge!

(I have 3 deprecation warnings when running the tests locally, you might want to fix those in a separate commit)

PR #228, opened 2023-02-16
minizinc time -> flatTime

Wout4:


cpmpy/solvers/minizinc.py
```diff
@@ -223,7 +223,7 @@ def _post_solve(self, mzn_result):
         # new status, translate runtime
         self.cpm_status = SolverStatus(self.name)
         if 'time' in mzn_result.statistics:
```
sourdough-bread, line 225: Don't you need to change 'time' as well ?

sourdough-bread (changes requested):

PR #229, opened 2023-02-16
227 false gets removed in transformations

IgnaceBleukx:
Wasn't really sure about how to do it in pysdd...


cpmpy/solvers/gurobi.py
```diff
@@ -281,6 +282,9 @@ def _post_constraint(self, cpm_expr):
             Solvers can raise 'NotImplementedError' for any constraint not supported after transformation
         """
         from gurobipy import GRB
+        # True or False
+        if is_bool(cpm_expr):
+            return self.grb_model.addConstr(cpm_expr)
```
Dimosts, line 287: Maybe better use the expr.is_bool() method, to be consistent with other similar checks that are done.

cpmpy/solvers/ortools.py
```diff
@@ -342,6 +342,9 @@ def _post_constraint(self, cpm_expr, reifiable=False):
 
         :param reifiable: if True, will throw an error if cpm_expr can not be reified by ortools (for safety)
         """
+        # True or False
+        if is_bool(cpm_expr):
```
Dimosts, line 346: same as before

IgnaceBleukx:
After discussion offline: will implement a `BoolVal` wrapper so `False` can become a CPMpy expression. This will be part of a new transformation which also unravels all nested lists. Then every transformation will simply get a flat list of constraints and all transformations should work as is with the new `BoolVal` expression.
Will still have to add the cas of posting `False` to the solver in the `_post_constraint` of all solvers.

cpmpy/solvers/z3.py
```diff
@@ -249,10 +250,7 @@ def transform(self, cpm_expr):
...
         # that also means we don't need to extract user variables here
         # we store them directly in `solver_var()` itself.
-        if is_any_list(cpm_expr):
-            return cpm_expr
-        else:
-            return [cpm_expr]
+        return make_cpm_expr(cpm_expr)
 
     def _post_constraint(self, cpm_expr):
         """
```
Wout4, line 256: the if is_any_list check in post_constraint is not needed anymore because of the make_cpm_expr call in transform, right?

cpmpy/transformations/flatten_model.py
```diff
@@ -110,21 +110,15 @@ def flatten_model(orig_model):
...
-    # base cases
-    if isinstance(expr, bool):
-        if expr:
-            return []
-        else:
-            return [expr]  # not sure about this one... means False is a valid FNF expression
-    elif isinstance(expr, _BoolVarImpl):
+    if isinstance(expr, _BoolVarImpl):
         return [expr]
     elif is_num(expr) or isinstance(expr, _NumVarImpl):
```
Wout4, line 123: I think this check could go in make_cpm_expr, so we can catch it as early as possible. Might be worse on performance but I don't really think so

Wout4 (changes requested):

tias:
We have a bunch of research code that expects `flatten_constraint()` to continue working also on a single expression.

I hence propose to make `flatten_constraint()` a semi-meta function that calls make_expr internally. That means solvers dont need to add the transform themselvs.  I'll play with this on the plane.

tias:
I reworked it some more. I think its looking nice, would be good to see this land rather soon, so we can base other branches on this.

It tested good on bigtest (ortools) too (no noticable timing differences though, a bit to my surprise)

@Wout4 and @IgnaceBleukx can you have a final look and merge?

cpmpy/transformations/normalize.py
```diff
@@ -2,27 +2,32 @@
...
-            unravel(e, append)
-        elif e is False:
-            append(BoolVal(e))
-        elif e is not True:  # if True: pass
-            append(e)
+        for e in lst:
+            if isinstance(e, Expression):
+                if isinstance(e, NDVarArray):  # sometimes does not have .name
+                    unravel(e.flat, append)
+                elif e.name == "and":
```
IgnaceBleukx, line 20: Here check for `merge_and`, will add

IgnaceBleukx:
Can confirm all tests run, for pysdd as well.
Ready to merge for me. @Wout4 waiting for your final thoughts before merging.

Wout4:
looks good to me!

Issue #210, opened 2023-01-16
Doubt on precedence in the constraint


ElFosco:
Writing these lines of code leads to an error

### 
``` python3
from cpmpy import *

model = Model()
test = intvar(0, 1000, name="test")
model += (test <= 100 & test >= 50)
model.solve()
```
I suppose it is due to the fact that it is evaluating 100 & test immediately, in fact if you change with 

``` python3
model += ((test <= 100) & (test >= 50))
```
It works properly

JoD:
From 'How to debug':

>   - when using `&` and `|`, make sure to always put the subexpressions in brackets. E.g. `(x == 1) & (y == 0)` instead of `x == 1 & y == 0`. The latter wont work, because Python will unfortunately think you meant `x == (1 & (y == 0))`.

When I try
```
from cpmpy import *

model = Model()
test = intvar(0, 1000, name="test")
model += (test <= 100) & (test >= 50)
print(model.solve())
```
it seems to work.

JoD:
It is related to https://github.com/CPMpy/cpmpy/issues/152

I will close this as a duplicate of that one.

Issue #198, opened 2022-12-20
Boolean operators on integers crash some solvers


Wout4:
Giving an integer as an argument for and, or and implies operators crash most solvers (gurobi, z3, minizinc, pysat)

example
Model( (bv[0] | 0)).solve("gurobi")

IgnaceBleukx:
I don't think we want to support this, right? We should just throw a CPMpy exception when encountering this.
Actually, this is another example where we want to introduce some typed-ness on the modelling side of the library.
We could then simply raise an exception when a user tries to construct this type of expression by checking in the constructor of "OR" wether both sides `.is_bool()`

Wout4:
pr #297 fixes this by throwing a cpmpy exception

Issue #152, opened 2022-11-14
constraint.implies(...) crashes if constraint is a simple True (or False)


rubenkindt:
solver: not relevant, all solvers
CPMpy: version v0.9.9 commit e79b3afedc934a9437c2ddb3a9f54d7e2d7bd3ee

Some easy constraints automatically become True (or False) when adding them to a model. Taking an implies of such constraint results in a crash, this could be confusing to the user if it is a more complicated crash.

```python 
from cpmpy import *

m = Model()
m += (1+1) > 3 # Imagine this to be complicated, but still resulting in a True or False
m += (m.constraints[0]).implies(True)
m.solve()  # AttributeError: 'bool' object has no attribute 'implies'
```

expected behavior: 
To not crash when taking an 'implies(...)' of a boolean.

possible solution:
By not adding a True or False to the model or by documenting that his could happen.

bug found while working on my master thesis.

tias:
This is a limitation of using operator overloading in Python, and Python not having an 'implies' operator.

We chose to add a function to our objects, but it does not exist on non-CPMpy objects (including pure python objects, or numpy objects). We could foresee a function `def implies(lhs, rhs)` instead of `lhs.implies(rhs)`, if that could better support this case?

Or would it be possible to attach the 'implies' function to the python bool object?

As to documenting this, you are right. Perhaps the best place is https://cpmpy.readthedocs.io/en/latest/modeling.html, comments welcome

IgnaceBleukx:
Apparently you could add the `implies` function to builtin python bools using the fishhook library:
https://pypi.org/project/fishhook/ altough I am not sure if we really want to do this.

```python 
@hook(bool)
def implies(self, other):
    if self is True:
        return other
    if self is False:
        return True
    # or alternatively
    return Operator("->", [self, other])
```

tias:
Oh, that could be pretty great actually... if we could give `.implies()` and `.value()` to basic types, it would make the modeling more consistent...

There are other quirks of mixed type overloading that would be great if we could solve them like this. They are (rightfully) surprising to users.

One is that `x = np.array([1,2,3]); m += x[intvar(1,9)]`; ideally we could hook onto numpy's `__in__` to catch it.

The other is that
```
>>> ~True
-2
>>> ~1
-2
>>> ~False
-1
>>> ~0
-1
```
because we use `~` as 'not' instead of as bitwise negation...
np.array([1,2,3])[intvar(

rubenkindt:
[...]

Issue #158, opened 2022-11-14
Importing a file and adding constrains can turn problems wrongly unsatisfiable


rubenkindt:
solver: not relevant
CPMpy: version v0.9.9 commit e79b3afedc934a9437c2ddb3a9f54d7e2d7bd3ee

When importing a file and adding new constraints with variables containing the same variable name (both generated or user-defined names) will cause the solver to treat the variable as one and the same, which can cause wrong solutions.

generating a file to import:
```python 
from cpmpy import *

i = intvar(lb=3, ub=6) # auto name IV0
m = Model()
m += i == 4
m.to_file("1")
```

```python 
from cpmpy import *

m2 = Model().from_file("1")
j = intvar(lb=0, ub=10) # auto name again IV0
m2 += j == 5
m2.solve()
print(m2.status())
```

expected behavior: 
Threat both variable as two separate variables and solve the problem.

bug found while working on my master thesis.

tias:
Interesting.

1. a variable's name is its unique identifier. If you give two variables the same name, you run into trouble. Even if we would support it at CPMpy side, things would quickly go wrong for different solvers...

I should document this more explicitly, e.g. in https://cpmpy.readthedocs.io/en/latest/modeling.html too.

2. the case you provide is actually not misuse... the user can not know they are reusing a variable name, because they are not actually naming anything. I think the proper fix here is to increase the magic variable counter when doing `from_file()`, based on the highest appearing IV* and BV* names in the 'fromfile' model... TODO ; )

PR #230, opened 2023-02-16
Fishhooking boolean operators to cover extreme cases

Dimosts:
Related to issues #198 #152 #210.

Based on the discussion of issue #152, and in order to cover both #152 and #198 (and #210 which is a duplicate), I introduced boolean operators for both our variables classes, and for int/bool objects. For the later, I used fishhooking.

The main changes are in variables.py (introducing the new functions for our variable classes, overloading logical operators) and in python_builtins.py, where I introduced the overloaded (hooked) logical operators for int and bool.

I also introduced a new "is_bool" function for ints. That is because we use it in many cases for expressions and their arguments (e.g. lines 64-72 of linearize.py). This was leading to errors if an argument was an integer, because of no built in function.  

The above cover all cases that have appeared in the mentioned issues, providing the appropriate CPMpyException messages in each case.

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -5,7 +5,7 @@
 ##
 """
     The `Expression` superclass and common subclasses `Expression` and `Operator`.
-    
+
     None of these objects should be directly created, they are automatically created through operator
     overloading on variables and expressions.
 
@@ -40,7 +40,7 @@
     - x ^ y         Xor([x,y])  # a global constraint
 
     Finally there are two special cases for logical operators 'implies' and '~/not'.
-    
+
     Python has no built-in operator for __implication__ that can be overloaded.
     CPMpy hence has a function 'implies()' that can be called:
 
@@ -62,7 +62,7 @@
                     subexpressions and doing the appropriate computation
                     this is used to conveniently print variable values, objective values
                     and any other expression value (e.g. during debugging).
-    
+
     ===============
     List of classes
     ===============
@@ -181,14 +181,8 @@ class Expression(object):
 
         # checking if both are bool
         # checking them separately, to give a more clear exception message to the user
-        if not self.is_bool():
-            raise CPMpyException(
-                f"Logical implication involving a non boolean argument ({type(self)}: {self}) is not allowed")
-
-        if not other.is_bool():
-            raise CPMpyException(
-                f"Logical implication involving a non boolean argument ({type(other)}: {other}) is not allowed")
-
+        assert self.is_bool(), f"Logical implication involving a non boolean argument ({type(self)}: {self}) is not allowed"
+        assert other.is_bool(), f"Logical implication involving a non boolean argument ({type(other)}: {other}) is not allowed"
 
         # other constant
         if other is True:
@@ -221,13 +215,8 @@ class Expression(object):
 
         # checking if both are bool
         # checking them separately, to give a more clear exception message to the user
-        if not self.is_bool():
-            raise CPMpyException(
-                f"Logical conjunction involving a non boolean argument ({type(self)}: {self}) is not allowed")
-
-        if not other.is_bool():
-            raise CPMpyException(
-                f"Logical conjunction involving a non boolean argument ({type(other)}: {other}) is not allowed")
+        assert self.is_bool(), f"Logical conjunction involving a non boolean argument ({type(self)}: {self}) is not allowed"
+        assert other.is_bool(), f"Logical conjunction involving a non boolean argument ({type(other)}: {other}) is not allowed"
 
 
         # some simple constant removal
@@ -237,19 +226,12 @@ class Expression(object):
             return False
 
         return Operator("and", [self, other])
-
-
     def __rand__(self, other):
 
         # checking if both are bool
         # checking them separately, to give a more clear exception message to the user
-        if not self.is_bool():
-            raise CPMpyException(
-                f"Logical conjunction involving a non boolean argument ({type(self)}: {self}) is not allowed")
-
-        if not other.is_bool():
-            raise CPMpyException(
-                f"Logical conjunction involving a non boolean argument ({type(other)}: {other}) is not allowed")
+        assert self.is_bool(), f"Logical conjunction involving a non boolean argument ({type(self)}: {self}) is not allowed"
+        assert other.is_bool(), f"Logical conjunction involving a non boolean argument ({type(other)}: {other}) is not allowed"
 
         # some simple constant removal
         if other is True:
@@ -263,14 +245,8 @@ class Expression(object):
 
         # checking if both are bool
         # checking them separately, to give a more clear exception message to the user
-        if not self.is_bool():
-            raise CPMpyException(
-                f"Logical disjunction involving a non boolean argument ({type(self)}: {self}) is not allowed")
-
-        if not other.is_bool():
-            raise CPMpyException(
-                f"Logical disjunction involving a non boolean argument ({type(other)}: {other}) is not allowed")
-
+        assert self.is_bool(), f"Logical disjunction involving a non boolean argument ({type(self)}: {self}) is not allowed"
+        assert other.is_bool(), f"Logical disjunction involving a non boolean argument ({type(other)}: {other}) is not allowed"
 
         # some simple constant removal
         if other is True:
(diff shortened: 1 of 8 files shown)


cpmpy/expressions/python_builtins.py
```diff
@@ -100,3 +102,61 @@ def sum(iterable):
     if not any(isinstance(elem, Expression) for elem in iterable):
         return np.sum(iterable)
     return Operator("sum", iterable)
+
+@hook(bool, int)
+def __and__(self, other):
+    if not (isinstance(other,_IntVarImpl) or isinstance(other, _BoolVarImpl) ):
```
IgnaceBleukx, line 108: Can we just check for `isinstance(other, Expression)`?

cpmpy/expressions/python_builtins.py
```diff
@@ -100,3 +102,61 @@ def sum(iterable):
...
+        return orig(self, other)
+    else:
+
+        if not self.is_bool():
+            raise CPMpyException(f"Logical conjunction involving an integer ({self}) is not allowed")
+
+        if isinstance(self,int):
```
IgnaceBleukx, line 115: Lets wait for this until we cleared out how to deal with ints and logical operators...

cpmpy/expressions/python_builtins.py
```diff
@@ -100,3 +102,61 @@ def sum(iterable):
...
+            self = (self == 1)  # convert to True or False
+
+        if not isinstance(other,_BoolVarImpl):
+            raise CPMpyException(f"Logical conjunction involving an IntVar ({other}) is not allowed")
+
+        return Operator("and", [self, other])
```
IgnaceBleukx, line 121: Making this code a bit shorter, we can actually do the following:

```python
@hook(bool, int)
def __and__(self, other):
    if isinstance(other, Expression):
        return other.__rand__(self)
    return orig(self, other)
```
This avoids a lot of duplication.

cpmpy/expressions/python_builtins.py
```diff
@@ -100,3 +102,61 @@ def sum(iterable):
...
+
+@hook(bool, int)
+def __or__(self, other):
```
IgnaceBleukx, line 124: Same as in `__and__`

cpmpy/expressions/variables.py
```diff
@@ -280,6 +281,16 @@ def __abs__(self):
             return self
         return super().__abs__()
 
+    def __and__(self, other):
```
IgnaceBleukx, line 284: This can probably go in a seperate commit to master. Very usefull!

IgnaceBleukx (commented):
This is a nice start, but does not cover the full potential of the `fishook` library I think.

For example, we can also hook into the `__getitem__` of  `list` and `np.ndarray` to eliminate the need of the `cpm_array` and create Element constraints very cleanly.
Although we need to be really carefull in how we implement this as we want to make sure we do not screw up the efficiency of __getitem__ of these classes...

Dimosts:
> This is a nice start, but does not cover the full potential of the `fishook` library I think.
> 
> For example, we can also hook into the `__getitem__` of `list` and `np.ndarray` to eliminate the need of the `cpm_array` and create Element constraints very cleanly. Although we need to be really carefull in how we implement this as we want to make sure we do not screw up the efficiency of **getitem** of these classes...

Yes, we can still do many more with fishhook library. This was just intended as a start, solving some specific issues. E.g., we can hook the value() functions in basic types (as suggested in #158) etc.

cpmpy/expressions/python_builtins.py
Dimosts: Yes true

cpmpy/expressions/python_builtins.py
Dimosts: I assume you mean 

`        return other.__and__(self)
`

and not 

`        return other.__rand__(self)`

cpmpy/expressions/python_builtins.py
Dimosts: done, checked if works properly and pushed

cpmpy/expressions/python_builtins.py
IgnaceBleukx: No, in Python, all `__` operators are also reversed.
As we [implemented both](https://github.com/CPMpy/cpmpy/blob/9fba76cc89845deb99f01a0b396021cc7f708924/cpmpy/expressions/core.py#L214), we can simply call the reversed operator from here and everything should work as is.

Dimosts: However, we still would need to check if the self part is bool or not, and also if the other part is bool (i.e., as it is an expression in this case, it can be an IntVar, where we throw an expression). So, it should call the appropriate function (like the newly added ones in _BoolVarImpl and _IntVarImpl

cpmpy/expressions/variables.py
```diff
@@ -334,6 +345,24 @@ def __ne__(self, other):
     def __abs__(self):
         return self
 
+    def __and__(self, other):
```
IgnaceBleukx, line 348: Why not add this logic to the [Expression class in core.py](https://github.com/CPMpy/cpmpy/blob/9fba76cc89845deb99f01a0b396021cc7f708924/cpmpy/expressions/core.py#L206)?

Dimosts: basically to avoid multiple checking of both sides and have a more, but can be moved to core.py if we also add a "self.is_bool()" check and throw the exception there (for cases of not boolean expressions, IntVars etc.).

Dimosts: moving it there now

cpmpy/expressions/python_builtins.py
```diff
@@ -100,3 +102,39 @@ def sum(iterable):
...
+            raise CPMpyException(f"Logical implication involving an IntVar ({other}) is not allowed")
+
+        return Operator("->", [self, other])
+
+@hook(int)
+def is_bool(self):
+    if self > 1 or self < 0:
+        return False
+    else:
+        return True
```
Dimosts, line 140: somehow this last function crashes the following:

```
a,b = boolvar(shape=2)
m = Model((a & b) <= 1)
m.solve(solver="z3")
```

Dimosts: btw I found this completely accidentally, I think we should add it to the tests

cpmpy/expressions/core.py
```diff
@@ -204,14 +218,39 @@ def __ge__(self, other):
...
+
+
         # some simple constant removal
         if other is True:
             return self
         if other is False:
             return False
 
         return Operator("and", [self, other])
+
```
tias, line 240: avoid cosmetic changes; we also put 'ror' right next after 'or'

cpmpy/expressions/python_builtins.py
```diff
@@ -19,9 +19,11 @@
         sum
 """
 import numpy as np
-from .variables import NDVarArray
+from fishhook import *
```
tias, line 22: this would require updating the requirements file then...


I would like all the fishooks to be in a new file 'cpmpy/expressions/fishooks..py' or similar. Also ideally the code would also work if somebody does not have fishook installed? or maybe not... but then definitely update requirements.

cpmpy/expressions/variables.py
```diff
@@ -50,6 +50,7 @@
 from collections.abc import Iterable
 import warnings # for deprecation warning
 import numpy as np
+from ..exceptions import CPMpyException
```
tias, line 53: undo changes to this file

cpmpy/solvers/z3.py
```diff
@@ -410,7 +410,7 @@ def _z3_expr(self, cpm_con, reify=False):
                 else:
                     lhs = self._z3_expr(lhs)
 
-                if is_boolexpr(rhs):
+                if is_boolexpr(rhs) and not isinstance(rhs,int):
```
tias, line 413: not sure where `is_boolexpr` is defined, but does it say 'yes' to an int? with fishook, this should be rhs.is_bool()?

cpmpy/transformations/flatten_model.py
```diff
@@ -127,7 +128,7 @@ def flatten_constraint(expr):
     elif isinstance(expr, _BoolVarImpl):
         return [expr]
     elif is_num(expr) or isinstance(expr, _NumVarImpl):
-        raise Exception("Numeric constants or numeric variables not allowed as base constraint")
+        raise CPMpyException(f"Numeric constants or numeric variables not allowed as base constraint: {expr}")
```
tias, line 131: I guess we should have used an assert here too.

cpmpy/expressions/core.py
```diff
@@ -177,11 +178,24 @@ def deepcopy(self, memodict = {}):
     # Python does not offer relevant syntax...
     # for double implication, use equivalence self == other
     def implies(self, other):
+
+        # checking if both are bool
+        # checking them separately, to give a more clear exception message to the user
+        if not self.is_bool():
```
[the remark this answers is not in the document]
Dimosts: They should be asserts indeed. They are independent of fishhooks, should I commit them directly to master? (as asserts and below the constant checks)

cpmpy/solvers/z3.py
Dimosts: I think this should be dealt in a different pull request, trying to make is_bool check consistent. Right now, we have many different ways of checking if an expression is bool (is_boolexpr(expr), is_bool(expr), expr.is_bool() etc.).

I can try and make it consistent in a different PR.

cpmpy/expressions/python_builtins.py
Dimosts: Yes we should update the requirements

Dimosts:
After discussion in-person, we retract from introducing fishhooks in CPMpy

PR #233, opened 2023-02-20
ite helper function and test

Wout4:


IgnaceBleukx:
Why not make it a global constraint? For example, z3 has an explicit `if-then-else` statement so we could post it directly?

Wout4:
> 
True, seems like a more consistent way to do this!

tias:
Wait...

Our 'global' is treating it as a logical function, e.g.
if bool_expr then bool_expr1 else bool_expr2

but z3 is using it as if bool_expr then num_expr else num_expr, so that the if-then-else can be used inside a sum for example; with returntype intvar... that is not what we intend; and not what the decomposition will 'return'? So there is a difference still?

Wout4:
Z3 allows for the second and third argument to be either bool or int, we only allow bool as it is now implemented.
So we are stricter and can still translate correctly to z3 i think.

In order to also allow int would make it a numerical global constraint, so then the question becomes do we make a separate global for that or type overload this one?

cpmpy/expressions/globalconstraints.py
```diff
@@ -289,7 +289,32 @@ def deepcopy(self, memodict={}):
 
     # TODO: value()
 
-# Numeric Global Constraints (with integer-valued return type)
+
+# syntax of the form 'if b then x == 9 else x == 0' is not supported
+# a little helper:
+class IfThenElse(GlobalConstraint):
+    def __init__(self, condition, if_true, if_false):
+        super().__init__("ite", [condition, if_true, if_false], is_bool=True)
```
IgnaceBleukx, line 299: We should typecheck the arguments here

cpmpy/expressions/globalconstraints.py
```diff
@@ -289,7 +289,32 @@ def deepcopy(self, memodict={}):
...
+
+    def value(self):
+        condition, if_true, if_false = self.args
+        condition_val = argval(condition)
+        true_val = argval(if_true)
```
IgnaceBleukx, line 304: Better to calculate the val of `true_val` and `false_val` lazily. I.e:
```python
if condition_val:
     return argval(if_true)
else:
     return argval(if_false)
```
As one of the arguments might be an incomplete function such as an element with weird indexes...

cpmpy/expressions/globalconstraints.py
```diff
@@ -289,7 +289,32 @@ def deepcopy(self, memodict={}):
...
+
+    def decompose(self):
+        condition, if_true, if_false = self.args
+        return (condition.implies(if_true) & \
+                (~condition).implies(if_false)
+                )
+
+    def __repr__(self):
+        condition, if_true, if_false = self.args
+        return "If {} Then {} Else {}".format(condition, if_true, if_false)
```
IgnaceBleukx, line 317: Maybe put brackets around the `if_true` and `if_false`?

IgnaceBleukx (commented):
Left some comments for a more robust implementation.

Should we add this global constraint to the `test_constraints` suite as well?

Wout4:
Ok so for the moment we are happy to keep it as is, just bool expressions

cpmpy/expressions/globalconstraints.py
JoD: After discussion, the general policy to put brackets in CPMpy is to do it for non-variable subexpressions. This is probably enough to disambiguate most operators (e.g., `-x+y` vs `-(x+y)`). However, as Wout remarks, the words in ITE work as well as brackets to disambiguate. For `condition` and `if_true` this is clear, for `if_false` there is no "closing" keyword, but as we do not have any post-fix operators (e.g., `if foo then bar else baz**'**`) I think `if_false` is fine too.

JoD: So yeah, agree with Wout that for now we don't need them.

JoD (commented):
Looks good

PR #239, opened 2023-02-23
Count globals

Wout4:


Wout4:
these are the last changes from pr #207

cpmpy/expressions/globalconstraints.py
```diff
@@ -560,3 +563,65 @@ def deepcopy(self, memodict={}):
         return Cumulative(*copied_args)
 
 
+class GlobalCardinalityCount(GlobalConstraint):
+    """
+        GlobalCardinalityCount(a,gcc): Collect the number of occurrences of each value 0..a.ub in gcc.
+    The array gcc must have elements 0..ub (so of size ub+1).
```
JoD, line 569: What is `a.ub`? E.g., if `a` is `x+y` with `x.ub=10` and `y.ub=10`?

Wout4: a is a list, so a.ub is the highest value that can be in the list. Confusing notation with the .ub because this makes it look like an intvar..

PR #242, opened 2023-03-01
Manage incomplete functions when calculating value

IgnaceBleukx:
Implemented a new type of exception to work with incomplete functions.
Changed `.value()` computation of `->` and `or` accordingly

### DIFF
--- a/cpmpy/expressions/utils.py
+++ b/cpmpy/expressions/utils.py
@@ -26,6 +26,9 @@ import numpy as np
 from collections.abc import Iterable # for _flatten
 from itertools import chain, combinations
 
+from cpmpy.exceptions import IncompleteFunctionError
+
+
 def is_int(arg):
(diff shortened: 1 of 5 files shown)


cpmpy/expressions/utils.py
```diff
@@ -76,7 +79,11 @@ def argval(a):
         
         We check with hasattr instead of isinstance to avoid circular dependency
     """
-    return a.value() if hasattr(a, "value") else a
+    try:
+        return a.value() if hasattr(a, "value") else a
+    except IncompleteFunctionError as e:
+        if a.is_bool(): return False
```
JoD, line 85: One problem that could arise is: `(x/0>0) == (x/0<0)` which seems a bit counter-intuitive.

JoD (approved):
If we accept the weirdness arising from treating incomplete function comparisons as `false`, then this MR looks good.

PR #245, opened 2023-03-02
Count globals rebased

JoD:


cpmpy/expressions/globalconstraints.py
```diff
@@ -582,3 +585,65 @@ def deepcopy(self, memodict={}):
         return Cumulative(*copied_args)
 
 
+class GlobalCardinalityCount(GlobalConstraint):
+    """
+        GlobalCardinalityCount(a,gcc): Collect the number of occurrences of each value 0..a.ub in gcc.
```
JoD, line 590: Can `a` contain arbitrary expressions, not only pure values? I guess yes, right?

Wout4: with the bounds calculations pr that would be possible, as is only values or intvars are allowed

JoD:
@Wout4 Review finished, let me know what you think!

Wout4:
I added a change to the bounds_calucation pr to calculate a.ub appropriatly for gcc
Also fixed a bug equivalent to the one in table decomposition

cpmpy/expressions/globalconstraints.py
```diff
@@ -582,3 +584,61 @@ def deepcopy(self, memodict={}):
...
+    def decompose(self):
+        a, gcc = self.args
+        ub = max([v.ub for v in a])
+        assert (len(gcc) == ub + 1), f"GCC: length of gcc variables {len(gcc)} must be ub+1 {ub + 1}"
+        return [Count(a, i) == v for i, v in enumerate(gcc)]
+
+    def value(self):
+        a, gcc = self.args
+        gval = [argval(y) for y in gcc]
+        return all([gval[i] == Count(a,i).value() for i in range(len(gcc))])
```
tias, line 605: I guess you could do `all(self.decompose()).value()`? Would be cleaner? or not possible?

cpmpy/expressions/globalconstraints.py
```diff
@@ -582,3 +584,61 @@ def deepcopy(self, memodict={}):
...
+
+    def __init__(self,arr,val):
+        super().__init__("count", [arr,val], is_bool=False)
+
+    def decompose_comparison(self, cmp_op, cmp_rhs):
+        """
+        Count(arr,val) can only be decomposed if it's part of a comparison
+        """
+        arr, val = self.args
+        return [eval_comparison(cmp_op, Operator('sum',[ai==val for ai in arr]), cmp_rhs)]
```
tias, line 629: bit clumsy formulation.
why not using 'sum' as in value?

smth like `cnt = sum(ai == val for ai in arr); eval_comparison(cmp_op, cnt, cmp_rhs)`?
could be more readable...

tias (approved):
looks good, left 2 readability comments but can also be merged as is.

cpmpy/expressions/globalconstraints.py
Wout4: Yes that works, changed it!

PR #246, opened 2023-03-02
Decompose globals

IgnaceBleukx:
New transformation to decompose any unsupported global constraints in flat expressions.
Code is functional but can be somewhat compacted I think...

Transformation should be run after flatten() for every solver so we do not have to decompose globals at a later stage.

Work in progress still

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -446,7 +446,7 @@ class Element(GlobalConstraint):
         from .python_builtins import any
 
         arr, idx = self.args
-        return [(idx == i).implies(Comparison(cpm_op, arr[i], cpm_rhs)) for i in range(len(arr))] + \
+        return [(idx == i).implies(eval_comparison(cpm_op, arr[i], cpm_rhs)) for i in range(len(arr))] + \
                [idx >= 0, idx < len(arr)]
 
     def __repr__(self):
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -434,7 +434,7 @@ class Element(GlobalConstraint):
             return argval(arr[idxval])
         return None # default
 
-    def decompose_comparison(self, cmp_op, cmp_rhs):
+    def decompose_comparison(self, cpm_op, cpm_rhs):
         """
             `Element(arr,ix)` represents the array lookup itself (a numeric variable)
             It is not a constraint itself, so it can not have a decompose().
@@ -445,8 +445,9 @@ class Element(GlobalConstraint):
         """
         from .python_builtins import any
 
-        arr,idx = self.args
-        return [any(eval_comparison(cmp_op, arr[j], cmp_rhs) & (idx == j) for j in range(len(arr)))]
+        arr, idx = self.args
+        return [(idx == i).implies(Comparison(cpm_op, arr[i], cpm_rhs)) for i in range(len(arr))] + \
+               [idx >= 0, idx < len(arr)]
 
     def __repr__(self):
         return "{}[{}]".format(self.args[0], self.args[1])
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -111,7 +111,7 @@ import warnings # for deprecation warning
 import numpy as np
 from ..exceptions import CPMpyException, IncompleteFunctionError
 from .core import Expression, Operator, Comparison
-from .variables import boolvar, intvar, cpm_array
+from .variables import boolvar, intvar, cpm_array, _NumVarImpl
 from .utils import flatlist, all_pairs, argval, is_num, eval_comparison, is_any_list, is_boolexpr
 from ..transformations.flatten_model import get_or_make_var
 
@@ -144,6 +144,13 @@ class GlobalConstraint(Expression):
             it does not create a circular dependency.
         """
         raise NotImplementedError("Decomposition for",self,"not available")
+
+    def is_total(self):
+        """
+            Returns whether the global constraint is a total function.
+            If true, its value is defined for all arguments
+        """
+        return True
 
 # Global Constraints (with Boolean return type)
 def alldifferent(args):
@@ -409,6 +416,14 @@ class Element(GlobalConstraint):
         arr, idx = self.args
         return [(idx == i).implies(eval_comparison(cpm_op, arr[i], cpm_rhs)) for i in range(len(arr))] + \
                [idx >= 0, idx < len(arr)]
+
+    def is_total(self):
+        arr, idx = self.args
+        if isinstance(idx, _NumVarImpl):
+            return idx.lb >= 0 & idx.ub < len(arr)
+        else:
+            raise NotImplementedError("Bound calculation for {idx} not supported yet, fix in pull request #224")
+
 
     def __repr__(self):
         return "{}[{}]".format(self.args[0], self.args[1])
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -369,7 +369,7 @@ class Minimum(GlobalConstraint):
 
         arr = argval(self.args)
         _min = intvar(-2147483648, 2147483647)
-        return all([any(x <= _min for x in arr), all(x >= _min for x in arr), eval_comparison(cpm_op, _min, cpm_rhs)])
+        return [any(x <= _min for x in arr), all(x >= _min for x in arr), eval_comparison(cpm_op, _min, cpm_rhs)]
 
 class Maximum(GlobalConstraint):
     """
@@ -403,7 +403,7 @@ class Maximum(GlobalConstraint):
 
         arr = argval(self.args)
         _max = intvar(-2147483648, 2147483647)
-        return all([any(x >= _max for x in arr), all(x <= _max for x in arr), eval_comparison(cpm_op, _max, cpm_rhs)])
+        return [any(x >= _max for x in arr), all(x <= _max for x in arr), eval_comparison(cpm_op, _max, cpm_rhs)]
 
 
 def element(arg_list):
--- a/cpmpy/transformations/decompose_global.py
+++ b/cpmpy/transformations/decompose_global.py
@@ -41,7 +41,7 @@ def decompose_global(lst_of_expr, supported=set(), supported_reif=set()):
 
         return True
 
-    assert supported_reif <= supported, "`supported` set is assumed to be a subset of the `supported_reified` set"
+    assert supported_reif <= supported, "`supported_reif` set is assumed to be a subset of the `supported` set"
     if not is_any_list(lst_of_expr):
         lst_of_expr= [lst_of_expr]
 
--- a/cpmpy/transformations/decompose_global.py
+++ b/cpmpy/transformations/decompose_global.py
@@ -1,3 +1,4 @@
+import copy
 
 from ..expressions.globalconstraints import GlobalConstraint
 from ..expressions.core import Expression, Comparison, Operator
@@ -12,6 +13,12 @@ def decompose_global(lst_of_expr, supported={}):
         Accepts a list of flat constraints as input
         Returns a list of flat constraints
     """
+    def _is_supported(cpm_expr):
+        if isinstance(cpm_expr, GlobalConstraint) and cpm_expr.name not in supported:
+            return False
+        if isinstance(cpm_expr, Comparison) and isinstance(cpm_expr.args[0], GlobalConstraint) and cpm_expr.name not in supported:
+            return False
+        return True
 
     if not is_any_list(lst_of_expr):
         lst_of_expr= [lst_of_expr]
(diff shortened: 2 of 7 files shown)


IgnaceBleukx:
Significantly cleaned up the code of the new transformation.
Last design decision we should agree upon is whether to also have an argument to the new transformation `supported_reified`. Now these globals are decomposed in `reify_rewrite` based on the `supported` list we get there.
I'm doubtfull as we also handle global constraint differently in other transformations such as in `only_numexpr_eq`, but no decompositions, so yeah, not sure...
What do you think @JoD, @tias and @Wout4?

cpmpy/expressions/globalconstraints.py
```diff
@@ -369,7 +369,7 @@ def decompose_comparison(self, cpm_op, cpm_rhs):
 
         arr = argval(self.args)
         _min = intvar(-2147483648, 2147483647)
-        return all([any(x <= _min for x in arr), all(x >= _min for x in arr), eval_comparison(cpm_op, _min, cpm_rhs)])
+        return [any(x <= _min for x in arr), all(x >= _min for x in arr), eval_comparison(cpm_op, _min, cpm_rhs)]
```
JoD, line 372: #224 touches this code too. First merge that?

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,78 @@
...
+        decomp_idx = None
+
+        if hasattr(cpm_expr, "decompose") and cpm_expr.is_bool() and cpm_expr.name not in supported:
+            cpm_expr = cpm_expr.decompose() # base boolean global constraints
+        elif isinstance(cpm_expr, Comparison):
+            lhs, rhs = cpm_expr.args
+            if cpm_expr.name == "==" and not _is_supported(lhs): # can be both boolean or numerical globals
+                decomp_idx = 0
+
+            if not _is_supported(cpm_expr):
```
JoD, line 38: can a `Comparison` be not-supported?

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,78 @@
...
+                cpm_expr = do_decompose(cpm_expr) # base global constraint in comparison
+                decomp_idx = None
+
+        elif isinstance(cpm_expr, Operator) and cpm_expr.name == "->":
+            lhs, rhs = cpm_expr.args
+            if not _is_supported(rhs):
+                decomp_idx = 1 # reified (numerical) global constraint, probably most frequent case
+            elif not _is_supported(lhs):
+                decomp_idx = 0
```
JoD, line 47: it cannot be the case that both `lhs` and `rhs` are not supported? In that case, ass an assert and a comment explaining why.

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,78 @@
...
+    for cpm_expr in lst_of_expr:
+        assert isinstance(cpm_expr, Expression), f"Expected CPMpy expression but got {cpm_expr}, run 'cpmpy.transformations.normalize.toplevel_list' first"
+        decomp_idx = None
+
+        if hasattr(cpm_expr, "decompose") and cpm_expr.is_bool() and cpm_expr.name not in supported:
+            cpm_expr = cpm_expr.decompose() # base boolean global constraints
+        elif isinstance(cpm_expr, Comparison):
+            lhs, rhs = cpm_expr.args
+            if cpm_expr.name == "==" and not _is_supported(lhs): # can be both boolean or numerical globals
+                decomp_idx = 0
```
JoD, line 36: how about `_is_supported(lhs)`? Add an assert and comment if this is not possible.

JoD: What about comparisons that are not inequalities? E.g., `x >= abs(y)`? If this case is not possible, add some assert and comment why.

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,78 @@
...
+
+def do_decompose(cpm_expr):
+    """
+        Helper function for decomposing global constraints
+        - cpm_expr: Global constraint or comparison containing a global constraint on lhs
+    """
+    if isinstance(cpm_expr, Comparison):
+        lhs, rhs = cpm_expr.args
+        if lhs.is_bool():
+            return [Comparison(cpm_expr.name, all(lhs.decompose()), rhs)]
```
JoD, line 74: `rhs` should never be decomposed? What happens if I write `max(x)==min(y)`? Add assert and comment.

JoD (commented):
Looks good, but the decompose algorithm has many edge cases that are not covered, probably correctly so. So I'd like some more comment and explanation why these do not need to be covered.

cpmpy/transformations/decompose_global.py
IgnaceBleukx: Yes, both helper functions look in the arguments if passed a comparison, otherwise they look at the top level constraint.

cpmpy/transformations/decompose_global.py
IgnaceBleukx: This is indeed never the case as the constraint should be in flat normal form.

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,78 @@
...
+            elif not _is_supported(lhs):
+                decomp_idx = 0
+
+        if decomp_idx is not None:
+            cpm_expr = copy.copy(cpm_expr)
+            cpm_expr.args[decomp_idx] = all(do_decompose(cpm_expr.args[decomp_idx]))
+            cpm_expr = [cpm_expr]
+
+        if isinstance(cpm_expr, list): # some decomposition happened, have to run again as decomp can contain new global
+            flat = flatten_constraint(cpm_expr)
```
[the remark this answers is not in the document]
IgnaceBleukx: Reifications will almost always have to be flattened again I think.
For example:
```python
p -> AllDifferent(a,b,c)
= p -> And([a != b, a !=c, b != c])
= [p -> a != b, a != c, b != c]
```

For some global constraints its actually even worse. For example, the `GlobalCardinalityCount` constraint decomposes into separate new `Count` constraints...

I do agree it would make a lot of sense to assume a decomposition is flat. BUT then what with nested constraints? Decomposing `AllDifferent(min(x,y,z), abs(x), max(x,y,z))` will create some interesting decomposition I think  : )
Ideally, we have a very efficient check in `flatten_constraint` that checks if a constraint is flat so we do not lose a lot performance when calling `flatten_consrtaint` on a flat constraint...

cpmpy/transformations/decompose_global.py
IgnaceBleukx: Again because we only allow for flat constraints, will add some docs

IgnaceBleukx:
Added some documentation and covered + tested more edge cases.
Most comments are resolved I think, except for merging "bound calculation" first.

IgnaceBleukx:
Merged master with bound calculations into this branch, now `Element` can do a proper check if it is total or not.

cpmpy/transformations/decompose_global.py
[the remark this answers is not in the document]
JoD: `AllDifferent(min(x,y,z), abs(x), max(x,y,z))` is not flattened, right? The flattened version which would appear here is `AllDifferent(a,b,c) & a=min(x,y,z) & b=abs(x) & c=max(x,y,z)`.

You make a good point though. It's ok that we flatten again for now, as long as flattening does not introduce any constraints to decompose.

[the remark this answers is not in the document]
JoD: For the `GlobalCardinalityCount` case: we should decompose until fixpoint, right? Where do we do this?

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,100 @@
...
+        Decomposes any global constraint not supported by the solver
+        Accepts a list of flat constraints as input
+        Returns a list of flat constraints
+
+        - supported: a set of supported global constraints or global functions
+        - supported_reified: a set of supported global constraints within a reification
+
+        Unsupported global constraints are decomposed into equivalent simpler constraints.
+        Numerical global constraints are left reified even if not part of the `supported_reified` set
+            as they can be rewritten using `cpmpy.transformations.reification.reify_rewrite`...
```
JoD, line 21: What is the advantage of this approach?

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,100 @@
...
+                ... reified partial functions are decomposed when unsupported by `supported_reified`
+
+    """
+    def _is_supported(cpm_expr, reified):
+        if isinstance(cpm_expr, GlobalConstraint) and cpm_expr.is_bool():
+            if reified and cpm_expr.name not in supported_reif:
+                return False
+            if reified and cpm_expr.name not in supported:
```
JoD, line 29: should this be `not reified` ? Where is the case for an unsupported, non-reified `alldiff(x,y,z)`?

cpmpy/transformations/decompose_global.py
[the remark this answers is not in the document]
IgnaceBleukx: We call `decompose_global` again on the next line :)

cpmpy/transformations/decompose_global.py
IgnaceBleukx: It means not decomposed and still reified.
The idea is the following:
if `max(x,y,z)` is a supported function in the solver (that is, `max` is in `supported`) then we can rewrite constraints as follows:
```python
bv -> max(x,y,z) <= 10
= [bv -> IV0 <= 10, max(x,y,z) == IV0]
```
So for numerical global constraints, it does not matter if they are reified or not. The rewrite happens in `reify_rewrite`. I do agree it is a bit messy but doing the `reify_rewrite` here also seems weird...

There is one edge case where we cannot do the rewrite trick: for non-total functions such as element as the following is not necessarily equivalent:
```python
bv -> [x,y,z][idx] <= 10
= [bv -> IV0 <= 10, [x,y,z][idx] == IV0]
```
as `idx` might be out of bounds if `bv` is `False`. So in that case we do have to decompose...

JoD: Also, what is a reified numerical global? Is it `count(...) > x` or `p -> count(...) > x`? Will this method decompose the latter?

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,100 @@
...
+                return False
+            if reified and not cpm_expr.args[0].is_total() and cpm_expr.args[0].name not in supported_reif:
+                # edge case for partial functions as they cannot be rewritten to non-reified versions
+                #  have to decompose to total functions
+                #  ASSUMPTION: the decomposition for partial global functions is total! (for now only Element)
+                return False
+
+        return True
+
+    assert supported_reif <= supported, "`supported` set is assumed to be a subset of the `supported_reified` set"
```
JoD, line 44: So something can be supported reified but not supported non-reified? That seems counter-intuitive.

IgnaceBleukx: Other way around.. The test is correct but the message is not

cpmpy/transformations/decompose_global.py
JoD: Aha, I got it now. Thanks for the explanation! Please add the above explanation to the comments - other developers and your future self will be grateful ;)

cpmpy/transformations/decompose_global.py
```diff
@@ -0,0 +1,100 @@
...
+    for cpm_expr in lst_of_expr:
+        assert isinstance(cpm_expr, Expression), f"Expected CPMpy expression but got {cpm_expr}, run 'cpmpy.transformations.normalize.toplevel_list' first"
+        decomp_idx = None
+
+        if hasattr(cpm_expr, "decompose") and not _is_supported(cpm_expr, reified=False):
+            cpm_expr = cpm_expr.decompose() # base boolean global constraints
+        elif isinstance(cpm_expr, Comparison):
+            lhs, rhs = cpm_expr.args
+            if cpm_expr.name == "==" and not _is_supported(lhs, reified=True): # rhs will always be const/var
+                decomp_idx = 0 # this means it is an unsupported reified boolean global
```
JoD, line 58: could it also be a numerical global? E.g., `count(...)=x`.

cpmpy/transformations/decompose_global.py
JoD: ok

cpmpy/transformations/decompose_global.py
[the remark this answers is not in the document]
JoD: excellent :)

cpmpy/transformations/decompose_global.py
JoD: ok

JoD (approved):
Allright, I'm pretty sure this is not the final word on decomposing globals, but it is a good step in the right direction. I've got two open comments left that should be easy to resolve, so the PR gets an approve.

Broader concerns:
- what about `abs`? This one should simply be a global constraint.
- this decomposition code assumes a lot about the form of the constraints (e.g., flattened). Based on these assumptions, we get a lot of if-then cases that are handled, and now we hope that we handled all possibilities. A different way of decomposing would be to traverse the expression trees, keeping track of whether we are in a reified or non-reified (top-level) expression. We return the appropriate decomposition to the parent, who then replaces its child. This is a generic way of decomposing, which to me looks less hard to maintain. But it is out of the scope of this MR.

IgnaceBleukx:
I agree we should make `abs` a global constraint indeed. We can then provide a `decompose_comparison` for that one as well.

Regarding decomposing globals in the expression tree, I thought about it as well but there are many edges cases which are unclear how to handle. `
For example lets assume we have the following situation: a solver supporting `AllDifferent` but not `Max`
Then how do we decompose the max constraints in the constraint below?
`AllDifferent(max(a,b,c), max(x,y,z), max(p,q,r))`? We have a `decompose_comparison` for numerical global functions like these but they do not appear in a comparison in this case! 
This leaves us with 2 options: decomposing the `AllDifferent` into pairwise inequalities, then do `decompose_comparison` for every `max` expression or flatten the constraint to [AllDifferent(IV0, IV1, IV2), max(a,b,c) = IV0, max(x,y,z) = IV1, max(p,q,r) = IV2]`

So it is to prevent these cases I opted for decomposing the globals after flattening

cpmpy/transformations/decompose_global.py
JoD: makes sense

PR #250, opened 2023-03-08
248 cannot deepcopy model with nested list of constraints

IgnaceBleukx:
Changed deepcopy functionality to work with `copy.deepcopy()`

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -174,6 +174,12 @@ class Expression(object):
         copied_args = self._deepcopy_args(memodict)
         return type(self)(self.name, copied_args)
 
+    # keep for backwards compatibility
+
+    def deepcopy(self, memodict={}):
+        warnings.warn("Deprecated, use copy.deepcopy() instead, will be removed in stable version", DeprecationWarning)
+        return copy.deepcopy(self, memodict)
+
     # implication constraint: self -> other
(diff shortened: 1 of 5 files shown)


tias:
Do we even need to overwrite the builtin deepcopy?

```
import copy
class cls:
    def __init__(self,a,b):
        self.a = a
        self.b = b
    
    def __repr__(self):
        return str((self.a, self.b))
x = cls("a", [1,2])
print(x)
print(copy.deepcopy(x))
```
this works... so, do we really need our own?

IgnaceBleukx:
Hmm, removing our custom `__deepcopy__` indeed does not break any tests.
I looked around a little bit and it seems there are two main reasons why you would want to override `__deepcopy__`:
1) To copy some attributes shallow and other deep (not applicable for us I think)
2) If all objects should be made by a `factory`. I.e., a custom function that creates the objects, potentially with some bookkeeping. This is much like our `intvar` or `boolvar` function which ups our magic counters every time it is called. However, when copying variables, we do **not** want the counter to be upped as the names should stay the same. So this case is also not applicable to our application.

So in summary, we can probably rely on the default implementation of `copy.deepcopy` and leave out our custom implementations...

tias:
Fantastic code reduction : ) Do push and get back to/close Helge's

PR #253, opened 2023-03-15
New advanced example: CPMpyXplain algorithm

VincentDeBoulle:


tias:
Thanks for the contribution; I confirm it runs fine. We will wait till after the project deadline to do a detailed review or changes, but it already looks solid as is.

IgnaceBleukx (changes requested):
Nice demonstration of the QuickXplain algorithm!
The logic looks good for the most part but the style of the example can use some love, I left some comments on what could be good to change before we merge it into master.

I really like the "iteration mode" as indeed the original paper relies heavily on the (actually arbitrary) ordering of the constraints...
When I use the "iteration mode", I get multiple of the same explanations, is that considered correct? Maybe you can check if the explanation is already found and not print it again?

Lastly: we have added a `.set_description()` option for constraints, maybe you can use it to nicely print the constraint relaxations here?

Dimosts:
Are there any news about it? It has been open for long time with minimal requested changes

tias:
I think (a later version of) the paper is now published in ICTAI24: https://ieeexplore.ieee.org/abstract/document/10849371/

So this would require checking if things changed, together with the style changes of Ignace' review.

More things changed since this example was proposed, most notably our XCP-Explain tutorial which now contains the original QuickXplain and many sorts of explanation algorithms. https://github.com/CPMpy/XCP-explain

Reworking this example seems to be low priority for everybody, so I'll just close it as unfinished then.

PR #257, opened 2023-03-21
255 comparison expression has no get bounds

Wout4:
default case for get_bounds


Wout4:
Seems straightforward to add this basecase to expressions, but asking for a review anyway :)

PR #266, opened 2023-03-29
Add chess_position example

KennyVDV:
The model makes a chess board that can be played in an actual game of chess from the probabilities of every piece and the squares where pieces are present.


examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+from enum import Enum
+from functools import total_ordering
+
+
+# Replace numerals by letters for easier interpretation
+def replace_value(x):
+    return dict.get(x, x)
+
+
+piece_probabilities = np.array([[-1.1558e+01, -1.8940e-01, -1.0785e+01, -8.4097e+00, -1.3034e+01, -8.7986e+00, -2.6694e+00,  1.2009e+01, -3.5726e+00, -9.1412e-01, -2.0600e+00, -5.9005e-01],
```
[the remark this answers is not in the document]
Dimosts: So, these are not probabilities. I assume it is the output of the NNs for the prediction. It is a bit weird for output, with the range of the values. It should be explained how you got them.

examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+                                True, False,  True,  True,  True, False, False, False,
+                               False, False, False,  True, False,  True, False,  True,
+                               False, False,  True, False,  True, False,  True, False,
+                                True, False, False, False,  True, False, False, False])
+
+# Variable to represent the whole board
+board = intvar(0, 12, (8,8), "board")
+
+# Variable to represent the pieces, with first row the probabilies and second row the indexes
+pieces = intvar(0, 10000000, (2, len(piece_probabilities)), name="pieces")
```
Dimosts, line 56: I understand that this is used, in the objective function, to give the value from the respective "probability". However, this is not the right way to do this. This slows down a lot the solving process, introducing so many variables with so large domains, including also all the constraints needed for this. See below for my comment on the objective function.

examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+
+# Make the probabilites into reasonable big values
+pieces_values = np.empty([len(piece_probabilities), len(piece_probabilities[0])], dtype=int)
+for i in range(len(piece_probabilities)):
+    for j in range(len(piece_probabilities[0])):
+        pieces_values[i,j] = int((piece_probabilities[i,j]*10000).item())
```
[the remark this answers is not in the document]
Dimosts: Basically you convert the "probabilities" (not really probabilities) in big values. Why not just divide with a small number defining the precision? e.g.

PRECISION = 1e-4
pieces_values = piece_probabilities / PRECISION
pieces_values = cpm_array(np.int64(lprobs)) # This is to convert to int + to convert the array to cpm_array, so that you can directly use it in the objective function

examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+
+# Shift all values in the array to be positive
+pieces_values = pieces_values - min_neg_arr
+
+pieces_values = pieces_values.astype(int)
+
+# All the constraints that a chess board has to follow
+model = Model(
+    # Both colors have only one king
+    (sum(board[l,c] == k for l in range(0,8) for c in range(0,8)) == 1),
```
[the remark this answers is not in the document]
Dimosts: This is just a  
Count(board, k) == 1,
Count(board, K) == 1,
...

examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+    (sum(board[l,c] == K for l in range(0,8) for c in range(0,8)) == 1),
+    # Both colors have maximum 16 pieces
+    (sum(board[l,c] < 6 for l in range(0,8) for c in range(0,8)) <= 16),
+    (sum(((board[l,c] >= 6) & (board[l,c] < 12)) for l in range(0,8) for c in range(0,8)) <= 16),
+    # Both colors have maximum 8 pawns
+    (sum(board[l,c] == p for l in range(0,8) for c in range(0,8)) <= 8),
```
Dimosts, line 92: Again, 
    Count(board, p) <= 8,
    Count(board, P) <= 8,

examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+    (sum(board[l,c] == P for l in range(0,8) for c in range(0,8)) <= 8),
+    # Pawns can't be on the edge of the board
+    (sum(board[0,c] == p for c in range(0,8)) == 0),
+    (sum(board[0,c] == P for c in range(0,8)) == 0),
+    (sum(board[7,c] == p for c in range(0,8)) == 0),
+    (sum(board[7,c] == P for c in range(0,8)) == 0),
```
Dimosts, line 98: In general the constraints can be written in a better way,  See previous comments

examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+    (board[1,6] == p).implies(~((board[0,7] == b))),
+    (board[1,1] == p).implies(~((board[0,0] == b))),
+    # If the pawns in column 2 or 7 didn't move, bishop of other color can't be in the corners of the board excpet when promoted
+    ((board[6,1] == P) & (board[7,0] == B)).implies(sum(board[l,c] == p for l in range(0,8) for c in range(0,8)) <= 7),
+    ((board[6,6] == P) & (board[7,7] == B)).implies(sum(board[l,c] == p for l in range(0,8) for c in range(0,8)) <= 7),
+    ((board[1,6] == p) & (board[0,7] == b)).implies(sum(board[l,c] == P for l in range(0,8) for c in range(0,8)) <= 7),
+    ((board[1,1] == p) & (board[0,0] == b)).implies(sum(board[l,c] == P for l in range(0,8) for c in range(0,8)) <= 7),
+)
+
+# Put the index of the maximum value into the second row of the pieces variable
```
Dimosts, line 133: The following 'for' is not needed, it is overcomplicated for something that can be done in a much more straightforward way, and it will make the solving process much slower too.

You try to match the probabilities with the pieces, while you already have the probabilities and you already have the pieces in different arrays, but you created new variables for the matching although it is 1 to 1.

You have 22 pieces in this example. why not match directly the 22 cells in the board that are predicted to have a piece (which you can just take by using board[occupancy_in_board]) with the 22 rows of pieces_values? You can do that in the objective function and avoid all these variables, and constraints.

examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+        | ((pieces[0,i] == pieces_values.item((i,6))) & (pieces[1,i] == 6)) | ((pieces[0,i] == pieces_values.item((i,7))) & (pieces[1,i] == 7)) | ((pieces[0,i] == pieces_values.item((i,8))) & (pieces[1,i] == 8)) | ((pieces[0,i] == pieces_values.item((i,9))) & (pieces[1,i] == 9)) | ((pieces[0,i] == pieces_values.item((i,10))) & (pieces[1,i] == 10)) | ((pieces[0,i] == pieces_values.item((i,11))) & (pieces[1,i] == 11)))
+
+# Put the pieces into the right place of the board
+index = 0
+for i in range(len(occupancy_in_board)):
+    if occupancy_in_board[i]:
+        model += (board[int(i/8), i%8] == pieces[1, index])
+        index += 1
+    else:
+        model += (board[int(i/8), i%8] == E)
```
Dimosts, line 145: The above for is not needed either.

To force the empty cells, you can just do 
model += board[occupancy_in_board == False] == E

and for the non empty cells, you can link directly with the probabilities in the objective function. So, you use only the board variables.

examples/chess_position.py
```diff
@@ -0,0 +1,159 @@
...
+
+# Maximize the probabilites of the pieces to get the most likely board
+model.maximize(sum(pieces[0]))
```
Dimosts, line 148: Based also on my comments before, regarding the variables for pieces that are not needed, I suggest to replace with

obj = sum(lp[v] for lp, v in zip(pieces_values, board[occupancy_in_board]))
model.maximize(obj)

This way, you directly linked the probabilities in pieces_values, with the cells in the board that need to be predicted.

Already tried the above and yield to the same solution.

And importantly, from 264 seconds that were needed to solve, it dropped to less than 0.5s.

Dimosts (changes requested):

Dimosts (approved):

PR #274, opened 2023-03-31
Fix 261

### LINKED ISSUE #261: negated_normal crashes when input contains a boolexpr compared with an numexpr.
Bug found with fuzztesting, running on flatten_constraint

example: iv = intvar(-8,8, shape=3)

cons = [(iv[0] == iv[1]) == (iv[2] + 3)]
negated_normal(cons[0])

example of how flatten leads to this bug:

cons = [(iv[0] == iv[2]).implies((iv[0] == iv[1]) == (iv[2] + 3)) == 0]
flatten_constraint(cons)

JoD:
Makes use of the invariant that reifications always have bool expressions on both sides of equality to check for them.

Fixes #261


Wout4:
Looks good, clean fix!

PR #275, opened 2023-04-12
add test for ortools version

IgnaceBleukx:
Added a test to check OR-tools version as version 9.5 has known bugs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 numpy
-ortools >= 9.0
+ortools >= 9.6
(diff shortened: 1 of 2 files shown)


tias:
bit unusual to add a test for it, which is only run by developers?

I think it would also be good to update or requirements.txt to require a higher version then.

IgnaceBleukx:
Hmm yes, the idea was to flag ourselves when other tests are failing that this could be due to an old OR-tools version.
I updated the requirements as well now

PR #277, opened 2023-04-17
minizinc exitstatus fix when time-out

Wout4:
written carefully with a lot of cases because we noticed earlier that minizinc sometimes gives time as an int and sometimes as a timedelta and has different names for time

### DIFF
--- a/cpmpy/solvers/minizinc.py
+++ b/cpmpy/solvers/minizinc.py
@@ -225,30 +225,11 @@ class CPM_minizinc(SolverInterface):
         self.cpm_status = SolverStatus(self.name)
         runtime = 0
         if 'time' in mzn_result.statistics:
-            time = mzn_result.statistics['time']
-            if isinstance(time, int):
-                self.cpm_status.runtime = time / 1000
-            elif isinstance(time, timedelta):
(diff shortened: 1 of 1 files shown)


cpmpy/solvers/minizinc.py
```diff
@@ -223,13 +223,34 @@ def _post_solve(self, mzn_result):
...
+        runtime = 0
         if 'time' in mzn_result.statistics:
             time = mzn_result.statistics['time']
             if isinstance(time, int):
                 self.cpm_status.runtime = time / 1000
             elif isinstance(time, timedelta):
                 self.cpm_status.runtime = time.total_seconds()  # --output-time
-
+        else:
+            if 'flatTime' in mzn_result.statistics:
```
IgnaceBleukx, line 234: Is `mzn_result.statistics` just a dictionary? 
If so you can reduce this code somewhat using:
```python
runtime += mzn_time_to_seconds(mzn_result.statistics.get('flatTime',0))
```
The second argument of the `.get` method of a dictionary is a default value for it the key is not present

IgnaceBleukx (commented):
Nice, left some suggestions to reduce the code

PR #282, opened 2023-04-24
check bounds of minizinc vars

Wout4:


IgnaceBleukx:
Looks good, maybe just move the new Exception up in the `exceptions.py` file so all minizinc exceptions are somewhat grouped?

Wout4:
Allright, will merge with the exceptions grouped together

PR #284, opened 2023-04-27
283 use BoolVal in stead of True and False

Wout4:
Had to do a little bit more work than just replacing the return true and false so made it into a pr

### DIFF
--- a/cpmpy/expressions/python_builtins.py
+++ b/cpmpy/expressions/python_builtins.py
@@ -20,7 +20,7 @@
 """
 import numpy as np
 from .variables import NDVarArray
-from .core import Expression, Operator
+from .core import Expression, Operator, BoolVal
 from .globalconstraints import Minimum, Maximum
 
 # Overwriting all/any python built-ins
@@ -35,7 +35,7 @@ def all(iterable):
     collect = [] # logical expressions
     for elem in iterable:
         if elem is False or elem is np.False_:
-            return False  # no need to create constraint
+            return BoolVal(False)  # no need to create constraint
         elif elem is True or elem is np.True_:
             pass
         elif isinstance(elem, Expression) and elem.is_bool():
@@ -46,8 +46,7 @@ def all(iterable):
         return collect[0]
     if len(collect) >= 2:
         return Operator("and", collect)
-    return True
-
+    return BoolVal(True)
 # any: listwise 'or'
 def any(iterable):
     """
@@ -59,7 +58,7 @@ def any(iterable):
     collect = [] # logical expressions
     for elem in iterable:
         if elem is True or elem is np.True_:
-            return True # no need to create constraint
+            return BoolVal(True) # no need to create constraint
         elif elem is False or elem is np.False_:
             pass
         elif isinstance(elem, Expression) and elem.is_bool():
@@ -70,7 +69,7 @@ def any(iterable):
         return collect[0]
     if len(collect) >= 2:
         return Operator("or", collect)
-    return False
+    return BoolVal(False)
 
 def max(iterable):
     """
--- a/cpmpy/expressions/python_builtins.py
+++ b/cpmpy/expressions/python_builtins.py
@@ -19,6 +19,8 @@
         sum
 """
 import numpy as np
+
+from .utils import is_false_cst, is_true_cst
 from .variables import NDVarArray
 from .core import Expression, Operator, BoolVal
 from .globalconstraints import Minimum, Maximum
@@ -34,9 +36,9 @@ def all(iterable):
     if isinstance(iterable, NDVarArray): iterable=iterable.flat # 1D iterator
     collect = [] # logical expressions
     for elem in iterable:
-        if elem is False or elem is np.False_:
+        if is_false_cst(elem):
             return BoolVal(False)  # no need to create constraint
-        elif elem is True or elem is np.True_:
+        elif is_true_cst(elem):
             pass
         elif isinstance(elem, Expression) and elem.is_bool():
             collect.append( elem )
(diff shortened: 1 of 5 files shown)


Wout4:
time for test suite:
3154 passed, 1 skipped, 4 warnings in 460.55s (0:07:40)

vs master
3154 passed, 1 skipped, 2 warnings in 512.05s (0:08:32) 
On my machine which I expect to not be very consistent

tias:
In retrospect I am not sure how reliable the testsuite is for time-benchmarking. I guess we would need a repo of exectuable python scripts that return a model, not even the pickles we have in bigtest, if we also want to benchmark model creation time... that is definitely for later ; )

Also you would have to compare the difference of just 1 commit to know the effect of that one change.


Anyway, on master all pass tests in ~3.5 minutes here, but this branch

FAILED tests/test_examples.py::test_examples[./examples/knapsack.py] - TypeError: int() argument must be a string, a bytes-like object or a number, not 'Operator'
FAILED tests/test_examples.py::test_examples[./examples/rcpsp.py] - AssertionError: IntVar upperbound must be integer <class 'cpmpy.expressions.core.Operator'> sum([0, 3, 2, 5, 4, 2, 3, 4, 2, 4, 6, 0])
FAILED tests/test_examples.py::test_examples[./examples/bus_schedule.py] - AssertionError: IntVar upperbound must be integer <class 'cpmpy.expressions.core.Operator'> sum([8, 10, 7, 12, 4, 4])
FAILED tests/test_examples.py::test_examples[./examples/jobshop.py] - AssertionError: IntVar upperbound must be integer <class 'cpmpy.expressions.core.Operator'> sum([3, 2, 2, 2, 1, 4, 0, 4, 3])
======================================= 4 failed, 3250 passed, 3 warnings in 143.88s (0:02:23) ========================================


e.g.
```
tests/test_examples.py:35: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
<frozen importlib._bootstrap_external>:790: in exec_module
    ???
<frozen importlib._bootstrap>:228: in _call_with_frames_removed
    ???
examples/rcpsp.py:22: in <module>
    start_time = intvar(0, max_duration, shape=nb_jobs)
cpmpy/expressions/variables.py:172: in intvar
    data = np.array([_IntVarImpl(lb,ub, name=_genname(name, idxs)) for idxs in np.ndindex(shape)]) # repeat new instances
cpmpy/expressions/variables.py:172: in <listcomp>
    data = np.array([_IntVarImpl(lb,ub, name=_genname(name, idxs)) for idxs in np.ndindex(shape)]) # repeat new instances
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[AttributeError("'_IntVarImpl' object has no attribute 'name'") raised in repr()] _IntVarImpl object at 0x7f1ac53c60d0>, lb = 0
ub = sum([0, 3, 2, 5, 4, 2, 3, 4, 2, 4, 6, 0]), name = None
```

I guess if it passes the tests we should not be too worried by another function call at creation time, at this point.

Wout4:
ah the examples don't run on my machine (and also not on the automated checks?)

cpmpy/expressions/python_builtins.py
```diff
@@ -34,9 +36,9 @@ def all(iterable):
     if isinstance(iterable, NDVarArray): iterable=iterable.flat # 1D iterator
     collect = [] # logical expressions
     for elem in iterable:
-        if elem is False or elem is np.False_:
-            return False  # no need to create constraint
-        elif elem is True or elem is np.True_:
+        if is_false_cst(elem):
+            return BoolVal(False)  # no need to create constraint
```
tias, line 40: I don't think we can/should in general change the return type of a Python builtin...

'any' can be used a lot in an arbitrary code-base that only includes cpmpy for a small part

maybe if we check that at least one element is a CPMpy expr, but for now I will just undo this

cpmpy/expressions/python_builtins.py
```diff
@@ -70,14 +71,14 @@ def any(iterable):
...
-    return False
+    return BoolVal(False)
 
 def max(iterable):
     """
         max() overwrites python built-in,
         checks if all constants and computes np.max() in that case
     """
-    if not any(isinstance(elem, Expression) for elem in iterable):
+    if is_false_cst(any(isinstance(elem, Expression) for elem in iterable)):
```
tias, line 81: this already shows that overwriting 'any' is not a good idea...

Note that I now implemented `__bool__()` for BoolVal, so the original code would work now (with not), but still, better without for now

cpmpy/expressions/python_builtins.py
```diff
@@ -97,6 +98,6 @@ def sum(iterable):
         otherwise, makes a sum Operator directly on `iterable`
     """
     iterable = list(iterable) # Fix generator polling
-    if not any(isinstance(elem, Expression) for elem in iterable):
+    if is_false_cst(any(isinstance(elem, Expression) for elem in iterable)):
```
tias, line 101: I think this is what is breaking the tests on my machine
(does this mean that the examples are not run on the github CI? nor on your computer?)

tias (approved):
see comments for change on built-in overwrites.

will merge

PR #286, opened 2023-04-28
Flatten does not decompose

Wout4:


Wout4:
the change to get_or_make_var is a bit hacky, maybe there is a nicer way to achieve the same result

Wout4:
Actually I do like it after considering it again, since ~bv == flatexpr is just as flat as bv == flatexpr, and this way we need not change any other code

tias:
if this is a cleanup, why did you close it unmerged? or it is superseded by another PR?

Wout4:
I extracted the only useful code into the get_or_make_var pr
This addressed the fact that we now decompose in flatten, but that will be resolved with the push down negation and recursive decomposition transformations

PR #287, opened 2023-04-28
265 xor of circuit global constraints crashes ortools

Wout4:
Works now after merging the flatten does not decompose branch, so review that one first ;)


Wout4:
This change is no longer needed after we made sure that flatten does not decompose globals. this way the decomposition is never used in a reification, because the only reified context that flatten now outputs is circuit == bv

PR #291, opened 2023-05-09
288 simplify boolean contstants

IgnaceBleukx:
Implemented transformation to convert boolean constants into integers.

Related but not implemented are conversions of simple boolean expressions such as BV >= 1 to True.
It can also be done in this transformation, but then it should get a new name such as `simplify_boolean` or something like that. 
The question is then how far do we want to go in this? In essence, we are checking if constraints posted to a solver are "satisfied by construction" and we can do this for many other expressions other than simple boolean expressions.


cpmpy/transformations/normalize.py
```diff
@@ -34,3 +35,70 @@ def unravel(lst, append):
...
+
+
+def simplify_bool_const(lst_of_expr, num_context=False):
+    """
+    removes boolean constants from all CPMpy expressions
+    only resulting boolean constant is literal 'false'
+    - list_of_expr: list of CPMpy expressions
+    """
+
+    def _is_false(expr):
```
tias, line 47: I just merged Wout's _is_false_cst in utils, which also checks for numpy bool and should be used here instead of defining new ones

tests/test_constraints.py
```diff
@@ -92,21 +92,25 @@ def comp_constraints(solver):
...
         for numexpr in numexprs(solver):
-            for rhs in [NUM_VAR, 1]:
+            for rhs in [NUM_VAR, 1, BoolVal(True)]:
                 yield Comparison(comp_name, numexpr, rhs)
 
     for comp_name in Comparison.allowed:
         for glob_expr in global_constraints(solver):
             if not glob_expr.is_bool():
-                for rhs in [NUM_VAR, 1]:
+                for rhs in [NUM_VAR, BoolVal(True)]:
```
tias, line 101: I think you wanted to keep '1' too as above

tias (commented):
for logical constraints (and, or, impl), this is currently already done in our overloadings of the python operators (at factory time).

This kind of constraint-level simplification can also be done after flatten.

I think the case we want more is the other case you are describing, where we are using a bool (const or var) as an int, in a comparison with another int, e.g. `bv >= 1`, `bv != 0` etc.

Also, I'm not sure that your code does a full traversal of the expression tree, e.g. I dont think you walk over the args of every expressions upto their leafs (variables or constants)?

perhaps we should do the other case in a separate pull request

cpmpy/transformations/normalize.py
IgnaceBleukx: Aha I missed that one, will change it

IgnaceBleukx:
Yes, I know for some operators we simpify at construction time (but for example not if you construct an `Operator` using its constructor directly). We yet have to converge on the question of what should be simplified at construction and what should be done later I think.

I forgot to add the case of global constraints, then it should actually traverse the entire tree [line 64](https://github.com/CPMpy/cpmpy/blob/efddba737fe39976612781ecbdfa466a081e4bdc/cpmpy/transformations/normalize.py#L64) and [line 99](https://github.com/CPMpy/cpmpy/blob/efddba737fe39976612781ecbdfa466a081e4bdc/cpmpy/transformations/normalize.py#L99)

The boolean simplification will basically be a switch statement summarizing the following table:

|x|<0 | 0 | 1 | > 1 |
|---|---|---|---|---|
|==|F | ~x|x | F|
|!= |T | x | ~x | T|
|>| T | x | F | F |
|<| F | F |~x | T|
|>=| T | T | x | F|
|<=| F | ~x | T | T

Maybe it does deserve to live in this pull request/transformation as well. I'll implement it and we can still remove it if this does not seem to be the right spot anyway.

IgnaceBleukx:
Updated with some more tests, I think it ready to merge now.

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,94 @@ def unravel(lst, append):
...
+            newlist.append(int(expr.value()) if num_context else expr)
+
+        elif isinstance(expr, Operator):
+            args = simplify_boolean(expr.args, num_context=not expr.is_bool())
+
+            if expr.name == "or":
+                if any(is_false_cst(arg) for arg in args):
+                    newlist.append(1 if num_context else BoolVal(True))
+                else:
+                    newlist.append(Operator("or", [arg for arg in args if not isinstance(arg, BoolVal)]))
```
tias, line 66: I think this would not clean a 'true' or np.true; I also think we are needlessly copying a lot of objects that do not change.

But I also don't have a solution straight away... will think about it

IgnaceBleukx: I could just check for `is_false_cst` instead of the `BoolVal` here? (The `is_false_cst` on line 63 should be `is_true_cst`!) That check will also hold for np constants.

Not sure if we can do anything about the copying... Except for changing args in place but some transformation will have to copy them at least once as the users constraints should remain unchanged

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,96 @@ def unravel(lst, append):
...
+
+        if isinstance(expr, bool):
+            # not sure if this should happen here or at construction time
+            expr = BoolVal(expr)
+
+        if isinstance(expr, BoolVal):
+            newlist.append(int(expr.value()) if num_context else expr)
+
+        elif isinstance(expr, Operator):
+            args = simplify_boolean(expr.args, num_context=not expr.is_bool())
```
JoD, line 60: Comparisons are Boolean, but their arguments can be int? So I guess we should upcast those too?

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,96 @@ def unravel(lst, append):
...
+                if any(is_true_cst(arg) for arg in args):
+                    newlist.append(1 if num_context else BoolVal(True))
+                else:
+                    newlist.append(Operator("or", [arg for arg in args if not isinstance(arg, BoolVal)]))
+
+            elif expr.name == "and":
+                if any(is_false_cst(arg) for arg in args):
+                    newlist.append(0 if num_context else BoolVal(False))
+                else:
+                    newlist.append(Operator("and", [arg for arg in args if not isinstance(arg, BoolVal)]))
```
JoD, line 72: If the and is empty, return true/1

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,96 @@ def unravel(lst, append):
...
+                    newlist.append(BoolVal(True))
+                else:
+                    newlist.append(~args[0])
+
+            else: # numerical expressions
+                newlist.append(Operator(expr.name, args))
+
+        elif isinstance(expr, Comparison):
+            lhs, rhs = simplify_boolean(expr.args, num_context=True)
+            if (isinstance(lhs, int) and isinstance(rhs, _BoolVarImpl)) or isinstance(lhs, BoolVal) and isinstance(rhs, _IntVarImpl):
```
JoD, line 98: Swap the comparison symbol too? Why swap?

cpmpy/transformations/normalize.py
JoD: I'm missing something here... args is empty for comparisons?

JoD: Comparisons are not operators :)

JoD: Maybe add a `TODO: check code in typecheck merge request. There may be an operator -> argument type array there`.

cpmpy/transformations/normalize.py
JoD: second case cannot happen, as `BoolVal` should be upcasted to `int` by line 97.

JoD: So only case where we can do something smart is when one of the arguments is `int` and the other is `BoolVal` - Ignace.

JoD: TODO: make sure comparison symbol is swapped too!

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,96 @@ def unravel(lst, append):
...
+                lhs, rhs = rhs, lhs
+            if isinstance(lhs, _BoolVarImpl) and isinstance(rhs, int):
```
JoD, line 100: The central idea is that a comparison of a Boolean variable to an integer value can always be simplified to a simple Boolean expression. E.g., `p <= 0 simplifies to `~p`.

cpmpy/transformations/normalize.py
JoD: TODO: note somewhere that we do not simplify comparisons between two integer expressions and between two Boolean expressions in general. We only simplify Boolean variables compared to integer values.

JoD:
After previous meeting: Tias suggested to not only check for Boolean variables compared to integer values, but also to Boolean expressions compared to integer values. Ignace suggested this would not be a lot of work and he would update the PR.

IgnaceBleukx:
Adapted as Jo described above.
Also added some documentation to make it a bit more clear what is happening.
TODO still is some extra testing simplifying boolean expressions

cpmpy/solvers/z3.py
```diff
@@ -250,7 +250,7 @@ def transform(self, cpm_expr):
         # Z3 supports nested expressions, so no transformations needed
         # that also means we don't need to extract user variables here
         # we store them directly in `solver_var()` itself.
-        return toplevel_list(cpm_expr)
+        return simplify_boolean(toplevel_list(cpm_expr))
```
tias, line 253: also z3 has its own simplifiers, we should just directly pass the model through

cpmpy/transformations/normalize.py
tias: currently all operators with bool output type have bool input types. I'm not sure where this should be documented... we also don't really check it here, it just influences whether '1' or True is returned.

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,118 @@ def unravel(lst, append):
...
+    only resulting boolean constant is literal 'false'
+    - list_of_expr: list of CPMpy expressions
+    """
+
+    newlist = []
+    for expr in lst_of_expr:
+
+        if isinstance(expr, bool):
+            # not sure if this should happen here or at construction time
+            expr = BoolVal(expr)
```
tias, line 54: there can be Py bools here yes.

you can check the num_context here and add to newlist directly, instead of relying on the next if to do that then (which can then become an elif)

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,118 @@ def unravel(lst, append):
...
+
+        if isinstance(expr, BoolVal):
+            newlist.append(int(expr.value()) if num_context else expr)
+
+        elif isinstance(expr, Operator):
+            args = simplify_boolean(expr.args, num_context=not expr.is_bool())
+
+            if expr.name == "or":
```
tias, line 62: since we are rewriting or/and/impl here anyway:
in a follow-up PR, we should take the optimisations of nested 'and' and implications of ands etc and move them here. That will significantly easy the flatten code.

that will be most powerful after push down negation and decompose is preceding this transfo; so we can do those first

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,118 @@ def unravel(lst, append):
...
+                    newlist.append(~args[0])
+
+            else: # numerical expressions
+                newlist.append(Operator(expr.name, args))
+
+        elif isinstance(expr, Comparison):
+            lhs, rhs = simplify_boolean(expr.args, num_context=True)
+            name = expr.name
+            if isinstance(lhs, int) and is_boolexpr(rhs): # flip arguments of comparison to reduct nb of cases
+                flipmap = {"==":"==", "!=":"!=", "<=":">=", "<":">"}
```
tias, line 108: that is just 2 ifs instead of creating a set and doing a lookup in it?

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,118 @@ def unravel(lst, append):
...
+            >  |  T	 x	 F	 F
+            <  |  F	 F	~x	 T
+            >= |  T	 T	 x	 F
+            <= |  F	~x	 T	 T
+            """
+            if is_boolexpr(lhs) and isinstance(rhs, int):
+                # direct simplification of boolean comparisons
+                if rhs < 0:
+                    newlist.append(BoolVal(name in  {"!=", ">", ">="})) # all other operators evaluate to False
+                if rhs == 0:
```
tias, line 126: these are cleaner to be elifs?

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,118 @@ def unravel(lst, append):
...
+                        newlist.append(~lhs)
+                    if name == ">":
+                        newlist.append(BoolVal(False))
+                    if name == "<=":
+                        newlist.append(BoolVal(True))
+                if rhs > 1:
+                    newlist.append(BoolVal(name in  {"!=", "<", "<="})) # all other operators evaluate to False
+            else:
+                newlist.append(eval_comparison(name, lhs, rhs))
+        elif hasattr(expr, "decompose"):
```
tias, line 148: a GlobalConstraint does not require to have a decompose function, please check for isinstance(GlobalConstraint)

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,118 @@ def unravel(lst, append):
...
+            expr = copy.deepcopy(expr)
```
tias, line 149: a deepcopy will copy the args and all children of the args.

But here, we 'recreate' the args ourselves, so a shallow copy where we overwrite the args is what we need (otherwise could be useless performance penalty)

cpmpy/transformations/normalize.py
```diff
@@ -34,3 +37,118 @@ def unravel(lst, append):
...
+            expr.args = simplify_boolean(expr.args) # TODO: how to determine boolean or numerical context?
+            newlist.append(expr)
+        else: # variables/constants
```
tias, line 152: # variables/constants/DirectConstraints

tias (commented):
some detailed feedback

cpmpy/solvers/minizinc.py
```diff
@@ -382,7 +382,7 @@ def transform(self, cpm_expr):
 
         :return: list of Expression
         """
-        return toplevel_list(cpm_expr)
+        return simplify_boolean(toplevel_list(cpm_expr))
```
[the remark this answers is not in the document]
Wout4: I added this because it fixes a bug where minizinc does not recognice 'False' when doing something like IV == False, but that can indeed be done with a small transformation as well

cpmpy/transformations/normalize.py
IgnaceBleukx: So for now we simply leave the check as is?

cpmpy/transformations/normalize.py
IgnaceBleukx: The 'decompose' function is part of the superclass `GlobalConstraint` so the check should actually be equivalent.
But I agree checking with isinstance is cleaner

cpmpy/transformations/normalize.py
JoD: Ok, will make a follow-up issue to not forget this.

JoD: See issue #345

PR #292, opened 2023-05-09
DirectConstraint (a part of #121 nativeCons)

tias:
only the parts relating to DirectConstraint from #121 

with some small fixes.

TODO:

- [x] nonogram: works (again) but the explanation is too short. What is a nonogram?  Also the visualisation can be more pedagogic, e.g. show what the image originally is, then what it looks like after solving. The nonogram part should be clear ; )
- [x] pysat: handle direct constraint, add test
- [x] minizinc:  handle direct constraint, add test
- [x] gurobi:  handle direct constraint, add test
- [x] z3:  handle direct constraint? (from functional API?), add test if relevant
- [x] PySDD:  handle direct constraint? add test if relevant
- [x] describe in 'solvers' doc + in each solver's doc, say on which object it does the lookup

@IgnaceBleukx can you update Nonogram? Any help with the solvers also appreciated : )


IgnaceBleukx:
Added some more documentation to the nonogram example and a smaller example of the chrome T-rex.
The plot now also shows the monogram constraints at the very bottom

tias:
I don't see any nonogram constraints at the bottom @IgnaceBleukx 

I do now realize that you are first taking the image, then generating the constraints (but not showing them!!) and then solving that constrained problem.

I will do a minor edit so I can merge it, but ideally after generating the constriants you can show them like people typically show a nonogram (however that might be ; )

PR #301, opened 2023-05-16
Multi dimentional element constraint

Dimosts:
Extended the element constraint to accept multiple dimentions, with multiple vars/expressions instead of only 1.

Basically, in NDVarArray's __getitem__, where the array to be considered is found (based on the constant values in the indices) and flattened, an new index expression is created based on the ones given and the dimentions of the array.

Then, this new expression is given as index to the Element constraint, along with the flattened array.

In addition, if the user writes [x1][x2] instead of [x1,x2], i.e., does not use comma-seperated indexing, an exception is thrown.

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -46,6 +46,7 @@
     Module details
     ==============
 """
+import math
 from collections.abc import Iterable
 import warnings # for deprecation warning
 import numpy as np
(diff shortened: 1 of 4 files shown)


cpmpy/expressions/variables.py
```diff
@@ -418,15 +417,17 @@ def __getitem__(self, index):
...
+            # for calculating the index combining the vars given
+            dim_lengths.append(1)
+            new_index = 0
             for i in range(len(index)):
                 if isinstance(index[i], Expression):
-                    index_rest[i] = Ellipsis # selects all remaining dimensions
-                    var.append(index[i])
-            assert (len(var)==1), "variable indexing (element) only supported with 1 variable at this moment"
-            # single var, so flatten rest array
+                    index_rest[i] = slice(0,dim_lengths[i]) # selects all remaining dimensions
```
IgnaceBleukx, line 426: This needs some documentation

IgnaceBleukx:
Certainly a nice addition, but two things:
- We should have a way to print the 2-d element constraint, now if a user prints it he will get something like `[0,1,....,8][3 * idx0 + idx1]` which might be confusing (not sure how to fix it tho...)
- The implementation of __get_item__ really took me a while to understand it, we should add some documentation and/or just completely rewrite it. I can do an effort in rewriting it. If I'm correct, it first checks if we can select a one-dimensional array from the matrix using numpy indexing (if the idx on some dimension are integer) and then proceeds to build a new index using CPMpy expressions to use on the flattened array.

IgnaceBleukx:
Hmm, I do like the idea of checking it in the `__repr__` actually.
But maybe we should look at another representation all together for multi-dimensional elements...
If I take a 2d matrix such as sudoku, this is still not looking very good imho.
```python3
    [e, e, e,  2, e, 5,  e, e, e],
    [e, 9, e,  e, e, e,  7, 3, e],
    [e, e, 2,  e, e, 9,  e, 6, e],

    [2, e, e,  e, e, e,  4, e, 9],
    [e, e, e,  e, 7, e,  e, e, e],
    [6, e, 9,  e, e, e,  e, e, 1],

    [e, 8, e,  4, e, e,  1, e, e],
    [e, 6, 3,  e, e, e,  e, 8, e],
    [e, e, e,  6, e, 8,  e, e, e]][idx0, idx1]
```

Dimosts:
I am not sure about the test. It will always be true (unless we change the repr) and also is not really readable for the user, because it actually prints ('[IV0 IV1 IV2 IV3 IV4 IV5 IV6 IV7 IV8 IV9 IV10 IV11][sum([1, 4, 2] * [IV14, '
 'IV12, IV13])]')

IgnaceBleukx:
Ok so we will leave the specialized representation for later (or never), will merge this as is.

PR #302, opened 2023-05-16
Adding Isin constraint to specify non-interval domains

Dimosts:
Adding Isin constraint to specify non-interval domains. Open to different names for the constraint (e.g. in_domain).

syntax: Isin(expr, array)

The decompose() method takes one by one the values in the domain of the expression given, and adds a constraint expr != val if val not in array.

This leads to a conjuction of != constraints.

An alternative would be to add a disjuntion of "=" constraints for the values in the array. This could be done by replacing the decompose function with the following:
```
    def decompose(self):
        from .python_builtins import any
        return [any(self.expr == i for i in self.arr)]
```

However, based on a discussion in-person with @JoD, we thought that this would not be efficient, and preferred to use the "!=" constraints, leading also to more efficient linearization.

I tried also to overload the built in "in", but this was not possible even if we overload __contains__, as "in" always has to return a bool (True or False) and cannot return an expression.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -435,7 +435,7 @@ class in_domain(GlobalConstraint):
         expr, arr = self.args
         expressions = any(isinstance(a, Expression) for a in arr)
         if expressions:
-            return [any(expr == a for a in arr)]
+            return [any(expr == a for a in arr)], []
         else:
             # find intervals in the domain given:
             arr = sorted(arr)
@@ -455,7 +455,7 @@ class in_domain(GlobalConstraint):
                 cons.extend([expr >= intervals[0][0], expr <= intervals[-1][1]])
             else:
                 cons.append(False)
-            return cons
+            return cons, []
 
     def value(self):
         return argval(self.args[0]) in argval(self.args[1])
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -451,8 +451,13 @@ class in_domain(GlobalConstraint):
 
             # create constraints based on the intervals
             # Create a conjunction of disjunctions
-            cons = [(expr <= intervals[i-1][1]) | (expr >= intervals[i][0]) for i in range(1, len(intervals))]
-            cons.extend([expr >= intervals[0][0], expr <= intervals[-1][1]])
+            cons = []
+            if len(intervals) > 0:
+                cons.extend([(expr <= intervals[i-1][1]) | (expr >= intervals[i][0]) for i in range(1, len(intervals))])
+                cons.extend([expr >= intervals[0][0], expr <= intervals[-1][1]])
+            else:
+                cons.append(False)
+                
             return cons
 
     def value(self):
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -339,7 +339,7 @@ class InDomain(GlobalConstraint):
     def __init__(self, expr, arr):
         assert not (is_boolexpr(expr) or any(is_boolexpr(a) for a in arr)), \
             "The expressions in the InDomain constraint should not be boolean"
-        super().__init__("InDomain", [expr, arr], is_bool=True)
+        super().__init__("InDomain", [expr, arr])
 
     def decompose(self):
         """
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -420,15 +420,15 @@ class Maximum(GlobalConstraint):
         bnds = [get_bounds(x) for x in self.args]
         return max(lb for lb,ub in bnds), max(ub for lb,ub in bnds)
 
-class in_domain(GlobalConstraint):
+class InDomain(GlobalConstraint):
     """
-        The "in_domain" constraint, defining non-interval domains for an expression
+        The "InDomain" constraint, defining non-interval domains for an expression
     """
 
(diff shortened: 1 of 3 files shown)


Wout4:
Maybe add some tests? Also I'm adding typechecks in the global constraints constructors atm, and I'm guessing here we want the expression and the values to not be boolean
I like the name 'in_domain', if I'd have to vote ;)

Dimosts:
Yes, I will add some tests and typechecks.

cpmpy/expressions/globalconstraints.py
```diff
@@ -417,6 +417,37 @@ def get_bounds(self):
         bnds = [get_bounds(x) for x in self.args]
         return max(lb for lb,ub in bnds), max(ub for lb,ub in bnds)
 
+class Isin(GlobalConstraint):
+    """
+        The "isin" constraint, defining non-interval domains for an expression
+    """
+
+    def __init__(self, expr, arr):
+        self.expr = expr
```
IgnaceBleukx, line 426: We typically don't store the individual arguments in the object.
If you need it later, you can simply do:
```python
expr, arr = self.args
````

File: cpmpy/expressions/globalconstraints.py [hunk 2]
```diff
@@ -417,6 +417,37 @@ def get_bounds(self):
...
+    def decompose(self):
+        cons =[]
+        lb, ub = self.expr.get_bounds()
+
+        expressions = all(isinstance(a, Expression) for a in self.arr)
+        if expressions:
+            from .python_builtins import any
+            cons = [any(self.expr == a for a in self.arr)]
+        else:
+            for i in range(lb,ub+1):
```
IgnaceBleukx, line 439: maybe cleaner with a comprehension as well:
````python
return [expr != val for val in range(lb,ub+1) if val not in arr]
````

**Review by IgnaceBleukx — COMMENTED** [review]
Nice, another possible name: "anyof"?

**Comment by Dimosts** [comment]
They do in 2 different ways: Either create directly a variable with a non-interval domain (e.g. ortools supports this) or by using the Table constraint, providing only 1 variable and the respective values in 1 row.

The first case is not supported by us, as when creating the variable we have to put interval domains.
The second case does not seem that efficient.

File: cpmpy/expressions/globalconstraints.py [hunk 3]
```diff
@@ -417,6 +417,33 @@ def get_bounds(self):
...
+
+    def decompose(self):
+        expr, arr = self.args
+        lb, ub = expr.get_bounds()
+        expressions = all(isinstance(a, Expression) for a in arr)
+        if expressions:
+            from .python_builtins import any
+            return [any(expr == a for a in arr)]
+        else:
+            return [expr != val for val in range(lb, ub + 1) if val not in arr]
```
JoD, line 439: The idea here is to distinguish between a purely numeric array and an array of arbitrary expressions. What if we have a mixed array though? E.g., `[3, y, z]`.

Proposal: if `arr` is purely values, figure out the intervals it has, add the conjunction of disjunctions that `expr` is in those intervals. E.g., `IsIn(x,[3,5,6,7,8])` has two intervals - `3..3` and `5..8`. Then we get `x>=3 & (x=<3 | x>=5) & (x=<8)`. This can be way more efficient than iterating over individual values.

tests/test_globalconstraints.py
```diff
@@ -158,6 +158,18 @@ def test_inverse(self):
         # constraint can be used as value
         self.assertTrue(inv.value())
 
+    def test_isin(self):
+        iv = cp.intvar(-8, 8)
+        iv_arr = cp.intvar(-8, 8, shape=5)
+        constraints = [cp.Isin(iv, iv_arr)]
+        model = cp.Model(constraints)
+        assert model.solve() == True
```
IgnaceBleukx, line 166: Use `self.assertTrue(model.solve())`

tests/test_globalconstraints.py
```diff
@@ -158,6 +158,18 @@ def test_inverse(self):
...
+        assert iv.value() in iv_arr.value()
```
IgnaceBleukx, line 167: Use `self.assertIn(iv.value(), iv_arr.value()`

tests/test_globalconstraints.py
```diff
@@ -158,6 +158,18 @@ def test_inverse(self):
...
+        vals = [1, 5, 8, -4]
+        constraints = [cp.Isin(iv, vals)]
+        model = cp.Model(constraints)
+        assert model.solve() == True
+        assert iv.value() in vals
```
IgnaceBleukx, line 173: Maybe add something fancy such as: `cp.Isin(min(ivs), vals)`?

tias (commented):
decomposition too complicated for constants? x != v propagates instantly...

cpmpy/expressions/globalconstraints.py
```diff
@@ -420,6 +420,49 @@ def get_bounds(self):
...
+    """
+        The "in_domain" constraint, defining non-interval domains for an expression
+    """
+
+    def __init__(self, expr, arr):
+        assert not (is_boolexpr(expr) or any(is_boolexpr(a) for a in arr)), \
+            "The expressions in the in_domain constraint should not be boolean"
+        super().__init__("in_domain", [expr, arr], is_bool=True)
+
+    def decompose(self):
```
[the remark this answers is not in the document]
Dimosts: that was the initiall decompose but after discussion with Jo we changed it to the intervals. Shall I revert it back?

Dimosts:
Seems ready now, shall we merge?

Issue #256, opened 2023-03-21
Global cardinality constraint is not general


JoD:
The global cardinality constraint described in http://sofdem.github.io/gccat/gccat/Cglobal_cardinality.html is more general than the one we have (https://github.com/CPMpy/cpmpy/blob/master/cpmpy/expressions/globalconstraints.py#L551).

Do we want to address this?

Dimosts:
You mean because in our case the user has to define the exact amount of times all values appear, and cannot define it only for a subset of them, right?

JoD:
Yes, exactly.

Dimosts:
I will prepare and alternative version for the more general case and I will open a pull request. Then we can see if we prefer that or the current one.

PR #303, opened 2023-05-16
Improving global constraints

Dimosts:
Based on the discussion in #256, changed the global cardinality constraint (GlobalCardinalityCount()) to make it more general.

Specifically,  in the previous version the user had to define the exact amount of times all values appear, and could not define it only for a subset of them.

The new GlobalCardinalityCount() of this PR is in align with http://sofdem.github.io/gccat/gccat/Cglobal_cardinality.html and also with what minizinc has https://www.minizinc.org/doc-2.5.5/en/lib-globals.html. In addition, the new version complies with hakan's description in http://www.hakank.org/bprolog/global_cardinality.pl

I also checked in all the examples, if we used anywhere the GlobalCardinalityCount constraint, in order to change to the new syntax, but did not find any case.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -594,7 +594,7 @@ class GlobalCardinalityCount(GlobalConstraint):
     """
 
     def __init__(self, vars, vals, occ):
-        super().__init__("gc", [vars,vals,occ])
+        super().__init__("gcc", [vars,vals,occ])
 
     def decompose(self):
         vars, vals, occ = self.args
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -589,26 +589,7 @@ class Cumulative(GlobalConstraint):
 
 class GlobalCardinalityCount(GlobalConstraint):
     """
-        GlobalCardinalityCount(a,gcc): Collect the number of occurrences of each value 0..a.ub in gcc.
-    The array gcc must have elements 0..ub (so of size ub+1).
-        """
-
-    def __init__(self, a, gcc):
-        ub = max([get_bounds(v)[1] for v in a])
-        assert (len(gcc) == ub + 1), f"GCC: length of gcc variables {len(gcc)} must be ub+1 {ub + 1}"
-        super().__init__("gcc", [a,gcc])
-
-    def decompose(self):
-        a, gcc = self.args
-        return [Count(a, i) == v for i, v in enumerate(gcc)]
-
-    def value(self):
-        from .python_builtins import all
-        return all(self.decompose()).value()
-
-class GlobalCardinality(GlobalConstraint):
-    """
-    GlobalCardinality(vars,vals,occ): The number of occurrences of each value vals[i] in the list of variables
+    GlobalCardinalityCount(vars,vals,occ): The number of occurrences of each value vals[i] in the list of variables
     must be equal to occ[i].
     """
 
@@ -622,6 +603,7 @@ class GlobalCardinality(GlobalConstraint):
     def value(self):
         from .python_builtins import all
         return all(self.decompose()).value()
+
 
 class Count(GlobalConstraint):
     """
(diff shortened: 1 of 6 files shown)


IgnaceBleukx:
I changed the pull request to be more general changes to global constraints.

cpmpy/expressions/globalconstraints.py
```diff
@@ -190,7 +190,7 @@ def __init__(self, *args):
         super().__init__("alldifferent_except0", flatlist(args))
 
     def decompose(self):
-        return [((var1 != 0) & (var2 != 0)).implies(var1 != var2) for var1, var2 in all_pairs(self.args)]
+        return [(var1 == var2).implies(var1 == 0) for var1, var2 in all_pairs(self.args)]
```
tias, line 193: is this true? Should it not be `((var1 == 0) | (var2 == 0))` ?

cpmpy/expressions/globalconstraints.py
```diff
@@ -209,7 +209,7 @@ def __init__(self, *args):
     def decompose(self):
         """Returns the decomposition
         """
-        return [var1 == var2 for var1, var2 in all_pairs(self.args)]
+        return [var1 == var2 for var1, var2 in zip(self.args[:-1], self.args[1:])]
```
tias, line 212: this is cryptic... can you add a comment about what you do (sufficient to post on pairs of forward/backward ordering of the array) and why (why? if this is true, then just the subsequent pairs in the sequence is also sufficient; but why not just post all?)

cpmpy/expressions/globalconstraints.py
```diff
@@ -274,18 +274,8 @@ def decompose_negation(self):
...
-                # different orders
-                AllDifferent(order)
-                    ]),
-                # not negating following constraints since they involve the auxiliary variables
-                # loop: first one is successor of '0'
-                order[0] == succ[0]
-                ] + [order[i] == succ[order[i - 1]] for i in range(1, n)]
-
+        decomp = self.decompose()
+        return [~all(decomp[:2])] + decomp[3:]
```
tias, line 278: this is cryptic and brittle...

cryptic: what does it do? why is it true?
how can others learn from this code?

brittle: what if somebody changes decomposition, or the order of the constraints in decompose...

cpmpy/expressions/globalconstraints.py
```diff
@@ -589,18 +579,16 @@ def value(self):
 
 class GlobalCardinalityCount(GlobalConstraint):
     """
-        GlobalCardinalityCount(a,gcc): Collect the number of occurrences of each value 0..a.ub in gcc.
-    The array gcc must have elements 0..ub (so of size ub+1).
-        """
+    GlobalCardinalityCount(vars,vals,occ): The number of occurrences of each value vals[i] in the list of variables
```
tias, line 582: list of variables `vars` must be...

tias (commented):
GCC is fine and can be fast-tracked in a separate merge if needed.

The changes to the existing decompositions are not improvements imho

cpmpy/expressions/globalconstraints.py
tias: basically the preferred decomposition is a single clause `(var1 == 0) | (var2 == 0) | (var1 != var2)` imho

IgnaceBleukx: Well if `var1 == var2` the if `var1 == 0`, it also holds that `var2 == 0`. 
I double-checked and the new decomposition is indeed equivalent.

The new one requires much less auxiliary variables when flattening:
```python
flatten_constraint(old_decomp)
>>[((BV29) and (BV30)) -> (BV31),
    (a != 0) == (BV29),
    (b != 0) == (BV30),
    ((a) != (b)) == (BV31)
]

flatten_constraint(new_decomp)
>>[((a) == (b)) -> (BV32)
    (a == 0) == (BV32)]
```

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: I thought this would be more efficient as the number of constraints we have to generate is now `n` instead of `n^2`.
Some documentation might indeed be useful tho...

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Hmm, yes this should probably be undone... 
I thought is was a good idea but maybe too cryptic indeed.
The main reasoning was: if we change the decomposition, we don't have to change it twice, but actually we do so yeah...

PR #305, opened 2023-05-19
283 return our new cpmpy expression of false

tias:
#284 was merged too quickly, these were commits that were supposed to be pushed on top, and only the end effect should have been merged...


tias:
DOn't merge this.

Ive pushed a new branch 'v0.9.13fix' with only the effect of these commits wrt v0.9.13, and will release it as an emergency v0.9.14 release

tias:
Fixed with v0.9.14 release

PR #306, opened 2023-05-19
inverse works with non vararrays

Wout4:


tias:
what is the source of the error? is it because of the way we use decompose? Then in the 'cpm_array' should only be in the decompose?

Now its just a bit weird that fwd is not wrapped while rev is; and I can only think this is because the non-described error is in decompose

Wout4:
it's because of the decompose yes, will move the cpm_array there then

PR #308, opened 2023-05-22
optimize typechecks (pr 297)

Wout4:


cpmpy/expressions/core.py
```diff
@@ -407,7 +407,7 @@ def __init__(self, name, arg_list):
         arity, is_bool = Operator.allowed[name]
         if is_bool:
```
IgnaceBleukx, line 408: This is actually a hack (which I also applied in #291...) The arguments of an operator can be non-boolean while the output is. Not currently but maybe future operators will be.
We could make a new function on the `Expression` object "args_are_bool" or something like that. Or we make it a helper function in utils?

tias: If we don't have a concrete example, we don't support it nor future-proof for it; it would lead us too far

tias: if such an operator arrives, we just refactor to support it at that time

cpmpy/expressions/globalconstraints.py
```diff
@@ -191,7 +192,7 @@ class AllDifferentExcept0(GlobalConstraint):
     """
     def __init__(self, *args):
         flatargs = flatlist(args)
-        if not (all(is_boolexpr(arg) for arg in flatargs) or not any(is_boolexpr(arg) for arg in flatargs)):
+        if not is_single_type_list(flatargs):
```
tias, line 195: I am not convinced we need any of these 'single_type_list' checks...

If you can use booleans as integers, then any of them should be valid with a mix of bool/int too... this is not a typecheck restriction. If a solver does not support it, we need to transform it later for that solver

tias: we are going from 'no type checking' to 'very strict' type checking... I would say to drop the 'is_single_type_list' checks (do keep the IfThenElse one)

tias (changes requested):

cpmpy/expressions/globalconstraints.py
Wout4: What about checking if we get lists where we expect lists and vice versa? did it for Count now, but might be overkill

PR #309, opened 2023-05-23
simplify output for get or make var in specific circumstances

Wout4:
I think this is a case that only started existing since the introduction of our not-operator, but the output of get_or_make_var can be reduced.

```python
get_or_make_var(Operator('not',[x]) == y)  : 
(BV4, [((BV3) == (y)) == (BV4), (~x) == (BV3)])
can be reduced to 
(BV3, [((~x) == (y)) == (BV3)])
```

Maybe we will have to wait for our push down negation transformation to see if this case will occur

### DIFF
--- a/cpmpy/transformations/flatten_model.py
+++ b/cpmpy/transformations/flatten_model.py
@@ -338,8 +338,7 @@ def get_or_make_var(expr):
 
         if isinstance(flatexpr,_BoolVarImpl):
             #avoids unnecessary bv == bv or bv == ~bv assignments
-            #return flatexpr,flatcons
-            pass
+            return flatexpr,flatcons
         bvar = _BoolVarImpl()
         return (bvar, [flatexpr == bvar]+flatcons)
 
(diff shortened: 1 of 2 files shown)


tias (changes requested):
yes, it is due to not.

But the fix is one level too low, the origin is in the normalized_boolexpr:

```python
import cpmpy as cp
x,y = cp.boolvar(shape=2, name="xy")
e = ~x == y
e = cp.expressions.core.Operator('not', [x]) == y
print(e)
cp.transformations.flatten_model.normalized_boolexpr(e)
```
the first 'e' returns the correct, the second does not and that is what you intend to fix

Wout4:
so we are putting some kind of pre-condition on get_or_make_var to never have a not(bv) expression as input?

then why not do the same for normalized_boolexpr and call it a day? :) 
(not sure if there is still a case where flatten triggers this, there used to be when i discovered this earlier)

Wout4:
ok so get_or_make_var calls normalize_boolexpr, but normalize_boolexpr also calls get_or_make_var.
The fix in get_or_make_var also fixes normalize_boolexpr (the example you give above runs fine on this branch) so i think it is in the right place

tias (approved):

Wout4:
Agreed that we should avoid the Operator constructor, but it does happen in some of our transformations

PR #310, opened 2023-05-23
298 implement push down negation

IgnaceBleukx:
Start of implemenation of push-down negation as a fully recursive function.

I kept the function which was "negated_normal" as several transformations call it.

Todo is to add it to the solver interfaces and do some in-depth testing.

### DIFF
--- a/cpmpy/transformations/negation.py
+++ b/cpmpy/transformations/negation.py
@@ -1,73 +1,107 @@
 import copy
+import numpy as np
 
 from .normalize import toplevel_list
-from ..expressions.core import Comparison, Operator, BoolVal
-from ..expressions.variables import _BoolVarImpl
-def push_down_negation(lst_of_expr):
+from ..expressions.core import Expression, Comparison, Operator, BoolVal
+from ..expressions.variables import _BoolVarImpl, _NumVarImpl
+from ..expressions.utils import is_any_list
+
+def push_down_negation(lst_of_expr, toplevel=True):
+    """
+        Transformation that checks all elements from the list,
+        and pushes down any negation it finds with the `recurse_negation()` function.
+
+        Assumes the input is a list (typically from `toplevel_list()`) en ensures the output is
+        a toplevel_list if the input was.
+    """
+    if isinstance(lst_of_expr, np.ndarray) and not (lst_of_expr.dtype == object):
+        # shortcut for data array, return as is
+        return lst_of_expr
 
     newlist = []
     for expr in lst_of_expr:
-        if expr.name == "not":
-            newlist.append(recurse_negation(expr.args[0], negative_context=True))
+        if is_any_list(expr):
+            # can be a nested list with expressions?
+            newlist.append(push_down_negation(expr, toplevel=toplevel))
+
+        elif not isinstance(expr, Expression) or isinstance(expr, (_NumVarImpl,BoolVal)):
+            # nothing to do
+            newlist.append(expr)
+
+        elif expr.name == "not":
+            # the negative case, negate
+            arg_neg = recurse_negation(expr.args[0])
+            if toplevel:
+                # make sure there is no toplevel 'and' (could do explicit check for and?)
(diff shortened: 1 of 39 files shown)


cpmpy/transformations/negation.py
```diff
@@ -0,0 +1,73 @@
...
+        return newexpr
+
+    elif isinstance(expr, Operator):
+        assert(not negative_context or expr.is_bool()), f"Can only negate boolean expressions but got {expr}"
+
+        if expr.name == "not":
+            return recurse_negation(expr.args[0], not negative_context)
+
+        if negative_context and expr.name == "->":
+            # XXX this might create a top-level and
```
JoD, line 53: What is the problem with creating a top-level and? The same is true for `or`?

IgnaceBleukx: It just means we have to run `topelvel_list` afterwards again. Not really a problem imho, but we could check for it here as well.

JoD: Yeah. Or flatten the top level list after pushing negations?

Wout4:
test failing now should be resolved by the decompose global pr.. (using both circuit decompose and decompose negation)

cpmpy/transformations/negation.py
IgnaceBleukx: Also possible! Then we just have to adjust recurse negation to return a list if gets a list as argument, otherwise return a CPMpy expression as it does now.

cpmpy/transformations/negation.py
```diff
@@ -0,0 +1,73 @@
...
+            return expr.args[0] & recurse_negation(expr.args[1], True)
```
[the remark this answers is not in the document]
IgnaceBleukx: Absolutely! I forgot about this case.

cpmpy/transformations/negation.py
```diff
@@ -0,0 +1,73 @@
...
+        if negative_context:
+            if   expr.name == "and": newexpr.name = "or"
+            elif expr.name == "or": newexpr.name = "and"
+            else:
+                raise ValueError(f"Unknown operator to negate {expr}")
+        return newexpr
+
+    # global constraints
+    if hasattr(expr, "decompose"):
+        return ~expr if negative_context else expr #TODO: recurse into arguments? This should be re-called after decompose_globals anyway...
```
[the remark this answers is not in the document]
IgnaceBleukx: Yes, it would make a lot of sense to first decompose the global constraints and then push negations.
Mainly as the decomposition of globals will probably introduce negations...

tias:
I've made it backward compatible with a warning.

Also, I think we agreed we should do negation before simplify? as it allows to simplify more? I did had to add a constructor check for it (and resulting cleanups elsewhere) to keep a case from the tests cleaner

I'm confident I addressed the remarks of Jo, so if he does not respond tomorrow feel free to merge as is.

IgnaceBleukx:
After having another think about it, I actually think it's better to first do `push_down_negation` and then `simplify_boolean`.
The reason is that `simplify_boolean` can actually introduce negations of arbitrary expressions.

For example:
```python
cons  = all(boolvar(shape=3)) <= 0
simplify_boolean(push_down_negation([cons])
>> [not([and([BV0, BV1, BV2])])]
push_down_negation(simplify_boolean([cons])
>> [or([~BV0, ~BV1, ~BV2])]
```

tias:
If we move (later) the bool optimisations from 'flatten' to 'simplify_bool', then they must be after pushing down negation to get the smallest encodings... 

I checked and there are 2 rare cases where simplify_bool can introduce a negation.

If we acknowledge that simplify_bool makes most sense to be run after push down negation, then we should make simplify_bool call push `recurse_negate` when it wishes to negate smth in those 2 rare cases... so that simplify_bool promises that if its input contains no negations, it will add none either.

tias:
nice, good to go imho (another import one!)

PR #312, opened 2023-05-24
solver interfaces: make add directly call transform, simpler and more explicit

tias:
removes the need for an inbetween _post_constraint function (well, ortools for now still needs it)

For some solvers I did a 'two space indent' hack to not make it look like all code has changed.

At a future refactor, we can use 4 spaces indent everywhere again...


Wout4:
I corrected some typo's, otherwise looks good to me!

IgnaceBleukx:
We should also update the TEMPLATE.py file. Otherwise looks good to me too

IgnaceBleukx:
Updated the template with the new layout, good to merge now.

PR #315, opened 2023-05-24
check for bv == iv in normalized boolexpr and only_bv_implies

Wout4:


cpmpy/transformations/flatten_model.py
```diff
@@ -436,7 +436,7 @@ def normalized_boolexpr(expr):
 
                 # this is a reified constraint, so lhs must be var too to be in normal form
                 (lhs, lcons) = get_or_make_var(lexpr)
-                if expr.name == '!=':
+                if expr.name == '!=' and is_boolexpr(rvar):
```
IgnaceBleukx, line 439: You can just do `rvar.is_bool()` which will be more efficient compared to checking `is_boolexpr`.
`rvar` is guaranteed to be a variable at this point.

IgnaceBleukx:
Logic is good, just  a small optimization that can be added.
And some tests should be added to cover these cases.

PR #316, opened 2023-05-24
pysat: accept reified cardinality constraints

### LINKED ISSUE #296: pysat reified cardinality
PySAT supports cardinality constraints.

I would expect that it should also be able to handle implied cardinality constraints then? @sourdough-bread ?

```
from cpmpy import *
bv = boolvar(5)
c = bv[0].implies(sum(bv[1:]) > 2)
Model(c).solve("pysat")
```
fails with not-supported.

Bug found as part of an extension of #295

tias:
refactor to directly call the transformations

split out cardinality encoding in a function so that it can also be used for implied constraints

fixes #296


IgnaceBleukx:
Small update to tests so negative boolvars are also tested.
Looks good to me!

Wout4:
good to merge

Issue #273, opened 2023-03-31
Minizinc crashes when True or False literals are compared with numeric expressions


Wout4:
expected behavior: treat true as 1 and false as zero in these comparison (as other solvers do)

```
iv = intvar(1,8, shape=3)
Model([iv[0] == False]).solve("minizinc")
```

error: undefined identifier `False'

PR #317, opened 2023-05-25
Minizinc bounds and false literals fix

Wout4:
fixes issues #273 and minizinc boundexceptions, after which minizinc should run without errors on bigtest


tias (approved):
OK for me to merge.

But I'm wondering: are we creating these huge domains with our current code? I think our code should cowardly refuse to create an auxiliary variable (in flatten) if it can not compute a meaningfully tight domain for it...
`get_or_make_var()` seems to directly use `get_bounds()` and all get_bounds seem non-maxint, so perhaps this is an artifact in the way the bigtest pickels were created?

Wout4:
These huge domains are not created by our code, but there are some models in bigtest that actually sum or multiply enough big variables to go beyond the minizinc bounds

PR #319, opened 2023-05-26
single file int2bool direct encoding meta solver

tias:
Single file so it does not depend on a CPMpy version, and so that it can be copied and used in side-projects without needing to be in master

Based on files Emilio sent me a month ago, refactored into a 'meta solver' that accepts any cpmpy solver as subsolver (meant for pysat and its incremental solving). I mostly refactored the top part, not the actual encodings.

Only tested on sudoku, this can benefit from a lot more testing...


IgnaceBleukx:
I made some small changes so you can use  `solverLookup.get("int2bool:subsolver")` as a normal solver.
This also allowed me to use it in the `test_constraints` test suite, which now spectacularly fails...

tias:
superseded by #653

Issue #138, opened 2022-10-10
Any suggestion about or-tools hyperparameter search 


asdqsczser:
In the [docs](https://cpmpy.readthedocs.io/en/latest/solver_parameters.html), a link is recommended but broken, "We documented some of the frequent once in our [CPM_ortools API](https://cpmpy.readthedocs.io/en/latest/cpmpy/solvers/ortools.py)".

In the tutorial, three hyperparameters are recommended.

```
params = {'cp_model_probing_level': [0,1,2,3],
          'linearization_level': [0,1,2],
          'symmetry_level': [0,1,2]}
```

Can you provide more critical parameters for the CP-SAT solver of or-tools? 

Thanks.

tias:
Hi,

I had the pleasure of having dinner with Laurent Perron from OR-Tools at a workshop today.

We went over the parameters together and he mentioned the following as those that could be worth trying.

These are rough notes, they can be mapped (in reverse order) to actual param names in the file you linked.

use branch in lp, optimize with core, search branching, bool enc level, linearization lvl, minimize core, cp model probing level, cp model presolve, clause ordering, clause partitioning, binary minimization algo, confl minimization algo, use phase saving

@wout can you add them to the docs and tool and examples/param tune notebook? Check with @IgnaceBleukx if unclear

PR #323, opened 2023-05-30
Updated tuner

IgnaceBleukx:
Update our hyperparameter tuner and added some more documentation.
Also implemented a more naive grid search so we can easily compare the implementations.
The new hyperparameters for OR-tools are chosen according to issue #138. However, for some parameters no upper bound is given by the [sat_parameters.proto](https://github.com/google/or-tools/blob/5425dedcfbb22cb74c636c1374a9b5ad684b1eb5/ortools/sat/sat_parameters.proto#LL469C4-L470C61) file... (i.e., for example `cp_model_probing_level`).

### DIFF
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -544,7 +544,7 @@ class CPM_ortools(SolverInterface):
     def tunable_params(cls):
         """
             Suggestion of tunable hyperparameters of the solver.
-            List compiled based on Laurents feedback (issue #138)
+            List compiled based on a conversation with OR-tools' Laurent Perron (issue #138).
         """
         return {
             'use_branching_in_lp': [False, True],
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -542,28 +542,40 @@ class CPM_ortools(SolverInterface):
 
     @classmethod
     def tunable_params(cls):
+        """
+            Suggestion of tunable hyperparameters of the solver.
+            List compiled based on Laurents feedback (issue #138)
+        """
         return {
-            'cp_model_probing_level': [0, 1, 2],
-            'preferred_variable_order': [0, 1, 2],
+            'use_branching_in_lp': [False, True],
+            'optimize_with_core' : [False, True],
+            'search_branching': [0,1,2,3,4,5,6],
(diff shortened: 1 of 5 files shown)


cpmpy/solvers/ortools.py
```diff
@@ -542,28 +542,40 @@ def get_core(self):
 
     @classmethod
     def tunable_params(cls):
+        """
+            Suggestion of tunable hyperparameters of the solver.
+            List compiled based on Laurents feedback (issue #138)
```
tias, line 547: 'on a conversation with OrTools' Laurent Perron'

cpmpy/tools/tune_solver.py
```diff
@@ -115,3 +115,60 @@ def _params_to_np(self,combos):
...
+        self.base_runtime = solver.status().runtime
+        self.best_runtime = self.base_runtime
+
+        # Add default's runtime as first entry in configs
+        combos = list(param_combinations(self.all_params))
+
+        i = 0
+        if max_tries is None:
+            max_tries = len(combos)
+        while len(combos) and i < max_tries:
```
tias, line 148: i dont really like the config_idx below and this 'while', it should just be a simple for? If you want randomized order, maybe something like `for i, param_dict in rand.shuffle(param_combinations(self.all_params)):`?

docs/solver_parameters.md
```diff
@@ -47,25 +47,45 @@ The parameter tuner is based on the following publication:
...
-    'symmetry_level': 2
+    "MIPFocus": 0,
+    "Method": -1,
+    "FlowCoverCuts": -1
 }
 
-tuner = ParameterTuner("ortools", model, tunables, defaults)
-best_params = tuner.tune(max_tries=100)
-best_runtime = tuner.best_runtime
+tuner = ParameterTuner("pysat", model, tunables, defaults)
```
tias, line 84: probably meant to be gurobi instead of PySAT?

tias (approved):
good, just some style feedback

cpmpy/tools/tune_solver.py
IgnaceBleukx: Ah yes, this tuner is indeed much simpler to code, updated

tias:
Nice, thanks!

PR #324, opened 2023-05-30
NegBoolView was using IntVarImpl init and not BoolVarImpl init

363734:
This creates the check (using the asserts during the init of BoolVarImpl) that bv should have a domain subset of {0,1}.
Using NegBoolVarView on an IntVarImpl with domain subset of {0,1} would still pass tho, but using it on other IntVar would now trigger assert.

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -349,7 +349,7 @@ class NegBoolView(_BoolVarImpl):
     def __init__(self, bv):
         #assert(isinstance(bv, _BoolVarImpl))
         self._bv = bv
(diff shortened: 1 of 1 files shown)


Wout4:
The constructor of BoolVarImpl calls the constructor of IntVarImpl so this would be slightly less efficient.

A user is never supposed to directly call NegBoolView, but has to use ~bv.
~iv already throws a type-error.

If we really want to do this check I think we can just uncomment the assert in the constructor of NegBoolView, but I'm sure it got uncommented because it's not really needed?

363734:
True!
Let's maybe put a comment telling that, since it is not common practice to not use the constructor of the direct parent.

Wout4:
Perfect, will merge!

PR #325, opened 2023-05-30
bounds of a numerical value should always be tight

363734:
When calling get_bounds on a boolean value, the bounds were loose. While calling the get_bounds on other numerical values was giving tight bounds

### DIFF
--- a/cpmpy/expressions/utils.py
+++ b/cpmpy/expressions/utils.py
@@ -134,7 +134,7 @@ def eval_comparison(str_op, lhs, rhs):
 
 
 def get_bounds(expr):
-    # can return floats, use floor and ceil when creating an intvar!
+    # returns appropriately rounded integers
     from cpmpy.expressions.core import Expression
     if isinstance(expr,Expression):
         return expr.get_bounds()
--- a/cpmpy/expressions/utils.py
+++ b/cpmpy/expressions/utils.py
@@ -139,4 +139,4 @@ def get_bounds(expr):
         return expr.get_bounds()
     else:
         assert is_num(expr), f"All Expressions should have a get_bounds function, `{expr}`"
-        return expr, expr
+        return int(expr), int(expr)
--- a/cpmpy/expressions/utils.py
+++ b/cpmpy/expressions/utils.py
@@ -139,6 +139,4 @@ def get_bounds(expr):
         return expr.get_bounds()
     else:
         assert is_num(expr), f"All Expressions should have a get_bounds function, `{expr}`"
-        if is_bool(expr):
-            return 0, 1
         return expr, expr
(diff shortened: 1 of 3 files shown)


IgnaceBleukx:
Now the `get_bounds` function can return a boolean constant instead of an integer when `expr` is a boolean constant.
I think what it should be is `return (int(expr), int(expr))`

363734:
I'm gonna change to that

Wout4:
The reason we didn't originally convert to the output of get_bounds to integer is exaclty because we want the bounds to be as tight as possible. It's now up to the caller of get_bounds to correctly use/cast the output.

If we would change it to always output integers it has to be floor(expr) and ceil(expr) to deal with float constants.
get_bounds(5.6) now gives (5,5), on master it gives (5.6, 5.6) but i guess you could argue it has to be (5,6) (i think keeping it tight is better though)

But for bools it should definitively give tight bounds, so then what we would need is this?
`        if is_bool(expr):
            return int(expr), int(expr)
        return expr, expr`

JoD:
> It all depends on what do we prefer (be correct in all cases or be the tightest and assuming we don't have operations leading to rounding errors)

If we are not correct in all cases, we will have issues in the future arising from these exceptional cases which are hard to resolve. Yes, we could document these cases and add enough case checks to cover them all, but that is a huge drag on future development. So I have a strong preference to be correct in all cases.

Wout4:
For divisions involving variables we only support floordivision, so we are already rounding down. The float bounds only exist temporarily when the user introduces float constants. when they get used for an auxiliary variable floor and ceil is used.
We also will never calculate the bounds of x/y (with x and y constants) recursively in the get_bounds function, since this python operation will just be evaluated first.
I'm not sure I can fabricate an example that would lead to rounding errors in our bound calculations.

something like 10000000000 - iv*0.0000000000001 < 10000000000 would be a candidate, but this will be flattened to the following:
10000000000 + IV1 < 10000000000 and IV1 = iv*-0.0000000000001 (where iv1 will get the rounded bounds -1,0)

however i will concede that get_bounds(10000000000 - iv*0.0000000000001) will not return correct bounds but not sure if this will ever happen

363734:
Let's then have, at least for now, correct but less tight bounds. And if the future we want the tightest bound possible we can check it again

Wout4:
only concern i have left is that we use ceil and floor in get_or_make_var, which now became redundant. Wanted to remove it but seems sketchy if we later alter get_bounds again to return floats.

363734:
If, for the moment, we set formally the spec of get_bounds to return ints only, then the .floor and .ceil in get_or_make_var could be removed. 

(If we change the specs later, when we are confident there is no rounding errors, we can just find and verify all usage of getbounds to check them, or even define a second method get_bounds_tight, which would return the float bound, and get_bounds would return the floor and ceil on get_bounds_tight, still matching its specs, leading to no verification of the usage to do)

PR #327, opened 2023-05-30
EQ and NEQ with bool and bool expr

363734:
The __eq__ and __neq__ in boolvarimpl should allow the value to be a float also (ie, BV == 1.0).

Also, I think this should be removed from this class and put into the __eq__ and __neq__ of the Expression (they are not handled symmetrically there, so I also thing cases are already missing).
(I didn't do it yet because I think there are more cases when it's tested on expression)

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -308,11 +308,11 @@ class _BoolVarImpl(_IntVarImpl):
     def __eq__(self, other):
         # (BV == 1) <-> BV
         # if other == 1: XXX: dangerous because "=="" is overloaded 
-        if (is_int(other) and other == 1) or \
+        if (is_num(other) and other == 1) or \
                 other is True or \
                 other is np.bool_(True):
             return self
-        if (is_int(other) and other == 0) or \
+        if (is_num(other) and other == 0) or \
                 other is False or \
                 other is np.bool_(False):
             return ~self
@@ -320,11 +320,11 @@ class _BoolVarImpl(_IntVarImpl):
     def __ne__(self, other):
         # (BV == 0) <-> BV
         # if other == 1: XXX: dangerous because "=="" is overloaded 
-        if (is_int(other) and other == 1) or \
+        if (is_num(other) and other == 1) or \
                 other is True or \
                 other is np.bool_(True):
             return ~self
-        if (is_int(other) and other == 0) or \
+        if (is_num(other) and other == 0) or \
                 other is False or \
                 other is np.bool_(False):
             return self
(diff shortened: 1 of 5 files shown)


IgnaceBleukx:
I would even argue many of these checks should go out of the constructor of our `Expression` objects as they alter the way the user wrote his constraint originally. We now have the `simplify_boolean` transformation which simplifies exactly these cases.
In the future, we could also have transformations merging n-ary constraints such as `sum(a,b,c) + sum(x,y,z)` which should become `sum(a,b,c,x,y,z)`. Although it will probably be more efficient to check for these things at construction time (which we do now but not entirely consistently), I think it makes sense to keep the `Expression` object as close to the formula the user wrote.

363734:
I completely agree. It could be confusing for the user to have an expression tree not consistent with what he is creating (ie, when printing it before simplification or when debugging). If there is a way to uniform it in one place it would be best

363734:
the __init__ of Operator is indeed doing some refactoring directly.

Wout4:
if we have these cases taken care of in simplify_boolean then these checks can just be removed from the constructor? Or are there cases/solvers where simplify_boolean does not get called and we still want these simplifications to happen?

Dimosts:
We have to remove them either from here or from simplify boolean. Doing the same thing in 2 different places is not ideal. I think simplify boolean is the right place to do it, I agree with what has been mentioned already, that we should not alter significantly in the constructor the way the user wrote his constraint originally. So, I suggest to completely remove them from the constructor, and also check for other similar cases to take care of in this PR.

Wout4:
turns out we cannot remove these cases after all, the reason being that objective functions also get flattened, but here simplify_boolean does not get called, hence the simplification are not done.
Should we make it so simplify_boolean is also called on objective functions?

Dimosts:
Interesting. I think that yes, simplify boolean should also be used in the objective.

IgnaceBleukx:
Added the removal of float constants in boolean expressions to `simplify_boolean` transformation and call `simplify_boolean` on the objective when its flattened.
This allows us to remove all of the extra code from the `BoolVar` class as its now just treated as any boolean expression. This DOES mean the expression `bv == 0` will remain `bv == 0` until after `simplify_boolean` But personally, I think thats fine.

This one should probably wait on #310 so the changes to `simplify_boolean` get merged nicely.

IgnaceBleukx:
Updated, should be ready to merge now

Wout4:
so you are also simplifying bv == x with x a float in the interval [0..1] (where only 1.0 is considered true)?

IgnaceBleukx:
Yes, the updated  `simplify_boolean` transformation removes all floating point values in comparisons with boolean expressions.
1.0 is considered as True, 0.0 is considered as False.
If you have a boolean expression `expr` and constraint `expr <= 0.42` it will be simplified to `~expr` etc

Wout4:
oh i see, also things outside the range get simplified away, then it's very nice!

PR #329, opened 2023-05-31
Improvement of the beginner_tutorial.rst (typos, new ref)

363734:
found some typos, added ref to the mooc, added space to uniform code example with PEP 8 standard

### DIFF
--- a/docs/beginner_tutorial.rst
+++ b/docs/beginner_tutorial.rst
@@ -10,7 +10,7 @@ To decide if a problem is feasible or finding the best one amongst all the optio
 
 Instead, the paradigm of **constraint programming (CP)** allow you to:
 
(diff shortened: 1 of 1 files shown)


tias:
Great.

I am thinking of cutting this page, in favor of reworking the 'modeling' and 'solving' pages into one big page that explains variables, constraints, objectives, solvers.

Because now, we have this; then we also have the 'modeling' doc page, and then also a 'solving' doc page. What do you think?

363734:
Yes, it would be better indeed. It would remove the redundancy of information. A potential new user would have access to the example (notebooks), if he just wants a quick view of the formulation and what programming in cpmpy looks like. And could go to the new reworked page to get the technical detail (which would point also to the ref currently here, if they really want to go into the theory of constraint programming)

PR #330, opened 2023-05-31
minor fix in documentation of variables.py

363734:
- fix deprecation warning redirecting all to the same method instead of the right ones (bad copy past when deprecation of the method)
- correction of spec for shape type (to match the shape type of numpy, which is int or tuples of int in their doc)
- few trivial comments added
- few space added to comply PEP 8


Wout4:
some nice cleanups, fine to merge for me

PR #331, opened 2023-05-31
minor changes to utils.py

363734:
- refactor the direct use of call to args[0] instead of calling API method .value() for bool var
- refactor is_bool, is_int, is_num so there is no code duplication
- add few space to comply with PEP 8


Wout4:
Looks good to me!

JoD:
Ok, let's merge then. @Wout4 feel free to merge immediately after having a look. One person's approval is plenty for these kind of changes.

PR #333, opened 2023-05-31
Class and function documentation

Dimosts:
Having basic documentation of the functionality of classes and functions would be quite nice.

Following discussions in person, I generated automatically documentation starting for the global constraint classes. The  documentation was auto generated using codiumAI (pycharm extention).

Let's see if it is good enough and if yes we can proceed with more functions and classes. 

In the first commit, only the text generated automatically is included. In following commits I will check the texts and maybe change some parts if needed

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -125,6 +125,26 @@ class GlobalConstraint(Expression):
         Overwrites the `.is_bool()` method. You can indicate
         in the constructor whether it has Boolean return type or not.
     """
+
+    """
+    Main functionalities:
+    The GlobalConstraint class is an abstract superclass that represents a global constraint in Constraint Programming. 
+    It has a name and a list of arguments, and can be either a Boolean or a numerical expression. It provides methods 
+    for decomposition, negation, and checking whether it is a total function.
+
+    Methods:
+    - __init__(self, name, arg_list, is_bool=True): initializes a GlobalConstraint object with a name, a list of 
+    arguments, and a Boolean flag indicating whether it is a Boolean expression or not.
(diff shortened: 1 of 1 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -494,6 +698,25 @@ class Xor(GlobalConstraint):
         The 'xor' exclusive-or constraint
     """
 
+    """
+    Main functionalities:
+    The Xor class is a subclass of the GlobalConstraint class and represents the exclusive-or constraint. It takes a list 
+    of Boolean arguments and enforces that exactly one of them is true.
```
JoD, line 704: This is not quite correct: "xor" is true iff the number of true variables in the given list is odd (with only two arguments, xor is indeed equivalent to 'exactly one is true', but we allow more).

cpmpy/expressions/globalconstraints.py
```diff
@@ -443,6 +628,25 @@ class Element(GlobalConstraint):
         a 'numeric' global constraint.
     """
 
+    """
+    Main functionalities:
+    The Element class is a CPMpy built-in global constraint that enforces that the result equals Arr[Idx] with 'Arr' an 
+    array of constants of variables (the first argument) and 'Idx' an integer decision variable, representing the index into
```
JoD, line 634: 'of variables' -> 'or variables'

JoD:
I both like and hate this: it is 95% correct, and figuring out where it is incorrect is time consuming. I'd allow this if we put an "AI-generated"-tag next to it. People can remove that tag once they checked it manually.

It reminds me of why StackOverflow forbade the use of AI-generated answers: it *looks* correct, but figuring out whether it *is* correct is hard.

Issue #184, opened 2022-11-29
Incomplete functions on rhs of implication


IgnaceBleukx:
Some expressions are incomplete, that is, there exist value for their arguments for which there is no function value defined.
Two cases I can think of and that raises errors currently:

### Element constraint index out of range:
``` python3
a = boolvar()
arr = intvar(0,3, shape=4)
idx = intvar(0,10)

cons = a.implies(arr[idx] >= 1)
assert Model([cons, idx == 10]).solve() 
cons.value() # index out of range error
```
This first case can/should be handled when calculating the value of a constraint by lazily evaluating the value of the tail in the constraint.

### Division by zero
```python3 
a = boolvar()
x = intvar(0,10)

from cpmpy.transformations.reification import reify_rewrite
cons = a.implies(10 / x <= 5)
reify_rewrite(cons) # results in division by 0 error
```
The second case is another annoying bounds calculation bug.

tias:
was still open...

we already have an 'argval' in expr/utils. If we replace arg_vals[0] by argval(self.args[0]) everywhere instead of computing it all upfront, and then we might automatically get it also for and, or, etc if Python does it lazily...

IgnaceBleukx:
Yes and no...
There are still cases where the value computation can go wrong:
`cond -> subexpr` with `cond = [1,2,3][l] == 2` and `subexpr = p`:
Valid solution (I think) is `l = 10` and `p = True`. But computing the value of `cond` will raise an IndexError.
What I think would be a good solution to this problem is to introduce a `IncompleteFunctionError` when calculating the value of `cond` and catch this when computing the value of the entire expression

JoD:
I'd like to reopen this issue. Supporting partial functions is a huge pain in the butt. If Microsoft with Z3 cannot do it, we should not either.

One way this manifests is that the way Element is handled right now leads to flattening not being equivalence preserving. E.g., right now, the constraint `not (x * element([1,2,3],y) >= 5)` is true if `y<0` or `y>2`. Unnesting yields the following constraints: `not (x * aux >= 5)` and `aux == element([1,2,3],y)`. If `y<0` or `y>2`, these constraints now evaluate to false.

[...]

PR #335, opened 2023-05-31
Decompose globals in tree

IgnaceBleukx:
Reimplemented the "decompose_globals" transformation to work at the expression tree level.
This allows us to remove all decomposing logic from the `z3` solver (and any future solver supporting an object-oriented API). 
Special care is taken for non-equivalent decompositions such as `Circuit` where instead of `bv == Circuit.decompose()` we introduce two constraints: `bv -> Circuit.decompose(), ~bv -> Circuit.decompose_comparison()`.

The pull request contains probably more than it should and some parts can be split off, such as changes to the global constraints themselves.

The transformation itself will probably also need some more iterations over it to be ready to merge.
I tried to add enough documentation to all of the cases, but please let me know if you have any doubts or concerns about which cases might have been missed.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -118,7 +118,6 @@ from ..exceptions import CPMpyException, IncompleteFunctionError, TypeError
 from .core import Expression, Operator, Comparison
 from .variables import boolvar, intvar, cpm_array, _NumVarImpl
 from .utils import flatlist, all_pairs, argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds
-from ..transformations.flatten_model import get_or_make_var
 
 # Base class GlobalConstraint
 class GlobalConstraint(Expression):
@@ -187,7 +186,8 @@ class AllDifferentExcept0(GlobalConstraint):
         super().__init__("alldifferent_except0", flatlist(args))
 
     def decompose(self):
-        return [((var1 != 0) & (var2 != 0)).implies(var1 != var2) for var1, var2 in all_pairs(self.args)],[]
+        # equivalent to (var1 == 0) | (var2 == 0) | (var1 != var2)
+        return [(var1 == var2).implies(var1 == 0) for var1, var2 in all_pairs(self.args)], []
 
     def value(self):
         vals = [a.value() for a in self.args if a.value() != 0]
@@ -206,7 +206,8 @@ class AllEqual(GlobalConstraint):
     def decompose(self):
         """Returns the decomposition
         """
-        return [var1 == var2 for var1, var2 in all_pairs(self.args)], []
+        # arg0 == arg1, arg1 == arg2, arg2 == arg3... no need to post n^2 equalities
+        return [var1 == var2 for var1, var2 in zip(self.args[:-1], self.args[1:])], []
 
     def value(self):
         return len(set(a.value() for a in self.args)) == 1
@@ -605,22 +606,19 @@ class Cumulative(GlobalConstraint):
 
 class GlobalCardinalityCount(GlobalConstraint):
     """
-        GlobalCardinalityCount(a,gcc): Collect the number of occurrences of each value 0..a.ub in gcc.
-    The array gcc must have elements 0..ub (so of size ub+1).
-        """
+    GlobalCardinalityCount(vars,vals,occ): The number of occurrences of each value vals[i] in the list of variables vars
+    must be equal to occ[i].
+    """
 
-    def __init__(self, a, gcc):
-        flatargs = flatlist([a, gcc])
+    def __init__(self, vars, vals, occ):
+        flatargs = flatlist([vars, vals, occ])
         if any(is_boolexpr(arg) for arg in flatargs):
             raise TypeError("Only numerical arguments allowed for gcc global constraint: {}".format(flatargs))
-        ub = max([get_bounds(v)[1] for v in a])
-        if not (len(gcc) == ub + 1):
-            raise TypeError(f"GCC: length of gcc variables {len(gcc)} must be ub+1 {ub + 1}")
-        super().__init__("gcc", [a,gcc])
+        super().__init__("gcc", [vars,vals,occ])
 
     def decompose(self):
-        a, gcc = self.args
-        return [Count(a, i) == v for i, v in enumerate(gcc)], []
+        vars, vals, occ = self.args
+        return [Count(vars, i) == v for i, v in zip(vals, occ)], []
 
     def value(self):
         from .python_builtins import all
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -110,7 +110,7 @@
 import copy
 import warnings # for deprecation warning
 import numpy as np
-from ..exceptions import CPMpyException, IncompleteFunctionError
+from ..exceptions import CPMpyException, IncompleteFunctionError, TypeError
 from .core import Expression, Operator, Comparison
 from .variables import boolvar, intvar, cpm_array, _NumVarImpl
 from .utils import flatlist, all_pairs, argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds
@@ -123,7 +123,7 @@ class GlobalConstraint(Expression):
 
         Like all expressions it has a `.name` and `.args` property.
         Overwrites the `.is_bool()` method. You can indicate
-        in the constructer whether it has Boolean return type or not.
+        in the constructor whether it has Boolean return type or not.
     """
     # is_bool: whether this is normal constraint (True or False)
     #   not is_bool: it computes a numeric value (ex: Minimum, Element)
@@ -223,8 +223,11 @@ class Circuit(GlobalConstraint):
     """The sequence of variables form a circuit, where x[i] = j means that j is the successor of i.
     """
     def __init__(self, *args):
-        super().__init__("circuit", flatlist(args))
-        if len(flatlist(args)) < 2:
+        flatargs = flatlist(args)
+        if any(is_boolexpr(arg) for arg in flatargs):
+            raise TypeError("Circuit global constraint only takes arithmetic arguments: {}".format(flatargs))
+        super().__init__("circuit", flatargs)
+        if len(flatargs) < 2:
             raise CPMpyException('Circuit constraint must be given a minimum of 2 variables')
 
     def decompose(self):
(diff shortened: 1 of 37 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -157,13 +157,14 @@ def decompose_negation(self):
         from .python_builtins import all
         return [~all(self.decompose())]
 
-    def is_total(self):
```
JoD, line 160: You can probably split off the removal of is_total, I'll happily do a swift review.

IgnaceBleukx:
Just had an offline discussion with @JoD  about decompose_comparison. 
It turns out our decompositions are actually not equivalence preserving!
I.e., `max(x,y,z) <= 10` is not equivalent to `[(x <= IV | y <= IV | z <= IV), x >= IV, y >= IV, z >= IV, IV <= 10]` as we now decompose the `max` constraint.
This is important as the last iteration of the code replaced the node `max(x,y,z) <= 10` with the conjunction of all the constraints in the decomposition.

We devised two ways to fix this issue:
1) We make `decompose_comparison` return two lists of constraints: one of which the conjunction should replace the node in the expression tree, and the other whose constraints should be added top-level so they are always true. 
    - pros: very general, (Actually, we can do the same for the Circuit constraint!) and will always work
    - cons: return type of decompose_comparison changes and is not in line with other decompositions anymore
2) We make the decomposition of the global functions we have equivalence preserving. For the example: `[x <= 10, y <= 10, z <= 10]`
    - pros: return type of decompose_comparison remains the same
    - cons: we may encounter global constraints for which this is actually not possible, for example when more than 1 auxillary variable is needed. Moreover, it will complicate the code for `decompose_comparison` as every comparison will have a different case.

I went with the first option for now, happy to change it if the second case seems favourable.
What are your thoughts on this @Wout4 and @tias ?

Wout4:
i think the cases of circuit and max are similar (use of auxiliary variables makes decompositions not equivalence preserving) so the approach should also be similar imo (adding aux constraints at top level) 
We are thinking of making a different class for numerical globals anyway, so then this return type issue wouldn't be so weird anymore as well

cpmpy/expressions/globalconstraints.py
```diff
@@ -463,19 +484,16 @@ def decompose_comparison(self, cpm_op, cpm_rhs):
...
 
         arr, idx = self.args
         return [(idx == i).implies(eval_comparison(cpm_op, arr[i], cpm_rhs)) for i in range(len(arr))] + \
-               [idx >= 0, idx < len(arr)]
-
-    def is_total(self):
-        arr, idx = self.args
-        lb, ub = get_bounds(idx)
-        return lb >= 0 & idx.ub < len(arr)
+               [idx >= 0, idx < len(arr)], []
```
JoD, line 496: These are redundant now that Element is total.

cpmpy/expressions/globalconstraints.py
```diff
@@ -378,11 +383,17 @@ def value(self):
     def decompose_comparison(self, cpm_op, cpm_rhs):
         """
         Decomposition if it's part of a comparison
+        Returns two lists of constraints:
+            1) constraint replacing node in expression tree
+            2) constraints which should be true at toplevel
         """
         from .python_builtins import any, all
+        if cpm_op == "==": # can avoid creating aux var
+            return [any(x == cpm_rhs for x in self.args), all(x >= cpm_rhs for x in self.args)]
```
JoD, line 392: `any(x =< cpm_rhs for x in self.args)` no need for the full equality.

cpmpy/expressions/globalconstraints.py
```diff
@@ -378,11 +383,17 @@ def value(self):
...
+
```
JoD, line 393: Optionally:
```
if cpm_op == ">=":
  return [all(x >= cpm_rhs for x in self.args)]
if cpm_op == "<=":
  return [any(x <= cpm_rhs for x in self.args)]
```

tests/test_transf_reif.py
```diff
@@ -48,30 +48,30 @@ def test_reif_element(self):
...
-                 (1,2,2),
-                 (1,3,3),
-                 (2,3,2),
-                ]
-
-        for (lb,ub,cnt) in cases:
-            idx = intvar(lb,ub, name="idx")
-            e = (rv == (arr[idx] != 1))
-            self.assertEqual(Model(e).solveAll(), cnt)
+        # # Another case to be careful with:
```
JoD, line 51: May just be removed?

cpmpy/expressions/globalconstraints.py
```diff
@@ -269,6 +270,10 @@ def value(self):
 
         return pathlen == len(self.args) and idx == 0
 
+
+    def equivalent_decomposition(self):
+        return False
+
     def decompose_negation(self):
```
tias, line 277: as discussed offline, this too will return a tuple of two expressions (like decompose_comparison below): the first with an expression containing the 'constraint' part; the second with the (total) definitions of any auxiliary variables defined that will be enforced toplevel

cpmpy/expressions/globalconstraints.py
tias, line 392: please make this always return a tuple (expr, expr) instead of [expr, expr], so we can actually check that it is a tuple and not one list of constraints.

cpmpy/transformations/decompose_global.py
```diff
@@ -1,107 +1,134 @@
...
     """
         Decomposes any global constraint not supported by the solver
-        Accepts a list of flat constraints as input
-        Returns a list of flat constraints
+        Accepts a list of CPMpy expressions as input and returns a list of CPMpy expressions,
+            otherwise, returns a list of CPMpy expressions and new constraints to be added toplevel.
 
         - supported: a set of supported global constraints or global functions
-        - supported_reified: a set of supported global constraints within a reification
+        - supported_nested: as set of supported global constraints/functions within other expressions
```
tias, line 18: I don't understand the need for this. All expressions in CPMpy can be used 'nested' (either reified or as part of a numeric operation)? why does this need special handling?

cpmpy/transformations/decompose_global.py
```diff
@@ -1,107 +1,134 @@
...
+                        _toplevel.extend([all(impl), all(neg_impl)])
+                else:
+                    newlist.append(expr)
+            else:
+                # numerical global function (e.g. min, max, count, element...) in non-comparison context
+                if (nested and expr.name not in supported_nested) or (not nested and expr.name not in supported):
+                    aux = intvar(*expr.get_bounds())
+                    newlist.append(aux)
+                    l1, l2 = expr.decompose_comparison("==", aux)
+                    newlist.extend(l1 + l2) # all constraints should be added toplevel, node replaced by aux var
```
tias, line 83: as the comment hints, I think this is meant to be added to _toplevel

cpmpy/transformations/decompose_global.py
```diff
@@ -1,107 +1,134 @@
...
+            #   while more optimally, its just `min(x,y,z).decompose_comparison('==', a)`
+            #    so have to be careful here and operate from 'one level above'
+
+            lhs, rhs = expr.args
+            if hasattr(lhs, "decompose_comparison"):
+                lhs = copy.copy(lhs)
+                lhs.args = decompose_in_tree(lhs.args, supported, supported_nested, _toplevel, nested=True)
+                lhs_supported = (nested and lhs.name in supported_nested) or (not nested and lhs.name in supported)
+                if not lhs_supported:
+                    newexpr, _toplevel[len(_toplevel):] = lhs.decompose_comparison(expr.name, rhs)
```
tias, line 99: the in-place iterator trick is very subtle, too few people will understand it, please separate it out as a separate `.extend(l2)`

cpmpy/transformations/decompose_global.py
```diff
@@ -1,107 +1,134 @@
...
+                    expr = decompose_in_tree(newexpr, supported, supported_nested, _toplevel, nested=nested)  # handle rhs
```
tias, line 100: lhs is already decomposed, so only rhs needs to be decomposed? not the entire expression again?

tias, line 100: lhs is already decomposed, so only rhs needs to be decomposed? not the entire expression again?

cpmpy/transformations/decompose_global.py
```diff
@@ -1,107 +1,134 @@
...
+                    newlist.append(all(expr))  # this can create toplevel-and
```
tias, line 101: then use `newlist.extend(expr)`?

cpmpy/transformations/decompose_global.py
```diff
@@ -1,107 +1,134 @@
...
+            else:
+                lhs = decompose_in_tree([lhs], supported, supported_nested, _toplevel, nested=True)
+                lhs = lhs[0]
+
+            if hasattr(rhs, "decompose_comparison"):
+                rhs = copy.copy(rhs)
+                rhs.args = decompose_in_tree(rhs.args, supported, supported_nested, _toplevel, nested=True)
+                rhs_supported = (nested and rhs.name in supported_nested) or (not nested and rhs.name in supported)
+                if not rhs_supported:
+                    # we now lhs is supported, so just flip comparison
```
tias, line 112: 'know'

cpmpy/transformations/decompose_global.py
```diff
@@ -1,107 +1,134 @@
...
+            # recreate original comparison
+            newlist.append(eval_comparison(expr.name, lhs, rhs))
+            continue
+
+        else:  # constants, variables, direct constraints
+            newlist.append(expr)
+
+    if nested is True:
+        return newlist
+    elif len(_toplevel):
```
tias, line 130: I prefer an explicit condition e..g len(topl)>0
(and then to have the ==0 case first ; )

note that in the else you do toplevel_list(newlist) while in the other you do newlist + ... without topleveling it first.

tias (commented):
we are progressing : ) this is an important piece to get right

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Yes, I'm not sure if we should do this... It actually will save some auxiliary variables indeed, but might blow up the codebase...

cpmpy/transformations/decompose_global.py
IgnaceBleukx: Solvers can support differnet global constraint depending on the context. I.e., AllDiff is supported by OR-tools toplevel but not in a nested (reified) context. Therefore, we need the distinction.

cpmpy/transformations/decompose_global.py
IgnaceBleukx: We can only do this toplevel. If expr is an argument of another expression, it should be replaced with a single expression here.

Hmm ok this makes me think if we actually should split it into a toplevel_decomposition and recursive_decomposition part...

cpmpy/expressions/globalconstraints.py
tias: I've removed this new restriction to total, to be discussed in another PR e.g. #184

tias (commented):
I think this is now looking good and ready to!

EXCEPT the changes to the tests, more specifically the removal of the Element partial test and some related code in `reifcation.py`. I think they should be left for a separate pull request where we can fine-tune our support and understanding of partial functions better. In the mean time, I would leave all the element code as is.

OK @IgnaceBleukx ? can you do this and do a final code review + merge? thx

IgnaceBleukx:
Ok, so re-added the reification code to decompose the partial version of Element.
All tests are running, as wel as bigtest. So will merge this into master.

PR #337, opened 2023-06-01
minor fix: globalconstraint.py and python_buildins.py

363734:
In python_buildins.py: minor fixes to comply with Pep 8
In globalconstraint.py: minor fixes to comply with Pep 8, some typos, one refactoring in the cumulative constraint to use numpy array operation instead of for loops


cpmpy/expressions/globalconstraints.py
```diff
@@ -577,14 +577,8 @@ def decompose(self):
...
         lb, ub = min(s.lb for s in start), max(s.ub for s in end)
-        for t in range(lb,ub+1):
-            demand_at_t = 0
-            for job in range(len(start)):
-                if is_num(demand):
-                    demand_at_t += demand * ((start[job] <= t) & (t < end[job]))
-                else:
-                    demand_at_t += demand[job] * ((start[job] <= t) & (t < end[job]))
-            cons += [capacity >= demand_at_t]
+        for t in range(lb, ub+1):
```
IgnaceBleukx, line 591: This should work but it doesn't... Our tests decomposing `Cumulative` are now failing.
It exposes another issue related to our implementation of numpy arrays.
I'll open a separate issue for that, but until that is fixed, we should revert this.

363734:
I revert the numpy formulation that was blocking. I'm opening an issue for it, so it can be fixed later. The rest can be pushed now

Issue #334, opened 2023-05-31
Subclasses for operation/comparison/subtype of binary/unary/nary expressions


363734:
Wouldn't it be better to have subclasses for each operation and comparison (ie, one for ==, one for sum,...)? 
I know it would probably increase the code size, but it would simplify the actual code containing lots of switches and cases applied only to one of the operations. It would help ease the testing of each type of expression, and helps delegate to each class the simplification, the decomposition,.... And thus also diminish runtime execution by having all the switches removed. And It would increase the maintainability of each operation and the addition of new ones.
And you could still have a static method to create the different operations with a switch if it is easier from the user's point of view.

It helps resolve the issue pointed out in PR #327

Wout4:
I think this does make sense, but it might take some serious refactoring.. Enhancement label seems appropriate ;)

PR #338, opened 2023-06-01
Minor fix in core.py

363734:
pep 8 compliance (space,...), some small refactoring (local imports only when needed, removal of redundant code), renaming of local variable to avoid using already used function names

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -208,23 +208,23 @@ class Expression(object):
         return Operator("or", [other, self])
 
     def __xor__(self, other):
-        # avoid cyclic import
-        from .globalconstraints import Xor
         # some simple constant removal
         if is_true_cst(other):
             return ~self
         if is_false_cst(other):
             return self
+        # avoid cyclic import
+        from .globalconstraints import Xor
         return Xor([self, other])
 
     def __rxor__(self, other):
-        # avoid cyclic import
-        from .globalconstraints import Xor
         # some simple constant removal
         if is_true_cst(other):
             return ~self
         if is_false_cst(other):
             return self
+        # avoid cyclic import
+        from .globalconstraints import Xor
         return Xor([other, self])
 
     # Mathematical Operators, including 'r'everse if it exists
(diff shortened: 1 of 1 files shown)


IgnaceBleukx:
Looks good, but a test in flatten fails because of the changes to the `__mod__` ufunc.

Now I've done I deep dive into the github workflows in #339 , it seems we can also automatically check for PEP 8 compatiity of our code and unused imports when pushing. Pycharm indeed throws some warnings but other IDE's might not.
Shoud we add this to the workflow? What do you think @tias ?

363734:
Do you have an idea why the flatten fail?

IgnaceBleukx:
It's just because of the optimization you did to `mod`.
If you run `pytest tests/test_flatten.py` you should get the error too.

tias:
Yes, as we mature it makes sense that we add linter checks (I guess that pep8 is that?) also coverage checks (some other branch mentioned that).

As we go along : ) If you know how, go ahead

cpmpy/expressions/core.py
```diff
@@ -272,21 +277,25 @@ def __rmul__(self, other):
...
         if is_num(other) and other == 1:
             return self
         return Operator("div", [self, other])
+
     def __rfloordiv__(self, other):
         return Operator("div", [other, self])
 
     def __mod__(self, other):
+        if is_num(other) and other == 1:
+            return self
```
JoD, line 296: This should be `return 0` - anything mod 1 is 0?

JoD: Or just drop the optimization. This kind of optimization may be better in an actual transformation instead of sprinkled throughout the code (though it is tempting! :D )

JoD: Maybe this is why the test fails?

363734: ho my bad yes... mod 1 = 0... This is what copy-paste leads to.. I'm updating the PR

363734: I also agree that this kind of optimization should be done elsewhere (linked to issue #334

JoD:
The mod test still was not happy, it required `str(a % 1 == 0)` to be `"(IV0) mod 1 == 0"` and not `"0"`. Since you agreed that this was not the right place for this optimization, and the test suggests the same, I just reverted it.

JoD:
Ok, all tests work, there's nothing controversial in this PR, so I've accepted it.

PR #339, opened 2023-06-01
Updated workflow when pushing

IgnaceBleukx:
Update github actions file for pushing.
Now the current branch should be ran instead of master. Also simplified the script somewhat to be more inline with what Github recommends: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -1,42 +1,26 @@
-name: Test-code
-# Triggers the workflow on push or pull request events
-on: [push, pull_request]
-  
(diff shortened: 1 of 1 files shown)


Wout4:
do you know if this will make the z3 tests run?

Wout4:
runtime could be cool. I think we had both on push and on pull request enabled with the idea of maybe doing a more extensive check only on pull requests. runtime and code coverage might be a bit much for every push? Or maybe not if it runs in the cloud anyway

PR #340, opened 2023-06-01
Add explicit requirements for jupyter notebooks examples

363734:
Add for all example notebooks the explicit list of Python packages the user should have (in addition to cpmpy) to make the notebooks run


363734:
If it is possible using the requirements file, it would be the best! It would simplify the life of the users and avoid the need to search all the commands to install all the packages separately

examples/nqueens_1000.ipynb
```diff
@@ -12,7 +12,9 @@
     "Somebody had written on a slide that CSP solving with backtracking search, arc consistency and a good variable ordering makes solving n-queens for N=1000 'feasible'.\n",
     "\n",
     "So, this notebook checks how 'feasible' n-queens for different N is for a modern CP solver like or-tools CP-SAT.\n",
-    "(technical note: we first tuned the hyper parameters on N=100 so that the solver is performing at its best)"
+    "(technical note: we first tuned the hyper parameters on N=100 so that the solver is performing at its best)\n",
+    "\n",
+    "This example requires you to install _numpy_."
```
IgnaceBleukx, line 17: Numpy is a requirement of CPMpy itself, so no need to mention it here.

Wout4:
The conclusion here was to not use the requirement file, since that would lead to many otherwise unneeded packages being installed (and we don't have time to figure out the optional requirements). So this should be good to merge now?

IgnaceBleukx:
Btw what we can also do is to install the packages from within the notebook. 
This way you do not need to put it in the `requirements` file, but also allows you to skip manual installation.

Actuallyt, any system-command can be executed using '!' before the command.
E.g., in the first cell of the notebook we can do `! pip install plotly geopy` for example. Is this something we want?

363734:
I would not recommend putting the pip install within the notebook. As the user could want to define its own conda environment to test the notebook. 

It can be merge

IgnaceBleukx:
I'm in the process of adding it to the notebooks, should be ready to merge by the end of the day

PR #347, opened 2023-06-12
Remove redundant test for gcc (leftover)

Dimosts:
For some reason we had 2 different test functions for gcc, the test_global_cardinality_count and the test_gcc. They were doing the same thing. I had only noticed and changed the test_global_cardinality_count in my PR that was merged, and the test_gcc redundant test was left same as in the previous version, and thus now fails with the new syntax. 

In this PR I just remove the redundant test test_gcc, as we already have tests in test_global_cardinality_count


Dimosts:
ok, changed the typecheck tests then to the new version.

Dimosts:
I now see what happened, the typechecks tests were not introduced when this PR opened, thus there were no typecheck tests to change then, this only occured when merged to main, where in the meanwhile these tests have been introduced.

IgnaceBleukx:
Ok, looks good so I'll merge. Once the new master is merged into other open PR's, github will stop complaining about failed tests there as well.

PR #355, opened 2023-06-21
catch beginner mistakes

tias:
from doc_rework branch

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -176,6 +176,9 @@ class Expression(object):
     # Boolean Operators
     # Implements bitwise operations & | ^ and ~ (and, or, xor, not)
     def __and__(self, other):
+        # catch beginner mistake
+        if is_num(other):
+            raise TypeError(f"{self}&{other} is not valid because {other} is a number, did you forgot to put brackets? E.g. always write (x==2)&(y<5).")
         # some simple constant removal
         if is_true_cst(other):
             return self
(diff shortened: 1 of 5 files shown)


Wout4:
Do we also want to catch invalid constraints that are given to the constructor?
Model([x + 7]).solve()
And maybe also a check for adding some none-expression numerical value (maybe less needed) (Model += 8)

tias:
that feels duplicate, maybe we should call self.cons = [] and then self += cons?

Wout4:
adding constraints in the constructor is commonly done I think, it's also done in our examples, so added the check in the constructor as well

Wout4:
true, that should work, but ran into the issue that the add function uses append (leading to nested list of constraints) where as the constructor was setting a flat list of constraints. Some test are failing because of this. When adding a list (or tuple) of constraints should we just add them as is (nested) or maybe unnest them first?

Wout4:
So i see 4 options here:
1. Don't test constraints in model init (original pr)
2. copy paste the tests to the init (silly)
3. call add within init to have the tests included
3a. make _add_ use extend in stead of append (no more nested constraints)
3b. change the tests that are now failing to expect nested lists of constraints. (always nested constraints)

Not sure what to do here, I can see how the add function creating nested lists of constraints keeps them tidy and organised. But the way it works now is that the constraints added in init will not be nested and the ones added with the add function will be nested, this is a bit counterintuitive.

Dimosts:
I think we should be consistent, and not flatten the list of constraints in the constructor, this can happen in transformations.

PR #356, opened 2023-06-21
test comparisons between bool and int variables

tias:
from doc_rework, the test is literally from the new docs (was failing for ortools, includes fix)

I had a look at adding it to test_constraints, but the generator to which to add these cases is not clear to me...

if there is a way to simply test this with all solvers, without having to add it to test_constraints, that would also be fine for me. Even these as is just for ortools is already something.

### DIFF
--- a/cpmpy/solvers/z3.py
+++ b/cpmpy/solvers/z3.py
@@ -25,10 +25,8 @@ from ..exceptions import NotSupportedError
 from ..expressions.core import Expression, Comparison, Operator, BoolVal
 from ..expressions.globalconstraints import GlobalConstraint, DirectConstraint
 from ..expressions.variables import _BoolVarImpl, NegBoolView, _NumVarImpl, _IntVarImpl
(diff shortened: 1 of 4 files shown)


IgnaceBleukx:
z3 was complaining quite a bit, updated the `__add__` code and `test_constraints.py` to include the new cases.

I had some issues testing Minizinc with the new comparisons, but probably due to my installation... Would be good if someone can rerun the tests for minizinc before merging

Wout4:
test_constraints.py runs fine for me

Wout4:
that's the only test file you needed double checked right?

Wout4:
i do have gecode 6.3 as the default solver though.. Don't think i can easily downgrade to test

IgnaceBleukx:
Ok, yes thats the one. 
Than this should be ready for review and merge.

Wout4:
looks good, maybe some obsolete tests now that it's added to test_constraints but more tests doesn't hurt

PR #357, opened 2023-06-22
abs as global function

Wout4:
abs as a global,  might have forgotten some cleanups?

### DIFF
--- a/tests/test_globalconstraints.py
+++ b/tests/test_globalconstraints.py
@@ -2,7 +2,7 @@ import copy
 import unittest
 import cpmpy as cp
 from cpmpy.expressions.globalconstraints import GlobalConstraint
-from cpmpy.expressions.globalfunctions import GlobalFunction
+from cpmpy.expressions.globalfunctions import GlobalFunction, Abs
 from cpmpy.exceptions import TypeError
 
 class TestGlobal(unittest.TestCase):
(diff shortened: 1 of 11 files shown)


tests/test_globalconstraints.py
```diff
@@ -2,7 +2,7 @@
 import unittest
 import cpmpy as cp
 from cpmpy.expressions.globalconstraints import GlobalConstraint
-from cpmpy.expressions.globalfunctions import GlobalFunction
+from cpmpy.expressions.globalfunctions import GlobalFunction, Abs
```
Dimosts, line 5: For the rest of the global constraints and global functions we do not import them and just use cp.function, e.g. cp.max. To be consistent, in this case we should not import abs (it is weird to be the only one imported) and just use cp.abs

tests/test_globalconstraints.py
```diff
@@ -470,6 +470,15 @@ def test_bounds_maximum(self):
         self.assertFalse(cp.Model(expr<lb).solve())
         self.assertFalse(cp.Model(expr>ub).solve())
 
+    def test_bounds_abs(self):
```
Dimosts, line 473: Should we also put normal tests for abs besides testing the bounds? I mean, asserting that the value is the expected one, as we do with the rest of the global constraints/functions

Dimosts (commented):
I think it is almost ok, easy to review, could be merged before freeze. Just a couple of comments for consistency.

JoD:
Ran `test_constraints.py` with the exact branch merged with this one. No issues, so good for me.

The only concern I have is that we might break bigtest even more?

Wout4:
aha that is indeed annoying.. I think the only way out would be to allow abs as an operator as well but that's a bit ugly

PR #364, opened 2023-06-26
flatten simplification fix

Wout4:


cpmpy/transformations/flatten_model.py
```diff
@@ -236,7 +236,7 @@ def flatten_constraint(expr):
                 rewritten = True
 
             # rewrite 'BoolExpr != BoolExpr' to normalized 'BoolExpr == ~BoolExpr'
-            if exprname == '!=' and lexpr.is_bool():
+            if exprname == '!=' and lexpr.is_bool() and is_boolexpr(rexpr):
```
JoD, line 239: why the different check between both (`is_bool` vs `is_boolexpr`)

JoD: In any case, it works, thanks!

PR #367, opened 2023-06-30
Allow description for a constraint

IgnaceBleukx:
For explanation techniques, its usefull to have a human-intepretable representation for a constraint.
This pull request allows a user to create a constraint and then set its description, when the constraint is printed, the description is taken instead of the `__repr__` method.

I guess we should take extra care for Minzinc for this one as we might convert an expression based on its string representation? But I could not see this happening anywhere.

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -107,13 +107,18 @@ class Expression(object):
         assert (is_any_list(arg_list)), "_list_ of arguments required, even if of length one e.g. [arg]"
         self.args = arg_list
 
-    def set_description(self, txt):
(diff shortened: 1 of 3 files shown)


Wout4:
Not really related to this pr, but do you think we could do something similar to make a parser? in stead of printing Not(x) the parse_str() could return Operator('not',[x]) or ~x. this would  then make parsing models from a file quite easy.

Wout4:
You could set a description for a variable as well, which i think is desirable. However some solvers use the string representation of a variable as the name to send to the solver. This should probably be changed to use the name (or uuid in another pr)

Issue #376, opened 2023-07-05
Not backwards compatible model init


Dimosts:
The new model init is not compatible with what was done before, when a decomposed global is used.

Instead of adding the constraints in model.constraints, it adds the list of constraints of the decomposition.

Minimal case;

```
import cpmpy as cp

ivs = cp.intvar(0,100, shape=10)

m = cp.Model(cp.AllDifferent(ivs).decompose())
C = m.constraints

print(C)
```

This prints:

`[[(IV0) != (IV1), (IV0) != (IV2), (IV0) != (IV3), (IV0) != (IV4), (IV1) != (IV2), (IV1) != (IV3), (IV1) != (IV4), (IV2) != (IV3), (IV2) != (IV4), (IV3) != (IV4)]]`

which is a list with 1 element, a list of constraints. 

In cases where the user wanted to take m.constraints and use it directly as a list of constraints this can create problems..

The problem occurs because now the decompose() function of globals returns a tuple of lists of constraints.

Wout4:
it's the decompose function that is not backwards compatible or the model init?

tias:
That is the right question Wout.

Decompose now returns a tuple of (constraints, var_decalring_constraints)
and so a tuple of a list and an emptylist for alldiff.

So the backward incompatible change is in alldiff.decompose(), which few people use directly.

So `Model()` is still backward compatible and we will not change it. So indeed you will have to do the changes in your constraint acquisition code Dimos...

Wout4:
The answer might actually be that both are not backwards compatible, since we (I) did also change the model constructor

PR #380, opened 2023-07-05
Backwards compatible model initialization

Dimosts:
Use toplevel_list from normalize to unravel list of constraints in model initialization. Solve #376.


tias:
This will also de-nest nested nests, which will make the constraint structure different from what the user input; if they create a model and then print it, it should look like what they wrote.

tias:
the code is backward compatible the way it was, so not merging

(bug #376 was a special case due to a change in decompose())

PR #384, opened 2023-07-06
skip tests in examples if required packages not installed

Wout4:
remove z3 test from abs

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -33,6 +33,23 @@ class TestExamples(unittest.TestCase):
     Args:
         example ([string]): Loaded with parametrized example filename
     """
+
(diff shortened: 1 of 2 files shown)


Wout4:
I'm sure there is some more elegant way to do this but it should work

tias (approved):
works now on fresh install.

I don't know if there is a way to catch all ModuleNotFoundErrors when the test is run, but this is also fine thx : )

Wout4:
ah yeah that makes more sense

Wout4:
messed up the commit message, but it now just checks for any modulenotfounderror

tias:
tested in fresh venv and original system, all fine now

PR #252, opened 2023-03-11
Add Inverse() global constraint

bradediger:
Adds `Inverse(fwd, rev)`, which constrains the arrays `fwd` and `rev` to be inverse maps. 

- CP-SAT uses native `AddInverse()`
- MiniZinc uses `inverse()`, with a bit of caution around 1-based indexing
- Other solvers use the standard decomposition.

Feedback welcome, thank you for the CPMpy library!

PR #391, opened 2023-07-10
fix: decompose_global: mark inverse as supported

bradediger:
Hi there, good day! In #252, I added `Inverse` as a global constraint, with or-tools support. Since then, ortools's `decompose_global` forgot that it's supported, and decomposes it instead of passing it through natively.

This change should restore support for native inverse. Thanks!


Wout4:
Thanks for catching this oversight, and sorry for any inconvience!
Pretty straightforward fix, will merge it right away

PR #393, opened 2023-07-11
Make element total

Wout4:


Wout4:
After discovering that solvers enforce the element bounds at toplevel, we should do the same. This also simplifies handling reified element constraints

Wout4:
We decided to enforce strict element bounds at construction time. This eliminates the need for checking the bounds in the decomposition, and any trouble with incomplete functions.
It comes however at a slight inconvenience to the user, that has to give precise bounds to the index variable.

Wout4:
Another option would be to allow the creation of such an element constraint, but enforcing the bounds at toplevel and throwing an error when trying to get the value of an element constraint with index out of bounds.

Wout4:
Enforced that element is Total at construction time, and because of that also enforce strict bounds for Inverse and Circuit global constraints. Added tests for this and altered the existing tests to use accurate bounds

Wout4:
@Dimosts can you maybe check the changes in the chess example? As you can see this is mostly changes to our tests and examples, so that they use strict bounds

cpmpy/expressions/globalconstraints.py
```diff
@@ -223,6 +223,12 @@ def __init__(self, *args):
         flatargs = flatlist(args)
         if any(is_boolexpr(arg) for arg in flatargs):
             raise TypeError("Circuit global constraint only takes arithmetic arguments: {}".format(flatargs))
+        lbs, ubs = zip(*[get_bounds(arg) for arg in flatargs])
+        lb = min(lbs)
+        ub = max(ubs)
+        if lb < 0 or ub >= len(flatargs):
+            raise TypeError("Circuit global constraint is only defined if all bounds are between 0 and the length of the array.")
```
IgnaceBleukx, line 230: I wouldn't call it a type error, but rather a PartialFunctionError or something?  I think we even have one already

IgnaceBleukx: Maybe make the error message more explicit too? 
Smt like: `elements in the array must have bounds 0 <= x <= len(flatargs) but got bounds [{lb},{ub}] for arg ...`

cpmpy/expressions/globalconstraints.py
```diff
@@ -278,6 +284,12 @@ def __init__(self, fwd, rev):
         if any(is_boolexpr(arg) for arg in flatargs):
             raise TypeError("Only integer arguments allowed for global constraint Inverse: {}".format(flatargs))
         assert len(fwd) == len(rev)
+        lbs, ubs = zip(*[get_bounds(arg) for arg in flatargs])
+        lb = min(lbs)
+        ub = max(ubs)
+        if lb < 0 or ub >= len(fwd):
+            raise TypeError("All arguments must be within array bounds for Inverse to have meaning over the whole domain!")
```
IgnaceBleukx, line 291: Also here can be more explicit

IgnaceBleukx (commented):
Looks good!

You can also link to the github issue you created in the error messages.
Then, if anyone has some questions about it they can comment on the issue?

IgnaceBleukx:
I'm wondering if we should have a "vectorized check" util function or something as we are duplicating code quite a bit for these checks.

It would take as input a (nested) list of elements and a boolean function which has to be checked for each of the args.
Then you can return the object vioating the condition if any? It would also simplify several type checks in other constraints

Wout4:
Yes that could be nice, it's a bit ugly now, the reason I put it there is so the work happens once you already know there is a bounds issue (with an easier check), so that you don't have the overhead on normal inputs. 
Seems marginal however.. Will check it out in another pr, since it will involve a lot of other globals

Wout4:
This will be replaced by our safening pr

PR #396, opened 2023-07-27
Exact solution hints

IgnaceBleukx:
Implemented solution hinting for exact.

Also added some tests to verify it works :)


JoD (approved):
Straightforward, looks good to me.

Issue #400, opened 2023-08-18
linearize alters number of solutions


Wout4:
Due to auxiliary variables being created on the rhs of an implication, the number of solutions can increase after linearize

Wout4:
example code:
`iv = intvar(1,2, shape=1)
bv = boolvar()
cons = [(Minimum([iv])!=1) == bv]
print(Model(cons).solveAll('ortools'))
print(Model(cons).solveAll('gurobi',solution_limit=3))` #more solutions according to gurobi
print(linearize_constraint(only_bv_implies(flatten_constraint(cons))))
cons = linearize_constraint(only_bv_implies(flatten_constraint(cons))) #its because of linearize (big m)
print(Model(cons).solveAll('ortools'))

PR #401, opened 2023-08-19
Updated linearize

IgnaceBleukx:
Updated linearize to exploit the "top-level-listness" of the input

Also added a fix for #400 which maps all new vars to their lb if the condition of a reification is False


Wout4:
I fixed the failing tests, I think it is fine like this, also adding a constraint to stop extra solutions is nice

PR #402, opened 2023-08-21
UUID

Wout4:
Added an ID fields to variables, using uuid to get unique id's.
Naming stays the same, for display/print reasons.
When posting to the solver I decided to use name + id because
- it will still be unique
- minizinc names have to start with a letter, this ensures that and makes it consistent for all solvers
- might make debugging easier as the var name is still part of the solver var name

Let me know if you think this is silly.

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -347,6 +347,7 @@ class NegBoolView(_BoolVarImpl):
         # it already comply with the asserts of the __init__ of _BoolVarImpl and can use 
         # __init__ from _IntVarImpl
         _IntVarImpl.__init__(self, 1-bv.ub, 1-bv.lb, name=str(self))
+        self.id = bv.id
 
     def value(self):
         """ the negation of the value obtained in the last solve call by the viewed variable
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -47,6 +47,7 @@
     ==============
 """
 import math
+import uuid
 from collections.abc import Iterable
 import warnings # for deprecation warning
 from functools import reduce
@@ -238,6 +239,7 @@ class _NumVarImpl(Expression):
         self.lb = lb
         self.ub = ub
         self.name = name
+        self.id = uuid.uuid4()
         self._value = None
 
     def is_bool(self):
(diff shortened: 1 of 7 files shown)


363734:
I think this will also be useful for the CSE, as you can then create easily a unique hash for each expression based on the type of operation and the argument of it

JoD:
This is hard to review because of the autoformatting changes. Would it be possible to add these in a separate MR? Also, provide an explanation for all collaborators on how to use the same autoformatter, so that the autoformatting does not get broken in future MRs.

Summarized: don't mix autoformatting with feature MRs.

JoD:
I'm lukewarm on using a `uuid`. The alternative is using a global variable counter. The advantage of a global counter is that it will also be faster: long string names of variables do have a measurable impact (both for memory and efficiency) in scenario's with lots of variables passed to a solver, so keeping them short (via a counter starting from 0) is better.

However, this is just an optimization issue, so if you think the current approach is fine, that's fine by me.

363734:
global variable counter does not work with pickled models, as the counter is not pickled. For example, you define a first model with variable id 1 to 10, counter is at 10, you pickle the model, fine. Then you unpickle it in another run, counter is resetted, you decide to add new var to model, var id 1 reused, conflict

JoD:
Ah, did not think of that. The counter is indeed part of the state and should be pickled too, but I can see that this may tip the scales in the balance between simplicity and efficiency.

Wout4:
I will try to make a case again for why we need unique identifiers :)  (to implement CSE)

After looking into python dicts/sets it turns out that in the event of a hash collision `__eq__` gets called on the keys, to decide if they are the same (so not the pointer, nor `is`)
Since we overwrite __eq__ for cpmpy expressions to return a new comparison object, __eq__ always returns a 'truthy' value.
So 2 expressions with the same hash can never co-exist in a dictionary where we use the expressions themselves as keys.
This means we will need unique hashes if we want to use expressions as keys. (-> uuid needed)

On the other hand, we could use something else as the key, to avoid the problem with overwriting __eq__. But then we would still need a unique property of expressions to use as the key, since the string representation could not be unique (when the user gives particular names or descriptions to the variables/expressions) (-> uuid needed)

PR #405, opened 2023-09-06
circumvent gurobi integer division to match cpmpy floordivision

Wout4:
There does not seem to be floordivision in gurobi, so I implemented a workaround using continuous (gurobi) variables.


JoD:
Interesting. So Gurobi supports integer division that rounds towards zero instead of floor? E.g., `-3/2` yields `-1` instead of `-2`? What does Gurobi give with `3/-2`?

What is the core idea of your fix? I did not immediately get it from the code - something with abs?

Would it make sense to just have a decomposition for `div`? E.g., `a/b` is decomposed to `q` with toplevel `q*b=a+r` and `0<=r<b`.

Wout4:
we fixed this by linearizing division

Issue #407, opened 2023-09-07
Solution hinting not working as intended when NDVarArray more than 1 dimension 


Dimosts:
If we define a variable array with more than 1 dimension, solution hinting is not working

Slightly changing the example we have in the docs for solution hinting (https://cpmpy.readthedocs.io/en/latest/solvers.html) to 2D var array, it is not working anymore.

Reproduce:
from cpmpy import *
x = intvar(0,10, shape=(2,2))
s = SolverLookup.get("ortools")
s += sum(x) <= 10
s.solution_hint(x, [[1,2],[3,4]])
s.solve()
print(x.value())

Error:
TypeError: unhashable type: 'NDVarArray' (line 244 of ortools.py)

However, the error actually comes from lines 522-524, where we call solver_var(), assuming that cpm_var is 1 dimention, using a for to iterate.

Wout4:
looks good to me, but I can't get ortools to actually give me a solution that is using the hints. Is that because the problem is just too easy?

Dimosts:
Probably yes, because it is too easy. Sme happens with the 1-D solution hinting we have in master

PR #408, opened 2023-09-07
Solution hinting fix, when giving N-D variables and values

Dimosts:
Connected to issue #407 

Flattening the variables and values first, before moving to the for loop of adding the hints.

Covering cases that vars and values are NDVarArrays, ndarrays or anything that has a .flat attribute + cases that the user may give lists for hinting.

### DIFF
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -588,6 +588,15 @@ class CPM_exact(SolverInterface):
         :param cpm_vars: list of CPMpy variables
         :param vals: list of (corresponding) values for the variables
         """
+
+        if not hasattr(cpm_vars, "flat"):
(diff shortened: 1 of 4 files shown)


IgnaceBleukx:
We actually have a very efficient `flatlist` utility function in `expressions.utils`, it seems to me we can just call `flatlist(vars)` and `flatlist(vals)` before posting the solution hints? Shouldn't that fix it?

Dimosts:
Indeed, it is cleaner that way, I didn't notice we had a function for that. Changed it to use flatlist.

IgnaceBleukx:
Did some cleanup and added a tests,good to merge now for me.

IgnaceBleukx (approved):

PR #411, opened 2023-09-13
remove unused code

Dimosts:
Some code in existing transformations is unused due to changed happened later. I found a part in linearize, I guess we have more. Opening this PR as draft so we can collect more cases before merging it


Wout4:
Will merge this already, we can always make a new pr if stuff comes up

PR #412, opened 2023-09-17
scip interface (first version)

tias:
with help of Mark Turner from SCIP.

Now... turns out that @IgnaceBleukx  already worked on a SCIP interface too, about a year ago (e.g. [SCIP branch](/CPMpy/cpmpy/tree/scip)

So maybe Ignace can now make a best of both worlds? : )
(current branch is up to date with current template, but the __add__ part is minimal, e.g. reified linear is not properly tested/supported at the least).


IgnaceBleukx:
Updated the interface somewhat, it seems SCIP does not have the fancy Min/Max/Abs constraints built-in in its API : /

Added the SOS1 and cardinality constraints to the interface too, these constraints are created when decomposing alldiff for example so definately usefull to have them imho. We should also add them to the Gurobi interface

SolveAll is in a weird state for PySCIPOpt: SCIP does allow for retrieving all feasible solutions, but the Python interface does not seem to allow for collecting the solutions which are found... Added the issues to track in the code so we can add the specialized implementation when these are resolved.

All tests are passing so ready for review I think.

Wout4:
I get an error for optimization problems:
```
\cpmpy\solvers\scip.py", line 154, in solve
    self.objective_value_ = scip_objective.getObjVal()
AttributeError: 'pyscipopt.scip.Expr' object has no attribute 'getObjVal'
```

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,419 @@
...
+        # translate exit status
+        if scip_status == "optimal":
+            if self.has_objective():
+                self.cpm_status.exitstatus = ExitStatus.OPTIMAL
+            else:
+                self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+        elif scip_status == "infeasible":
+            self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+        elif scip_status.endswith("limit"):  # timelimit, nodelimit, ...
+            self.cpm_status.exitstatus = ExitStatus.FEASIBLE
```
Wout4, line 131: This should only be feasible if at least one solution is found, is there a way to check this? Otherwise should be set to unknown to avoid trying to print a solution that doesn't exist later on

Wout4 (dismissed):

cpmpy/solvers/scip.py
Opt-Mucca: To check if a valid solution exists in storage you can call:
```if self.scip_model.getNSols() > 0```

File: cpmpy/solvers/scip.py [hunk 2]
```diff
@@ -0,0 +1,419 @@
...
+        elif scip_status == "unknown":
+            self.cpm_status.exitstatus = ExitStatus.UNKNOWN
+        # TODO: "userinterrupt" and check if there is a solution?
```
Opt-Mucca, line 134: Same as above. Check if there are any solutions in storage and then to obtain the solution call:
```scip_sol = self.scip_model.getBestSol()```

File: cpmpy/solvers/scip.py [hunk 3]
```diff
@@ -0,0 +1,419 @@
...
+            # fill in variable values
+            for cpm_var in self.user_vars:
+                solver_val = self.scip_model.getVal(self.solver_var(cpm_var))
+                if cpm_var.is_bool(): 
+                    cpm_var._value = solver_val >= 0.5
+                else:
+                    cpm_var._value = int(solver_val)
+            # set _objective_value
+            if self.has_objective():
+                self.objective_value_ = scip_objective.getObjVal()
```
Opt-Mucca, line 154: This should raise an error. The ```getObjVal()``` function can only be called for ```scip_model```, not for the objective expression. So I'd say that you can also remove the line: 
```scip_objective = self.scip_model.getObjective()```

File: cpmpy/solvers/scip.py [hunk 4]
```diff
@@ -0,0 +1,419 @@
...
+
+        self.scip_model.freeTransform()  # Mark Turner from SCIP told me you need to do this if you want to support adding additional vars/constraints later...
```
Opt-Mucca, line 156: So a tiny summary for this: SCIP transforms the problem that you are actual solving into something it believes is easier to solve (presolve techniques etc). All solutions when found are then transformed back to the original space, because that is obviously what the user modelled and can interpret. Allowing the user to change the model during solving gets messy when balancing these things however, so it's forbidden. What ```self.scip_model.freeTransform()``` does is to remove all information from the transformed space, and just leave the original model. All solutions are kept, and the user can now change the model as they will. The downside is that potentially useful information for speeding up the next optimisation call is thrown out.

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,419 @@
...
+        # translate solution values (of user specified variables only)
+        self.objective_value_ = None
+        if has_sol:
+            # fill in variable values
+            for cpm_var in self.user_vars:
+                solver_val = self.scip_model.getVal(self.solver_var(cpm_var))
+                if cpm_var.is_bool(): 
+                    cpm_var._value = solver_val >= 0.5
+                else:
+                    cpm_var._value = int(solver_val)
```
Opt-Mucca, line 151: I'm going to assume here that continuous variables just aren't used for CPMpy. If so then this should be fine.

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,419 @@
...
+                if (isinstance(lhs, Operator) and lhs.name == "sum" and all(a.is_bool() and not isinstance(a, NegBoolView) for a in lhs.args)):
+                    if rhs == 1: # special SOS1 constraint?
+                        self.scip_model.addConsSOS1(self.solver_vars(lhs.args))
+                    else: # cardinality constraint
+                        self.scip_model.addConsCardinality(self.solver_vars(lhs.args), rhs)
+                else:
+                    sciplhs = self._make_numexpr(lhs)
+                    self.scip_model.addCons(sciplhs <= sciprhs)
+
+            elif cpm_expr.name == '>=':
```
Opt-Mucca, line 314: Just a general interest: Due to how the constraints are transformed are you certain that these consraints can't be multiplied by -1 to become cardinality constraints?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,419 @@
...
+                self.scip_model.addCons(sciplhs >= sciprhs)
+            elif cpm_expr.name == '==':
+                if isinstance(lhs, _NumVarImpl) \
+                        or (isinstance(lhs, Operator) and (lhs.name == 'sum' or lhs.name == 'wsum' or lhs.name == "sub")):
+                    # a BoundedLinearExpression LHS, special case, like in objective
+                    sciplhs = self._make_numexpr(lhs)
+                    self.scip_model.addCons(sciplhs == sciprhs)
+
+                elif lhs.name == 'mul':
+                    assert len(lhs.args) == 2, "scip only supports multiplication with 2 variables"
```
Opt-Mucca, line 325: I'm not sure what is meant here?
SCIP does support more than just quadratic constraints. So you can any order of degree constraint. For example:
```scip_model_addCons(a * b * c * d * e * f * g == sciprhs)``` should work fine

File: cpmpy/solvers/scip.py [hunk 8]
```diff
@@ -0,0 +1,419 @@
...
+                    a, b = self.solver_vars(lhs.args)
+                    #self.scip_model.setParam("NonConvex", 2)  # gurobi leftover
```
Opt-Mucca, line 327: This can be removed for SCIP. SCIP allows many more non-linear type formulations than just quadratic.
If you want to give options to SCIP on how to handle quadratic constraints then you can change the existing parameters.

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,419 @@
...
+                    self.scip_model.addCons(a * b == sciprhs)
+
+                elif lhs.name == 'div':
+                    assert is_num(lhs.args[1]), "scip only supports division by constants"
```
Opt-Mucca, line 331: This is also wrong I think..... You can just throw in the constraint regardless if both lhs and rhs are expressions:
```self.scip_model.addCons(a / b == sciprhs)```

File: cpmpy/solvers/scip.py [hunk 10]
```diff
@@ -0,0 +1,419 @@
...
+                    lhs = -lhs
+                self += cond.implies(lhs <= -rhs)
+
+            elif sub_expr.name == "==": # split into <= and >=
+                self += [cond.implies(lhs <= rhs), cond.implies(lhs >= rhs)]
+            else:
+                raise Exception(f"Unknown linear expression {sub_expr} name")
+
+        # True or False
+        elif isinstance(cpm_expr, BoolVal):
```
Opt-Mucca, line 378: Don't really understand what this constraint means, always struggle a bit with the CP language. If you can explain then I can see if there's a more natural MIP approach.

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,419 @@
...
+                        default/None: nothing displayed
+                - time_limit: stop after this many seconds (default: None)
+                - solution_limit: stop after this many solutions (default: None)
+                - call_from_model: whether the method is called from a CPMpy Model instance or not
+                - any other keyword argument
+
+            Returns: number of solutions found
+        """
+
+        warnings.warn("Solution enumeration is not implemented in PyScipOPT, defaulting to CPMpy's naive implementation")
```
Opt-Mucca, line 411: So I mentioned this is an email. I think it's just that very few MIP practitioners actually need such functionality. To mimic most of this:
- Call optimize for the ```self.scip_model```.
- After solving is completed. Call ```self.scip_model.getSols()```. 
The potential problems with this is that SCIP automatically deletes the "worst" solutions in storage after it has found x many. This is by default 10. See ```limits/maxsol``` and ```limits/maxorigsol```. So to save memory SCIP will at times remove older solutions that it knows are suboptimal because it has clearly better ones. The number of solutions found in total can be retrieved with ```scip_model.getNSolsFound()```, but unless you increase the limits you will not be able to retrieve all of them.

Opt-Mucca:
> I get an error for optimization problems:
> 
> ```
> \cpmpy\solvers\scip.py", line 154, in solve
>     self.objective_value_ = scip_objective.getObjVal()
> AttributeError: 'pyscipopt.scip.Expr' object has no attribute 'getObjVal'
> ```

This should now be addressed in one of my points above

Opt-Mucca:
@IgnaceBleukx @tias I haven't run anything, and truthfully haven't tried to understand the larger code base, but I've gone through and given comments that I think should make the interface work.
For a final point: While min / max expressions probably won't work, SCIP can handle the absolute value expressions fairly easily. I'm just not sure where to add them in the current code without some guidance.

cpmpy/solvers/scip.py
IgnaceBleukx: It did : ) Fixed it and will push it soon

cpmpy/solvers/scip.py
IgnaceBleukx: Jup, indeed we only allow for integer and boolean variables

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,419 @@
...
+                                                     binvar=self.solver_var(cond), activeone=True)
+
+            elif sub_expr.name == ">=": # change sign
+                if lhs.name == "sum":
+                    lhs = Operator("wsum", [[-1]*len(lhs.args), lhs.args])
+                elif lhs.name == "wsum":
+                    lhs = Operator("wsum", [[-w for w in lhs.args[0]], lhs.args[1]])
+                else:
+                    lhs = -lhs
+                self += cond.implies(lhs <= -rhs)
```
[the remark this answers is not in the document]
IgnaceBleukx: Yes, this line sends the new constraint `cond => lhs <= -rhs` through the entire `__add__` pipeline again.
But coming to think of it, we probably pay quite a penalty by doing so as it also sends the constraint through the transformation pipeline...
I'll refactor this somewhat so we avoid unnecessary transformations...

cpmpy/solvers/scip.py
IgnaceBleukx: This is a very uncommon case where a user posts the constraint "False" to the solver, in essence making the solver status trivialy infeasible. 
We run into this case sometimes when our expressions are simplified to "False"

cpmpy/solvers/scip.py
IgnaceBleukx: Aha no, actually we are not... I will do a check for it here

IgnaceBleukx: Aha actually no... I'll add a check for it here

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+            return version("pyscipopt")
+        except PackageNotFoundError:
+            return None
+
+    def __init__(self, cpm_model=None, subsolver=None):
+        """
+        Constructor of the native solver object
+
+        Arguments:
+        - cpm_model: a CPMpy Model()
```
hbierlee: missing arg description (although I'd say if the doc is the same as the superclass, not including it will inherit that doc)

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        if not self.supported():
+            raise ModuleNotFoundError(
+                "CPM_scip: Install SCIPOptSuite and the python package 'pyscipopt' to use this solver interface.")
+        assert subsolver is None, "SCIP does not support subsolvers"
+        import pyscipopt as scip
+
+        self.scip_model = scip.Model("From CPMpy")
+        self.scip_model.setParam("display/verblevel", 0) # remove solver logs from output
+        # initialise everything else and post the constraints/objective
+        # it is sufficient to implement __add__() and minimize/maximize() below
```
hbierlee: superfluous comment(s)?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+    def native_model(self):
+        """Returns the solver's underlying native model (SCIP Model) for direct solver access."""
+        return self.scip_model
+
+    def solve(self, time_limit=None, solution_callback=None, **kwargs):
+        """
+            Call the SCIP solver
+
+            Arguments:
+            - time_limit:  maximum solve time in seconds (float, optional). Persists across solve() calls until overridden.
```
hbierlee, line 118: is this persistence the same behaviour as in other solvers? Any way to make this uniform? I also don't really understand what is meant with it (with time_limit=10, if the first call is 6 seconds, will the next call have a limit of 4 seconds, or 10 again, or will setting time_limit=None overwrite it?)

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+            - kwargs:      any keyword argument, sets parameters of solver object (e.g. "limits/time" overrides time_limit).
+
+            Arguments that correspond to solver parameters (passed via setParams):
+            Examples: limits/nodes=1000, limits/solutions=1, parallel/maxnthreads=4,
```
hbierlee: how does one pass an arg "limits/nodes" as kwarg?

hbierlee: I suppose as `**dict` may be worth indicating

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                display/verblevel=4, separating/maxrounds=0.
+            For a full list see https://www.scipopt.org/doc/html/PARAMETERS.php
+        """
+        if time_limit is not None:
+            if time_limit <= 0:
+                raise ValueError("Time limit must be positive")
+            self.scip_model.setParam("limits/time", float(time_limit))
+
+        if solution_callback is not None:
+            raise NotSupportedError("SCIP: solution callback not (yet?) implemented")
```
hbierlee: Meaning SCIP does not support it or CPMpy? If latter, should it be in this PR?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        self.scip_model.setParams(kwargs)
+        _ = self.scip_model.optimize()
+
+        scip_status = self.scip_model.getStatus()
+
+        # new status, translate runtime
+        self.cpm_status = SolverStatus(self.name)
+        self.cpm_status.runtime = self.scip_model.getSolvingTime()
+
+        unknown_stati = {"unknown", "userinterrupt", "unbounded", "inforunbd", "terminate"}
```
hbierlee: stati?

hbierlee: apparently statuses is more correct/widely used

hbierlee: not reused, could just be inline in 161

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        if scip_status == "optimal":
+            if self.has_objective():
+                self.cpm_status.exitstatus = ExitStatus.OPTIMAL
+            else:
+                self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+        elif scip_status == "infeasible":  # proven unsat
+            self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+        else:
+            n_sols = self.scip_model.getSols()
+            n_sols = len(n_sols) if isinstance(n_sols, list) else n_sols
```
hbierlee: on what does this dpeend? I'd expect a comment on why this is necessary

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+            if scip_status.endswith("limit"):  # timelimit, nodelimit, etc.
```
hbierlee: I'd list the options here (although painful it will lead to less unexpected behavior)

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                self.cpm_status.exitstatus = ExitStatus.FEASIBLE if n_sols > 0 else ExitStatus.UNKNOWN
+            elif n_sols > 0:  # at least one feasible solution found, not proven optimal
+                self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+            elif scip_status in unknown_stati:
```
hbierlee: I feel like this whole section can be just matching on all these statuses, then checking n_sols, no need for all these casees

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        # use best solution object so values are correct (getVal can be wrong after transform)
+        self.objective_value_ = None
+        if has_sol:
+            best_sol = self.scip_model.getBestSol()
+            for cpm_var in self.user_vars:
+                if cpm_var not in self._varmap:
+                    continue  # variable never posted (e.g. unused); skip to avoid addVar after solve
+                scip_var = self._varmap[cpm_var]
+                if best_sol is not None:
+                    solver_val = self.scip_model.getSolVal(best_sol, scip_var)
```
hbierlee: might want to check if we can access multiple vars at once?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        has_sol = self._solve_return(self.cpm_status)
+
+        # translate solution values (of user specified variables only)
+        # use best solution object so values are correct (getVal can be wrong after transform)
+        self.objective_value_ = None
+        if has_sol:
+            best_sol = self.scip_model.getBestSol()
+            for cpm_var in self.user_vars:
+                if cpm_var not in self._varmap:
+                    continue  # variable never posted (e.g. unused); skip to avoid addVar after solve
```
hbierlee: this is dangerous as then that user variable is unassigned. Since we add all model variables, this should be an assert

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                scip_var = self._varmap[cpm_var]
```
hbierlee: why not via self.solver_var() ?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                if best_sol is not None:
+                    solver_val = self.scip_model.getSolVal(best_sol, scip_var)
+                else:
+                    solver_val = self.scip_model.getVal(scip_var)
+                if cpm_var.is_bool():
+                    cpm_var._value = solver_val >= 0.5
+                else:
+                    cpm_var._value = int(solver_val)
```
hbierlee: this one should also be round  like said in #884

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        # From SCIP's Mark Turner:
+        # SCIP transforms the problem that you are actual solving into something it believes
+        # is easier to solve (presolve techniques etc). All solutions when found are then 
+        # transformed back to the original space, because that is obviously what the user 
+        # modelled and can interpret. Allowing the user to change the model during solving 
+        # gets messy when balancing these things however, so it's forbidden. 
+        # What self.scip_model.freeTransform() does is to remove all information from the 
+        # transformed space, and just leave the original model. All solutions are kept, 
+        # and the user can now change the model as they will. The downside is that potentially 
+        # useful information for speeding up the next optimisation call is thrown out.
```
hbierlee, line 189: I think a) this comment should be just summarized, also b) if freeTransform is necessary before you incrementally solve anyway (unless you re-solve the exact same problem) then this downside seems unavoidable? Is there any useful way to incrementally solve without calling freeTransform in-between calls?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        if minimize:
+            self.scip_model.setObjective(scip_obj, sense='minimize')
+        else:
+            self.scip_model.setObjective(scip_obj, sense='maximize')
+
+    def has_objective(self):
+        try:
+            obj = self.scip_model.getObjective()
+            return obj is not None and (hasattr(obj, 'terms') and len(obj.terms) != 0)
+        except (AttributeError, TypeError):
```
hbierlee: unnecessary caught exceptions?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        scip_obj = self._make_numexpr(obj)
+        if minimize:
+            self.scip_model.setObjective(scip_obj, sense='minimize')
+        else:
+            self.scip_model.setObjective(scip_obj, sense='maximize')
+
+    def has_objective(self):
+        try:
+            obj = self.scip_model.getObjective()
+            return obj is not None and (hasattr(obj, 'terms') and len(obj.terms) != 0)
```
hbierlee: could be `obj is not None and getattr(obj, 'terms', False)`

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        except (AttributeError, TypeError):
+            return False
+
+    def _make_numexpr(self, cpm_expr):
+        """
+            Turns a numeric CPMpy 'flat' expression into a solver-specific
+            numeric expression
+
+            Used especially to post an expression as objective function
```
hbierlee: it's used equally when adding constraints ; superfluous doc imo

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        """
+        import pyscipopt as scip
+
+        if is_num(cpm_expr):
+            return cpm_expr
+
+        # negated bool (e.g. in objective): 1 - bv
+        if isinstance(cpm_expr, NegBoolView):
+            return 1 - self.solver_var(cpm_expr._bv)
```
hbierlee, line 272: needed? in gurobi, we remove all NegBoolViews

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+            a,b = self.solver_vars(cpm_expr.args)
+            return a - b
+        # wsum
+        if cpm_expr.name == "wsum":
+            return scip.quicksum(w * self.solver_var(var) for w, var in zip(*cpm_expr.args))
+
+        # GlobalFunction: abs, mul (PySCIPOpt supports these in constraints/objective)
+        if isinstance(cpm_expr, GlobalFunction):
+            if cpm_expr.name == "abs":
+                return abs(self._make_numexpr(cpm_expr.args[0]))
```
hbierlee, line 291: but they will still be needlessly flattened right? so this might be another usecase for #912

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        cpm_cons = only_numexpr_equality(cpm_cons, supported=frozenset(["sum", "wsum", "sub"]), csemap=self._csemap)
+        cpm_cons = linearize_reified_variables(cpm_cons, min_values=2, csemap=self._csemap)
+        cpm_cons = only_bv_reifies(cpm_cons, csemap=self._csemap)
+        cpm_cons = only_implies(cpm_cons, csemap=self._csemap)
+        cpm_cons = linearize_constraint(cpm_cons, supported=frozenset({"sum", "wsum", "sub", "->", "sum!=", "wsum!="}) | self.supported_global_constraints, csemap=self._csemap)
+        cpm_cons = only_positive_bv(cpm_cons, csemap=self._csemap)
+        return cpm_cons
+
+    def add(self, cpm_expr_orig):
+        """
```
hbierlee: I think this entire doc is duplicated so can be inherited

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        :param cpm_expr_orig: CPMpy expression, or list thereof
+        :type cpm_expr_orig: Expression or list of Expression
+
+        :return: self
+        """
+        get_variables(cpm_expr_orig, collect=self.user_vars)
+        # Ensure every user var has a solver variable (so we get values after solve even if
+        # the constraint was simplified away and the var never appears in transformed constraints)
+        for cpm_var in self.user_vars:
+            if not is_num(cpm_var):
```
hbierlee: is_num already checked in solver_var

I think just one-liner suffices with : self.solver_Vars(self.user_vars)

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+        for cpm_expr in self.transform(cpm_expr_orig):
+            if isinstance(cpm_expr, Comparison):
+                lhs, rhs = cpm_expr.args
+                lhs_is_operator = isinstance(lhs, Operator)
+                sciprhs = self.solver_var(rhs)
+
+                if cpm_expr.name == '!=':
+                    # sum(x) != k  =>  (lhs <= k-1) or (lhs >= k+1); use SCIP's native disjunction (no extra binary)
+                    sciplhs = self._make_numexpr(lhs)
+                    self.scip_model.addConsDisjunction([
```
hbierlee, line 341: this tells me we should leave in just.. disjunctions?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                        sciplhs <= sciprhs - 1,
+                        sciplhs >= sciprhs + 1
+                    ])
+                    continue
```
hbierlee: just using if .. elif .. else would avoid continue's I'd say

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+
+                if cpm_expr.name == '<=':
+                    if (lhs_is_operator and lhs.name == "sum" and all(a.is_bool() and not isinstance(a, NegBoolView) for a in lhs.args)):
```
hbierlee, line 346: we should consider adding SOS constraints for gurobi as well! Just making this comment to be converted to issue if agreed

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                    raise NotImplementedError(
+                        "Not a known supported scip comparison '{}' {}".format(cpm_expr.name, cpm_expr))
+
+            elif isinstance(cpm_expr, Operator) and cpm_expr.name == "->":
+                cond, sub_expr = cpm_expr.args
+                assert isinstance(cond, _BoolVarImpl), f"Implication constraint {cpm_expr} must have BoolVar as lhs"
+                assert isinstance(sub_expr, Comparison), "Implication must have linear constraints on right hand side"
+
+                lhs, rhs = sub_expr.args
+                lhs_is_globalfunc = isinstance(lhs, GlobalFunction)
```
hbierlee, line 370: these variables can be avoided at no cost I think

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                lhs_is_operator = isinstance(lhs, Operator)
+                if lhs_is_globalfunc and lhs.name == "abs":
+                    (arg,) = lhs.args
```
hbierlee, line 373: the neccessity here took some time to understand, maybe add a comment. So for reified abs, we make an ad-hoc transformation here?

hbierlee: the others do work reified?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                assert is_num(rhs), f"linearize should only leave constants on rhs of comparison but got {rhs}"
+
+                if sub_expr.name == ">=":
+                    if lhs_is_operator and lhs.name == "sum":
+                        lhs = Operator("wsum", [[-1] * len(lhs.args), lhs.args])
+                    elif lhs_is_operator and lhs.name == "wsum":
+                        lhs = Operator("wsum", [[-w for w in lhs.args[0]], lhs.args[1]])
+                    else:
+                        lhs = Operator("wsum", [[-1], [lhs]])
+                    sub_expr = lhs <= -rhs
```
hbierlee, line 394: why no indicator constraint? Seems supported https://pyscipopt.readthedocs.io/en/latest/tutorials/constypes.html

hbierlee: ok it's used below. I'm a bit lost as to what is exactly posted here?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+
+                if sub_expr.name == "<=":
```
hbierlee, line 396: why is the <= case so different from the >= case?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                                                         binvar=self.solver_var(cond), activeone=True)
+                elif sub_expr.name == "==":
+                    self.add([cond.implies(lhs <= rhs), cond.implies(lhs >= rhs)])
+                else:
+                    raise Exception(f"Unknown linear expression {sub_expr} name")
+
+            elif isinstance(cpm_expr, BoolVal):
+                if cpm_expr.args[0] is False:
+                    bv = self.solver_var(boolvar())
+                    self.scip_model.addCons(bv <= -1)
```
hbierlee, line 413: no e.g. 0 <= -1  possible?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                cpm_expr.callSolver(self, self.scip_model)
+
+            elif isinstance(cpm_expr, GlobalConstraint):
+                if cpm_expr.name == "xor":
+                    # SCIP addConsXor(vars, rhs): rhs is the value the xor must equal (True for top-level)
+                    args = [a for a in cpm_expr.args if not is_false_cst(a)]
+                    if not args:
+                        # only false constants: xor() is False -> no solution
+                        if not hasattr(self, "_scip_infeasible_aux"):
+                            self._scip_infeasible_aux = self.scip_model.addVar(vtype='B', lb=0, ub=0, name='_false')
```
hbierlee, line 425: just can add(BoolVal(False)), avoiding a bit of code here

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                            self.scip_model.addCons(self._scip_infeasible_aux >= 1)
+                        continue
+                    if any(is_true_cst(a) for a in args):
+                        if not hasattr(self, "_scip_true_var"):
+                            self._scip_true_var = self.scip_model.addVar(vtype='B', lb=1, ub=1, name='true')
```
hbierlee, line 430: _true

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                if cpm_expr.name == "xor":
+                    # SCIP addConsXor(vars, rhs): rhs is the value the xor must equal (True for top-level)
+                    args = [a for a in cpm_expr.args if not is_false_cst(a)]
+                    if not args:
+                        # only false constants: xor() is False -> no solution
+                        if not hasattr(self, "_scip_infeasible_aux"):
+                            self._scip_infeasible_aux = self.scip_model.addVar(vtype='B', lb=0, ub=0, name='_false')
+                            self.scip_model.addCons(self._scip_infeasible_aux >= 1)
+                        continue
+                    if any(is_true_cst(a) for a in args):
```
hbierlee, line 428: needless if check? Can't we just have line 471 to replace constants and variables by scip constants and variables?

hbierlee: this whole block: `[self._scip_true_var if is_true_cst(a) else self.solver_var(a) for a in args if not is_false_cst(a)]` seems more readable and faster

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,514 @@
...
+                - any other keyword argument
+
+            Returns: number of solutions found
+        """
+
+        warnings.warn("Solution enumeration is not implemented in PyScipOPT, defaulting to CPMpy's naive implementation")
+        """
+            Issues to track for future reference:
+                https://github.com/scipopt/PySCIPOpt/issues/549 and
+                https://github.com/scipopt/PySCIPOpt/issues/248
```
hbierlee, line 470: both these issues are now closed so perhaps we can do it natively now?

hbierlee (changes requested):
@tias I've reviewed the PR, I have many small comments and a few larger ones. Is the idea now that I'll resolve them and nominate the PR for review?

cpmpy/solvers/scip.py
tias: I would expect it means that it stays 10 seconds for subsequent calls, just doc'ing it is OK for now I think

cpmpy/solvers/scip.py
tias: idd

cpmpy/solvers/scip.py
tias: it means CPMpy has not implemented it; a PR choses its own scope so it can remain unimplemented

cpmpy/solvers/scip.py
tias: This is a quote from SCIP/Mark directly, to avoid misinterpreting it is can be good to keep it as is.

we do it because we want to support incrementality; if you don't need incrementality you don't need it...
its to document the motivation; e.g. somebody at some point might consider adding a flag to see if solve has already been called before and only then do this. documenting is future-proofing

cpmpy/solvers/scip.py
tias: it seems to ensure that lhs is a wsum (the one at 430 is needlessly recreating a wsum)

cpmpy/solvers/scip.py
tias: silly agents...

cpmpy/solvers/scip.py
tias: No need to extend the PR...

tias:
> @tias I've reviewed the PR, I have many small comments and a few larger ones. Is the idea now that I'll resolve them and nominate the PR for review?

Normally the one that created the PR would be the one fixing the comments.

In this case, the PR was written partly by Ignace, me and more recently an agent. Hence possible inconsistencies and overcomplications.

So, if you would be willing to go over it and clean it up (I commented on the more open questions), then that would be much appreciated!

hbierlee:
yep perfect that's what I had in mind.

cpmpy/solvers/scip.py
hbierlee: this is a bit subtle, note to self, I'll add a solver interface test for this.

cpmpy/solvers/scip.py
hbierlee: doesn't seem to be possible

cpmpy/solvers/scip.py
hbierlee: this can just be updated in #912

cpmpy/solvers/scip.py
hbierlee: nop

cpmpy/solvers/scip.py
hbierlee: yup, I'll add a new issue though:

With https://github.com/scipopt/PySCIPOpt/issues/549 and https://github.com/scipopt/PySCIPOpt/issues/248 now resolved, we can enumerate solutions for SCIP natively using `getSols` rather than with the (presumably) slower CPMpy generic implementation.

cpmpy/solvers/scip.py
hbierlee: I believe this would require partial flattening e.g. what #912 is trying to accomplish

hbierlee:
SCIP PR ready for review with some notable update:

- Came across some additional sensible cases for solver_interface tests ; these are on a separate branch
- simplified XOR handling, which is currently untested in CI, but will be tested by Xor simplify pr. I confirmed it worked manually.

Most notably, I ran pytest-cov to get a coverage on what lines of codes are actually touched by the CI. It's very instructive. If a line of code is not run by CI, either the test for it is missing, or more likely, it is dead code. This left to the following simplifications:

- removed: SCIP supports `abs` at root, but not reified, so there was code to ad-hoc transform it. This code never fired, since reified abs is rewritten (in flatten I think).
- removed: `sub` support ; never fired because linearize always rewrites this (i.e. it does not check supported for it). I didn't want to add support for it in this PR, because we should think about whether it is even useful for linears?
- removed: `!=` support: there was code to to ad-hoc transform this (using the native `or` handling of SCIP, saving a binary variable), but similarly as `sub`, linearize does not check `supported` for it. Then I added it as supported but this meant we now have to support `b -> ... != .. `  as well (and I cannot indicate, support `!=` only when non-reified, at least, that will get very messy). Handling reified `!=` is a little complicated to make work with that ad-hoc transform. The solution IMO is to add proper support for `or` if SCIP supports it (and thus NOT transform `!=` but rewrite it into `(.. > ..) | (.. < ..)`). This all felt very out of scope so I think for the SCIP interface, we should just do `!=` similar to other ILP solvers.

Recommend checking out the reports below for an older code cov  (htmlcov-bak/*scip*.html) report (from running `pytest --solver=scip`) which revealed this issues, and the new one which shows only "unimportant" code is not covered (although missing coverage for both `DirectConstraint` and `FEASIBLE` status when a proper subset of all solutions have been enumearted (e.g. within timeout) may still be worth a look).

[coverages.zip](https://github.com/user-attachments/files/26940084/coverages.zip)

cpmpy/solvers/scip.py
hbierlee: interestingly because the SCIP docs actually said only `<=` was allowed in indicators! Yet this is actually not true. I reported this incorrect doc https://github.com/scipopt/PySCIPOpt/issues/1207

cpmpy/solvers/scip.py
hbierlee: actually, no global is left in reified form, so this code turned out to be dead as revealed by code coverage

cpmpy/solvers/scip.py
hbierlee: documented now

cpmpy/solvers/scip.py
hbierlee: To be handled in https://github.com/CPMpy/cpmpy/issues/932

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,387 @@
...
+        # What self.scip_model.freeTransform() does is to remove all information from the 
+        # transformed space, and just leave the original model. All solutions are kept, 
+        # and the user can now change the model as they will. The downside is that potentially 
+        # useful information for speeding up the next optimisation call is thrown out.
+        self.scip_model.freeTransform()
+
+        return has_sol
+
+
+    def solver_var(self, cpm_var):
```
tias, line 190: do we do this function the same elsewhere?

its not written with 'fast path' in mind... an alterantive would be
```
solver_var = self._varmap.get(cpm_var, None)

if solver_var = None:
  if is_num...
  elif is _BoolVarImpl:
    if NegBool: raise
    else:
       solver_var = ...
  elif _IntVarImple:
  self._varmap[cpm_var] = solver_var

return solver_var
```

if other solvers do similar like we do here, we should fix this in a separate PR

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,387 @@
...
+
+        scip_obj = self._make_numexpr(obj)
+        if minimize:
+            self.scip_model.setObjective(scip_obj, sense='minimize')
+        else:
+            self.scip_model.setObjective(scip_obj, sense='maximize')
+
+    def has_objective(self):
+        obj = self.scip_model.getObjective()
+        return obj is not None and getattr(obj, 'terms', False)  # obj could be `Expr({})`
```
tias: I don't get this one; if it is Expr, is it not valid?

ah, or specifically if it is an Expr of empty list? Can you tighten your check?

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def _make_numexpr(self, cpm_expr):
+        """
+            Turns a numeric CPMpy 'flat' expression into a solver-specific numeric expression
+        """
+        import pyscipopt as scip
+
+        if is_num(cpm_expr):
+            return cpm_expr
+        elif isinstance(cpm_expr, NegBoolView):  # negated bool (e.g. in objective): 1 - bv
```
tias: check can go, it would match the next case, where it would be caught in solver_var

cpmpy/solvers/scip.py
```diff
@@ -0,0 +1,387 @@
...
+                raise Exception(f"Unknown linear expression {sub_expr} name")
+
+        elif isinstance(cpm_expr, BoolVal):
+            if cpm_expr.args[0] is False:
+                self.scip_model.addConsXor([], True)  # easiest way to post False to SCIP (e.g. 0 <= -1 is not allowed, bv <= -1 requires adding a dummy variables, ...)
+
+        elif isinstance(cpm_expr, DirectConstraint):
+            cpm_expr.callSolver(self, self.scip_model)
+
+        elif isinstance(cpm_expr, GlobalConstraint):
```
tias, line 348: I think it is a tiny-tiny bit neater for this one to appear before the isisnstance BoolVal

tias (commented):
very nice and clean, made some minor tweaks and minor comments

(also a less minor comment that is just triggered by the review, need not be for the PR)

cpmpy/solvers/scip.py
hbierlee: It does seem indeed more or less copy-paste from gurobi, but it's not the fast way. I agree it's pretty sub-optimal, coverage revealed the is_num is never run, similarly NegBoolView should never run as long as the transforms are correct.

https://github.com/CPMpy/cpmpy/blob/626bfa80c6507bf2ada8aa5d665766ceb031af75/cpmpy/solvers/gurobi.py#L252-L276

I'll convert this comment to issue

hbierlee:
@tias fixed comments (and I can't request your review since you started the PR)

tias (commented):
I approve very much, but can't formally approve because I started the PR indeed : )

But its your earlier review that is blocking it, so please approve if you approve and then merge ; )

hbierlee (approved):

PR #413, opened 2023-09-18
Adding choco

Dimosts:
Adding support to Choco solver.

Also added some tests.

Compared to ortools, choco does not need transformation only_bv_implies, because it supports complete reification. However, I had to split the part where the bool var is sent to the left part and the boolexpr to the right part from only_bv_implies, creating a new transformation, i.e., only_bv_reifies, keeping in only_bv_implies only the part that the reification is split into 2 implications. We can also put the only_bv_reifies part inside the reify_rewrite function, which would make sense to me.

Choco also it supports reification in all constraints as after creating a constraint, this can be either posted or be reified. Thus all bool globals are included in reify_rewrite. 

That is also why I renamed the function _post_constraint() to _get_constraint(), as I use it also to get constraints and reify them. The actual post function is called in __add__.

Also, note that Choco has many restrictions on the parameters each constraint accepts. Some constraints accept only variables in some arguments, so if we have an integer we must create a variable that will be equal to that integet in order to use this constraint.
These checks and creation of variables is currently in _get_constraint(). Should it be in a different transformation? Not sure, it is not generic enough for this.

In addition to these restrictions, choco needs all num expressions to be in canonical form (variables left side, constant right side), thus a new transformation is introduced to deal with it, using a part from linearize. Also, changed linearize to called
this transformation inside instead of reusing the same code.

Also, currently the "implies" constraints and the reification are not fully supported in the python API, so, until they are added in the python API which will happen soon, they are added with different ways to the solver.

Finally, the python API for Choco currently has a bug on the time limit (which will be soon resolved) and thus it is not supported currently.


cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,582 @@
...
+                return self.chc_model.member(expr, table)
+            elif cpm_expr.name == "cumulative":
+                start, dur, end, demand, cap = self.solver_vars(cpm_expr.args)
+                # Everything given to cumulative in Choco needs to be a variable.
+                # Convert demands to variables
+                if is_num(demand):  # Create list for demand per task
+                    demand = [demand] * len(start)
+                if isinstance(demand, _NumVarImpl):
+                    demand = self.solver_vars(demand)
+                else:
```
IgnaceBleukx, line 531: Demand in the cumulative can be either a list of demands for each task or be static and equal for all of the tasks.
Do you cover the list of NumVars case here?

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,582 @@
...
+                    demand = [self.chc_model.intvar(d, d) for d in demand]  # Create variables for demand
+                # Create task variables. Choco can create them only one by one
+                tasks = [self.chc_model.task(s, d, e) for s, d, e in zip(start, dur, end)]
+                # Convert capacity to variable
+                # Choco needs result to be a variable
+                if isinstance(cap, int):
+                    capacity = self.chc_model.intvar(cap, cap)  # convert to "variable"
```
IgnaceBleukx, line 538: Could be usefull to make a helper function for making variables from ints? Such as `def to_var(self, int)`

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,582 @@
...
+                    f"Unknown global constraint {cpm_expr}, should be decomposed! If you reach this, please report on github.")
+
+        # unlikely base case: Boolean variable
+        elif isinstance(cpm_expr, _BoolVarImpl):
+            return self.chc_model.and_([self.solver_var(cpm_expr)])
+
+        # unlikely base case: True or False
+        elif isinstance(cpm_expr, BoolVal):
+            # Choco does not allow to post True or False. Post "certainly True or False" constraints instead
+            if cpm_expr:
```
IgnaceBleukx, line 571: `if cpm_expr.args[0] is True`

tests/test_solveAll.py
```diff
@@ -28,8 +28,6 @@ def test_solveall_no_obj(self):
             else:
                 count = solver.solveAll(solution_limit=1000, display=add_sol)
             self.assertEqual(3, count)
-            self.assertSetEqual(sols,
```
IgnaceBleukx, line 31: No no no, this check should remain! Ensur the `value` of Boolean variables is converted to True/False instead of `int`

tests/test_transf_reif.py
```diff
@@ -12,26 +12,6 @@ def setUp(self):
         _IntVarImpl.counter = 0
         _BoolVarImpl.counter = 0
 
-    def test_only_bv_implies(self):
```
IgnaceBleukx, line 15: These tests should remain in some form or another, but probably this will get split into another pull request anyway

IgnaceBleukx (commented):
Very nice addition! Tests also seem to run very fast (faster than OR-tools?)!
I have many style comments and some "soundness" comments but nothing major.
The transformation changes should live in another pull request tho, this one should only contain the new Choco interface

Dimosts:
Opened seperate PR for the new transformations

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,571 @@
...
+            raise Exception("Install the python 'pychoco' package to use this solver interface")
+
+        import pychoco as chc
+
+        assert (subsolver is None)
+
+        # initialise the native solver objects
+        self.chc_model = chc.Model()
+        self.chc_solver = chc.Model().get_solver()
+        self.helper_var = self.chc_model.intvar(0, 2)
```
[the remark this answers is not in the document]
Dimosts: have changed the bounds of the helper var to 0,0 to not alter no of solutions.

Using it for True or False cases, to put surely True or False constraints in the solver. Maybe the surelt True constraint is not needed, but it is needed for the False case.

cpmpy/solvers/choco.py
Dimosts: nope, will include

cpmpy/solvers/choco.py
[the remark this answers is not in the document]
IgnaceBleukx: I see, had to do something similar in the scip interface...
But maybe its overkill to create the variable upfront? The usecase of adding `True` or `False` is extremely rare so we can also just create the variable when the case comes up

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,582 @@
...
+
+        :return: list of Expression
+        """
+
+        cpm_cons = toplevel_list(cpm_expr)
+        supported = {"min", "max", "abs", "count", "element", "alldifferent", "alldifferent_except0", "allequal",
+                     "table", "InDomain", "cumulative", "circuit", "gcc", "inverse"}
+        supported_reified = {"alldifferent", "alldifferent_except0", "allequal",
+                     "table", "InDomain", "cumulative", "circuit", "gcc", "inverse"}
+        cpm_cons = decompose_in_tree(cpm_cons, supported, supported_reified)
```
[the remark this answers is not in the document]
Dimosts: the 2 sets are not equal. In the supported_reified I dont have the globalfunctions, as they need to be in equality

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,582 @@
...
+
+        # sum or weighted sum
+        if isinstance(lhs, Operator):
+            if lhs.name == 'sum':
+                return self.chc_model.sum(self.solver_vars(lhs.args), op, self.solver_var(rhs))
+            elif lhs.name == "sub":
+                a, b = self.solver_vars(lhs.args)
+                return self.chc_model.arithm(a, "-", b, op, self.solver_var(rhs))
+            elif lhs.name == 'wsum':
+                w = [int(wght) for wght in lhs.args[0]]
```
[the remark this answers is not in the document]
Dimosts: If it is already a numpy array yes. If it is a list, does it worth to convert it to np array and then back to a list that way?

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,582 @@
...
+        cpm_cons = toplevel_list(cpm_expr)
+        supported = {"min", "max", "abs", "count", "element", "alldifferent", "alldifferent_except0", "allequal",
+                     "table", "InDomain", "cumulative", "circuit", "gcc", "inverse"}
+        supported_reified = {"alldifferent", "alldifferent_except0", "allequal",
+                     "table", "InDomain", "cumulative", "circuit", "gcc", "inverse"}
+        cpm_cons = decompose_in_tree(cpm_cons, supported, supported_reified)
+        cpm_cons = canonical_comparison(cpm_cons)
+        cpm_cons = flatten_constraint(cpm_cons)  # flat normal form
+        cpm_cons = only_numexpr_equality(cpm_cons, supported=frozenset(["sum", "wsum", "sub"]))  # support >, <, !=
+        cpm_cons = reify_rewrite(cpm_cons, supported=frozenset(["sum", "wsum", "alldifferent", "alldifferent_except0", "allequal",
```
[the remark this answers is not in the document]
Dimosts: used the supported reified test and made it cleaner indeed

cpmpy/solvers/choco.py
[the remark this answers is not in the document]
IgnaceBleukx: Somehow it is still faster if you have a list of Python ints to first convert it to numpy and then go back to Python using the `.tolist()`... Don't ask me how that is possible

Extra method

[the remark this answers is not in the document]
Dimosts: Intereting.. Changed it to this then

Dimosts:
All comments resolved or answered, added also the new functionalities with the new release of pychoco.

I think it is ready for a new review

Wout4:
I get an error when using an integer + var as the index of an element constraint.
This results in 1 - iv[0] == aux being sent to choco, but this is not in a format that choco accepts.

```
iv = intvar(-8,8, shape=5)
cons = [iv[1-iv[0]] == 2]
Model(cons).solve('choco')
```

Wout4:
Don't know if that is a canonical comparison bug or it has to be fixed somewhere else, but I found another bug that in canonical comparison as well: #418

Dimosts:
> I get an error when using an integer + var as the index of an element constraint. This results in 1 - iv[0] == aux being sent to choco, but this is not in a format that choco accepts.
> 
> ```
> iv = intvar(-8,8, shape=5)
> cons = [iv[1-iv[0]] == 2]
> Model(cons).solve('choco')
> ```

I see, it is due to the fact that canonical_comparison was moved before flatten for choco, but as it is not recursive it should be after flatten.

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,585 @@
...
+        # Comparisons: both numeric and boolean ones
+        # numexpr `comp` bvar|const
+        # Choco accepts only int32, not int64
+        elif isinstance(cpm_expr, Comparison):
+            for i in range(len(cpm_expr.args)):
+                if isinstance(cpm_expr.args[i], np.integer):
+                    cpm_expr.args[i] = int(cpm_expr.args[i])
+            lhs, rhs = cpm_expr.args
+
+            if lhs.is_bool() and rhs.is_bool(): #boolean equality -- Reification
```
Wout4, line 411: I noticed that canonical comparison can output things that would be normalised away in simplify_bool such as `bv > 0`, but since it is now again after flatten they can appear here.
So we need to either use is_boolexpr(lhs) and is_boolexpr(rhs) here (which will treat bv > 0 as a numexpr, but that seems to work). Or change canonical comparison to output bv in stead of bv > 0 (might be harder i guess)

Wout4: i.e _canonical_comparison(sum(bv) > 0)_ outputs` bv > 0 `

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,585 @@
...
+                    lhs.name == 'sum' or lhs.name == 'wsum' or lhs.name == "sub")):
+                # a BoundedLinearExpression LHS, special case, like in objective
+                chc_numexpr = self._make_numexpr(cpm_expr)
+                return chc_numexpr
+            elif cpm_expr.name == '==':
+                chcrhs = self.solver_var(rhs)
+                # NumExpr == IV, supported by Choco (thanks to `only_numexpr_equality()` transformation)
+                if lhs.name == 'min':
+                    if isinstance(rhs, int):     # Choco does not accept an int in rhs
+                        chcrhs = self.to_var(rhs)  # convert to "variable"
```
Wout4, line 427: Choco also does not accept ints in any of the arguments of min, max, abs,.. something like Max(iv,0) also needs the 0 to become a variable.

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,585 @@
...
+                    elif not isinstance(rhs, _NumVarImpl):
+                        raise Exception(f"Choco does not accept {rhs} with type {type(rhs)} as rhs of expression {lhs.name}")
+                    return self.chc_model.absolute(chcrhs, self.solver_var(lhs.args[0]))
+                elif lhs.name == 'count':
+                    if isinstance(rhs, int):  # Choco does not accept an int in rhs
+                        chcrhs = self.to_var(rhs)  # convert to "variable"
+                    elif not isinstance(rhs, _NumVarImpl):
+                        raise Exception(
+                            f"Choco does not accept {rhs} with type {type(rhs)} as rhs of expression {lhs.name}")
+                    arr, val = self.solver_vars(lhs)
```
Wout4, line 449: This does not seem to work? lhs is not a var here but a Count

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,585 @@
...
+                    return self.chc_model.count(val, arr, chcrhs)
+                elif lhs.name == 'mul':
+                    return self.chc_model.times(self.solver_vars(lhs.args[0]), self.solver_vars(lhs.args[1]),
+                                                chcrhs)
+                elif lhs.name == 'div':
+                    # Choco needs divisor to be a variable
```
Wout4, line 455: what about lhs.args[0]? 2/iv also gives an error for me

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,585 @@
...
+                elif lhs.name == 'mod':
+                    divisor = lhs.args[1]
+                    if not isinstance(rhs, _NumVarImpl): # if divisor is variable then result must be variable too
+                        rhs = self.to_var(rhs)  # convert to "variable"
+                    else:
+                        rhs = self.solver_vars(rhs)  # get choco variable
+                    return self.chc_model.mod(self.solver_var(lhs.args[0]), self.solver_var(divisor), rhs)
+                elif lhs.name == 'pow':
+                    if isinstance(rhs, int):     # Choco does not accept an int in rhs
+                        chcrhs = self.to_var(rhs)  # convert to "variable"
```
Wout4, line 488: same as for div, 2 mod IV also gives an error, so all arguments need to be converted to 'variable'

Wout4 (changes requested):
We need to deal with some more cases where integer constants are in the constraints

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,585 @@
...
+
+        return len(sols)
+
+    def solver_var(self, cpm_var):
+        """
+            Creates solver variable for cpmpy variable
+            or returns from cache if previously created
+        """
+        if is_num(cpm_var):  # shortcut, eases posting constraints
+            return cpm_var
```
Wout4, line 221: seems like choco crashes on ints outside of int32 bounds (just like minizinc) so we should add a check for that here and throw an error. 
```
xx, yy = intvar(0,5041000,2)
Model(xx * yy == 504100*504100).solve("choco")
```
but this does not crash:
```
xx, yy = intvar(0,5041000,2)
Model(xx * yy == 504100).solve("choco")
```
but this does crash:
```
xx, yy = intvar(0,504100*504100,2)
Model(xx * yy == 504100).solve("choco")
```

cpmpy/solvers/choco.py
Wout4: Yes, but simplify bool does not simplify away sum(bv) >0, and then canonical comparison outputs bv > 0 which gives an error when using .is_bool.
I guess you could argue that's a bug in simplify_bool actually but leaving .is_bool would require you to be very sure to cover all cases

Dimosts: I see

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,664 @@
...
+                    return self.chc_model.absolute(chcrhs, self.solver_var(lhs.args[0]))
+                elif lhs.name == 'count':
+                    if isinstance(rhs, int):  # Choco does not accept an int in rhs
+                        chcrhs = self.to_var(rhs)  # convert to "variable"
+                    elif not isinstance(rhs, _NumVarImpl):
+                        raise Exception(
+                            f"Choco does not accept {rhs} with type {type(rhs)} as rhs of expression {lhs.name}")
+                    arr, val = lhs.args
+                    chcarr = []
+                    for v in arr:  # Choco accepts only variables in lhs
```
Wout4, line 478: this is a lot of duplicate code, for min max abs element count div mod. Maybe it can just happen upfront, since it seems its always needed? or make a helper function

Dimosts: Yes, was thinking the same. And not only for these, but in almost all constraints. Was thinking to add the whole process to to_var() helper function.

Dimosts:
Requested changed resolved. Duplicare code went to to_var and to_vars + added checks for int32 + specific ChocoTypeException

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,617 @@
...
+                    raise ChocoTypeException(f"Choco cannot accept alldifferent_except0 with: {vars}")
+                return self.chc_model.all_different_except_0(vars)
+            elif cpm_expr.name == 'allequal':
+                vars = self.to_vars(cpm_expr.args)
+                if any(c is None for c in vars):
+                    raise ChocoTypeException(f"Choco cannot accept allequal with: {vars}")
+                return self.chc_model.all_equal(vars)
+            elif cpm_expr.name == 'table':
+                assert (len(cpm_expr.args) == 2)  # args = [array, table]
+                array, table = self.solver_vars(cpm_expr.args)
```
Wout4, line 565: This is also gonne have to become to_vars in stead of solver_vars for the array component. Not sure about the table component though..

Wout4: maybe not logical but we allow something like Table([var,5],[(2,5),(3,6)])

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,617 @@
...
+                wgt, x = lhs.args
+                w = np.array(wgt).tolist()
+                x = self.solver_vars(lhs.args[1])
+                return self.chc_model.scalar(x, w, op, self.solver_var(rhs))
+
+        raise NotImplementedError("Choco: Not a known supported numexpr {}".format(cpm_expr))
+
+    def to_var(self, val):
+        from pychoco.variables.intvar import IntVar
+        if isinstance(val, int):
```
Wout4, line 311: us our expressions.utils is_int(), or you will miss np integers and bools here

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,617 @@
...
+                elif isinstance(display, list):
+                    print([v.value() for v in display])
+                else:
+                    display()  # callback
+
+        return len(sols)
+
+    def solver_var(self, cpm_var):
+        """
+            Creates solver variable for cpmpy variable
```
Wout4, line 218: So this one allows integers to not be converted to a choco var, but i think there is still a problem with np.integers not being supported in choco, so I guess those have to be converted here to a python int

Wout4 (commented):
minor bugs remaining :)

cpmpy/solvers/choco.py
Dimosts: tbh I am not sure we should. first args should be vars always, second one the allowed tuples

Dimosts: Choco table constraint only accepts list of list of ints as second argument, so we don't want to use to_var as it will make it a list of list of vars

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,617 @@
...
+
+        raise NotImplementedError("Choco: Not a known supported numexpr {}".format(cpm_expr))
+
+    def to_var(self, val):
+        from pychoco.variables.intvar import IntVar
+        if is_int(val):
+            if val < -2147483646 or val > 2147483646:
+                raise ChocoBoundsException(
+                    "Choco does not accept integer literals with bounds outside of range (-2147483646..2147483646)")
+            var = self.chc_model.intvar(val, val)  # convert to "variable"
```
Wout4, line 315: also here cast to int()

cpmpy/solvers/choco.py
Wout4: I think you are right, will make a pr to put a check in the constructor of table

IgnaceBleukx (commented):
It looks good but quite a lot of cleanup needs to happen still.
As we now have fixed many issues such as the recursiveness of `canonical_comparison` this pull request should become a lot easier.
Can you also make a map of features->pychoco Github issues so we can keep track when they add the functionality we want?

cpmpy/solvers/choco.py
Wout4: pr (https://github.com/CPMpy/cpmpy/pull/428) ;)

cpmpy/solvers/choco.py
```diff
@@ -0,0 +1,617 @@
...
+        """
+
+        cpm_cons = toplevel_list(cpm_expr)
+        supported = {"min", "max", "abs", "count", "element", "alldifferent", "alldifferent_except0", "allequal",
+                     "table", "InDomain", "cumulative", "circuit", "gcc", "inverse"}
+        supported_reified = {"alldifferent", "alldifferent_except0", "allequal",
+                     "table", "InDomain", "cumulative", "circuit", "gcc", "inverse"}
+        cpm_cons = decompose_in_tree(cpm_cons, supported, supported_reified)
+        cpm_cons = flatten_constraint(cpm_cons)  # flat normal form
+        cpm_cons = canonical_comparison(cpm_cons)
```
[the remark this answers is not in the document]
IgnaceBleukx: It's not merged apparently...

Wout4:
All tests are succeeding now, also added pychoco to the github tests.
So choco is ready to become a tier 2 solver ;)

PR #414, opened 2023-09-19
Canonical_comparison and only_bv_reifies

Dimosts:
2 new transformations, based on work done while adding choco:

canonical_comparison: all num expressions converted in canonical form (variables left side, constant right side), This was done inside linearize before, but as choco also needs it, it is now a seperate transformation, which is also called inside linearize.

only_bv_reifies:  the bool var is sent to the left part and the boolexpr to the right part in reifications. This was done in only_bv_implies, which also converted complete reifications to 2 implications. I splitted the 2 parts, as the first one is needed in Choco, while the second one noe (it supports complete reification). First thought was to put it reify_rewrite(), but pysat doesn't need the additional stuff happening there, while it needs only_bv_implies.

Also, as now only_bv_implies is only assuring that there are no "==" and transforms them to implications, it is renamed to only_implies.

### DIFF
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -33,7 +33,7 @@ from ..transformations.flatten_model import flatten_constraint, flatten_objectiv
 from ..transformations.get_variables import get_variables
 from ..transformations.decompose_global import decompose_in_tree
 from ..transformations.linearize import linearize_constraint, only_positive_bv
-from ..transformations.reification import only_bv_implies, reify_rewrite
+from ..transformations.reification import only_implies, reify_rewrite, only_bv_reifies
 from ..transformations.normalize import toplevel_list
 from ..expressions.globalconstraints import DirectConstraint
 from ..exceptions import NotSupportedError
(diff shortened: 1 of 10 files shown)


Wout4:
I am extracting some of the optimizations from flatten into a normalize transformation (they avoid creating aux variables). Ignace pointed out that something that is now missing there is that comparisons like a+b == c + d create an aux var in stead of just throwing all the variables to one side. Maybe the Canonical representation thing can be way higher in the transformation stack, since it is also useful in simplifying the flattening?

Wout4:
It turns out canonical comparison only brings constants to the rhs, but not variables to the lhs, is that intended?

Dimosts:
It should bring variables to lhs, at least it does in my examples. In which cases it does not do it?

Dimosts:
I saw what you mean, am fixing it

Dimosts:
Should work normally now

Dimosts:
> I am extracting some of the optimizations from flatten into a normalize transformation (they avoid creating aux variables). Ignace pointed out that something that is now missing there is that comparisons like a+b == c + d create an aux var in stead of just throwing all the variables to one side. Maybe the Canonical representation thing can be way higher in the transformation stack, since it is also useful in simplifying the flattening?

With the new changes now, I tried to use it before flattenning in choco and works normally, indeed avoiding some aux variables that are introduced in flattenning in some cases.

Wout4:
I wrote some tests, mostly to better my own understanding of canonical comparison.
Everything seems to work fine for use within linearize, so I'm fine with this being merged already.

We could do more though, and should when using it before flatten.
a + b < b + c does not simplify away the b (flatten now reifies rhs)
(a + b) * 5 < c is not supported (flatten normally makes a wsum out of this)
(a + b) < c // 5 does not turn into a wsum (flatten now reifies this)

Dimosts:
Yes, for using it before flatten it should be extended more, first of all it should be made recursive! I think we can merge it as it is for now, so also choco can be supported, and we see in a next PR if we move it

PR #416, opened 2023-09-22
adding QuickXplain

IgnaceBleukx:
I implemented the original basic version of the QuickXplain algorithm of Junker:
> Junker, Ulrich. "Preferred explanations and relaxations for over-constrained problems." AAAI-2004. 2004.

### DIFF
--- a/cpmpy/tools/mus.py
+++ b/cpmpy/tools/mus.py
@@ -96,13 +96,17 @@ def mus_naive(soft, hard=[], solver="ortools"):
 
 def quickxplain(soft, hard=[], solver="ortools"):
     """
+        Find a preferred minimal unsatisfiable subset of constraints, based on the ordering of constraints.
+
+        A total order is imposed on the constraints using the ordering of `soft`.
+        Constraints with lower index are preferred over ones with higher index
+
+        Assumption-based implementation for solvers that support s.solve(assumptions=...) and s.get_core()
+        More naive version available as `quickxplain_naive` to use with other solvers.
+
         CPMpy implementation of the QuickXplain algorithm by Junker:
             Junker, Ulrich. "Preferred explanations and relaxations for over-constrained problems." AAAI-2004. 2004.
             https://cdn.aaai.org/AAAI/2004/AAAI04-027.pdf
-
-        Find a preferred minimal unsatisfiable subset of constraints
-        A partial order is imposed on the constraints using the order of `soft`.
-        Constraints with lower index are preferred over ones with higher index
     """
 
     soft = toplevel_list(soft, merge_and=False)
@@ -147,17 +151,19 @@ def quickxplain(soft, hard=[], solver="ortools"):
 
 def quickxplain_naive(soft, hard=[], solver="ortools"):
     """
+        Find a preferred minimal unsatisfiable subset of constraints, based on the ordering of constraints.
+
+        A total order is imposed on the constraints using the ordering of `soft`.
+        Constraints with lower index are preferred over ones with higher index
+
+        Naive implementation, re-solving the model from scratch.
+        Can be slower depending on the number of global constraints used and solver support for reified constraints.
+
         CPMpy implementation of the QuickXplain algorithm by Junker:
             Junker, Ulrich. "Preferred explanations and relaxations for over-constrained problems." AAAI-2004. 2004.
             https://cdn.aaai.org/AAAI/2004/AAAI04-027.pdf
-
-        Naive implementation of the quickXplain algorithm without assumptions.
-        Can be faster when using many global constraints and a solver not support reification of globals.
-
-        Find a preferred minimal unsatisfiable subset of constraints
-        A partial order is imposed on the constraints using the order of `soft`.
-        Constraints with lower index are preferred over ones with higher index
     """
+
 
     soft = toplevel_list(soft, merge_and=False)
     assert cp.Model(hard + soft).solve(solver) is False, "The model should be UNSAT!"


examples/advanced/quickxplain.py
```diff
@@ -0,0 +1,83 @@
+"""
+    CPMpy implementation of the QuickXplain algorithm by Junker.
+        Junker, Ulrich. "Preferred explanations and relaxations for over-constrained problems." AAAI-2004. 2004.
+"""
```
sourdough-bread, line 4: Add url of paper because it's hard to find: 
https://cdn.aaai.org/AAAI/2004/AAAI04-027.pdf

examples/advanced/quickxplain.py
```diff
@@ -0,0 +1,83 @@
...
+
+    soft = toplevel_list(soft, merge_and=False)
+    assump = cp.boolvar(shape=len(soft))
+    solver = cp.SolverLookup.get(solver, cp.Model(hard))
+    solver += assump.implies(soft)
+
+    assert solver.solve(assumptions=assump) is False, "The model should be UNSAT!"
+
+    dmap = dict(zip(assump, soft))
+    core = recurse_explain(list(assump),[],[],order, dmap, solver=solver)
```
sourdough-bread, line 32: Useless remark that can be ignored:
Since it does not change anything to the execution, but to be as close as possible to the original algorithm, delta should be = list(assump)
```
core = recurse_explain(list(assump), [], list(assump), order, dmap, solver=solver)
```

examples/advanced/quickxplain.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def quickXplain(soft, hard=[], order=lambda x : 0, solver="ortools"):
+    """
+        Find a preferred minimal unsatisfiable subset of constraints
+        A partial order is imposed on the constraints using the `order` argument
+        Constraints with lower order are preferred to ones with higher order
+
+        :param: soft: soft constraints, list of expressions
+        :param: hard: hard constraints, optional, list of expressions
```
sourdough-bread, line 18: Something about hard = background constraints that cannot be relaxed (B) and soft = conflict (C) (that can be relaxed)

examples/advanced/quickxplain.py
```diff
@@ -0,0 +1,83 @@
...
+        :param: solver: name of a solver, see SolverLookup.solvernames()
+            "z3" and "gurobi" are incremental, "ortools" restarts the solver
+    """
+
+    soft = toplevel_list(soft, merge_and=False)
+    assump = cp.boolvar(shape=len(soft))
+    solver = cp.SolverLookup.get(solver, cp.Model(hard))
+    solver += assump.implies(soft)
+
+    assert solver.solve(assumptions=assump) is False, "The model should be UNSAT!"
```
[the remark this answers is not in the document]
sourdough-bread: here is a naive impelementation of the QuickXplain algo.

```Python
def quickXplain_naive(hard, soft, order=lambda x : 0, solver="ortools"):
    if len(soft) == 0: return []

    soft = toplevel_list(soft, merge_and=False)
    assert cp.Model(hard + soft).solve() is False, "The model should be UNSAT!"

    return recurse_explain_naive(hard, hard, soft, order)

def recurse_explain_naive(hard, delta, soft, order):

    if len(delta) != 0 and cp.Model(hard).solve() is False:
        # conflict is in hard constraints, no need to recurse
        return []

    if len(soft) == 1:
        # conflict is not in hard constraints, but only 1 soft constraint
        return list(soft) # base case of recursion

    soft = sorted(soft, key=lambda a : order(a))
    split = len(soft) // 2 # determine split point
    more_preffered, less_preffered = soft[:split], soft[split:] # split constraints into two sets

    # treat more preferred part as hard and find extra constants from less preferred
    delta2 = recurse_explain_naive(hard + more_preffered, more_preffered, less_preffered, order=order)
    # find which preferred constraints exactly
    delta1 = recurse_explain_naive(hard + delta2, delta2, more_preffered, order=order)
    return delta1 + delta2
```

sourdough-bread (changes requested):
Nice example, small improvements can be added for readibility and being closer to the original algorithm.

examples/advanced/quickxplain.py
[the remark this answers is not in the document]
IgnaceBleukx: I think we should only add this in case the quickXplain algo becomes a "tool" in CPMpy, but not really needed for an example in my opinion.
Should we add it as a tool?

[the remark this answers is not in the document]
sourdough-bread: Can be added to both examples and tools. I think it’s a rather compact example that would fit nicely in the tools.

examples/advanced/quickxplain.py
IgnaceBleukx: Hmm not sure about this one, I like the current version as it gives a clear view on the initialization of the recursion.
But we might add a comment saying why we do not put the hard constraints here (we added them to the solver).

About the order of arguments, it really depends on which paper you read and I would like to keep it close to the interface of the `mus` tool

sourdough-bread: Sounds reasonable to me :-)

examples/advanced/quickxplain.py
```diff
@@ -0,0 +1,82 @@
...
+
+    dmap = dict(zip(assump, soft))
+    core = recurse_explain(list(assump),[],[], solver=solver)
+    return [dmap[a] for a in core]
+
+
+def recurse_explain(soft, hard, delta, solver):
```
sourdough-bread, line 36: Is there a reason why you recurse_explain (resp. do_recursion)  in the tools is embedded and not here ?
Change the function name to be consistent across both files?

tests/test_tools_mus.py
```diff
@@ -2,10 +2,15 @@
 from unittest import TestCase
 
 from cpmpy import *
```
sourdough-bread, line 4: Change `import cpmpy as cp`, especially write cp.all for the “all” method that is sometimes overloaded sometimes not.

tests/test_tools_mus.py
```diff
@@ -48,34 +53,64 @@ def test_bug_191_many_soft(self):
...
+        self.naive_func = quickxplain_naive
+
+    def test_prefered(self):
+
+        a,b,c,d = [boolvar(name=n) for n in "abcd"]
+
+        mus1 = [b,d]
+        mus2 = [a,b,c]
+
+        hard = [~all(mus1), ~all(mus2)]
```
sourdough-bread, line 103: cp.all

cpmpy/tools/mus.py
```diff
@@ -76,18 +77,111 @@ def mus_naive(soft, hard=[], solver="ortools"):
...
+
+
+def quickxplain(soft, hard=[], solver="ortools"):
+    """
+        CPMpy implementation of the QuickXplain algorithm by Junker:
+            Junker, Ulrich. "Preferred explanations and relaxations for over-constrained problems." AAAI-2004. 2004.
+            https://cdn.aaai.org/AAAI/2004/AAAI04-027.pdf
+
+        Find a preferred minimal unsatisfiable subset of constraints
+        A partial order is imposed on the constraints using the order of `soft`.
```
sourdough-bread, line 104: I would make it more explicit that the user needs to order it himself.

sourdough-bread (commented):
Some small improvements mostly for readability. Ready to be merged after changes.

examples/advanced/quickxplain.py
IgnaceBleukx: Not really, I wanted to stay very close to the original pseudocode of the paper in the example, while the tools-version has some subtle optimizations, plus embedding the recursive call ensures users cannot call the helper function themselves by accident.

PR #417, opened 2023-09-26
no simplify bool in minizinc and z3

Wout4:
Turns out we don't need it anymore :)


Dimosts (approved):
It passes all the tests and seems correct.

Issue #420, opened 2023-10-05
Canonical comparison does not handle reification of comparisons


Wout4:
Flatten_constraint() allows nested output in the case of a reified comparison i.e `(IV1 == IV2) == BV`

This case is not handled by canonical comparison. Althought the example above does not give an error when using choco, the following does:

```
iv = intvar(-8,8, shape=5)
bv = boolvar(shape=3)
cons = [(((4 + iv[0]) == iv[1]) == bv[0])]
Model(cons).solve('choco')
```

Dimosts:
Indeed. Created a PR handling that

Dimosts:
hmm, indeed. So we need to cover both rhs and lhs, or move canonical comparison after reify_rewrite. But as the plan is to move it before/in flatten at some point, we should just cover both cases.

Dimosts:
This was solved with 2 PRs already merged

PR #421, opened 2023-10-06
Handle reification of comparison in canonical_comparison

Dimosts:
Solves issue #420.


Wout4:
I guess this makes it more recursive than strictly necessary but it does the job, merging it now (and back to testing choco a bit more ;)

PR #422, opened 2023-10-06
Canonical comparison Implied

Dimosts:
again tackles issue #420 with the implication case revealed.


Wout4:
Looks good, will merge it with the choco branch as well to test some more..

PR #425, opened 2023-10-13
395 xor decomposition is very inefficient

Wout4:
Came up with this one, it avoids nested conjunctions/disjunctions, in favour of nested sums.

Seems to lead to a smaller amount of constraints after flattening.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -387,9 +387,12 @@ class Xor(GlobalConstraint):
         super().__init__("xor", flatargs)
 
     def decompose(self):
-        # there are multiple decompositions possible
-        iv = intvar(0, (len(self.args) // 2))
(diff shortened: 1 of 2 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -387,16 +387,11 @@ def __init__(self, arg_list):
...
-        cons = (a0 | a1) & (~a0 | ~a1)  # one true and one false
-
-        # for more than 2 variables, we cascade (decomposed) xors
-        for arg in self.args[2:]:
-            cons = (cons | arg) & (~cons | ~arg)
-        return [cons], []
+        # there are multiple decompositions possible, Recursively using sum allows it to be efficient for all solvers.
+        decomp = [sum(self.args[:2]) == 1]
+        if len(self.args) > 2:
+            decomp = Xor([decomp,self.args[2:]]).decompose()[0]
```
Dimosts, line 400: Ah, you made it recursive, nice one. It's indeed clear that it will create far less constraints

Dimosts (approved):

PR #428, opened 2023-10-19
test table constraint first argument.

Wout4:


Wout4:
Array argument of Table global constraint can be restricted to only allow vars/expressions, there is no use cases where something else would be needed.

Dimosts (approved):

PR #431, opened 2023-10-27
XCSP3 parser

IgnaceBleukx:
It would be a nice addition to CPMpy to be able to read XCSP files into a CPMpy model.
One of the main reasons is we can load competition models and benchmark our changes to the library on it.


Wout4:
as it stands we can translate 50% of the CSP23 problems, will put the parser on hold a bit for now. Probably better to first implement the new globals in #433 since we will need them for the parser anyway.

PR #432, opened 2023-10-30
Canonical recursive

IgnaceBleukx:
Made canonical comparison fully recursive.
This covers some dangling edge cases I found such as having a sum with only one argument.
Also cleaned up the code a bit : )

### DIFF
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -277,17 +277,28 @@ def canonical_comparison(lst_of_expr):
             lhs, rhs = cpm_expr.args
 
             # put all vars to lhs
-            if not is_num(rhs) and isinstance(lhs, _NumVarImpl) or lhs.name =="sum" or lhs.name == "wsum":
+            if not is_num(rhs) and (isinstance(lhs, _NumVarImpl) or lhs.name =="sum" or lhs.name == "wsum"):
+
+                if isinstance(rhs, Operator):
+                    if rhs.name == "sum":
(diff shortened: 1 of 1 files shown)


IgnaceBleukx:
After discussing offline: an alternative can be to merge this code into `simplify_bool` which would then become `simplify`.
Hard-ish part will be to ensure we do not run `simplify` again after flattening.
E.g., the constraint `max(a + b, c,d) != 10` will be flattened to `max(IV0,c,d) != 10 & a + b == IV0`. Thereby introducing a non-canonical comparison...

Wout4:
do we want to cover division as well? I guess it would make sense to do.

cpmpy/transformations/linearize.py
```diff
@@ -255,71 +255,87 @@ def only_positive_bv(lst_of_expr):
...
+                        extra_vars = rhs.args
+                    elif rhs.name == "wsum":
+                        extra_weights=  [-w for w in rhs.args[0]]
+                        extra_vars = rhs.args[1]
+                elif isinstance(rhs, _NumVarImpl):
+                    extra_weights = [-1]
+                    extra_vars = [rhs]
 
+                if lhs.name == "sum":
+                    lhs_weights = [1]*len(lhs.args) + extra_weights
```
Wout4, line 294: extra_weights will not be defined if the rhs is an operator other than sum or wsum (like div or mul or whatever)

tias:
probably closable?

PR #434, opened 2023-11-06
get_bounds() for arrays

IgnaceBleukx:
Added the get_bounds() functionality for NDVarArrays and (nested) lists in general.
Returns two arguments with arg0 being a list of lowerbounds for each of the elements and the second return argument being the upperbounds.


IgnaceBleukx:
Tests are failing but I don't think they have anything to do with this pull request...
It is some test to count the number of intermediate solutions in OR-tools @tias or @Wout4 could you have a look?

Wout4:
ran fine on my machine, here as well after rerunning..

Wout4:
Very nice addition, I think we could already simplify a lot of bound calculations in existing code. Not necessary for this pr though

PR #435, opened 2023-11-10
Changes to Cumulative constraint

IgnaceBleukx:
- Restructured __init__ for readability
- When demand is numeric, it is automatically converted to a list of equal demands (no solver actually support constant demand so better to convert it here than in each solver interface separately)
- Fixed bug in the decomposition

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -417,28 +417,24 @@ class Cumulative(GlobalConstraint):
     """
     def __init__(self, start, duration, end, demand, capacity):
         assert is_any_list(start), "start should be a list"
-        start = flatlist(start)
         assert is_any_list(duration), "duration should be a list"
-        duration = flatlist(duration)
-        for d in duration:
-            if get_bounds(d)[0]<0:
-                raise TypeError("durations should be non-negative")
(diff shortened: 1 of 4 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -468,7 +464,8 @@ def decompose(self):
                     demand_at_t += demand * ((start[job] <= t) & (t < end[job]))
                 else:
                     demand_at_t += demand[job] * ((start[job] <= t) & (t < end[job]))
-            cons += [capacity >= demand_at_t]
+
```
Wout4, line 467: why does the order matter here?

IgnaceBleukx: I'm not sure __exactly__ why it matters but in some cases the comparison was not evaluated and just become "True"... I think it has something to do if `capacity` is a numpy something vs just Python int... Anyway this seems to fix it in all cases. I will add a test where the old version failed still

IgnaceBleukx: Ok so it is indeed the Numpy thingy...
I will add a test to ensure we do not encounter this again

IgnaceBleukx:
This should be ready to merge now I think

PR #437, opened 2023-11-24
Update on tools/subsets

IgnaceBleukx:
Based in @tias suggestions to incorporate MSS/MCS tools into main library (apart from the XCP one).

- Refactored assumption-making (maybe putting it into /tools/utils is a bit overkill?)
- Added MCS/MSS
- Updated docs for debugging UNSAT models


tias:
great, but I think that the make_assump should not be in an utils file... it is specific to these forms of explanations, so should be within the mus.py file (even though the file name will no longer cover the base).

The alt is that we make a file mus.py, mcs.py, mss.py, quickxplain.py in tools; and that we add a explanation_utils.py...
@Dimosts opinions?

tias:
Actually I expect this section of the utils to grow, as we clean up/stabilize the implementations of the CPExplain tutorial.

So, I think we should add a `tools.explain` submodule (directory), there we can put a utils.py file, and then a mus.py, mss.py (e.g. we might want to add mss_grow at some point), etc...
most future-proof?
(the mus.py file would continue to exist for backward compatibility reasons)

IgnaceBleukx:
Refactoring into different files based on each type of subset might indeed be a good idea!
In that case I would certainly add the `grow` based MCS algorithm as well.

What other algorithms do you envision should get a place here? From the XCP-explain I can think of the inverse optimization method, but other than that I am not sure.
Slack-based methods are really specific to the problem at hand and I do not think we want to go as far as automatically finding linear equations and applying slack to those? Or do we?
For step-wise explanations we need quite a toolkit to make it somewhat efficient (e.g., filtering algorithms and efficient maximal propagation)

tias:
I am mostly thinking about what is most future proof. And also gives visibility that we don't just have a mus impl, but also quickxplain, mss, mcs etc.

I don't think we should expand it beyond this pull request now, but if we make it a submodule of tools, e.g. `tools.explain`, we have the freedom to have this 'utils.py' in it, and to have the different methods in different files. Then later we might for example move marco from the advanced examples here, have ous, have more advanced mcs techniques; including stuff by external contributors.

IgnaceBleukx:
Updated the pull request with a batch of explanation algorithms and refactored into multiple files.
For MSS/MCS the extra alias linking to the `_opt` variants might be a bit overkill, but to me it made sense.
Also added the grow variants with and without assumptions.

This we will probably want to add at some point but out of scope of this pr:
- Marco enumeration of MUSes
- Automatic relaxation for linear inequalities?

Also added some documentation on assumpion-based solving and how to correct models

IgnaceBleukx:
I'm thinking if we should also add smus/ocus in this pull request?

sourdough-bread:
There is already a lot in this PR. What about adding SMUS/OCUS in another one?

sourdough-bread:
PyCryptoSAT makes the tests fail, because it's missing from the test environment, @Wout4 could you check for it?

sourdough-bread (approved):
PyCryptoSAT makes the tests fail, because it's missing from the test environment, @Wout4 could you check for it?

Outside the failing tests, it's ready to be merged for me. @IgnaceBleukx

Wout4:
fixed it on main, will merge this

PR #440, opened 2023-12-22
Docs on solver statistics

Wout4:
update docs


Wout4:
You can check the docs for this pr here: https://cpmpy.readthedocs.io/en/docs-search_info/

docs/modeling.md
```diff
@@ -538,8 +538,34 @@ s.solve(cp_model_probing_level = 2,
...
+import cpmpy as cp
+x = cp.intvar(0,10, shape=3) 
+s = cp.SolverLookup.get("ortools")
+# we are operating on the ortools interface here
+s += (cp.sum(x) <= 5)
+s.solve()
+print(s.ort_solver.NumBranches())
+```
 
+Other solvers, like Minizinc do not have the distinction between model and solver objects, like ortools ort_model and ort_solver.
```
IgnaceBleukx, line 555: Maybe mention that each solver interfaces documents which solver native objects it initializes?
Might need some update in the docstring to make this somwhat more explicit. For OR-tools for example the docstring on ReadTheDocs is not as readible as it is in the code itself:

image

vs

image

IgnaceBleukx: I'm not sure how to make readthedocs understand this is an enumeration

363734: I think it needs some indentation to be understood as a list

cpmpy/solvers/minizinc.py
```diff
@@ -135,6 +135,7 @@ def __init__(self, cpm_model=None, subsolver=None):
         self.mzn_model.add_string("% Generated by CPMpy\ninclude \"globals.mzn\";\n\n")
         # Prepare solve statement, so it can be overwritten on demand
         self.mzn_txt_solve = "solve satisfy;"
+        self.mzn_result = None
```
363734, line 138: the docstring of minizinc interface needs to be updated to contain this new element, like the one of or-tool has, no?

Wout4: yes, updated it, but now i see that minizinc is not even in our list of solver api's and i don't know why..

363734: yes, it should be. Is there another missing?

Wout4: it wasn't in the init, fixed that, now trying to figure out linebreaks :(

Wout4:
Ok it now actually looks nice, yay! also added some missing links.
You have to force refresh if you had the page already open, before seeing the changes ;)

cpmpy/solvers/minizinc.py
```diff
@@ -184,6 +186,7 @@ def solve(self, time_limit=None, **kwargs):
         import minizinc.error
         try:
             mzn_result = mzn_inst.solve(**mzn_kwargs)
+            self.mzn_result = mzn_result
```
Dimosts, line 189: Why not directly
`self.mzn_result = mzn_inst.solve(**mzn_kwargs)?`

cpmpy/solvers/ortools.py
```diff
@@ -41,8 +41,7 @@
 from ..transformations.comparison import only_numexpr_equality
 
 class CPM_ortools(SolverInterface):
-    """
-    Interface to the python 'ortools' CP-SAT API
+    """Interface to the python 'ortools' CP-SAT API
```
Dimosts, line 44: Why?

docs/modeling.md
```diff
@@ -538,8 +538,35 @@ s.solve(cp_model_probing_level = 2,
 
 See [the API documentation of the solvers](api/solvers.html) for information and links on the parameters supported. See our documentation page on [solver parameters](solver_parameters.html) if you want to tune your hyperparameters automatically. 
 
+## Accessing the underlying solver object
```
IgnaceBleukx, line 541: Maybe better to put this into a new file and add a link to it in the section "Solving a model" as it already mentions our own status object?

cpmpy/solvers/minizinc.py
Wout4: mzn_results gets used a couple of times more in the function, does it make more sense to then use self.mzn_result there in stead?

cpmpy/solvers/ortools.py
Wout4: Ah yes I tried everything to get the formatting right, there was something in the sfinx documentation about things being on the first line but that turned out to not matter. Reversed it

docs/modeling.md
Wout4: a new file under advanced guides you mean?

cpmpy/solvers/minizinc.py
Dimosts: I think so yes

docs/modeling.md
Wout4: I actually like that it is just grouped together with the other use cases for using the solver interface

PR #441, opened 2024-01-02
Full propagation with exact

IgnaceBleukx:
Adapted the advanced example showing maximal propagation to use Exact.

Features highlighted:
- direct solver access
- `make_assump_model` from explanation tools
- subclassing solvers to add functionality


Dimosts:
Quite nice! I would also want that in the tools, having something like "ExactPropagator" to get and use.

Also maybe name the class something like that in the example too, PropagatorSolver seems to be generic, like being able to use any solver, but it is specific (subclassing) Exact.

No more comments, very nice.

Dimosts (approved):

PR #453, opened 2024-02-26
Cnf parser and writer

IgnaceBleukx:
I added a tool for writing and reading to and from `.cnf` files.
This format is supported by all SAT-solvers so a user can call a SAT solver externally based on the CPMpy model.

For reading cnf models, we can do very easy experimentation for example with hyperparameter configurations for SAT solvers. This allows to import any benchmark and convert it automatically to CNF.

Lastly, by being able to import these (often huge) SAT models we can also benchmark our transformation pipeline

### DIFF
--- a/cpmpy/tools/cnf.py
+++ b/cpmpy/tools/cnf.py
@@ -1,0 +1,96 @@
+import cpmpy as cp
+
+from cpmpy.expressions.variables import _NumVarImpl, _BoolVarImpl, NegBoolView
+from cpmpy.expressions.core import Operator, Comparison
+
+from cpmpy.transformations.normalize import toplevel_list
+from cpmpy.transformations.to_cnf import to_cnf
+from cpmpy.transformations.get_variables import get_variables
+
+def write_cnf(model):
+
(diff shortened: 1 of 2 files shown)


tias:
Nice addition; also a student group already needs it : )

Tools is the right place.

Light review by @Wout4  and merge?

JoD:
The format is tricky: newlines do *not* denote the end of clauses. This has two surprising consequences when parsing via getline.

```
p 2 1
1
2
0
```
has 3 solutions.

```
p 2 1

1 2 0
```
Also has 3 solutions.

Please add these test cases. 

I did not check this in 2013 and Bart and I unfairly won the SAT competition because a family of instances that contained empty lines was trivially determined as UNSAT by our solver, which was correct *by accident*, and no other solver had solved these. 😅

IgnaceBleukx:
Aha, good to know indeed ;)
I rewrote the parser to work at character level, without relying on newlines now, and added your testcases

Wout4:
I'm not familiar with the format, but I trust we now have enough testcases to merge it :)

PR #455, opened 2024-03-01
use floor and ceil to ensure int bounds

Wout4:
Bound calculations are not guaranteed to give integers, so convert before making an intvar


Dimosts:
Which bugs does this tackle? I also saw the same thing in PR #454, so I guess you encountered, but I am not sure which get_bounds can return non ints.

Dimosts (approved):

tias:
One more thing; can you check whether it is non-integer, and in that case throw a warning? or is that excessive?

Wout4:
I don't think there will ever be a surprise float appearing without the user writing it? Maybe a warning that we only have intvars and not realvars makes sense

PR #456, opened 2024-03-01
Is bool boolval

### LINKED ISSUE #452: Utils is_bool() does not include cpmpy BoolVal
is_bool() only checks python and numpy bool, but we also have the cpmpy BoolVal now.
This leads to a downstream bug in canonical comparison

Wout4:
fixes #452

### DIFF
--- a/cpmpy/expressions/utils.py
+++ b/cpmpy/expressions/utils.py
@@ -26,9 +26,6 @@ import numpy as np
 import math
 from collections.abc import Iterable # for _flatten
 from itertools import chain, combinations
-
-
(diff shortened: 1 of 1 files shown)


Dimosts:
Why include BoolVal locally and not globally?

My comment has to do in general with the structure of the file and its imports. We import BoolVal locally 3 times, while I guess needing it 3 times is enough to do it globally.

At the same time, we include globally things we use only once (iterable, combinations, math) or dont even use (chain).

Dimosts (approved):

PR #457, opened 2024-03-04
Userwarning when using floats.

Wout4:
Is a message like this what you were thinking about?


Wout4:
Also did some pep formatting, sorry

PR #458, opened 2024-03-10
Updated dimacs parser

IgnaceBleukx:
Apparently we were a bit too quick merging #453 into master.
After reworking the parser using Jo's comments, the parser did not work for more than 10 vars...

Updated the parser again and added some more tests to the testsuite to cover these cases.

### DIFF
--- a/cpmpy/tools/dimacs.py
+++ b/cpmpy/tools/dimacs.py
@@ -42,25 +42,8 @@ def write_dimacs(model, fname=None):
         if isinstance(cons, _BoolVarImpl):
             cons = Operator("or", [cons])
 
-        if isinstance(cons, Operator) and cons.name == "->":
-            # implied constraint
-            cond, subexpr = cons.args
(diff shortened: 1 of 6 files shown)


cpmpy/tools/dimacs.py
```diff
@@ -101,29 +86,22 @@ def read_dimacs(fname, sep=None):
         cnf = "\n".join(lines[i+1:]) # part of file containing clauses
```
Dimosts, line 86: It is mentioned that "If the header is omitted in the file, the number of variables and constraints are inferred."

But we basically do not allow it to be omitted in the current version, so one of the 2 should change

IgnaceBleukx: We do infer the number of variables and constraints, but the comment should be worded more carefully. The header line should always be present (that is, somewhere in the file you should have `p cnf`) but it does not necessarily need to include the number of variables and consrtaints

Dimosts (approved):

PR #462, opened 2024-04-29
Proper translation of count to minizinc

### LINKED ISSUE #461: count inside wsum crashes minizinc
Inside the CPM_minizinc _convert_expression function count only get's correctly handled when it occurs in a comparison,
this case does not occurs when count is nested in a wsum (when it's nested in other globals, these get decomposed since it is not supported nested, maybe it could be after we fix this)
When count(list,val) gets generically handled by our code, the arguments get put in a list: count([list, val]), this nested list is invalid minizinc syntax.

to reproduce:
```
iv = intvar(-8,8, shape=5)
ivv = intvar(1,2, shape=4)
con = sum([Count([ivv[0],ivv[1]], 3), 3*ivv[2]]) > iv[0]
model = Model(con)
sat = model.solve(solver='minizinc')
```

IgnaceBleukx:
Fixes #461

### DIFF
--- a/cpmpy/solvers/minizinc.py
+++ b/cpmpy/solvers/minizinc.py
@@ -498,20 +498,6 @@ class CPM_minizinc(SolverInterface):
             args_str = [self._convert_expression(e) for e in expr.args]
             return "alldifferent_except_0({})".format(args_str)
 
-        # count: we need the lhs and rhs together
-        if isinstance(expr, Comparison) and expr.args[0].name == 'count':
-            name = expr.name
-            lhs, rhs = expr.args
-            c = self._convert_expression(rhs)  # count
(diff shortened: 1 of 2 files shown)


cpmpy/solvers/minizinc.py
```diff
@@ -608,6 +608,12 @@ def zero_based(array):
         elif expr.name == "abs":
             return "abs({})".format(args_str[0])
 
+        elif expr.name == "count":
+            vars, val = expr.args
+            vars = self._convert_expression(vars)
+            val = self._convert_expression(val)
+            return "count({},{})".format(vars, val)
+
```
Dimosts, line 616: Good. So, now we should not need the one at line 502 right?

Wout4:
Aha we actually reduced code, good :) ready to merge

PR #463, opened 2024-05-02
Save all uservars

IgnaceBleukx:
In some edge cases, variables were unknown to the solver when constraints are simplified by our transformations.
A prominent example is the `simplify_bool` transformation which translates something like `cp.any(a,b,c) <= 10` to `True`.

I now added a "solver_vars" call before the solve call with all user variables.
It remains to be seen how this affects performance tho... But from a user perspective this is very welcome.

### DIFF
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -107,6 +107,8 @@ class CPM_choco(SolverInterface):
             - kwargs:      any keyword argument, sets parameters of solver object
 
         """
+        # ensure all vars are known to solver
+        self.solver_vars(list(self.user_vars))
 
         # call the solver, with parameters
         self.chc_solver = self.chc_model.get_solver()
@@ -338,6 +340,7 @@ class CPM_choco(SolverInterface):
         """
         # add new user vars to the set
         get_variables(cpm_expr, collect=self.user_vars)
+        # ensure all vars are known to solver
 
         # transform and post the constraints
         for con in self.transform(cpm_expr):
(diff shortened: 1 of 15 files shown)


cpmpy/solvers/gurobi.py
```diff
@@ -83,9 +83,9 @@ def __init__(self, cpm_model=None, subsolver=None):
         Arguments:
         - cpm_model: a CPMpy Model()
         """
-        if not self.supported():
-            raise Exception(
-                "CPM_gurobi: Install the python package 'gurobipy' and make sure your licence is activated!")
+        # if not self.supported():
```
Wout4, line 86: We want to throw this exception right, is this still leftover from testing?

IgnaceBleukx: Yes, this is leftover from testing... Licence was expired ;)

Wout4:
This is nice, we needed it.
I added a solveAll to the tests, to check if it affects this, but seems to fail for pysdd and choco..

IgnaceBleukx:
Aha yes indeed, nice catch.
I corrected it for Choco and added some exceptions for the other solvers.

For Pysdd I'm too unfamiliar witht the code to fix it I'm afraid

Wout4:
looks like pysdd internally uses more literals than we give it, and that is messing up the solution count. Will try to fix

cpmpy/solvers/choco.py
```diff
@@ -338,6 +343,7 @@ def __add__(self, cpm_expr):
         """
         # add new user vars to the set
         get_variables(cpm_expr, collect=self.user_vars)
+        # ensure all vars are known to solver
```
Dimosts, line 346: ?

Dimosts (approved):
Seems good to me too. Will just remove a commented line that does not make sense in choco.

PR #464, opened 2024-05-02
Extending automatic teststuite

IgnaceBleukx:
Extended the automatic `test_constraints` testsuite which uncovers several bugs (some of which are fixed in this PR)

Main changes to test file:
- Inspect "globalconstraints.py" and "globalfunctions.py" files to automatically generate all global constraints
- All Boolean expressions, including global constraints are also tested as numerical expressions now
- Option to do a `solveAll` which tests wheter all solutions to the constraint are valid. Untested for now as it takes a long time for some consrtaints, but potentially usefull before doing a release?
- General cleanup

Bugs and fixes:
- [fixed in #463] Some vars were unknown to the solver
- [fixed] Simplify Boolean did not translate comparisons to 0/1 when in numerical context
- [fixed] When partial functions are in a toplevel constraint, should be translated to False and not throw an error (I am unsure about this fix, it produced a lot of duplicate code, potential for cleanup here)
- [open] incorrect decomposition of `Inverse` consrtraint when nested

The last is also why the automatic tests fail for now.


Dimosts:
Is this a draft?

Dimosts:
Quite some errors in Choco too, regarding GCC. Choco accepts only ints for values, and not variables

Wout4:
The inverse bug is just because it uses element probably?
Maybe disable inverse if we want to merge this, but indeed there should be a way to get rid of all that duplicate code..

Wout4:
We could have argval return an 'undefined' value in stead of raising an error

cpmpy/expressions/globalconstraints.py
```diff
@@ -287,9 +300,16 @@ def decompose(self):
...
+        try:
+            fwd = [argval(a) for a in self.args[0]]
+            rev = [argval(a) for a in self.args[1]]
+        except IncompleteFunctionError:
+            return False
+        # args are fine, now evaluate actual inverse cons
+        try:
+            return all(rev[x] == i for i, x in enumerate(fwd))
+        except IndexError: # partiality of Element constraint
+            raise IncompleteFunctionError
```
Wout4, line 312: This should not throw an exception? just return False

IgnaceBleukx:
I'm not very convinced by the use of the NaN's anymore...
There are quite a bit of inconsistencies and hard things to catch.
E.g., nan != False returns True...

I think we should look for a more elegant solution

Wout4:
well NaN==anything is False so NaN!=anything should be True. Using NaN saves us a little bit of effort but using another undefined object is also fine for me

Wout4:
we could also make another argval function that turn NaN's into False, but then you need to use this one in boolean contexts and the other one in nonboolean contexts..

Wout4:
Fine to be merged by me, but some conflict to be resolved first

IgnaceBleukx:
Hmm not sure if I like having two value functions... Using argval seems like a good solution tho!
Actually, if we assume reasonable use by the user. That is, a user does not set values to a constraint himself, we might not even need to wrap the .value() function at all!

Take the following (convoluted) example:

```python
m = cp.Model()
arr = cp.intvar(0,10, shape=3, name="arr")
idx = cp.intvar(0,10, name="idx")
x = cp.boolvar(name="x")
y = cp.boolvar(name="y")

m += cp.AllEqual([cp.AllDifferent([arr[idx], 11, 12]),x,y]) 
m += idx == 7
```
Then, `arr[idx]` is undefined, hence will throw the exception.
As AllDifferent is the nearest Boolean context to the undefinedness, it should be set to False. But this is done by the `argval` call originating from `AllEqual`
E.g., the constraint becomes AllEqual(False, x,y) which is a perfectly valid constraint and can proceed as normal.

So basically what I'm saying is: the Boolean expression catching the undefinedness will never be at the top-level of the constraint model when the user is interested in the value of the decision variables. This is true as the model will be UNSAT when a constraint at toplevel is set to False because of the undefinedness so no values are assigned to the variables anyway.

So, as long as the value of the arguments of any expression are calculated using `argval` we do not need to put any logic about the exceptions into the individual expressions. This means we can revert the changes of the `safe_value()` and just put that logic into `argval` again like it used to be.

Wout4:
yeah exactly, `.value()` will only fail if it is called on a toplevel constraint with an undefined value directly underneath, because at all the other levels `argval `will take care of the undefinedness.
The extra safe_value function does feel quite useless, but we actually need it to be a class function because then it knows if it is being called on a boolean expression or not. `argval` can not know this..
I guess we could make it hidden `_safe_value() `and tell the user only about `argval`

IgnaceBleukx:
@Wout4 A lot of PRs depoend on this one now. I think it is ready to merge?

Wout4:
Indeed, if it also looks good to you will merge it

PR #466, opened 2024-05-06
Pycsp3 parser

Wout4:
using the parser that is available in PyCSP3, this branch contains the interface to cpmpy, so it can be used to parse XCSP3 to cpmpy. (requires pycsp3)


Wout4:
closing this, as we can extract what we need from the competition branch

PR #469, opened 2024-05-10
ShortTable Global

Wout4:
Table constraint with wildcards.
I made the string '*' the wildcard character, maybe that's not very elegant but making a new cpmpy expression for it seems unnecessary

### DIFF
--- a/cpmpy/expressions/__init__.py
+++ b/cpmpy/expressions/__init__.py
@@ -21,7 +21,7 @@
 # others need to be imported by the developer explicitely
 from .variables import boolvar, intvar, cpm_array
 from .variables import BoolVar, IntVar, cparray # Old, to be deprecated
-from .globalconstraints import AllDifferent, AllDifferentExcept0, AllDifferentExceptN, AllEqual, AllEqualExceptN, Circuit, Inverse, Table, Xor, Cumulative, \
+from .globalconstraints import AllDifferent, AllDifferentExcept0, AllDifferentExceptN, AllEqual, AllEqualExceptN, Circuit, Inverse, Table, ShortTable, Xor, Cumulative, \
     IfThenElse, GlobalCardinalityCount, DirectConstraint, InDomain, Increasing, Decreasing, IncreasingStrict, DecreasingStrict, \
     LexLess, LexLessEq, LexChainLess, LexChainLessEq, Precedence, NoOverlap, \
     NegativeTable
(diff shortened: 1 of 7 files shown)


IgnaceBleukx:
Nice, but I would decompose the ShortTable to a normal Table instead. E.g., add a row for each value in the domain of a variable corresponing to a wildcard.
But @363734 will surely be able to elaborate on this : )

Wout4:
I tried that first but there are instances in xcsp3 with thousands of wildcards, so just doing the decomposition would already take forever.. Maybe it can be decomposed into (a disjunction of) multiple Tables, so we never have to unwind the wildcards

IgnaceBleukx:
I see, then just need to add it to the solvers that support it. 
Probably only Choco and MiniZinc

363734:
choco has them, minizinc no. 
Gecode has them though but as we do through minizinc we cannot put them right?

Wout4:
I don't see any other types of table constraints in the python interface, maybe it's missing from pychoco?

Wout4:
Confirmed from their docs that it's only available in java currently, i opened an issue on pychoco

Wout4:
choco updated python interface so added shorttable.
We still have to wait for them to do a release though.

Also the way the star value is handled is a bit ugly because it is just a string and not a cpmpy object, maybe it's better to turn it into a cpmpy expression

363734:
I totally agree with creating a STAR cpmpy expression instead of string.
Another way to do it, to avoid dealing with mixed type arrays (mainly used in strong typed languages) is to have an out-of-domain value for the star and specify it as an argument of the expression. 
"shorttable(var, table, starvalue)"
ex: var x1, x2, x3, with dom {0,1,2}, table (1,1,-1),(2,-1,0),..., star value = -1

IgnaceBleukx:
Not sure if we want to make "star" an expression, we could define the constant in the utils or globalconstraints file?

E.g., `STAR = "*"`
Then a user can use it as:
```python
import cpmpy as cp
...
model += cp.ShortTable(iv, [[0,1,cp.STAR], [1,cp.STAR,2]])
```

IgnaceBleukx:
Updated this PR with the above comments.
@363734 I had to chose a filtering algo explicitly in Choco, could you have a look if there are better ones? : )
Other than that, I think this is ready to be merged in

Wout4:
Allright thanks for looking at it again, will merge this!

PR #470, opened 2024-05-10
remove type restriction for InDomain

Wout4:
We excluded booleans for this but it can also make sense for booleans.


Dimosts (approved):
Simple and good, all tests passed in all solvers locally, approving it

PR #471, opened 2024-05-10
Increasing and decreasing global constraints

Dimosts:
Increasing and decreasing (+ strict cases seperately) global constraints.

Implemented decomposition with all pairs (not only consecutive ones), should I change it to consequtive ones as it has the same effect with less constraints?

+ For choco, it normally supports all globals reified, but it has a bug at increasing and decreasing propagation in negative context. They fixed it, but until the next release the bug is there.

So, I had to remove the beautiul supported_reified = supported, and put the list of the rest constraints.

### DIFF
--- a/cpmpy/expressions/__init__.py
+++ b/cpmpy/expressions/__init__.py
@@ -5,13 +5,13 @@
     List of submodules
     ==================
     .. autosummary::
+        python_builtins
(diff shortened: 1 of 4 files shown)


Dimosts:
Changed the decomposition, using consequtive pairs only

PR #473, opened 2024-05-10
AllDiffExceptN, AllEqualExceptN

Wout4:


cpmpy/expressions/globalconstraints.py
```diff
@@ -215,7 +234,24 @@ def decompose(self):
...
+    def __init__(self, arr, n):
+        flatarr = flatlist(arr)
+        super().__init__("allequal_except_n", [flatarr, n])
+
+    def decompose(self):
+        return [((var1 == self.args[1]) | (var1 == var2) | (var2 == self.args[1])) for var1, var2 in all_pairs(self.args[0])], []
+
+    def value(self):
+        vals = [argval(a) for a in self.args[0] if argval(a) != argval(self.args[1])]
+        return len(set(vals)) == 1 or len(set(vals)) == 0
```
Dimosts, line 254: here has the logic of alldifferent_except0

cpmpy/expressions/globalconstraints.py
```diff
@@ -193,7 +193,26 @@ def decompose(self):
...
+        super().__init__("alldifferent_except_n", [flatarr, n])
+
+    def decompose(self):
+        if self.args[1] == 0:
+            return [AllDifferentExcept0(self.args[0])], []
+        # equivalent to (var1 == n) | (var2 == n) | (var1 != var2)
+        return [(var1 == var2).implies(var1 == self.args[1]) for var1, var2 in all_pairs(self.args[0])], []
+
+    def value(self):
+        vals = [argval(a) for a in self.args[0] if argval(a) != argval(self.args[1])]
```
Dimosts, line 215: why different .value() than the alldifferent_except0?

Which of the 2 is more efficient? which should keep in both the more efficient one!

Wout4: I don't see what you mean, it's the same?

Dimosts: Ah idd, probably looked at the decomposed() before

Dimosts:
all tests passed in all solvers. Am pushing a fix in the .value() of increasing decreasing similar to the one you have here for the allequal .value() etc.

Wout4:
I think you need to call argval() on each of the vars, in stead of on the comparison

Wout4:
it's like .value() but if it is a value already it also works

Dimosts:
Not necessarily, you can just create the expression and eval this (it has a .value() too)

Dimosts:
Seems more clean to use it once  than using it several times

Wout4:
i see, cool :)

Wout4:
So in exceptN, N can be a list of values, not just one value. Will need to update this pr. Also alldifferentlists can have an except version with the tuples that are excepted.

Dimosts:
Let's have a separate PR for the ones you mention, and merge this, so you can focus on the parser etc.

Wout4:
Sure shouldn't be too much work I already have the alldiff

Dimosts:
Finished, but now I cannot merge it as I commited, please revew the last changes

cpmpy/expressions/globalconstraints.py
```diff
@@ -98,6 +98,7 @@ def my_circuit_decomp(self):
 
         AllDifferent
         AllDifferentExcept0
+        AllDifferentExceptN
```
Wout4, line 101: allequalexceptN

cpmpy/expressions/globalconstraints.py
```diff
@@ -244,6 +257,23 @@ def decompose(self):
     def value(self):
         return len(set(argvals(self.args))) == 1
 
+class AllEqualExceptN(GlobalConstraint):
+    """
+    All arguments except those equal to n have a distinct value.
+    """
```
Wout4, line 263: this one can also be a list..

tests/test_globalconstraints.py
```diff
@@ -114,6 +114,37 @@ def test_alldifferent_except0(self):
         bv = cp.boolvar()
         self.assertTrue(cp.Model([cp.AllDifferentExcept0(iv[0], bv)]).solve())
 
+    def test_alldifferent_except_n(self):
```
Wout4, line 117: maybe also test with a list except

363734 (approved):
Seems good to me, Wout's comment have been adressed

Wout4:
Changed the name of solutionsets because I copy pasted that from circuit test.. good to merge after tests run

PR #474, opened 2024-05-10
Nvalues except

IgnaceBleukx:
Implemented the NValuesExcept constraint. Currently no solver supports it, but we need it for XCSP3 parsing

### DIFF
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -363,3 +363,59 @@ class NValue(GlobalFunction):
         Returns the bounds of the (numerical) global constraint
         """
         return 1, len(self.args)
+
+
+class NValueExcept(GlobalFunction):
+
+    """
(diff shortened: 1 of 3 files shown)


cpmpy/expressions/globalfunctions.py
```diff
@@ -362,4 +362,60 @@ def get_bounds(self):
...
+        arr = cpm_array(arr)
+        lbs, ubs = get_bounds(arr)
+        lb, ub = min(lbs), max(ubs)
+
+        constraints = []
+
+        # introduce boolvar for each possible value
+        bvars = boolvar(shape=(ub + 1 - lb))
+        idx_of_n = n - lb
+        if idx_of_n >= 0:
```
Wout4, line 403: and n < ub? to avoid index out of bound if big n is given

IgnaceBleukx: Ah yes indeed

PR #476, opened 2024-05-10
Among constraint

IgnaceBleukx:
Implemented the Among global constraint, similar to count


Wout4:
Bit messy for now because it builds on the testconstraints pr, and tests fail because of a bug in that one where the code in is_supported for gurobi is commented out. Among global looks good though :)

cpmpy/expressions/globalconstraints.py
```diff
@@ -334,10 +334,13 @@ def __init__(self, condition, if_true, if_false):
 
     def value(self):
         condition, if_true, if_false = self.args
-        if argval(condition):
-            return argval(if_true)
-        else:
-            return argval(if_false)
+        try:
```
Dimosts, line 337: Why this in this PR?

Dimosts (commented):
All good for the Among constraint. It includes an irrelevant things regarding the ifthenelse constraint though, that has to do with partial functions. Why is it needed, what bug does it fix?

PR #479, opened 2024-05-10
Closed version of the GCC constraint

IgnaceBleukx:
Added the optional parameter "closed" to the GCC constraint.
When true, all variables in the constraint should take a value from the values given.


cpmpy/expressions/globalconstraints.py
```diff
@@ -505,16 +541,20 @@ class GlobalCardinalityCount(GlobalConstraint):
...
     def decompose(self):
         from .globalfunctions import Count
         vars, vals, occ = self.args
-        return [Count(vars, i) == v for i, v in zip(vals, occ)], []
+        constraints = [Count(vars, i) == v for i, v in zip(vals, occ)]
+        if self.closed:
+            constraints += [InDomain(v, vals) for v in vars]
+        return constraints, []
 
     def value(self):
```
Wout4, line 559: Value should also be update?

PR #480, opened 2024-05-12
MiniZinc executable check

ThomSerg:


cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
 
 
+class SubCircuit(GlobalConstraint):
+    """
+        The sequence of variables form a subcircuits, where x[i] = j means that j is the successor of i.
+        Contrary to Circuit, there is no requirement on all nodes needing to be part of the circuit.
+        Nodes which aren't part of the subcircuit, should self loop i.e. x[i] = i.
+        The size of the subcircuit should be strictly greater than 1, so not all stops can selfloop.
+        When a startIndex is provided, it will be treated as the start of the subcircuit and will be garuanteed
+        to be inside the subcircuit.
```
IgnaceBleukx, line 283: Maybe put a link to the global constraint catalog here?

cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
+    """
+
+    def __init__(self, *args, startIndex:int=None):
```
IgnaceBleukx, line 286: style: we normally use `snake_case`, so here: `start_index = None`

cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
+        flatargs = flatlist(args)
+        # Ensure all args are integer successor values
+        if any(is_boolexpr(arg) for arg in flatargs):
+            raise TypeError("SubCircuit global constraint only takes arithmetic arguments: {}".format(flatargs))
+        # Ensure startIndex is an integer
+        if not isinstance(startIndex, int) and startIndex is not None:
+            raise TypeError("SubCircuit global constraint's startIndex argument must be an integer: {}".format(startIndex))
+        super().__init__("subcircuit", flatargs)
+        self.startIndex = startIndex
+        # Ensure there are at least two stops to create a circuit with
```
IgnaceBleukx, line 296: This one should go above the super() call, so all checks are done before creating any objects

cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
+
+    def decompose(self):
+        """
+            Decomposition for SubCircuit
+
+            A mix of the above circuit implementation, with elements from the Minizinc implementation for the support of optional visits:
+            https://github.com/MiniZinc/minizinc-old/blob/master/lib/minizinc/std/subcircuit.mzn
+        """
+
+        from .python_builtins import all as cpm_all
```
IgnaceBleukx, line 308: Nice : )

cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
+        
+        startIndex = self.startIndex
+        succ = cpm_array(self.args) # Successor variables
+        n = len(succ)
+        order = intvar(0, n-1, shape=n) # Order variables of the stops within the subcircuit (if part of it, otherwise a remaining value in the domain to satisfy AllDiff).
+        instantiated = boolvar(shape=n) # Whether a stop is part of the subcircuit.
+        endIndex = intvar(0, n-1) # Index of the last stop in the subcircuit, before looping back to the startIndex.
+         # If a startIndex is not supplied, create an additional auxilary variable
+        if startIndex is None: startIndex = intvar(0, n-1)
```
IgnaceBleukx, line 317: newline

cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
+            https://github.com/MiniZinc/minizinc-old/blob/master/lib/minizinc/std/subcircuit.mzn
+        """
+
+        from .python_builtins import all as cpm_all
+        
+        startIndex = self.startIndex
+        succ = cpm_array(self.args) # Successor variables
+        n = len(succ)
+        order = intvar(0, n-1, shape=n) # Order variables of the stops within the subcircuit (if part of it, otherwise a remaining value in the domain to satisfy AllDiff).
+        instantiated = boolvar(shape=n) # Whether a stop is part of the subcircuit.
```
IgnaceBleukx, line 314: this can be named better: maybe smt like: "is_part_of_circuit"?

cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
+        if startIndex is None: startIndex = intvar(0, n-1)
+
+        constraining = []
+        constraining += [AllDifferent(succ)] # All stops should have a unique successor.
+        constraining += [AllDifferent(order)] # All stops should have a unique order.
+        constraining += list( instantiated.implies(succ < len(succ)) ) # Successor values should remain within domain.
+
+        defining = []
+        defining += [order[startIndex] == 0] # The ordering starts at the startIndex stop.   
+        defining += [cpm_all(((instantiated[i] == True), (i != endIndex))).implies(order[succ[i]] == (order[i] + 1)) for i in range(0,n)] # If a stop is on the subcircuit and it is not the last one, than its successor should have +1 as order.
```
IgnaceBleukx, line 326: This is a bit long for a list comprehension ;)

Can you make this a normal for-loop instead? Also, use `&` instead of wrapping two expressions into an "All" this makes it more readable imho

cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
+        defining += [(instantiated[i] == (order[endIndex] >= order[i])) for i in range(0, n-1)] # A stop is either part of the subcircuit or has an order value larger than the length of the subcircuit (the order of the last element endIndex).
+        defining += [instantiated[startIndex] == True, instantiated[endIndex] == True] # Both the start and end stops should be part of the circuit.
```
IgnaceBleukx, line 328: "True" constants not required, but fine to keep for readability

cpmpy/expressions/globalconstraints.py
```diff
@@ -272,6 +273,105 @@ def value(self):
...
+
+        constraining = []
+        constraining += [AllDifferent(succ)] # All stops should have a unique successor.
+        constraining += [AllDifferent(order)] # All stops should have a unique order.
+        constraining += list( instantiated.implies(succ < len(succ)) ) # Successor values should remain within domain.
+
+        defining = []
+        defining += [order[startIndex] == 0] # The ordering starts at the startIndex stop.   
+        defining += [cpm_all(((instantiated[i] == True), (i != endIndex))).implies(order[succ[i]] == (order[i] + 1)) for i in range(0,n)] # If a stop is on the subcircuit and it is not the last one, than its successor should have +1 as order.
+        defining += [(instantiated[i] == (order[endIndex] >= order[i])) for i in range(0, n-1)] # A stop is either part of the subcircuit or has an order value larger than the length of the subcircuit (the order of the last element endIndex).
```
IgnaceBleukx, line 327: Comment can be more clear: its a double reification, so it says even more: if `i` is not part of a subcircuit, the order is strictly greater than the end index! (I.e., the node is put "after the end of the circuit"

cpmpy/solvers/choco.py
```diff
@@ -535,6 +536,15 @@ def _get_constraint(self, cpm_expr):
                 # Create task variables. Choco can create them only one by one
                 tasks = [self.chc_model.task(s, d, e) for s, d, e in zip(start, dur, end)]
                 return self.chc_model.cumulative(tasks, demand, cap)
+            elif cpm_expr.name == "subcircuit":
+                # Successor variables
+                succ = self.solver_vars(cpm_expr.args)
+                # Start index must be in subcircuit and thus cannot selfloop
+                if cpm_expr.startIndex is not None:
+                    self += [(cpm_expr.args[cpm_expr.startIndex] != cpm_expr.startIndex)]
```
IgnaceBleukx, line 544: Coming to think of it: this won't work in reified context... You'll always require the start index constraint to be true at the toplevel of the constraint model.

E..g, take the consrtaint `bv -> SubCircuit(x,y,z, start_idx=0)`
Then, valid solutions are:
```
b x y z
1 1 0 2
1 2 1 0
0 0 0 0
0 1 0 0
0 2 0 0
...
```
But the solution `0 0 0 0` for example is not allowed by this formulation. The restriction should be part of the reification (i.e., be in the _constraining_ part somehow)

cpmpy/solvers/choco.py
```diff
@@ -535,6 +536,15 @@ def _get_constraint(self, cpm_expr):
...
+                # Add an unused variable for the subcircuit length, assume that subcircuit should consist of at least 2 nodes.
+                subcircuit_length = self.chc_model.intvar(2, len(succ))
```
IgnaceBleukx, line 546: Don't make integer variables on the Choco solver directly, create a CPMpy intvar and go through `self.solver_var` instead

cpmpy/solvers/minizinc.py
```diff
@@ -572,10 +584,20 @@ def zero_based(array):
             return txt
 
         # rest: global constraints
-        elif expr.name.endswith('circuit'):  # circuit, subcircuit
+        elif expr.name == 'circuit': 
             # minizinc is offset 1, which can be problematic here...
             args_str = ["{}+1".format(self._convert_expression(e)) for e in expr.args]
 
+        elif expr.name == 'subcircuit':
```
IgnaceBleukx, line 591: Euh, this function should return a Minizinc string, but it returns nothing here?

cpmpy/solvers/ortools.py
```diff
@@ -472,17 +473,30 @@ def _post_constraint(self, cpm_expr, reifiable=False):
...
+                N = len(x)
+                arcvars = boolvar(shape=(N,N))
+                # post channeling constraints from int to bool
+                self += [b == (x[i] == j) for (i,j),b in np.ndenumerate(arcvars)]
+                # post the global constraint
+                # posting arcs on diagonal (i==j) allows for subcircuits
+                ort_arcs = [(i,j,self.solver_var(b)) for (i,j),b in np.ndenumerate(arcvars) if not ((i == j) and (i == cpm_expr.startIndex))] # The start index cannot self loop and thus must be part of the subcircuit.
+                # If no startIndex is given, garuantee that subcircuit has at least length 2
+                if cpm_expr.startIndex is None:
+                    self += [cpm_any(cpm_array(x) != np.arange(N))]
```
IgnaceBleukx, line 498: Here it does not matter the restriction is put to toplevel as OR-tools does not support reified subcircuit anyway, so fine to keep it here BUT needs to change in Choco

tests/test_globalconstraints.py
```diff
@@ -169,6 +169,70 @@ def test_not_circuit(self):
...
+            sat, x = model_and_solve(n, not_included, startIndex, negative=False, decompose=decompose)
+            self.assertTrue(sat)
+            self.assertEqual(len(included(x)), (n - len(not_included)))
+            self.assertTrue(cp.SubCircuit(x))
+            for i in not_included:
+                assert(i not in included(x))
+            if startIndex is not None:
+                assert(startIndex in included(x))
+
+        def assert_negative(n, not_included, startIndex, decompose):
```
IgnaceBleukx, line 212: If you use the Choco solver here, it will probably fail

IgnaceBleukx (changes requested):
Nice addition! I actually did not even know OR-tools had subcircuit.

Left some style-related comments but the reification in Choco and Minizinc is not correct, this really needs to be addressed before its merged

cpmpy/solvers/choco.py
ThomSerg: This was I think an issue that you tried to convey earlier in the day. In my previous implementation I had two global constraint classes for SubCircuit and thus a two-stage approach. One that enforces the start index and always decomposes (so that start index restriction can be put in constraining) and a helper global constraint that actually implements the subcircuit (not to be accessed by the user). This way we can decompose for the start index restriction (into the start index constraint and the helper global constraint) but still support the solver's subcircuit. Is this something you're looking for or is there a better way to add to the constraining part afterwards?

cpmpy/expressions/globalconstraints.py
ThomSerg: I just replicated the structure of Circuit (it also checks for the number of arguments after the super call). Should I fix that one too then?

cpmpy/solvers/minizinc.py
ThomSerg: Here I do the exact same as circuit, which is relying on the return at the complete end of `_convert_expression`. I could copy it to the subcircuit case (and at the same time also for circuit?).

cpmpy/solvers/choco.py
ThomSerg: For now I had not yet listed subcircuit in the list of supported_reified for choco

tests/test_globalconstraints.py
ThomSerg: No, since the negation of the subcircuit (the pure choco one, without the constraint for the start_index which for now is wrongly added at top level) does not contradict the start_index constraint (that the start_index stop cannot self-loop) 

For example [1, 0, 3, 2] satisfies choco's ~subcircuit, since two subcircuits are not allowed, and satisfies the start_index constraint if 0 was chosen as start_index.

What I think you hint at is the following example:
```python
n = 6
start_index = 0
x = cp.intvar(0, n-1, n)

c = []

a = cp.boolvar()
c += [~cp.SubCircuit(x, start_index=start_index)]
c += [x[start_index] == start_index]

model = cp.Model(c)
s = cp.SolverLookup.get("choco", model)
assert(s.solve())

print(x.value())
```
Here I enforce that the start_index must self-loop, which contradicts the start_index condition. In a correct implementation, this should not cause any issues since the start_index constraint is inside subcircuit which appears in a negated setting. But currently the start_index condition gets posted at top level and as you suspected the problem becomes UNSAT.

IgnaceBleukx:
Ok I see now where the issues come from with the implementation of the start-index variant.
I think indeed adding two version of th global will be the cleanest solution: `SubCircuit` and `SubCircuitWithStart` or something.
The latter can be decomposed into `SubCircuit(x) /\ start_idx != x[start_idx]`. This will internally be posted correctly as the previous expression will replace the constraint node `SubCircuitWithStart` in the expression tree. So it will be irrelevant of the context it occurs in and will get translated correctly to the solver

ThomSerg:
I made some of the changes I planned, but now I have an issue with the test-suite. With Ortools, everything works fine (both posting the global constraint directly and using my decomposition), but the other solvers fail. Choco is broken, as can be seen by issue #1085 in their repo (they claim they fixed it, but pychoco is not yet updated). But when I disable the support for subcircuit, the decomposition now also fails (on the more exotic tests like "test_comparison_constraints[choco-(subcircuitwithstart(IV26,IV27,IV28,IV29)) != (l)] - assert False", which previously did work on ortools + their are). Minizinc also fails on one test: "test_reify_imply_constraints[minizinc-((subcircuitwithstart(IV131,IV132,IV133,IV134)) == (p)) -> (p)]"

Dimosts:
oh damn it is not through a CPMpy branch so I cannot check the code locally and push commits to it

Dimosts (changes requested):
Left several comments in the code. It is more complex than I expected.

The big series of inline comments in the code more distract than help I think, better make them more concise + remove the unused code before merge. Also there are some confusing parts with variables naming that use the same name as in circuit but representing something different.

I agree with Ignace about elegant and maintainable code over minor performance issues, if they are shown to be minor. 

The Minizinc executable thing should be in a different PR I think.

Still we need the tests to pass though to merge it to master, will check tomorrow what tests are failing

cpmpy/expressions/globalconstraints.py
```diff
@@ -259,7 +261,9 @@ def value(self):
         visited = set()
         arr = argvals(self.args)
 
-        while idx not in visited:
```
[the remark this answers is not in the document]
ThomSerg: I don't know why I did that, it was late... Indeed this didn't have to change.

cpmpy/expressions/globalconstraints.py
```diff
@@ -270,7 +274,227 @@ def value(self):
...
+        """
+        from .python_builtins import min as cpm_min
+        from .python_builtins import all as cpm_all
+        
+        # Input arguments
+        succ = cpm_array(self.args) # Successor variables
+        n = len(succ)
+
+        # Decision variables
+        order = intvar(0, n-1, shape=n) # Order variables for the stops within the subcircuit (if part of it, otherwise a remaining value in the domain as to satisfy AllDiff).
```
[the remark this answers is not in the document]
ThomSerg: Yes, the AllDiff had been removed. Will change the comment

cpmpy/expressions/globalconstraints.py
```diff
@@ -270,7 +274,227 @@ def value(self):
...
+        # The distribution of the order values within the range gets decided by the successor values (see previous constraint). Stops not placed on the subcircuit get the 
+        # the order[end_index] + 1 assigned to them. This is needed for removing the degeneracy caused by unconstrained assignment of the remaining order values to the
+        # non-subcircuit stops. This seems like a better approach than an AllDiff over 'order' and trying to add an ordering over the 'order' values of the non-subcircuit stops,
+        # i.e. (~is_part_of_circuit[i] and ~is_part_of_circuit[j]) -> ( ( (i > j) and (order[i] > order[j]) ) or ( (i < j) and (order[i] < order[j]) ) ).
+        # Since we don't know which nodes will be part of the subcircuit, this would have had to be formulated over all pairs of nodes.
+        # We additionally have to check for the empty case. Otherwise, forcing all 'order' values to be 0 for symmetry-breaking reasons would force `is_part_of_circuit` 
+        # to be true for all.
+        defining += [ (is_part_of_circuit[i] == ((~empty) & (order[end_index] + 1 != order[i]))) for i in range(n)]
```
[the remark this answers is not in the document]
ThomSerg: yes indeed, to detect which stops are part of the subcircuit and which aren't

cpmpy/expressions/globalconstraints.py
```diff
@@ -270,7 +274,227 @@ def value(self):
...
+        visited = set([start_index])
+        idx = succ[start_index]
+        for i in range(len(succ)):
+            if idx ==start_index:
+                break
+            else:
+                if idx in visited:
+                    return False
+            # Something is wrong, certain variables didn't get values
+            if idx is None: return False
```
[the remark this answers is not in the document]
ThomSerg: It's an artifact from copying from the value() implementation of Circuit. Indeed, when there is not solution and the variables have value None, indexing `succ` with a None `start_index` at line 401 will result in an error first. And this has already been covered for at line 392.

cpmpy/expressions/globalconstraints.py
```diff
@@ -270,7 +274,227 @@ def value(self):
...
+            SubCircuitWithStart simply gets decomposed into SubCircuit and a constraint
+            enforcing the start_index to be part of the subcircuit.
+        """
+        # Get the arguments
+        start_index = self.start_index
+        succ = cpm_array(self.args) # Successor variables
+
+        constraining = []
+
+        # Optional improvement by passing start_index to SubCircuit.
```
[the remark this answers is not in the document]
ThomSerg: yes, as discussed with Ignace it's better to leave some performance on the table than have a very 'hacky' implementation where one global constraint changes the private fields of another

cpmpy/solvers/minizinc.py
```diff
@@ -74,16 +74,26 @@ class CPM_minizinc(SolverInterface):
     required_version = (2, 8, 2)
     @staticmethod
     def supported():
-        return CPM_minizinc.installed() and not CPM_minizinc.outdated()
+        return CPM_minizinc.installed() and CPM_minizinc.executable_installed() and not CPM_minizinc.outdated()
```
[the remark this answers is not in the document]
ThomSerg: Yes, it is a separate commit so I will pull request it separately.

cpmpy/solvers/ortools.py
```diff
@@ -472,17 +472,37 @@ def _post_constraint(self, cpm_expr, reifiable=False):
...
+                x = cpm_expr.args
+                N = len(x)
+                arcvars = boolvar(shape=(N,N))
+                # post channeling constraints from int to bool
+                self += [b == (x[i] == j) for (i,j),b in np.ndenumerate(arcvars)]
+                # post the global constraint
+                # posting arcs on diagonal (i==j) allows for subcircuits
+                ort_arcs = [(i,j,self.solver_var(b)) for (i,j),b in np.ndenumerate(arcvars)] # Allows for empty subcircuits
+                return self.ort_model.AddCircuit(ort_arcs)
+            elif cpm_expr.name == "subcircuitwithstart":
```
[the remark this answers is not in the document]
ThomSerg: No, because SubCircuitWithStart decomposes into SubCircuit and a constraint requiring that the start_index is in the subcircuit and thus does not self-loop. And SubCircuit posts all pairwise node-node links (also all self-loops) in the graph to the solver. Whilst for ortools, the api allows to just never post the link between start_index and itself in the first place. So less `ort_arcs` posted and one constraint less.

tests/test_constraints.py
```diff
@@ -194,6 +194,15 @@ def global_constraints(solver):
...
             expr = Cumulative(s, dur, e, demand, cap)
+        elif name == "Circuit":
+            S = intvar(0, 9, shape=10)
+            expr = Circuit(S)
+        elif name == "SubCircuit":
+            S = intvar(0, 9, shape=10)
+            expr = SubCircuit(S)
+        elif name == "SubCircuitWithStart":
+            S = intvar(0, 9, shape=10)
+            expr = SubCircuitWithStart(S)
```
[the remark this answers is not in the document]
ThomSerg: It defaults to 0 when no argument is provided, as required by the xcsp3 specification.

cpmpy/solvers/ortools.py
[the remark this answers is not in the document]
ThomSerg: ORTools actually supports SubCircuitWithStart and doesn't need the decomposition though SubCircuit

cpmpy/solvers/minizinc.py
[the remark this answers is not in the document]
Wout4: making a pr for it

tias:
no immediate need, closing but kept in tracker

PR #483, opened 2024-05-15
Update TEMPLATE.py

IgnaceBleukx:
Some small changes to Template.py which should make it more clear for solver developers what and where to implement.
In doubt still if we want to write out the `__add__` a bit more with the catches such as "==" being also used for reification

### DIFF
--- a/cpmpy/solvers/TEMPLATE.py
+++ b/cpmpy/solvers/TEMPLATE.py
@@ -88,7 +88,7 @@ class CPM_template(SolverInterface):
 
         # initialise everything else and post the constraints/objective
         # [GUIDELINE] this superclass call should happen AFTER all solver-native objects are created.
-        #           internally, constructor relies on __add__ which uses solver native objects
+        #           internally, the constructor relies on __add__ which uses solver native objects
         super().__init__(name="TEMPLATE", cpm_model=cpm_model)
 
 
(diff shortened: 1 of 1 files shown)


tias (approved):
Good.

Also totally agree on showing in the '==' case how to (easily?) separate out Boolean reification rather than numeric equality

IgnaceBleukx:
Things that need to be more clear in template cfr discssion with Maarten (Delft)
- __init__: super call must be after all solver-native obj because internally __add__ is called
- Time limit = wallclock time in float seconds, not process time 
- cpm_con = cpm_expr in __add__
- in transform: cpm_expr = cpm_cons
- make sure comparison case is clear
- notimplementederror in __add__ should be in else case or return statements during posting
- numexpr not required if solver does not have "sum" expression - document clearly w/ link to Choco/Exact for examples

IgnaceBleukx:
Reworked large part of the`__add__` function and made some more verbose comments and clarification to other parts.
@tias if you could have another look?

IgnaceBleukx:
I think it would make sense to merge this into master before we start adding GCS?

tias:
yes, I went over the changes and made some minor tweaks directly on github.

Will merge now, very useful resource to compare an ongoing interface agains.

PR #486, opened 2024-05-21
Scheduling constraints

IgnaceBleukx:
Currently still a draft.
PR to add Precedence and NoOverlap (+variants) constraints.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -538,7 +538,7 @@ class Precedence(GlobalConstraint):
         for s,t in zip(precedence[:-1], precedence[1:]):
             if vals[0] == t: return False
             for j in range(len(args)):
-                if vals[0] == t and sum(args[:j] == s) == 0:
+                if vals[j] == t and sum(vals[:j] == s) == 0:
                     return False
         return True
 
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -538,7 +538,7 @@ class Cumulative(GlobalConstraint):
 class Precedence(GlobalConstraint):
     """
         Constraint enforcing some values have precedence over others.
-        Given an array of variables X and and a list of precedences P:
+        Given an array of variables X and a list of precedences P:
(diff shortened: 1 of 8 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -503,6 +503,67 @@ def value(self):
...
+        vals = np.array(argvals(args))
+        for s,t in zip(precedence[:-1], precedence[1:]):
+            if vals[0] == t: return False
+            for j in range(len(args)):
+                if vals[j] == t and sum(vals[:j] == s) == 0:
+                    return False
+        return True
+
+
+class NoOverlap(GlobalConstraint):
```
Dimosts, line 574: From what I understand this is equivalent with the minizinc disjunctive constraint, right? (https://github.com/MiniZinc/libminizinc/blob/2.5.5/share/minizinc/std/disjunctive.mzn#L14-L23)

cpmpy/expressions/globalconstraints.py
```diff
@@ -503,6 +503,67 @@ def value(self):
...
+
+    def __init__(self, start, dur, end):
+        super().__init__("no_overlap", [start, dur, end])
```
Dimosts, line 577: Hmm, should end be made optional? Or we should leave it like that now, and make it optional later along with the cumulative end.

Dimosts: Also, need to make sure that the lists given have the same size

cpmpy/solvers/minizinc.py
```diff
@@ -584,6 +585,15 @@ def zero_based(array):
 
             return format_str.format(args_str[0], args_str[1], args_str[3], args_str[4])
 
+        elif expr.name == "precedence":
+            return "value_precede_chain({},{})".format(args_str[1], args_str[0])
+
+        elif expr.name == "no_overlap":
+            start, dur, end = expr.args
+            durstr = self._convert_expression([s + d == e for s, d, e in zip(start, dur, end)])
+            format_str = "forall(" + durstr + " ++ [disjunctive({},{})])"
```
Dimosts, line 609: I see

Dimosts (approved):

PR #488, opened 2024-05-28
Addingglobals channel

363734:
Addition of the channel constraint for xcsp3 competition

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -358,8 +358,11 @@ class InverseOne(GlobalConstraint):
         return [all(arr[x] == i for i, x in enumerate(arr))], []
 
     def value(self):
-        valsx = [argval(a) for a in self.args[0]]
-        return all(valsx[x] == i for i, x in enumerate(valsx))
+        valsx = argvals(self.args[0])
+        try:
+            return all(valsx[x] == i for i, x in enumerate(valsx))
+        except IndexError: # partiality of Element constraint
+            return False
 
 class Channel(GlobalConstraint):
     """
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -113,7 +113,7 @@ import numpy as np
 from ..exceptions import CPMpyException, IncompleteFunctionError, TypeError
 from .core import Expression, Operator, Comparison
 from .variables import boolvar, intvar, cpm_array, _NumVarImpl, _IntVarImpl
-from .utils import flatlist, all_pairs, argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds
+from .utils import flatlist, all_pairs, argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds, is_int
 from .globalfunctions import * # XXX make this file backwards compatible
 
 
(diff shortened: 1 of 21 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -270,15 +270,23 @@ class Inverse(GlobalConstraint):
...
         if any(is_boolexpr(arg) for arg in flatargs):
             raise TypeError("Only integer arguments allowed for global constraint Inverse: {}".format(flatargs))
-        assert len(fwd) == len(rev)
-        super().__init__("inverse", [fwd, rev])
+        if len(fwd) > len(rev):
+            raise TypeError("len(fwd) should be equal to len(rev) for the symmetric inverse, or smaller than len(rev) for the asymmetric inverse")
+        if len(fwd) == len(rev):
+            name = "inverse"
+        else:
+            name = "inverseAsym"
```
Dimosts, line 288: How and where is this handled?

cpmpy/expressions/globalconstraints.py
```diff
@@ -291,6 +299,52 @@ def value(self):
         rev = [argval(a) for a in self.args[1]]
         return all(rev[x] == i for i, x in enumerate(fwd))
 
+class InverseOne(GlobalConstraint):
+    """
+       Inverse (aka channeling / assignment) constraint but with only one array.
```
Dimosts, line 304: Which constraint from the global constraint catalog does this correspond to? Why does it need a separate class?

cpmpy/expressions/globalconstraints.py
```diff
@@ -291,6 +299,52 @@ def value(self):
...
+        arr = cpm_array(arr)
+        return [all(arr[x] == i for i, x in enumerate(arr))], []
+
+    def value(self):
+        valsx = [argval(a) for a in self.args[0]]
+        return all(valsx[x] == i for i, x in enumerate(valsx))
+
+class Channel(GlobalConstraint):
+    """
+        Channeling constraint. Channeling integer representation of a variable into a representation with boolean
```
Dimosts, line 328: Nice. Which constraint from the global constraint catalog does it correspond to?
Is there a more appropriate name? e.g. ChannelToBool or something, because it is not clear right now that it concerns int to bool represenation, and was confused seeing the name what is the difference with inverse

cpmpy/expressions/globalconstraints.py
```diff
@@ -291,6 +299,52 @@ def value(self):
...
+       Equivalent to Inverse(x,x)
+
+                arr[i] == j  <==>  arr[j] == i
+
+    """
+    def __init__(self, arr):
+        flatargs = flatlist([arr])
+        if any(is_boolexpr(arg) for arg in flatargs):
+            raise TypeError("Only integer arguments allowed for global constraint Inverse: {}".format(flatargs))
+        super().__init__("inverseOne", [arr])
```
Dimosts, line 314: How and where is this handled?

tests/test_globalconstraints.py
```diff
@@ -775,6 +775,26 @@ def test_inverse(self):
         self.assertFalse(cp.Model([cp.Inverse([x,y,x],[x,y,x])]).solve())
         self.assertRaises(TypeError,cp.Inverse,[a,b],[x,y])
         self.assertRaises(TypeError,cp.Inverse,[a,b],[b,False])
+        self.assertFalse(cp.Model([cp.Inverse([x, y, x], [x, y, x, y, x])]).solve())
+
+    def test_inverseOne(self):
```
Dimosts, line 780: Nice that it has tests here, but should also be included in the tests of the test_constraints file

Dimosts (changes requested):

cpmpy/expressions/globalconstraints.py
363734: the Channel(X) from the xcsp3 competition where X[i] =j <-> X[j]=i
New global because I ask (CFR slack conversation) and wout told me to do an other global instead of just extending Inverse(X,X) as not all the solvers accepts twice the same var as args

cpmpy/expressions/globalconstraints.py
363734: In XCSP3, all of these correspond to Channel... 
I can't seem to find this one in the global constraint catalog (closest is link_set_to_boolean but do not do that exactly)
The InverseOne is Inverse in the catalog.. The currently named inverse (symmetric) is the inverse_set in the global constraint catalog. Inverse asymetric is nowhere in the catalog.

cpmpy/expressions/globalconstraints.py
363734: for the moment decomposition only

tests/test_globalconstraints.py
363734: yep, noticed while reviewing other PR that this file was to be modified to, currently doing it

cpmpy/expressions/globalconstraints.py
```diff
@@ -181,7 +181,7 @@ def decompose(self):
         return [var1 != var2 for var1, var2 in all_pairs(self.args)], []
 
     def value(self):
-        return len(set(argvals(self.args))) == len(self.args)
+        return len(set(a.value() for a in self.args)) == len(self.args)
```
Wout4, line 184: why not argvals? We switched to argvals everywhere for consistency with the partial functions

Wout4 (commented):
Is there no solver interfaces to be changed, (do none support the additional kinds of channeling constraints?)

tias:
Adds 'Channel' (between bool and int), 'InverseOne' (x[i]==j == x[j]==i) and allows 'Inverse' to be one-sided in a subtle way (one array can be lager).

I'm not a fan of including these 3... solvers don't support them, their semantics are not all that clear from their name, and judging from the commits there is some partiality to take into account while using them.

I will close the PR, if needed we can always still include one of them when we have a good use case...

PR #489, opened 2024-05-28
Nooverlap2d

IgnaceBleukx:
Can remain a draft for now. Will merge directly into xcsp3_exectuable.


tias:
In https://github.com/CPMpy/cpmpy/blob/master/cpmpy/tools/xcsp3/globals.py

PR #491, opened 2024-05-28
Negative table

Wout4:
not sure if other solvers also support it


363734:
gecode supports it i think, but the minizinc interface do not have it I think

PR #492, opened 2024-05-31
Utility functions for NDVarArrays

IgnaceBleukx:
There was an issue with our overriding of numpy-functionality for NDVarArrays with higher dimensions.
E.g., the following happened:
```python
bvs = cp.boolvar(shape=(3,3,2))
res = bvs.any(axis=2)
```
Now `res` was a single clause with 18 elements, while doing the same with a numpy array results a 3x3 matrix, which is correct. This PR fixes that.

[Feature] Also added the option to create common global constraints directly from the array.
E.g., a user can now do `puzzle.alldifferent(axis=1)` to post the alldifferent constraints over the columns of a Sudoku.

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -555,6 +555,40 @@ class NDVarArray(np.ndarray, Expression):
 
         return np.apply_along_axis(cpm_all, axis=axis, arr=self)
 
+    def alldifferent(self, axis=None):
+        from .globalconstraints import AllDifferent
+        return np.apply_along_axis(lambda arr : AllDifferent(arr), axis=axis, arr=self)
+
+    def alldifferent_except0(self, axis=None):
+        from .globalconstraints import AllDifferentExcept0
+        return np.apply_along_axis(lambda arr: AllDifferentExcept0(arr), axis=axis, arr=self)
+
+    def allequal(self, axis=None):
+        from .globalconstraints import AllEqual
+        return np.apply_along_axis(lambda arr: AllEqual(arr), axis=axis, arr=self)
+
+    def circuit(self, axis=None):
+        from .globalconstraints import Circuit
+        return np.apply_along_axis(lambda arr: Circuit(arr), axis=axis, arr=self)
+
+    def increasing(self, axis=None):
+        from .globalconstraints import Increasing
+        return np.apply_along_axis(lambda arr: Increasing(arr), axis=axis, arr=self)
+
+    def increasing_strict(self, axis=None):
+        from .globalconstraints import IncreasingStrict
+        return np.apply_along_axis(lambda arr: IncreasingStrict(arr), axis=axis, arr=self)
+
+    def decreasing(self, axis=None):
+        from .globalconstraints import Decreasing
+        return np.apply_along_axis(lambda arr: Decreasing(arr), axis=axis, arr=self)
+
+    def decreasing_strict(self, axis=None):
+        from .globalconstraints import DecreasingStrict
+        return np.apply_along_axis(lambda arr: DecreasingStrict(arr), axis=axis, arr=self)
+
+
+
     def get_bounds(self):
         lbs, ubs = zip(*[get_bounds(e) for e in self])
         return cpm_array(lbs), cpm_array(ubs)
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -487,7 +487,7 @@ class NDVarArray(np.ndarray, Expression):
         if out is not None:
             raise NotImplementedError()
 
-        if axis is None:  # simple case where we want the maximum over the whole array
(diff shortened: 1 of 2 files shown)


cpmpy/expressions/variables.py
```diff
@@ -468,131 +468,125 @@ def sum(self, axis=None, out=None):
...
-        if axis is None:    # simple case where we want the Minimum over the whole array
-            arr = self.flatten()
-            return Minimum(arr)
-
-        arr = self.__axis(axis=axis)
-
-        out = []
-        for i in range(0, arr.shape[0]):
-            out.append(Minimum(arr[i, ...]))
+        if axis is None:    # simple case where we want the maximum over the whole array
```
Wout4, line 517: minimum

cpmpy/expressions/variables.py
```diff
@@ -468,131 +468,125 @@ def sum(self, axis=None, out=None):
...
-        # return the NDVarArray that contains the Minimum global constraints
-        return out
+        return cpm_array(np.apply_along_axis(cpm_min, axis=axis, arr=self))
 
     def any(self, axis=None, out=None):
         """
             overwrite np.any(NDVarArray)
         """
-        from .python_builtins import any
+        from .python_builtins import max as cpm_any
```
Wout4, line 526: wrong import

cpmpy/expressions/variables.py
```diff
@@ -468,131 +468,125 @@ def sum(self, axis=None, out=None):
...
 
         if out is not None:
             raise NotImplementedError()
 
-        if axis is None:    # simple case where we want the .any() over the whole array
-            arr = self.flatten()
-            return any(arr)
-
-        arr = self.__axis(axis=axis)
+        if axis is None:    # simple case where we want the maximum over the whole array
```
Wout4, line 534: update comment

cpmpy/expressions/variables.py
```diff
@@ -468,131 +468,125 @@ def sum(self, axis=None, out=None):
...
-        # return the NDVarArray that contains the any() constraints
-        return out
 
     def all(self, axis=None, out=None):
         """
             overwrite np.any(NDVarArray)
         """
 
-        from .python_builtins import all
+        from .python_builtins import max as cpm_all
```
Wout4, line 545: wrong import

cpmpy/expressions/variables.py
```diff
@@ -468,131 +468,125 @@ def sum(self, axis=None, out=None):
...
+        if any(not is_boolexpr(x) for x in self.flat):
+            raise TypeError("Cannot call .any() in an array not consisting only of bools")
+
         if out is not None:
             raise NotImplementedError()
 
-        if axis is None:    # simple case where we want the .all() over the whole array
-            arr = self.flatten()
-            return all(arr)
+        if axis is None:  # simple case where we want the maximum over the whole array
```
Wout4, line 553: .all() not maximum

Wout4 (commented):
some copy-paste errors, other than that i would maybe add a test case with solve call to make sure it behaves as expected. Same for the global with axis, looks nice though!

IgnaceBleukx:
Fixed the copy-paste errors

cpmpy/expressions/variables.py
```diff
@@ -468,131 +468,125 @@ def sum(self, axis=None, out=None):
...
+
+    def alldifferent(self, axis=None):
+        from .globalconstraints import AllDifferent
+        return cpm_array(np.apply_along_axis(lambda arr : AllDifferent(arr), axis=axis, arr=self))
+
+    def alldifferent_except0(self, axis=None):
+        from .globalconstraints import AllDifferentExcept0
+        return cpm_array(np.apply_along_axis(lambda arr: AllDifferentExcept0(arr), axis=axis, arr=self))
+
+    def allequal(self, axis=None):
```
Wout4, line 566: All of these crash when axis == None (which is the default value).

I guess they could all be valid without specifying an axis right. (would be the same as just calling the global directly)

tias:
I find it quite far-reaching to put all global constraints also in the variables... I've been avoiding this mixing of functionalities and pushing things into the expression objects (e.g. transforms could also be pushed in, which I think is harder to read/maintain, at least I had a lot of diffculty understanding how NumberJack worked in that way)

For numpy we have good reasons: we want to be numpy compatible, so people can treat our arrays as numpy arrays and call numpy functions on them and we want them to produce expressions (do we also call max and min?). But adding our global constraints is not for supporting standard numpy, its "numpy-like" programming of cpmpy.

so, how about only doing the numpy ones?

IgnaceBleukx:
Yes, adding all of the globals was a bit far fetched indeed... I'll adapt it to only include those overriding numpy's builtins

Wout4 (approved):
This is ready to merge now, axis argument works on all, any, max, min, sum and prod

PR #493, opened 2024-06-17
remove alldifferent Lists

Wout4:
was only here for the competition and we don't really expect anyone to use it


IgnaceBleukx:
Trivial, seems good to me

PR #495, opened 2024-07-04
linearize objectives

Wout4:
Remove NegBoolViews from objectives for linear solvers


IgnaceBleukx:
Hmm not sure I like this hacky way tbh...
I would expect a `linearize_objective` to take as input any numerical expression and return a weighted sum and some constraints in return...
Now, it makes a new var and relies on the transformations in the solver interface to linearize everything

tias:
yes, its weird because `3 + 2*~x` is simply `3 + 2*(1-x)` so there is a simple way to rewrite the linear expression...

this was done for exact and gurobi, but do those solvers really need it?

```python
import cpmpy as cp
m = cp.Model()
b = cp.boolvar()
m.maximize(3+2*~b)
m.solve("gurobi")
```

works; I don't have a working exact atm, can you test @IgnaceBleukx and close if it is no longer an issue?

tias:
thanks @OrestisLomis, that looks nicer.

But my question remains: do the solvers (still) actually need this? If you run your new testcase on master, does gurobi fail? does exact fail?

IgnaceBleukx:
> works; I don't have a working exact atm, can you test @IgnaceBleukx and close if it is no longer an issue?

Yes, your test does work on master, because we flatten the expression to:
```python
flat_expr, extra_cons = flatten_objective(3+2*~b)
print(flat_expr)
>> 3 + (IV0)
print(extra_cons)
>> [(sum([2] * [~b])) == (IV0)]
```
So there, the negative boolvar is transformed away as we add that constraint to the solver.

But, the problem remains when there is a flat expression that contains a negative boolvar, e.g., `x + ~b + y` or some weighted sum.
So this PR is certainly still required : )

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +437,46 @@ def only_positive_bv(lst_of_expr):
 
     return newlist
 
+def only_positive_bv_sub(expr):
+    """
+        Replaces an pure expression containing NegBoolView with an equivalent expression using only BoolVar.
```
tias, line 442: should be 'a pure'
but what does 'pure' mean? not a list?
so 'If Expression expr contains NegBoolView, rewrites it to an equivalent expression over only BoolVar's.'?

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +437,46 @@ def only_positive_bv(lst_of_expr):
...
+        expr is expected to be linearized.
```
tias, line 443: instead of 'expected to be linearized', be explicit: only accepts variables, sums, wsums; but this does not seem to be true as your bottom block is generic and mentions min/max?

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +437,46 @@ def only_positive_bv(lst_of_expr):
...
+        
+        Returns a tuple of the new expression as a wsum and a constant, which can be added to the wsum or subtracted from the rhs (depending on the usecase).
+    """
```
tias, line 446: in the code below, you will copy each expression, even if it does not contain a NegBoolView... that is unnecessarily expensive.

You can start with smth like
`nbv_sel = [isinstance(a, NegBoolView) for a in expr.args]`
to compute that once.

Then: 
if sum(nbv_sel) == 0: return expr

and then continue, where you now don't have to do the if isinstance every time in the code, but you can now do it smarter

I do realize now that if expr.name == "wsum" you should check expr.args[1] and not just expr.args, todo ; )

cpmpy/transformations/linearize.py
```diff
@@ -572,3 +604,26 @@ def only_positive_coefficients(lst_of_expr):
...
+    - Decision variable: Var
+    - Linear: sum([Var])                                   (CPMpy class 'Operator', name 'sum')
+              wsum([Const],[Var])                          (CPMpy class 'Operator', name 'wsum')
+              no negated BoolVars
+
+    Only difference with flatten_objective is that no negated boolvars are allowed.
+    """
+    flatexpr, flatcons = flatten_objective(expr, supported=supported)
+    pos_expr, const = only_positive_bv_sub(flatexpr)
+    if isinstance(const, list):
```
tias, line 618: that is not possible per your doc of that function?

tias: or wait, it checks for const to be a list? what does that mean? there is some hidden (unmaintainable) logic here...

cpmpy/transformations/linearize.py
```diff
@@ -572,3 +604,26 @@ def only_positive_coefficients(lst_of_expr):
...
+         return pos_expr, flatcons + const
+    pos_expr = Operator("wsum", [pos_expr.args[0]+[1], pos_expr.args[1]+[const]])
```
tias, line 620: euh... what?

if your function will always return a wsum, you should be more explicit about that.

then, again you create a new copy while it might not be needed...
if const != 0:
  assert pos_expr.name == "wsum", "linearize_obj: expected wsum"
  pos_expr = Operator("wsum", ...

tias (changes requested):
alright, some hidden assumptions to be made more explicit, see comments

cpmpy/transformations/linearize.py
OrestisLomis: 'pure' is indeed supposed to mean no list, but also not a 'comparison'. Not sure about the correct terminology.

cpmpy/transformations/linearize.py
OrestisLomis: If I skip inputs without NegBoolViews, the .name field of the returned expr will be ambiguous right? So I'm assuming that's okay then?

cpmpy/transformations/linearize.py
tias: well, 'pure' is unclear, so try something else... you can always give an explicit list of supported expressions for example

cpmpy/transformations/linearize.py
tias: I don't understand your question. A transformation only changes the expressions it has to, so you can always return an expression as is...

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +437,46 @@ def only_positive_bv(lst_of_expr):
...
+        expr = Operator("wsum", [list(nw), list(na)]) # force making wsum, even for arity = 1
+        const = sum(weights[i] for i in idxes)
+
+    else:
+    # other operators in comparison such as "min", "max"
+        expr = copy.copy(expr)
+        const = []
+        for i,arg in enumerate(list(expr.args)):
+            if isinstance(arg, NegBoolView):
+                new_arg, cons = get_or_make_var(1 - arg)
```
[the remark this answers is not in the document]
OrestisLomis: Wait where? If I change arg on line 470 with ar._bv a bunch of tests fail; basically all are the same and of the form:
`FAILED tests/test_constraints.py::test_comparison_constraints[exact-((x) * (y)) > (l)] - AttributeError: '_IntVarImpl' object has no attribute '_bv'`

ALL are Exact specific!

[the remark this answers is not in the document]
IgnaceBleukx: I mean in `get_or_make_var(1 - arg)`. This should become `get_or_make_var(1 - arg._bv)`

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +435,47 @@ def only_positive_bv(lst_of_expr):
...
+        which can be added to the wsum or subtracted from the rhs (depending on the usecase).
+        Otherwise the original expression is returned with a constant of 0.
+    """
+    if isinstance(expr, _NumVarImpl):
+        if isinstance(expr,NegBoolView):
+            expr, const = Operator("wsum",[[-1], [expr._bv]]), 1
+        else:
+            expr, const = expr, 0
+
+    elif expr.name == "sum":
```
tias, line 453: your code here does 4 loops over the args with the same isinstance check...

as in my previous comment, please do `nbv_sel = [isinstance(a, NegBoolView) for a in expr.args]` once and reuse that

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +435,47 @@ def only_positive_bv(lst_of_expr):
...
+        # count number of negboolviews
+            const = sum([1 for a in expr.args if isinstance(a, NegBoolView)])
+            expr = Operator("wsum",[[-1 if isinstance(arg, NegBoolView) else 1 for arg in expr.args], [arg._bv if isinstance(arg, NegBoolView) else arg for arg in expr.args]])
+        else:
+            expr, const = expr, 0
+
+    elif expr.name == "wsum":
+        weights, args = expr.args
+        if any(isinstance(a, NegBoolView) for a in args):
+            idxes = {i for i, a in enumerate(args) if isinstance(a, NegBoolView)}
```
tias, line 464: this is more in the direction of what I suggested, but you are doing it after the 'any' that already loops over everything, and you are using a dict which seems wastefull here... you can just have a boolean vector that you check on in your loops...

cpmpy/transformations/linearize.py
```diff
@@ -572,3 +603,34 @@ def only_positive_coefficients(lst_of_expr):
...
+    Only difference with flatten_objective is that no negated boolvars are allowed.
+    """
+    
+    flatexpr, flatcons = flatten_objective(expr, supported=supported)
+    
+    if isinstance(flatexpr, Operator) and flatexpr.name not in {"sum","wsum"}:
+        if flatexpr.name not in supported:
+            raise Exception(f"{flatexpr} is not linear or is not supported. Please report on github")
+        # other operators in expression such as "min", "max"
+        posexpr = copy.copy(flatexpr)
```
tias, line 623: also here please first check if there are NegBoolViews at all, and only copy in that case.

and your ifs on line 619 and 631 are so related but yet trickily different; better
'if isinsatnce(...) or flatexpr.name in {"sum", "wsum"}:
  ...only_pos_bv_sub()...
else:
   ...the manual code...

tias (changes requested):
needs some more cleaning

cpmpy/transformations/linearize.py
OrestisLomis: Is the comment about removing the loops completely (after the first loop) or about removing the isinstance() check in the loops? The suggested change of creating the boolvector seems to just remove the isinstance() checks, but more or less keeping the loops. I'm not sure what the expensive part here is.

single loop approach:

```
has_nbv_flag = false
weights = []
new_args = []
const = 0
for arg in args:
    if nbv:
        new_args.append(arg._bv)
        weights.append(-1)
        has_nbv_flag = true
        const += 1
    else:
       new_args.append(arg)
       weights.append(1)
 if has_nbv_flag : construct wsum
 else : return orig
```

The downside is that this has larger space complexity, since you have to create the copy even when not necessary.

I'm gonna go ahead and work on the change as suggested (and interpreted), i.e. removing the isinstance() checks with the checks on the boolvector, but feel free to correct me once again!

tias: its about reducing the number of isinstance checks indeed; that is the duplicate work (and in cpmpy it adds up...)

tias (approved):
Yes, the linearize.py changes are now good, clear implementations.

There is still Ignace' request regarding the tests, e.g. the first ones are very copy-pasty now, less maintainable then a list of tests...

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +435,63 @@ def only_positive_bv(lst_of_expr):
 
     return newlist
 
+def only_positive_bv_sub(expr):
+    """
+        Replaces a linear expression containing NegBoolView with an equivalent expression using only BoolVar.
+        expr is expected to be a var, sum or wsum.
+        
+        If the expression has a NegBoolView, the function returns a tuple of the new expression as a wsum and a constant, 
+        which can be added to the wsum or subtracted from the rhs (depending on the usecase).
```
tias, line 444: this sentence still unclear... it just returns a (wsum, tuple) constant, no need to specify what one 'could' do with it (e.g. the two ways in which you used it); that is not relevant, just what it returns.

Could also be more formal with

Arguments:
  expr:  Expression, variable, sum or wsum
  
  Returns:
    (wsum, const)  where wsum is a wsum where all bools are BooLVar (no negboolviews), and const is a constant value that must be added to the expression (or removed from the rhs of a comparison, as fit)
    
  (in correct rst style)
  
  note that I did end up saying smth about what can be done with it, but now it follows from the meaning and role of the constant: if you ignore it you are not equivalent!

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +435,66 @@ def only_positive_bv(lst_of_expr):
 
     return newlist
 
+def only_positive_bv_sub(expr):
```
IgnaceBleukx, line 438: I would change the name of this helper function, now the name can be confused as "it only works for subtraction"

Maybe smt like `only_postive_bv_num`, or `only_positive_bv_linear`

cpmpy/transformations/linearize.py
```diff
@@ -423,6 +409,10 @@ def only_positive_bv(lst_of_expr):
                         new_arg, cons = get_or_make_var(1 - arg)
```
IgnaceBleukx, line 409: I know it is not part of this PR currently, but it should be fixed; and this seems like a good place to do it. This should be `1 - arg._bv`

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +435,66 @@ def only_positive_bv(lst_of_expr):
 
     return newlist
 
+def only_positive_bv_sub(expr):
+    """
+        Replaces a linear expression containing :class:`~cpmpy.expressions.variables.NegBoolView` with an equivalent expression 
+        using only :class:`~cpmpy.expressions.variables.BoolVar`. 
+        
+        Arguments:
+        - `expr`: linear expression (sum, wsum, var)
```
IgnaceBleukx, line 444: suggestion: rename to `cpm_expr` for consistency

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +435,66 @@ def only_positive_bv(lst_of_expr):
...
+        
+        Returns tuple of:
+        - `expr`: linear expression (sum, wsum, var) without NegBoolView
```
IgnaceBleukx, line 447: suggestion: rename to `pos_expr` to distinguish with input

cpmpy/transformations/linearize.py
```diff
@@ -445,6 +435,66 @@ def only_positive_bv(lst_of_expr):
...
+        - `const`: constant to be added to the the expression (or removed from the rhs of a comparison, as fit)
```
IgnaceBleukx, line 448: This is what you should do with it, not __what__ it is exactly.
What I think it means: `cpm_expr == `expr + const`?

I.e., the difference between the original input expression and the transformed expression, resulting from rewring the expression using only positive Boolean variables

cpmpy/transformations/linearize.py
```diff
@@ -572,3 +622,39 @@ def only_positive_coefficients(lst_of_expr):
...
+
+    Only difference with flatten_objective is that no negated boolvars are allowed.
+    """
+    
+    flatexpr, flatcons = flatten_objective(expr, supported=supported)
+    
+    if isinstance(flatexpr, _NumVarImpl) or flatexpr.name in {"sum","wsum"}:
+        pos_expr, const = only_positive_bv_sub(flatexpr)
+        if (const != 0):
+            assert isinstance(pos_expr, Operator) and pos_expr.name == "wsum", f"unexpected expression, should be wsum but got {pos_expr}"
```
IgnaceBleukx, line 641: Not in line with the documentation of `sub` (nor with it's implementation); if the input expression does not have any negative Boolvars, it will just return as is, so this check will fail!

Better to also explicitely test for this as well in your test-suite. I.e., when the transformation is supposed to do nothing, ensure it indeed does nothing ; )

cpmpy/transformations/linearize.py
```diff
@@ -572,3 +622,39 @@ def only_positive_coefficients(lst_of_expr):
...
+            pos_expr = Operator("wsum", [pos_expr.args[0]+[1], pos_expr.args[1] + [const]])
+        return pos_expr, flatcons
+    
+    else: 
+        if flatexpr.name not in supported:
+            raise Exception(f"{flatexpr} is not linear or is not supported. Please report on github")
```
IgnaceBleukx, line 647: minor: I would prefer to refactor this to:

```python
...
else:
    assert flatexpr.name in supported, ("Unexpected numerical expression, {flatexpr} is not linear and not supported, please report on github)
   ...
```

As this means something went wrong in flatten_objective

IgnaceBleukx (changes requested):
Almost there, a few documentation/implementation inconsistencies, and 1 case that will definately fail, and hence should be added to the test-cases as well.

Also: there are a lot of unrelated files added in this PR... please remove them from this branch

cpmpy/transformations/linearize.py
OrestisLomis: I don't think anything is wrong here, as in that case const == 0, so we don't enter this code block and return the obj as is. I did add a testcase to be sure though

IgnaceBleukx: Aha yes, you're right! I missed the if-check here.

cpmpy/transformations/linearize.py
```diff
@@ -401,28 +401,18 @@ def only_positive_bv(lst_of_expr):
...
-                lhs = Operator("wsum", [list(nw), list(na)]) # force making wsum, even for arity = 1
-                rhs -= sum(weights[i] for i in idxes)
-
             if isinstance(lhs, Operator) and lhs.name not in {"sum","wsum"}:
             # other operators in comparison such as "min", "max"
                 lhs = copy.copy(lhs)
                 for i,arg in enumerate(list(lhs.args)):
                     if isinstance(arg, NegBoolView):
-                        new_arg, cons = get_or_make_var(1 - arg)
+                        new_arg, cons = get_or_make_var(1 - arg._bv)
```
tias, line 409: ok this pull request is getting a lot of attention : )

Still, while going over it again, I was wondering... so there is a `~bv` in some random expression;
and then we will replace it by 1-bv; which will become a wsum (1*1 + -1*bv) which is an int expression, which will become `iv = wsum([1,1], [1,bv])` and `~bv` will be replaced by iv with a domain of 0..1.

So... I think we should stay in Bool space, and do:
```
aux = cp.boolvar()
lhs.args[i] = aux
new_cons += (aux == arg)
```
?

cpmpy/transformations/linearize.py
```diff
@@ -572,3 +623,38 @@ def only_positive_coefficients(lst_of_expr):
...
+        return pos_expr, flatcons
+    
+    else: 
+        assert flatexpr.name in supported, (f"Unexpected numerical expression, {flatexpr} is not linear and not supported, please report on github")
+        # other operators in expression such as "min", "max"
+        new_cons = []
+        nbv_sel = [isinstance(a, NegBoolView) for a in flatexpr.args]
+        if any(nbv_sel):
+            posexpr = copy.copy(flatexpr)
+            for i,arg in enumerate(list(posexpr.args)):
```
tias, line 653: you shouldn't loop over something that you change inside; which you hide by taking a copy with list(), but you already have a copy... namely the flatexpr; plus you don't need to do the isinstance again, because you have it in nbv_sel...

```
if any(nbv_sel):
   newargs = []
   for i,arg in enumerate(flatexpr.args):
     if nbv_sel[i]:
        # same comment as above about making an intvar where a boolvar could do
        newargs.append(aux)
      else:
        newargs.append(arg)
      posexpr = copy
      posexpr.update_args(newargs)  # the only recommended way to update args of a fresh copy!
```

tias (changes requested):
some more things that I noticed

cpmpy/transformations/linearize.py
tias: which would be a linear constraint with a negbool, so might not work.. so aux + arg._bv == 1, which is probably what get_or_make_var returns? if it returns an intvar, then we better make the aux boolvar and that constraint ourself...

OrestisLomis: Isn't this the case everywhere then, since the simple objective ~p also gets transformed to the wsum : 1*1+(-1)*p?

OrestisLomis: I'm not sure if I understand the suggestion correctly, but consider the expression/objective `a*~b*c`, then how from my understanding of your suggestion this would become `a*aux*c` with a new constraint `aux+b==1`. But then `a*aux*c` is still not linear right? This is essentially why get_or_make_var is called. It essentially linearizes the constraint by plugging in a auxiliary for the whole expression see this unit test

`
    def test_linearize_objective_non_linear(self):
        a, b, c = [cp.boolvar(name=n) for n in "abc"]
        obj = linearize_objective(~a * b * c)
        self.assertEqual(str(obj), "(IV6, [((IV5) * (c)) == (IV6), ((~a) * (b)) == (IV5)])")
`

These aux' could/should arguably be BVs, but then that would need to change in get_or_make_var no?

tias: my comment is about this line of code (409), which is currently `new_arg, cons = get_or_make_var(1 - arg._bv)`  with 'arg' one argument inside an Operator inside a Comparison...

to replace it by `newarg = cp.boolvar(); cons = [(newarg + arg._bv == 1)]  # newarg = 1-arg._bv`

your example is handled elsewhere in this file?

tias:
There were still multiple odd things... so I rewrote it a bit myself, which revealed more odd things.

Please see the diff of the lastest commit.

I also think 'linearize_objective' is odd. when a solver does 'linearize_constraint', there is no guarantee that it does not contain NegBoolViews. But when it does 'linearize_objective' suddenly we do promise that?

If a solver doesn't support a negboolviews, it should call both linearize() and only_bv(); so, when in the objective they have the same, they should also call only_bv() on it directly? That would be more consistent?

OrestisLomis:
I agree that it would make more sense for linearize_objective to basically just be line 641 in `def linearize_objective`.

Then the rest of the function could be a separate function `only_positive_bv_objective` that gets called during initialisation of the solvers that require positive BVs (or however this is done in CPMpy.. not sure)

tias:
I was tricked by the subtle fact that it returns a tuple...

So we do need a function for the objective, but it should be name differently. If you could finish this PR Orestis...

PR #496, opened 2024-07-10
fix uservars in minizinc

Wout4:
All solvervars were being used as uservars in minizinc, from a time before we needed decompositions for minizinc.


Wout4:
Not sure if we need to uncomment the get_variables in objective() (line 401)
If there can be vars in the objective that are not in the constraints we need to?

IgnaceBleukx (approved):
Seems good, we indeed need to keep the line adding unknown variables in occuring in the objective

Wout4:
keep it? it's not there at the moment i think

Wout4:
Seems like putting variable in the objective that is not in the constraints doesn't work anyway
```
m = Model(x1 == 6)
m.minimize(x2)
m.solve('minizinc')
```

x2 will be None after solve even if we add it to the uservars.
Merging this

PR #500, opened 2024-07-15
Exact2

IgnaceBleukx:
Extracted changes from `xcsp3_executable` and refactored a little bit.

### DIFF
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -103,6 +103,9 @@ class CPM_exact(SolverInterface):
         options = [(opt[0], str(opt[1])) for opt in options] # Ensure values are also strings
         self.xct_solver = xct(options)
 
+        # can override encoding of variables
+        self.encoding = None
+
         # for solving with assumption variables,
         self.assumption_dict = None
         self.has_obj = False
@@ -284,7 +287,10 @@ class CPM_exact(SolverInterface):
             self.xct_solver.addVariable(revar)
         elif isinstance(cpm_var, _IntVarImpl):
             lb, ub = cpm_var.get_bounds()
-            encoding = "order" if ub-lb < 8 else "log" # heuristic bound
+            if self.encoding is None:
+                encoding = "order" if ub-lb < 8 else "log" # heuristic bound
+            else:
+                encoding = self.encoding # can also force it
             self.xct_solver.addVariable(revar, lb, ub, encoding)
         else:
             raise NotImplementedError("Not a known var {}".format(cpm_var))
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -161,7 +161,7 @@ class CPM_exact(SolverInterface):
             assump_vals = [int(not isinstance(v, NegBoolView)) for v in assumptions]
             assump_vars = [self.solver_var(v._bv if isinstance(v, NegBoolView) else v) for v in assumptions]
             self.assumption_dict = {xct_var: (xct_val,cpm_assump) for (xct_var, xct_val, cpm_assump) in zip(assump_vars,assump_vals,assumptions)}
-            self.xct_solver.setAssumption(list(zip(assump_vars,assump_vals)))
+            self.xct_solver.setAssumptions(list(zip(assump_vars,assump_vals)))
 
         # call the solver, with parameters
         start = time.time()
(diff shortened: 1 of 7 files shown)


cpmpy/solvers/exact.py
```diff
@@ -94,22 +93,13 @@ def __init__(self, cpm_model=None, subsolver=None):
         from exact import Exact as xct
 
         # initialise the native solver object
-        self.xct_solver = xct()
-        self.xct_solver.setOption("inp-purelits", "0")    # no dominance breaking to preserve solutions
-        self.xct_solver.setOption("inp-dombreaklim", "0") # no dominance breaking to preserve solutions
+        options = list(kwargs.items()) # options is a list of string-pairs, e.g. [("verbosity","1")]
+        options = [(opt[0], str(opt[1])) for opt in options] # Ensure values are also strings
```
ThomSerg, line 97: The "-" as separator will cause issues. The "options" come from optional keyword arguments via "kwargs", which does not allow for "-" inside the names of the method arguments. As an alternative, I tried requiring the arguments to be supplied to cpmpy with a "_" as seperator and then transforming them back using:
```python 
options = [(opt[0].replace("_", "-"), str(opt[1])) for opt in options] # Python does not support "-" in parameter names
```
But then we need to add a disclaimer to the documentation (and not just solely refer to Exact's list of arguments)

IgnaceBleukx: That should be fine - for other solvers this happens as well, but you can do the following hack:
```python
solver = CPM_Exact(**{"cool-argument-with-dashes" : 42})
```

Wout4:
Looks like solveAll gets stuck

tias:
great. Some tests failing though?

IgnaceBleukx:
Test were failing due to some syntax errors left from mergeing... should be good to go now

cpmpy/solvers/exact.py
```diff
@@ -227,7 +211,8 @@ def solve(self, time_limit=None, assumptions=None, **kwargs):
 
     def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from_model=False, **kwargs):
         """
-            Compute all solutions and optionally display the solutions.
+            Compute all solutions and optionallycpmpy/solvers/exact.py
```
ThomSerg, line 214: Something got accidentally pasted here?

cpmpy/solvers/exact.py
ThomSerg: I see you added some documentation for it.

ThomSerg (commented):
Looks good to me. Exact2 does however not seem to be available for Windows.

cpmpy/solvers/exact.py
```diff
@@ -61,9 +60,7 @@ class CPM_exact(SolverInterface):
         - xct_solver: the Exact instance used in solve() and solveAll()
         - assumption_dict: maps Exact variables to (Exact value, CPM assumption expression)
     to recover which expressions were in the core
-        - solver_is_initialized: whether xct_solver is initialized
-        - self.objective_given: whether an objective function is given to xct_solver
-        - self.objective_minimize: the direction of the optimization (if false then maximize)
+        - objective_: the objective function given to Exact, initially None
     as Exact can only minimize
```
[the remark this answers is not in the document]
IgnaceBleukx: Yeah.. it seems some merge issues caused very weird stuff to happen with docs and randomly copy-pasted stuff...

cpmpy/solvers/exact.py
```diff
@@ -91,7 +90,12 @@ def __init__(self, cpm_model=None, subsolver=None):
 
         Arguments:
         - cpm_model: Model(), a CPMpy Model() (optional)
-        - subsolver: None, not used
+        - subsolver: None
+
+        Additional keyword arguments:
+        The Exact solver parameters are defined by https://gitlab.com/JoD/exact/-/blob/main/src/Options.hpp#L207
```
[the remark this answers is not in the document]
IgnaceBleukx: Fixed.

Wout4:
tests pass, also tested locally, so I think this can be merged!

ThomSerg:
Still not able to download exact2 on Windows though (both through pip and through an editable install from the repo). Exact 2 is not even listed as an option to install when calling pip:
```
Could not find a version that satisfies the requirement exact==2 (from versions: 0.1.0, 0.2.0, 0.2.1, 0.3.0, 0.3.1, 0.3.2, 0.3.3, 0.4.0, 0.5.0, 0.5.1, 0.5.2, 0.5.3, 1.0.3)
```

PR #503, opened 2024-07-16
python_builtins: follow python signature

tias:
our 'min' and 'max' didn't allow multiple arguments, while Python's does

in general we were hesitating whether our min/max/sum should mimic Python's or NumPys. Now it was a mix in the middle... we were somewhat like Python but did np.max() inside (same for sum, messing up the return type); and we didnt yet get to implementing 'axis' like numpy either.

With this pull request, I propose that the least confusing thing for people that do `from cpmpy import *` is to not break any existing python code... that is, follow the Python signature of min/max/sum (and fall back to that built-in when no decision variables are given).

### DIFF
--- a/cpmpy/expressions/python_builtins.py
+++ b/cpmpy/expressions/python_builtins.py
@@ -111,7 +111,7 @@ def min(*iterable, **kwargs):
     return Minimum(iterable)
 
 
-def sum(iterable, **kwargs):
+def sum(*iterable, **kwargs):
     """
         sum() overwrites the python built-in to support decision variables.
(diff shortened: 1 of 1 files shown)


tias:
something to maybe add to this...

we don't overwrite 'abs', but using abs(decis_var) just works. However, if you do `import cpmpy as cp` then `cp.abs()` does not work, which is very very odd.

So maybe we should fake-override abs too, e.g. just `def abs(*args, **kwargs): builtins.abs(*args, **kwargs)` ?

PR #504, opened 2024-07-16
Minizinc print

ThomSerg:
When working on the XCSP3 competition, I needed to access the Minizinc and Flatzinc models to debug our decompositions. @IgnaceBleukx explained to me how I should access the underlying native solve and how to print the models. It is not that straight forward, especially for flatzinc due to the `_TemporaryFileWrapper` that the minizinc library returns. I thought it might be useful to add the retrieval of these "minizinc model strings" as a build-in method of the minizinc solver object.


Wout4:
Very convenient indeed, looks good!

PR #508, opened 2024-07-18
Pysat updates

Wout4:
fix the cryptosat dependency

### DIFF
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -81,7 +81,7 @@ class CPM_pysat(SolverInterface):
         from pysat.solvers import SolverNames
         names = []
         for name, attr in vars(SolverNames).items():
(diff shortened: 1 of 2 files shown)


cpmpy/solvers/pysat.py
```diff
@@ -81,7 +81,7 @@ def solvernames():
         from pysat.solvers import SolverNames
         names = []
         for name, attr in vars(SolverNames).items():
-            if not name.startswith('__') and isinstance(attr, tuple):
+            if not name.startswith('__') and isinstance(attr, tuple) and not 'crypto' in name:
```
tias, line 84: can you add why you are blocking it? maybe also add its name in full?
are we sure this is not an upstream bug btw, maybe we should report it while also (temporarily?) blocking it here?

Wout4: linked the github issue

PR #509, opened 2024-07-20
Accept a sequence of variable names

### LINKED ISSUE #507: Accept a sequence of variable names
While creating an array of variables `x0, x1, x2 = intvar(0, 10, shape=3, name='x')`, it's fine to have the name of `x0` as `'x[0]'`, `x1` as `'x[1]'` and so on.

However, if the name is given as a sequence (of the same length of the array), it would be convenient to assign the name to each variable accordingly, rather than use indexing.

E.g., `x, y = intvar(0, 10, shape=2, name=['x', 'y'])` should give `x` the name `'x'` and `y` the name `'y'`, but we have `"['x', 'y'][0]"` and `"['x', 'y'][1]"` currently, which is less readable...

sin3000x:
fix #507

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -118,7 +118,12 @@ def boolvar(shape=1, name=None):
         raise NullShapeError(shape)
     if shape == 1:
         return _BoolVarImpl(name=name)
-    data = _get_boolvar_data(shape=shape, name=name)
+
+    # collect the `names` of each individual decision variable
+    names = _gen_var_names(name, shape)
+
+    # create np.array 'data' representation of the decision variables
+    data = np.array([_BoolVarImpl(name=n) for n in names])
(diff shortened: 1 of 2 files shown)


tias:
Thanks for the nice patch and care in documentation too.

I felt a little uneasy with adding two very similar functions who are also each called only from one location, so I refactored it into a shared function centered around generating the names (this might have been a bit invisible in the original code).

This also allows for testing just that function. I used the occasion to test chatgpts capability of generating test cases, and (as many people already experienced) I agree that it is really great for that; for other CPMpy developers to take note ; )

Are you okay with this version? Then we will commit it

sin3000x:
Sure, it seems great, feel free to commit the version. Appreciate your effort!

tias:
Thanks, I much appreciate your efforts too.

Kind regards,
Tias

PR #510, opened 2024-07-24
Extension to tools: MARCO and SMUS

IgnaceBleukx:
I added the SMUS algorithm (with optional weights resulting in OUS) and the MARCO algorithm to tools.
A few questions:
- For now I made the "Grow" in the naive version of OUS an optional argument (which uses mss_grow_naive by default), is this smt we want?
- Do we need a naive version of MARCO as well?
- Does MARCO need to live in it's own file? For now its in tools/explain/mus

### DIFF
--- a/cpmpy/tools/explain/marco.py
+++ b/cpmpy/tools/explain/marco.py
@@ -1,0 +1,55 @@
+"""
+ Re-implementation of MUS-enumeration using MARCO.
+"""
+
+import cpmpy as cp
+from cpmpy.tools import make_assump_model
+
+
+
+def marco(soft, hard=[], solver="ortools", map_solver="ortools"):
+    """
+        Enumerates all MUSes in the unsatisfiable problem
+        Iteratively generates a subset of constraints (the seed) and checks whether it is SAT.
+        When the seed is SAT, it is grown to an MSS to derive an MCS.
+        This MCS is then added to the Map solver as a set to hit
+        In case the seed is UNSAT, the seed is shrunk to a real MUS and returned.
+        The Map-solver is instructed to not generated any superset of this MUS as next seeds
+
+        Based on:
+            Liffiton, Mark H., et al. "Fast, flexible MUS enumeration." Constraints 21 (2016): 223-250.
+    """
+
+    model, soft, assump = make_assump_model(soft, hard)
+    dmap = dict(zip(assump, soft))
+    s = cp.SolverLookup.get(solver, model)
+
+    map_solver = cp.SolverLookup.get(map_solver)
+    map_solver += cp.any(assump)
+    map_solver.solution_hint(assump, [1]*len(assump)) # we want large subsets, more likely to be a MUS
+
+    while map_solver.solve():
+
+        seed = [a for a in assump if a.value()]
+
+        if s.solve(assumptions=seed) is True: # SAT subset, find corr subsets
+            sat_subset = seed
+            new_mcs = [a for a, c in zip(assump, soft) if a.value() is False and c.value() is False]
+            map_solver += cp.any(new_mcs)
+            sat_subset += new_mcs
+            while s.solve(assumptions=sat_subset) is True:
+                new_mcs = [a for a, c in zip(assump, soft) if a.value() is False and c.value() is False]
+                sat_subset += new_mcs  # extend sat subset with new MCS
+                map_solver += cp.any(new_mcs)
+
+        else: # UNSAT, shrink to MUS, re-use MUSX
+            core = s.get_core()
+            mus = []
+            for i in range(len(core)):
+                subassump = mus + core[i + 1:]  # all but the 'i'th constraint
+                if s.solve(assumptions=subassump):
+                    mus.append(core[i])
+
+            yield [dmap[a] for a in mus]
+            # block in map solver
+            map_solver += ~cp.all(mus)
--- a/cpmpy/tools/explain/marco.py
+++ b/cpmpy/tools/explain/marco.py
@@ -11,7 +11,8 @@ from .utils import make_assump_model
 
 def marco(soft, hard=[], solver="ortools", map_solver="ortools", return_mus=True, return_mcs=True):
     """
-        Enumerates all MUSes in the unsatisfiable problem
+        Enumerating minimal unsatisfiable subsets (MUSes) and minimal correction sets (MCSes)
+         of unsatisfiable constraints.
         Iteratively generates a subset of constraints (the seed) and checks whether it is SAT.
         When the seed is SAT, it is grown to an MSS to derive an MCS.
         This MCS is then added to the Map solver as a set to hit
@@ -31,7 +32,7 @@ def marco(soft, hard=[], solver="ortools", map_solver="ortools", return_mus=True
     map_solver += cp.any(assump)
     hint = [1] *len(assump)
     map_solver.solution_hint(assump, hint) # we want large subsets, more likely to be a MUS
-    
+
     deletion_order = {a : -len(get_variables(dmap[a])) for a in assump} # avoid recomputing
 
     while map_solver.solve():
@@ -40,6 +41,8 @@ def marco(soft, hard=[], solver="ortools", map_solver="ortools", return_mus=True
 
         if s.solve(assumptions=seed) is True:
             # SAT, grow, to full MSS
+            # Assumptions encode indicator constraints a -> c, find all true assumptions
+            #    and those that could just as well be made true given the current solution
             mss = [a for a,c in zip(assump, soft) if a.value() or c.value()]
             for to_add in set(assump) - set(mss):
                 if s.solve(assumptions=mss + [to_add]) is True:
--- a/cpmpy/tools/explain/marco.py
+++ b/cpmpy/tools/explain/marco.py
@@ -29,6 +29,9 @@ def marco(soft, hard=[], solver="ortools", map_solver="ortools", return_mus=True
     map_solver = cp.SolverLookup.get(map_solver)
     map_solver += cp.any(assump)
     map_solver.solution_hint(assump, [1]*len(assump)) # we want large subsets, more likely to be a MUS
+    hint = [1] *len(assump)
+    map_solver.solution_hint(assump, hint) # we want large subsets, more likely to be a MUS
+
 
     while map_solver.solve():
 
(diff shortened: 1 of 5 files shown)


cpmpy/tools/explain/mus.py
```diff
@@ -108,6 +111,55 @@ def do_recursion(soft, hard, delta):
     return [dmap[a] for a in core]
 
 
+def optimal_mus(soft, hard=[], weights=1, solver="ortools", hs_solver="ortools"):
+    """
+        Find an optimal MUS according to a linear objective function.
+        Works by iteratively generating correction subsets and computing optimal hitting sets to those enumerated sets.
+        For better performance of the algorithm, use an incemental solver to compute the hitting sets such as Gurobi.
```
tias, line 118: I guess we should support something like hs_solver=["gurobi", "ortools"] where solverLookup.get(list) would return the first available one? (for a separate pull request clearly... this is just a request for comment on the idea)

cpmpy/tools/explain/mus.py
```diff
@@ -108,6 +111,55 @@ def do_recursion(soft, hard, delta):
...
+
+        Assumption-based implementation for solvers that support s.solve(assumptions=...)
+        More naive version available as `optimal_mus_naive` to use with other solvers.
+
+        CPMpy implementation loosely based on the "SMUS" algorithm from
```
tias, line 123: is it? the OCUS paper is, but if it is only indirectly through the OCUS paper then I think citing that is enough?

We might want to more explicitly say that without overwriting 'weights' this will find the smallest mus... e.g. in the top of the doc?

cpmpy/tools/explain/mus.py
```diff
@@ -108,6 +111,55 @@ def do_recursion(soft, hard, delta):
...
+
+    model, soft, assump = make_assump_model(soft, hard)
+    dmap = dict(zip(assump, soft))
+
+    s = cp.SolverLookup.get(solver, model)
+    # s.solution_hint(assump, [1]*len(assump)) # causes weirdness in OR-Tools: https://github.com/google/or-tools/issues/4324
+    assert s.solve(assumptions=assump) is False
+
+    # initialize hitting set solver
+    if is_num(weights):
```
tias, line 141: add some comment that single scalar to array of required dimension

cpmpy/tools/explain/mus.py
```diff
@@ -108,6 +111,55 @@ def do_recursion(soft, hard, delta):
...
+        weights = np.ones(len(assump), dtype=int)
+    hs_solver = cp.SolverLookup.get(hs_solver)
+    hs_solver.minimize(cp.sum(assump * np.array(weights)))
```
tias, line 144: you already assert s.solve() is false, so we know of one unsat core/mss; could that be used to initialize the hitting set solver?
Otherwise the simplest mcs to initialize with is simply all softs; you will need to hit at least one of those; this might 'save' one iteration (as I think now it will call the HS with no constraints so it will set all assumps to 0, instead of at least one to 1?

cpmpy/tools/explain/mus.py
```diff
@@ -108,6 +111,55 @@ def do_recursion(soft, hard, delta):
...
+
+    while hs_solver.solve() is True:
+        hitting_set = [a for a in assump if a.value()]
+        if s.solve(assumptions=hitting_set) is False:
+            break
+
+        # SAT, use hitting set as start for disjoint MCS enumeration
+        sat_subset = list(hitting_set)  # make new list
+        while s.solve(assumptions=sat_subset) is True:
```
tias, line 153: you are calling solve again with the same assumptions in the first iteration...
I think you should first add the 'just found' MCS, and then comment that you will also search for and extend with disjoint ones...
If you dont like the code duplication (this implementation is exceptionally clean currently...), you can also make it a while True:, with the solve call at the end and from which you break if False?

I guess I would favor something like:
mcs = [...]
hs_solver +=

if add_disjoint_mcs:
  sat_subset = hs + mcs...              <--- maybe a comment of why this works ; )
  while s.solve() is True:
   ...
   (

cpmpy/tools/explain/mus.py
```diff
@@ -108,6 +111,55 @@ def do_recursion(soft, hard, delta):
...
+            International Conference on Principles and Practice of Constraint Programming. Cham: Springer International Publishing, 2015.
+        and OCUS from:
+            Gamba, Emilio, Bart Bogaerts, and Tias Guns. "Efficiently explaining CSPs with unsatisfiable subset optimization."
+            Journal of Artificial Intelligence Research 78 (2023): 709-746.
+
+    """
+
+
+    model, soft, assump = make_assump_model(soft, hard)
+    dmap = dict(zip(assump, soft))
```
tias, line 134: maybe 'cmap'? with comment  # map of assumption variable to constraint

cpmpy/tools/explain/mus.py
```diff
@@ -184,3 +236,85 @@ def do_recursion(soft, hard, delta):
...
+        weights = np.ones(len(bvs), dtype=int)
+    hs_solver = cp.SolverLookup.get(hs_solver)
+    hs_solver.minimize(cp.sum(bvs * np.array(weights)))
+
+    while hs_solver.solve() is True:
+
+        hitting_set = [c for bv, c in zip(bvs, soft) if bv.value()]
+        if cp.Model(hard + hitting_set).solve(solver=solver) is False:
+            break
+        # SAT, how to do grow without assumptions?
```
tias, line 258: I'm confused... we don't grow in the above version either? though we could...
I think to mimic the above you just have to copy the disjoint part?

cpmpy/tools/explain/mus.py
```diff
@@ -184,3 +236,85 @@ def do_recursion(soft, hard, delta):
...
+            Liffiton, Mark H., et al. "Fast, flexible MUS enumeration." Constraints 21 (2016): 223-250.
+    """
+
+    model, soft, assump = make_assump_model(soft, hard)
+    dmap = dict(zip(assump, soft))
+    s = cp.SolverLookup.get(solver, model)
+
+    map_solver = cp.SolverLookup.get(map_solver)
+    map_solver += cp.any(assump)
+    map_solver.solution_hint(assump, [1]*len(assump)) # we want large subsets, more likely to be a MUS
```
tias, line 289: careful, I think this is ortools specific... maybe check if interface has this function

cpmpy/tools/explain/utils.py
```diff
@@ -14,6 +15,8 @@ def make_assump_model(soft, hard=[], name=None):
     assump = cp.boolvar(shape=(len(soft2),), name=name)
 
     # hard + implied soft constraints
+    if isinstance(hard, Expression):
```
tias, line 18: you can also run toplevel_list, that will do that for you...

cpmpy/tools/explain/mus.py
```diff
@@ -108,6 +111,55 @@ def do_recursion(soft, hard, delta):
     return [dmap[a] for a in core]
```
tias, line 113: maybe add a `def smus(): optimal_mus()` shorthand... I think we would use that

tias (changes requested):
I think Marco should live in a separate file indeed... its mus and mcs enumeration, not just single mus.

Seeing this now, I realize we already have a Marco... its in advanced examples:
https://github.com/CPMpy/cpmpy/blob/master/examples/advanced/marco_musmss_enumeration.py

not sure if you based in on that?


I also see a mention of ocus there...

cpmpy/tools/explain/mus.py
IgnaceBleukx: Hmm, not sure how we can use the core to init the set of MCSes? We cannot derive an MCS from a core...
Agree with saving 1 iteration tho, probably cleanest is to add a constraint `cp.sum(assump) >= 1` to the hs solver at the start.
Ah but... then what if the hard constraint themselves are UNSAT? Should we check for that? It's now done implicitely in iteration 1

cpmpy/tools/explain/mus.py
IgnaceBleukx: Made it default to "None" which is more intuitive

cpmpy/tools/explain/mus.py
tias: hmm. I guess with the current call, if the assumps are all false it must be in the hards?

cpmpy/tools/explain/mus.py
IgnaceBleukx: Refactored this to make it two clear steps.

cpmpy/tools/explain/mus.py
IgnaceBleukx: In the other's it's also dmap... Not sure why tho ;)

cpmpy/tools/explain/mus.py
IgnaceBleukx: We do implicitely grow in the assumption version, by looking at the assignment found  by the solver.
I.e., there may be more assumption set to True, besides `hitting_set`. Actually, that's the premise of why the correction-subset enumeration works in that algo.
But here, we don't get any "free" extra constraints in the sat_subset

cpmpy/tools/explain/mus.py
IgnaceBleukx: Hmm, not sure. 
As a novice it's nice to have the likely best solver to be selected but it might unwanted for a more experienced user to have the algo silently default to ortools if there is a problem with the first solver.

cpmpy/tools/explain/utils.py
```diff
@@ -17,6 +17,7 @@
 """
 
 import cpmpy as cp
+from cpmpy.expressions.core import Expression
```
Dimosts, line 20: Is this really needed?

cpmpy/tools/explain/marco.py
```diff
@@ -0,0 +1,81 @@
...
+                mss = [a for a,c in zip(assump, soft) if a.value() or c.value()]
+                # grow to full MSS
+                for to_add in set(assump) - set(mss):
+                    if s.solve(assumptions=mss + [to_add]) is True:
+                        mss.append(to_add)
+                mcs = [a for a in assump if a not in set(mss)]
+                map_solver += cp.any(mcs)
+                yield "MCS", [dmap[a] for a in mcs]
+
+            else:
```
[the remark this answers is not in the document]
Dimosts: was cleaned

cpmpy/tools/explain/marco.py
```diff
@@ -0,0 +1,81 @@
...
+        Enumerates all MUSes in the unsatisfiable problem
+        Iteratively generates a subset of constraints (the seed) and checks whether it is SAT.
+        When the seed is SAT, it is grown to an MSS to derive an MCS.
+        This MCS is then added to the Map solver as a set to hit
+        In case the seed is UNSAT, the seed is shrunk to a real MUS and returned.
+        The Map-solver is instructed to not generated any superset of this MUS as next seeds
+
+        Based on:
+            Liffiton, Mark H., et al. "Fast, flexible MUS enumeration." Constraints 21 (2016): 223-250.
+    """
```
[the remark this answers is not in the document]
Dimosts: it is indeed better now, can leave as is

PR #513, opened 2024-08-28
cover cp.abs() using the builtin

Dimosts:
To cover the case of cp.abs().

Currently, we do not override abs(), so it fails, although if we do from cpmpy import * just using directly abs() works (because we override it directly in the variables).

just including from builtins import abs in __init__.py of expressions tackles it

### DIFF
--- a/cpmpy/expressions/python_builtins.py
+++ b/cpmpy/expressions/python_builtins.py
@@ -136,9 +136,7 @@ def abs(element):
         if the element given is not a CPMpy expression, the built-in is called
         else an Absolute functional global constraint is constructed.
     """
-    if is_any_list(element):
-        raise CPMpyException('abs does not accept iterables')
-    if not isinstance(element, Expression):
+    if is_any_list(element) or not isinstance(element, Expression):
         return builtins.abs(element)
 
     return Abs(element)
--- a/cpmpy/expressions/python_builtins.py
+++ b/cpmpy/expressions/python_builtins.py
@@ -136,10 +136,13 @@ def abs(element):
         if the element given is not a CPMpy expression, the built-in is called
         else an Absolute functional global constraint is constructed.
     """
-    if is_any_list(element):
+    if is_any_list(element):  # compat: not allowed by builtins.abs(), but allowed by numpy.abs()
         return cpm_array([abs(elem) for elem in element])
 
-    if not isinstance(element, Expression):
-        return builtins.abs(element)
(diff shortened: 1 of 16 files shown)


IgnaceBleukx:
Maybe just put a comment on why specifically we import it here? 
And ideally also a test : )

tias:
this is too sneaky for me, it will be a hidden feature not present in our docs or visible to standard users.

The right way is a new function in python_builtins, with a one-liner documentation, that calls the `Abs()` global function directly.

Will also make it easier to (generate) tests for...

Dimosts:
Random not related test is failing .... Will dig into it

Wout4:
Looks good, thanks for adding tests as well, can be merged

cpmpy/expressions/python_builtins.py
```diff
@@ -125,3 +127,19 @@ def sum(*iterable, **kwargs):
...
+    """
+        abs() overwrites the python built-in to support decision variables.
+
+        if the element given is not a CPMpy expression, the built-in is called
+        else an Absolute functional global constraint is constructed.
+    """
+    if is_any_list(element):
+        return cpm_array([abs(elem) for elem in element])
+
+    if isinstance(element, Expression):
```
tias, line 142: is this missing a 'not'??? (or returning the wrong thing)

tias (changes requested):

cpmpy/expressions/python_builtins.py
Wout4: ah this actually works (for the same reason `from cpmpy import *` works, because our expressions override the builtin). but of course the not should be there, so we use the builtin for non-expressions.

This makes me think that we can maybe remove the abs code from the expressions themselves, since everything will go through this function now

Wout4:
Do we still want to allow using the builtin abs on our expressions, or should we enforce cpmpy.abs? I guess it's fine to allow it

cpmpy/expressions/python_builtins.py
Dimosts: indeed the not was missing

Wout4:
@tias looks like you have to approve this now that the 'not' is added :)

tias (approved):
changed to make the global in the if (without not), seems more logical

also added a comment on the 'any_list', which I still find a bit weird but not a blocker : ) e.g. why it has to be a cpm_array... could have done map(abs, arg) or so, but maybe not worth any further thought

PR #514, opened 2024-09-03
Linearize modulo

IgnaceBleukx:
Cherry picked the linearization of modulo from the Exact2 branch and added some tests.

### DIFF
--- a/tests/test_trans_linearize.py
+++ b/tests/test_trans_linearize.py
@@ -178,26 +178,34 @@ class TestVarsLhs(unittest.TestCase):
 
     def test_mod(self):
 
-        x,y = cp.intvar(1,5, name="x"), cp.intvar(1,5,name="y")
-        rhs = cp.intvar(-5,5, name="rhs")
+        x,y = cp.intvar(1,3, name="x"), cp.intvar(1,3,name="y")
 
-        cons = (x % y) <= rhs
+        cons = (x % y) <= 2
         sols = set()
-        cp.Model(cons).solveAll(display=lambda : sols.add((x.value(), y.value(), rhs.value())))
-        lincons = linearize_constraint([cons], supported={"sum", "wsum", "mod"})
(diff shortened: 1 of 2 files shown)


tests/test_trans_linearize.py
```diff
@@ -176,6 +176,30 @@ def test_impl(self):
...
 
+    def test_mod(self):
+
+        x,y = cp.intvar(1,5, name="x"), cp.intvar(1,5,name="y")
+        rhs = cp.intvar(-5,5, name="rhs")
+
+        cons = (x % y) <= rhs
+        sols = set()
+        cp.Model(cons).solveAll(display=lambda : sols.add((x.value(), y.value(), rhs.value())))
+        lincons = linearize_constraint([cons], supported={"sum", "wsum", "mod"})
```
Dimosts, line 187: Are you sure this tests what it should? "mod" in supported is not used anywhere, what was added uses "mul" in supported.

tests/test_trans_linearize.py
```diff
@@ -176,6 +176,30 @@ def test_impl(self):
...
+
+        linsols = set()
+        cp.Model(lincons).solveAll(display=lambda : linsols.add((x.value(), y.value(), rhs.value())))
+        self.assertSetEqual(sols, linsols)
+        # what about reified?
+        bv = cp.boolvar(name="bv")
+        cons = bv.implies((x % y) <= rhs)
+        sols = set()
+        cp.Model(cons).solveAll(display=lambda: sols.add((bv.value(), x.value(), y.value(), rhs.value())))
+        lincons = linearize_constraint([cons], supported={"sum", "wsum", "mod"})
```
Dimosts, line 197: Same as before. I actually tried to see if it ever reached the new code for mode, but it does not. And if I replace mod with mul in supported, then the test fails

Dimosts:
It looks good now.

Dimosts (approved):

PR #516, opened 2024-09-30
Streamline modulo and division accross solvers

Wout4:
should solve division for exact?

### DIFF
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -470,8 +470,8 @@ class CPM_ortools(SolverInterface):
                                                      self.solver_vars(lhs.args[0]), ortrhs)
                 elif lhs.name == 'mod':
                     # catch tricky-to-find ortools limitation
-                    if lhs.args[0].lb <= 0 or lhs.args[1].lb <= 0:
-                        raise NotSupportedError("Both arguments of a modulo Operator must be positive.")
+                    if lhs.args[1].lb <= 0:
+                        raise NotSupportedError("First argument of a modulo Operator must be positive.")
                     return self.ort_model.AddModuloEquality(ortrhs, *self.solver_vars(lhs.args))
                 elif lhs.name == 'pow':
                     # only `POW(b,2) == IV` supported, post as b*b == IV
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -4,7 +4,12 @@
 ## ortools.py
 ##
 """
-    Interface to ortools' CP-SAT Python API
+    Interface to OR-Tools' CP-SAT Python API. 
+    
+    The 'ortools' python package is bundled by default with CPMpy.
+    It can be installed through `pip`:
+
+        $ pip install ortools
 
     Google OR-Tools is open source software for combinatorial optimization, which seeks
     to find the best solution to a problem out of a very large set of possible solutions.
@@ -22,6 +27,10 @@
         :nosignatures:
 
         CPM_ortools
+
+    ==============
+    Module details
+    ==============
 """
 import sys  # for stdout checking
 import numpy as np
@@ -42,7 +51,7 @@ from ..transformations.comparison import only_numexpr_equality
 
 class CPM_ortools(SolverInterface):
     """
-    Interface to the python 'ortools' CP-SAT API
+    Interface to the Python 'ortools' CP-SAT API
 
     Requires that the 'ortools' python package is installed:
     $ pip install ortools
@@ -79,7 +88,7 @@ class CPM_ortools(SolverInterface):
 
         Arguments:
         - cpm_model: Model(), a CPMpy Model() (optional)
-        - subsolver: None
+        - subsolver: None, not used
         """
         if not self.supported():
             raise Exception("Install the python 'ortools' package to use this solver interface")
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -98,6 +98,12 @@ class CPM_ortools(SolverInterface):
 
         # initialise everything else and post the constraints/objective
         super().__init__(name="ortools", cpm_model=cpm_model)
+    @property
+    def native_model(self):
+        """
+            Returns the solver's underlying native model (for direct solver access).
+        """
+        return self.ort_model
 
 
     def solve(self, time_limit=None, assumptions=None, solution_callback=None, **kwargs):
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -470,9 +470,11 @@ class CPM_ortools(SolverInterface):
                                                      self.solver_vars(lhs.args[0]), ortrhs)
                 elif lhs.name == 'mod':
                     # catch tricky-to-find ortools limitation
-                    if get_bounds(lhs.args[1])[0] <= 0:
-                        raise NotSupportedError("First argument of a modulo Operator must be positive.")
-                    return self.ort_model.AddModuloEquality(ortrhs, *self.solver_vars(lhs.args))
+                    x,y = lhs.args
+                    if get_bounds(y)[0] <= 0: # not supported, but result of modulo is agnositic to sign of second arg
+                        y, link = get_or_make_var(-lhs.args[1])
+                        self += link
+                    return self.ort_model.AddModuloEquality(ortrhs, *self.solver_vars([x,y]))
                 elif lhs.name == 'pow':
                     # only `POW(b,2) == IV` supported, post as b*b == IV
                     if not is_num(lhs.args[1]):
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -470,6 +470,8 @@ class CPM_ortools(SolverInterface):
                                                      self.solver_vars(lhs.args[0]), ortrhs)
                 elif lhs.name == 'mod':
                     # catch tricky-to-find ortools limitation
+                    if lhs.args[0].lb <= 0 or lhs.args[1].lb <= 0:
+                        raise NotSupportedError("Both arguments of a modulo Operator must be positive.")
                     return self.ort_model.AddModuloEquality(ortrhs, *self.solver_vars(lhs.args))
                 elif lhs.name == 'pow':
                     # only `POW(b,2) == IV` supported, post as b*b == IV
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -41,7 +41,7 @@ from ..expressions.core import Expression, Comparison, Operator, BoolVal
 from ..expressions.globalconstraints import DirectConstraint
 from ..expressions.variables import _NumVarImpl, _IntVarImpl, _BoolVarImpl, NegBoolView, boolvar
 from ..expressions.globalconstraints import GlobalConstraint
-from ..expressions.utils import is_num, eval_comparison, flatlist, argval, argvals
+from ..expressions.utils import is_num, eval_comparison, flatlist, argval, argvals, get_bounds
 from ..transformations.decompose_global import decompose_in_tree
 from ..transformations.get_variables import get_variables
 from ..transformations.flatten_model import flatten_constraint, flatten_objective, get_or_make_var
@@ -470,7 +470,7 @@ class CPM_ortools(SolverInterface):
                                                      self.solver_vars(lhs.args[0]), ortrhs)
                 elif lhs.name == 'mod':
                     # catch tricky-to-find ortools limitation
-                    if lhs.args[1].lb <= 0:
+                    if get_bounds(lhs.args[1])[0] <= 0:
                         raise NotSupportedError("First argument of a modulo Operator must be positive.")
                     return self.ort_model.AddModuloEquality(ortrhs, *self.solver_vars(lhs.args))
                 elif lhs.name == 'pow':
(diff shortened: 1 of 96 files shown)


Wout4:
ah we added linearization of modulo on main, should merge that here

cpmpy/transformations/linearize.py
```diff
@@ -139,10 +144,29 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
                             raise NotImplementedError("Modulo with a divisor domain containing 0 is not supported. Please safen the expression first.")
                         k = intvar(*get_bounds((x - rhs) // y))
                         mult_res, newcons = get_or_make_var(k * y)
-                        newlist += linearize_constraint([rhs < abs(y)]+newcons, supported, reified=reified)
+                        # (abs of) modulo rhs is smaller than (abs of) the divisor y, but also needs to be of same sign as x.
+                        newlist += linearize_constraint(only_implies(only_bv_reifies(flatten_constraint([Abs(rhs) < Abs(y), (x > 0).implies(rhs >= 0), (x < 0).implies(rhs <= 0)]))) + newcons, supported, reified=reified)
```
Wout4, line 148: Should we make this nicer? also confused why we don't have to decompose the abs?

Wout4:
Added tests for modulo and division, added a fix for modulo and added linearization of division.

There also was a small optimisation with a helper function in the xcsp3 branch, to keep track of newly created vars in stead of having to call get_variables but I decided to omit it here. The variables are used to make sure we don't mess up solution counts for aux vars that are not totally defined, but most of the time they are totally defined so it seemed unnecessary to always use the helper function.

cpmpy/transformations/linearize.py
IgnaceBleukx: Well, I'm not sure what this is doing for currently?
It also ties back to the division-saga. 

As we decided to do integer division, do we need to decide how to implement Modulo too? Or is this independent?

cpmpy/transformations/linearize.py
```diff
@@ -139,10 +144,29 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
...
+                        raise TypeError(
+                            f"Can't divide by a domain containing 0, safen the expression first")
+                    r = intvar(-(max(abs(lb) - 1, abs(ub) - 1)), max(abs(lb) - 1, abs(ub) - 1)) # remainder can be both positive and negative (round towards 0, so negative r if a and b are both negative)
+                    cpm_expr = [eval_comparison(cpm_expr.name, a, b * rhs + r)]
+                    cond = [Abs(r) < Abs(b), Abs(b * rhs) <= Abs(a)]
+                    decomp = toplevel_list(decompose_in_tree(cond))  # decompose abs
+                    cpm_exprs = toplevel_list(decomp + cpm_expr)
+                    exprs = linearize_constraint(flatten_constraint(cpm_exprs), supported=supported)
+                    newlist.extend(exprs)
+                    continue
```
IgnaceBleukx, line 168: Some elaborate comment on what this is doing exactly would be very helpful.
I think it's doing this, but not sure:
```
x  / y = z :: x  =  y * z + r
```
with `r` either -1,0 or 1, right?
But I'm very confused about the `abs` constraints here

cpmpy/transformations/linearize.py
```diff
@@ -316,10 +341,12 @@ def canonical_comparison(lst_of_expr):
                                     if isinstance(b, _NumVarImpl)], \
                                     sum(-a * b for a, b in zip(rhs.args[0], rhs.args[1])
                                     if not isinstance(b, _NumVarImpl))
+                # 2) add collected variables to lhs
                 if isinstance(lhs, Operator) and lhs.name == "sum":
                     lhs, rhs = sum([1 * a for a in lhs.args] + lhs2), rhs
                 elif isinstance(lhs, _NumVarImpl) or (isinstance(lhs, Operator) and lhs.name == "wsum"):
-                    lhs, rhs = lhs + lhs2, rhs
+                    if len(lhs2) != 0:
+                        lhs, rhs = lhs + lhs2, rhs
```
IgnaceBleukx, line 349: Why add this check? The re-assignment of rhs also seems a little weird

IgnaceBleukx (changes requested):
I'm not really a fan of the many side-conditions in linearize. Do we really need them? I feel their might be easier ways to do it

cpmpy/transformations/linearize.py
Wout4: conceptually it's dependent but you can implement it independantly. You can define modulo as the remainder after floordivision (which is how it's mostly defined mathematically I think), but our solvers (using integer division) seem to define modulo as the remainder after integer division (so we mimic this to stay in line with the our integer division decision, also only z3 does it different but also not according to the mathematical definition as far as i can tell)

Wout4: this line says the remainder has to be smaller than the divisor (otherwise we could fit another time the divisor). Had to add the constraints that determine the sign of the remainder, otherwise you can always have multiple solutions (one negative and one positive).

Wout4: but these constraint have to go through our transformations again.. I guess we could manually linearize them as well

cpmpy/transformations/linearize.py
Wout4: ah the second constraint didn't need abs indeed!
r is just the remainder after (integer) division. I added comments

Wout4:
what do you mean with side conditions? we are mostly just going case by case on the different expression types

Wout4:
It doesn't seem to break anything yet, but our bound calculations are also wrong, we have to use the integer modulo there as well

Wout4:
@IgnaceBleukx can you maybe look at this again sometime?
Would be nice to have, also solves some minor bugs and adds tests.

cpmpy/solvers/ortools.py
```diff
@@ -470,14 +470,8 @@ def _post_constraint(self, cpm_expr, reifiable=False):
...
                     # catch tricky-to-find ortools limitation
-                    divisor = lhs.args[1]
-                    if not is_num(divisor):
-                        if divisor.lb <= 0 and divisor.ub >= 0:
-                            raise Exception(
-                                    f"Expression '{lhs}': or-tools does not accept a 'modulo' operation where '0' is "
-                                    f"in the domain of the divisor {divisor}:domain({divisor.lb}, {divisor.ub}). "
-                                    f"Even if you add a constraint that it can not be '0'. You MUST use a variable "
-                                    f"that is defined to be higher or lower than '0'.")
+                    if lhs.args[0].lb <= 0 or lhs.args[1].lb <= 0:
```
Wout4, line 473: First argument can be negative, just not the second

cpmpy/transformations/linearize.py
```diff
@@ -146,26 +152,69 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
...
+                            side_cons.append(abs_of_z < -y)
+                        # sign(x) = sign(z)
+                        lbx,ubx = get_bounds(x)
+                        if lbx >= 0:
+                            side_cons.append(rhs >= 0)
+                        elif ubx <= 0:
+                            side_cons.append(rhs <= 0)
+                        else: # x can be pos or neg
+                            x_is_pos, x_is_neg = cp.boolvar(), cp.boolvar()
+                            side_cons += [(~x_is_pos).implies(x <= 0), (~x_is_neg).implies(x >= 0),
```
Wout4, line 189: I think this doesn't totally define x_is_pos and x_is_neg. Can't we accomplish the same with only 1 boolvar and totally defining it?
_(~x_is_pos).implies(x <= 0),
(x_is_pos).implies(x >= 0),
x_is_pos.implies(rhs >= 0), 
(~x_is_pos).implies(rhs < 0)_

cpmpy/transformations/linearize.py
```diff
@@ -146,26 +152,69 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
...
+                        # k * y + z == x
+                        k = intvar(*get_bounds((x - rhs) // y))
+                        mult_res, side_cons = get_or_make_var(k * y)
                         cpm_expr = (mult_res + rhs) == x
+                        # |z| < |y|
+                        abs_of_z = cp.intvar(*get_bounds(abs(rhs)))
+                        if "abs" in supported:
+                            side_cons.append(abs(rhs) == abs_of_z)
+                        else: # need to linearize abs...
+                            z_is_pos = cp.boolvar()
```
Wout4, line 173: I like doing this here, in stead of calling the whole transformation stack

tests/test_constraints.py
```diff
@@ -78,8 +78,8 @@ def numexprs(solver):
             operator_args = [list(range(len(NUM_ARGS))), NUM_ARGS]
         elif name == "div" or name == "pow":
             operator_args = [NN_VAR,3]
-        elif name == "mod":
-            operator_args = [NN_VAR,POS_VAR]
+        elif solver == "ortools" and name == "mod":
+            operator_args = [POS_VAR, NN_VAR]
```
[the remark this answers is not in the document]
IgnaceBleukx: Hmm tests were failing when I made the first one negative I think... Will check again

cpmpy/transformations/linearize.py
```diff
@@ -257,6 +326,25 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
...
+    assert isinstance(lhs, GlobalFunction) and lhs.name == "abs", f"Expected abs expr but got {lhs}"
+
+    lb, ub = get_bounds(lhs)
+    if lb >= 0: # always positive
+        return [lhs == rhs]
+    if ub <= 0: # always negative
+        return [lhs == -rhs]
+
+    lhs_is_pos = cp.boolvar()
+    return [lhs_is_pos.implies(lhs >= 0), (~lhs_is_pos).implies(lhs < 0),
```
Wout4, line 345: lhs < 0 is not linear right? I think lhs <= 0 or lhs <= -1 are both correct here since 0 = -0 anyway.

cpmpy/transformations/linearize.py
IgnaceBleukx: Aha yes indeed! Adapted it.

cpmpy/transformations/linearize.py
IgnaceBleukx: Hmm you're right indeed `<` is not linear, it should not really matter here as we pass it through linearize either way. But when we allow it to be `<= 0` the `lhs_is_pos` is not fully defined...
I'll set it to `lhs <= -1`.

IgnaceBleukx:
I did a pass over the modulo and division code, I think the idea's behind the transformations are now a little more clear and they also avoid another pass through the transformation stack.

This PR ended up touching many more parts of the code then I expected, as tests kept on failing in various places...
I'll rename this PR accordingly ;)

Wout4:
Nice, starting to look good!
Only the z3 division thing with negative numbers remains a todo, will just disable it in the tests for now

IgnaceBleukx:
Yes, for z3 it's quite annoying still... For division it seems fixable, but modulo is another question... As far as I know z3 always rounds towards infinity...

Wout4:
Only way would be do decompose mod/div for z3 I think, but that's a big ask

IgnaceBleukx:
Turned out it was not that hard after all to get it right in the z3 interface : )
Needed some clever negation of the arguments/result to get it consistent, but I think it should be correct like this.

Ready to merge as far as I'm concerned.

cpmpy/solvers/z3.py
```diff
@@ -379,7 +379,12 @@ def _z3_expr(self, cpm_con):
                 elif cpm_con.name == "mul":
                     return x * y
                 elif cpm_con.name == "div":
-                    return x / y
+                    # z3 rounds towards negative infinity, need this hack when result is negative
+                    return z3.If(z3.And(x >= 0, y >= 0), x / y,
+                           z3.If(z3.And(x <= 0, y <= 0), -x / -y,
+                           z3.If(z3.And(x >= 0, y <= 0), -(x / -y),
+                           z3.If(z3.And(x <= 0, y >= 0), -(-x / y), 0))))
```
Wout4, line 386: ah damn that's nice!
Do we ever reach that 0 case?

Wout4:
can we do the same for the negative modulus in ortools? if it's negative negate it. I tried with adding an abs but that didn't immediately work out

cpmpy/solvers/z3.py
IgnaceBleukx: Normally no, as we have safened away the 0-s in the denominator

cpmpy/transformations/linearize.py
```diff
@@ -146,33 +157,108 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
...
+                        newlist.append(x + rhs == 0)
+                    else:
+                        lhs_is_pos = cp.boolvar()
+                        newcons = [lhs_is_pos.implies(x >= 0), (~lhs_is_pos).implies(x <= -1),
+                                   lhs_is_pos.implies(x == rhs), (~lhs_is_pos).implies(x + rhs == 0)]
+                        newlist += linearize_constraint(newcons, supported=supported, reified=reified)
+                    continue # all should be linear now
+
+
+            elif isinstance(lhs, GlobalFunction) and lhs.name not in supported:
```
Wout4, line 261: should this be globalconstraint? globalfunction does not inherit from globalconstraint anymore so maybe should be both?

Wout4: Ah i see it's still at the bottom of the file, it looks deleted here

PR #517, opened 2024-10-01
Ensure value of expressions is Python int

### LINKED ISSUE #484: int64 not json serializable
Some faciliyt location model, a CPMpy expression, when calling `exp.value()` on it, it returned an int64 (which was not serializable)

Why do we create non-standard types? Actually we don't, but numpy does somewhere, and we go along with it...

@kostis-init can provide the model

IgnaceBleukx:
Fixes #484

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -329,10 +329,11 @@ class Expression(object):
 
     def __pow__(self, other, modulo=None):
         assert (modulo is None), "Power operator: modulo not supported"
-        if other == 0:
-            return 1
-        elif other == 1:
-            return self
+        if is_num(other):
+            if other == 0:
+                return 1
+            if other == 1:
+                return self
         return Operator("pow", [self, other])
 
     def __rpow__(self, other, modulo=None):
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -541,7 +541,7 @@ class Operator(Expression):
         if any(a is None for a in arg_vals): return None
         # non-boolean
         elif self.name == "sum": return sum(arg_vals)
-        elif self.name == "wsum": return sum(arg_vals[0]*np.array(arg_vals[1]))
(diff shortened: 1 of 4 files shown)


cpmpy/expressions/core.py
```diff
@@ -329,10 +329,11 @@ def __rmod__(self, other):
 
     def __pow__(self, other, modulo=None):
         assert (modulo is None), "Power operator: modulo not supported"
-        if other == 0:
-            return 1
-        elif other == 1:
-            return self
+        if is_num(other):
```
Wout4, line 332: Why?

cpmpy/solvers/ortools.py
```diff
@@ -217,7 +217,9 @@ def solve(self, time_limit=None, assumptions=None, solution_callback=None, **kwa
 
             # translate objective
             if self.has_objective():
-                self.objective_value_ = self.ort_solver.ObjectiveValue()
+                ort_obj_val = self.ort_solver.ObjectiveValue()
+                assert int(ort_obj_val) == ort_obj_val, "Objective value should be integer, please report on github"
```
Wout4, line 221: Does ortools return numpy ints? Do we need this for other solvers as well?

cpmpy/expressions/core.py
IgnaceBleukx: The test `if other == 0` will always go through when `other` is an expression (will return a new expression). Alternatively we can do `if other == 0 is True` but I think this is more clear

cpmpy/solvers/ortools.py
IgnaceBleukx: No, it returns floats (but always `.0` of course as everything is int)

IgnaceBleukx: I checked for the other solvers, they all return `int` except for Gurobi, which is also patched now

cpmpy/expressions/core.py
Wout4: oh wauw that's a pretty serieus bug actually :o

PR #518, opened 2024-10-01
Testing incrementality

IgnaceBleukx:
Added some tricky cases for incremental solving.
Especially the use of an objective function proved to be tricky, as some solvers add the cuts during optimization as hard constraints to the solver, and let it run until it reports UNSAT internally.

Currently, there is no straightforward way to get around this restriction in Choco... Incremental solving without objective does work as expected.


Wout4:
How should we let users know not to use choco incementally with objective?

Wout4:
Oke the ellaborate error should be good enough ;)

PR #519, opened 2024-10-07
Cumulative decomposition fix

Dimosts:
Decomposition of cumulative did not allow nested expressions instead of variables, due to direct access of variables bounds.

Fixed by using get_bounds instead.

Also added a test for that


IgnaceBleukx (approved):
Fairly trivial. Good to merge for me

PR #523, opened 2024-10-24
Clear values when UNSAT

IgnaceBleukx:
There was some inconcistency regarding what to do when the solver found UNSAT.
Z3 already cleared the values of all known variables, which I think made sense.

Extended this behaviour to all other solvers too and updated documentation

### DIFF
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -175,6 +175,9 @@ class CPM_choco(SolverInterface):
             # translate objective
             if self.has_objective():
                 self.objective_value_ = sol.get_int_val(self.solver_var(self.obj))
+        else: # clear values of variables
+            for cpm_var in self.user_vars:
+                cpm_var.clear()
 
         return has_sol
 
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -215,6 +215,11 @@ class CPM_choco(SolverInterface):
         self.cpm_status = SolverStatus(self.name)
         self.cpm_status.runtime = end - start
 
+        # if no solutions, clear values of variables
+        if len(sols) == 0:
+            for var in self.user_vars:
+                var.clear()
+
         # display if needed
         if display is not None:
             for sol in sols:
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -177,7 +177,7 @@ class CPM_choco(SolverInterface):
                 self.objective_value_ = sol.get_int_val(self.solver_var(self.obj))
         else: # clear values of variables
             for cpm_var in self.user_vars:
-                cpm_var.clear()
+                cpm_var._value = None
 
         return has_sol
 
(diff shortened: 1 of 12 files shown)


cpmpy/solvers/choco.py
```diff
@@ -175,6 +175,9 @@ def solve(self, time_limit=None, **kwargs):
             # translate objective
             if self.has_objective():
                 self.objective_value_ = sol.get_int_val(self.solver_var(self.obj))
+        else: # clear values of variables
+            for cpm_var in self.user_vars:
+                cpm_var.clear()
```
Dimosts, line 180: I guess we also have to put None to objective value??

Also, it is possible that vars have already value, and then the user changes something and uses solveAll(). So, in that case we also need to empty them? As I see in the solveAll() code, it normally has the value of the last solution found...

cpmpy/solvers/exact.py
```diff
@@ -127,7 +127,7 @@ def _fillVars(self):
         if not self.xct_solver.hasSolution():
             self.objective_value_ = None
```
Dimosts, line 128: yes, here we do empty the objective value, so we should do it also in all solvers

cpmpy/solvers/z3.py
```diff
@@ -190,9 +190,9 @@ def solve(self, time_limit=None, assumptions=[], **kwargs):
                 obj = self.z3_solver.objectives()[0]
                 self.objective_value_ = sol.evaluate(obj).as_long()
 
-        else:
+        else:  # clear values of variables
             for cpm_var in self.user_vars:
-                cpm_var._value = None # XXX, maybe all solvers should do this...
```
Dimosts, line 195: maybe yes, indeed :)

tests/test_solvers.py
```diff
@@ -753,6 +753,21 @@ def test_incremental_objective(self):
...
+
+        for name, cls in cp.SolverLookup.base_solvers():
+            if cls.supported() is False: # solver not supported
+                continue
+            self.assertTrue(sat_model.solve(solver=name))
+            for v in (x,y,z):
+                self.assertIsNotNone(v.value())
+            self.assertFalse(unsat_model.solve(solver=name))
+            for v in (x,y,z):
+                self.assertIsNone(v.value())
```
Dimosts, line 771: tests also with solveAll() and objective after they are done?

Dimosts (changes requested):
Good, indeed needed! 

My comments mainly have to do that we have to be consistent also with objective value (we should clear that too) and also when we do solveAll() and the count is 0, we should again clear the values.

cpmpy/solvers/exact.py
IgnaceBleukx: It is actually already there for all solvers, right before the translation of the solution

tias (commented):
Yes, but why pay the cost of an extra 'clear()' lookup and function call?

A few lines above each added for loop, we already directly access ._value; I think we better set v._value = None given that the code above already assumes you know it exists, and then it is more lightweight...

PR #528, opened 2024-10-31
Better naive grow

IgnaceBleukx:
Small change to `mcs_naive` to support _some_ incrementality.

If the solver at hand (e.g, Gurobi) does not support assumption variables, but is incremental, this addition will keep the state of the solver as long the result is SAT. 
Solver is reset to scratch at each failure (before it was reset at each solve call)

### DIFF
--- a/cpmpy/tools/explain/mss.py
+++ b/cpmpy/tools/explain/mss.py
@@ -82,13 +82,15 @@ def mss_grow_naive(soft, hard=[], solver="ortools"):
 
     to_check = list(soft)
     sat_subset = []
+    solver = cp.SolverLookup.get(solver, cp.Model(hard))
 
     while len(to_check):
         c = to_check.pop()
-        if cp.Model(sat_subset + [c] + hard).solve(solver=solver) is True:
-            sat_subset += [c]
+        solver += c
+        if solver.solve() is True:
+            sat_subset.append(c)
         else:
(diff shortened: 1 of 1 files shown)


tias (approved):
Yes, all good.

The SolverLookup with Model creation does look a bit weird... You don't need to create an explicit model object is all you want to do is add constraints to it; e.g.
s = cp.SolLookup.get(solvername)
s += hard  
avoids creating that throwaway model object

cpmpy/tools/explain/mss.py
```diff
@@ -82,13 +87,17 @@ def mss_grow_naive(soft, hard=[], solver="ortools"):
...
-            sat_subset += [c]
+        s += c
+        if s.solve() is True:
+            sat_subset.append(c)
         else:
-            # UNSAT, cannot add to sat subset
-            pass
+            # UNSAT, cannot add to sat subset, reset solver to just sat subset
+            s = cp.SolverLookup.get(solver)
+            s += hard
```
tias, line 101: sorry to re-open, but should be hard + sat_subset?

tias (commented):
correctness to check

cpmpy/tools/explain/mss.py
IgnaceBleukx: Yes, you're right, I'll push the fix to main - pretty surprising the tests did not catch this...

PR #529, opened 2024-11-06
Reals in objective function

IgnaceBleukx:
It seems #517 was merged a little too eager.
In newer versions of OR-Tools, you can actually use a real coefficient in the objective function, which was prohibited by my earlier fix.

Updated tests accordingly and made some small changes to Gurobi to make things more streamlined

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -542,7 +542,7 @@ class Operator(Expression):
         if any(a is None for a in arg_vals): return None
         # non-boolean
         elif self.name == "sum": return sum(arg_vals)
-        elif self.name == "wsum": return int(sum(arg_vals[0]*np.array(arg_vals[1])))
+        elif self.name == "wsum": return np.dot(arg_vals[0], arg_vals[1]).item()
         elif self.name == "mul": return arg_vals[0] * arg_vals[1]
         elif self.name == "sub": return arg_vals[0] - arg_vals[1]
         elif self.name == "mod": return arg_vals[0] % arg_vals[1]
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -542,7 +542,11 @@ class Operator(Expression):
         if any(a is None for a in arg_vals): return None
         # non-boolean
         elif self.name == "sum": return sum(arg_vals)
-        elif self.name == "wsum": return np.dot(arg_vals[0], arg_vals[1]).item()
+        elif self.name == "wsum": 
+            val = np.dot(arg_vals[0], arg_vals[1]).item()
+            if round(val) == val: # it is an integer
+                return int(val)
+            return val # can be a float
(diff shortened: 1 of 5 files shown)


cpmpy/expressions/core.py
```diff
@@ -542,7 +542,7 @@ def value(self):
         if any(a is None for a in arg_vals): return None
         # non-boolean
         elif self.name == "sum": return sum(arg_vals)
-        elif self.name == "wsum": return int(sum(arg_vals[0]*np.array(arg_vals[1])))
+        elif self.name == "wsum": return np.dot(arg_vals[0], arg_vals[1]).item()
```
Wout4, line 545: so this will return a float (even if it equals an int exactly), but in the solver code you do convert to an int in those cases, just a small inconsistency that I don't know what to do with.

cpmpy/solvers/gurobi.py
```diff
@@ -181,10 +181,10 @@ def solve(self, time_limit=None, solution_callback=None, **kwargs):
             # set _objective_value
             if self.has_objective():
                 grb_obj_val = grb_objective.getValue()
-                if grb_obj_val != int(grb_obj_val):
-                    self.objective_value_ =  grb_obj_val # can happen with DirectVar using floats
-                else:
+                if round(grb_obj_val) == grb_obj_val: # it is an integer?:
                     self.objective_value_ = int(grb_obj_val)
```
Wout4, line 185: do we need this? this example shows the inconsistency between our value function and objective value calculation:
m = Model()
m += 2 * x + 3 * y == 2
m.maximize(5*x + 0.7*y)
m.solve()
s = SolverLookup.get('ortools', m)
print(m.objective_value())
s.solve()
print(s.objective_value())

This happens because m.objective_value() just does m.objective_.value(), it does not know about the converted objective value in s.

Wout4 (commented):
Just one small thing I noticed, not sure if it's better to convert everything to ints or if it's fine to return floats as objective values even when they have no decimal part. (did the chatbot struggle with that or something?)

cpmpy/expressions/core.py
IgnaceBleukx: Aha, yes ok dan we need to convert it here too.
So will add a "round" check as I did in the objective translation. The "item" is apparently a good way of getting the Python value instead of the numpy value

cpmpy/solvers/gurobi.py
IgnaceBleukx: This should be fixed now with the extra check.

IgnaceBleukx: If we do not return an integer here, you sometimes get quite confusing errors afterwards.
For example, if I want to make a model UNSAT by restriciting the objective value

```python
s = cp.SolverLookup.get("gurobi")
obj = ...
s.minimize(obj)
assert s.solve()
s += obj.value() < s.objective_value()
assert s.solve() is False # will throw an error if objective_value() returns a float
```

PR #530, opened 2024-11-08
Propose some edits to the docs

hbierlee:
Just what I found while reading through the guides. Mostly edits concerning grammar/spelling, some pedantic, some a bit subjective (feel free to reject if you disagree), and not too much content-wise. This will change the modeling to modelling page link, so might break some user bookmarks.


docs/unsat_core_extraction.md
```diff
@@ -2,7 +2,7 @@
 
 When a model is unsatisfiable, it can be desirable to get a better idea of which Boolean variables make it unsatisfiable. Commonly, these Boolean variables are 'switches' that turn constraints on, hence such Boolean variables can be used to get a better idea of which _constraints_ make the model unsatisfiable.
 
-In the SATisfiability literature, the Boolean variables of interests are called _assumption_ variables and the solver will assume they are true. The subset of these variables that, when true, make the model unsatisfiable is called an unsatisfiable _core_.
+In the satisfiability literature, the Boolean variables of interests are called _assumption variables_ and the solver will assume they are true. The subset of these variables that, when true, makes the model unsatisfiable is called an unsatisfiable _core_.
```
IgnaceBleukx, line 5: Technically, they are literals right?

IgnaceBleukx (commented):
Great, looks good

docs/unsat_core_extraction.md
hbierlee: Yes, I guess so, especially since we're saying the assumptions are "true", we should call them literals. IPASIR2 also calls them assumption literals.

hbierlee:
I've changed it now so that this section is technically correct (and I think we should call them "assumptions" so they can be literals or even e.g. integers in the future). Let me know if it's worth it to change this throughout the rest of the doc/code.

docs/unsat_core_extraction.md
hbierlee: I've changed it now so that this section is technically correct (and I think we should call them "assumptions" so they can be literals or even e.g. integers in the future). Let me know if it's worth it to change this throughout the rest of the doc/code.

tias: as you wish

IgnaceBleukx:
Can you merge `orgin/master` again on this branch? It's on a local fork on your account so I can't do it myself : )

PR #532, opened 2024-11-18
Has subexpr optimisation

Wout4:


cpmpy/expressions/variables.py
```diff
@@ -263,11 +266,23 @@ def __init__(self, lb, ub, name):
...
     def is_bool(self):
         """ is it a Boolean (return type) Operator?
         """
         return False
 
+    def is_leaf(self):
+        """ Is it the leaf of an expression tree?
+            This is only the case for decision variables (and constants).
+        """
+        return True
```
tias, line 284: we can build an cpm_array of expressions... I don't think it is generally true to be a leaf?

cpmpy/expressions/core.py
```diff
@@ -137,12 +159,44 @@ def __repr__(self):
...
+
+        return self._has_subexpr
+
     def is_bool(self):
         """ is it a Boolean (return type) Operator?
             Default: yes
         """
         return True
 
+    def is_leaf(self):
```
tias, line 194: since we have `has_subexpr()` maybe we should name it `is_leafexpr()` because 'leaf' as a word is not very descriptive

cpmpy/expressions/core.py
```diff
@@ -107,7 +107,29 @@ def __init__(self, name, arg_list):
...
-        self.args = arg_list
+        self._args = arg_list
+
+
+    @property
+    def args(self):
+        return self._args
+
+    @property
+    def vars(self):
```
tias, line 118: vars() returns _args?

cpmpy/expressions/core.py
tias: actually, what `Expressions` are leafs? Only variables? so should be `is_var()`?

cpmpy/expressions/core.py
```diff
@@ -137,12 +159,44 @@ def __repr__(self):
...
+                    if not el.is_leaf():  # NDVarArrays are Expr and any_list, so they are covered too (allows early-exit for decision var arrays)
+                        return True
+                elif is_any_list(el) and recursive_has_subexpr(el): # recursively call on list-like
+                    return True
+            return False
+
+        # If not an Expression (e.g. list-like) or _has_subexpr has not been computed before / has been reset
+        if not hasattr(self, '_has_subexpr'):
+            # args can have lists of lists... -> need for recursive implementation
+            self._has_subexpr = recursive_has_subexpr(self.args)
```
tias, line 184: hmm, I'm hesitant. I think the 'hot path' should not call a recursive function, just do the loop here.

Then, if an arg is a 'any_list', actually there is one optimisation: if its an NDVarArray and type != object, you don't even need to walk it.
Otherwise... I guess a recursive function, or you do it with a stack, just keep popping till its emtpy (and add to it if you encounter a list)

it might not actually have an impact on efficiency, so still hesitation from my side ; )

tias (commented):
intermediate review

cpmpy/expressions/core.py
Wout4: Ah I forgot some parts that use this, this is the default value. Only get's overwritten in tables for now, because there are a lot of constants in tables. makes get_variables a lot faster. Maybe we can do it for weighted sums as well

cpmpy/expressions/core.py
Wout4: or a lists of vars/constants yes

cpmpy/expressions/variables.py
Wout4: this is another part of the code, not ndvararray

Wout4:
fuzztester did not find any bugs on this branch so far

cpmpy/transformations/flatten_model.py
```diff
@@ -167,7 +165,7 @@ def flatten_constraint(expr):
                 if expr.args[1].name == 'and':
                     a1s = expr.args[1].args
                     a0 = expr.args[0]
-                    newlist.extend(flatten_constraint([a0.implies(a1) for a1 in a1s]))
+                    lst_of_expr.extend([a0.implies(a1) for a1 in a1s])
```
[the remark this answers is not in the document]
Wout4: newlist is the output, lst_of_expr is stack to be flattened, so we are adding to the stack in stead of recursing. Maybe not relevant to this pr..

tias:
That's a nice speedup...

To my surprise without isleaf a bit slower, but neglegible. 

Did a final review, I think there is one unintended merge/line change that is wrong, rest looks good.


BTW, 117 seconds? how many test instances do you have???
If I run it in the fuzz repo, I have ~1.02 seconds on master and ~1.06 seconds with has_subexpr; so a small slowdown (surprisingly?) but acceptable... very far from 117 seconds though (will open a comment in the fuzz repo)

Wout4:
I regenerated instances, because they now have _args in stead of args, there are 4800 instances and almost 6000 constraints. The additional import might make the version without is_leaf slightly slower but not sure.

cpmpy/transformations/flatten_model.py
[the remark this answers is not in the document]
Wout4: i moved this to a different pr, still think we prefer not to recurse, right?

Wout4:
Another thing to keep in mind: if we ever introduce a new expression that is a leaf, we now have to change the has_subexpr function and add it to that check. This is a bit ugly but also probably won't happen.
(maybe something like a table wildcard, ..)

PR #533, opened 2024-11-21
Blocks world

IgnaceBleukx:
Added the "blocks world" problem as a planning problem example.

It unveiled an interesting shortcoming in the 2-d element constraint which is also fixed in this PR.


cpmpy/expressions/variables.py
```diff
@@ -441,10 +441,23 @@ def __getitem__(self, index):
...
+            if len(index) != self.ndim:
+                raise NotImplementedError("CPMpy does not support returning an array from an Element constraint. Provide an index for each dimension. If you really need this, please report on github.")
+
             # find dimension of expression in index
-            expr_dim = next(dim for dim,idx in enumerate(index) if isinstance(idx, Expression))
-            arr = self[tuple(index[:expr_dim])] # select remaining dimensions
-            index = index[expr_dim:]
+            expr_dim = [dim for dim,idx in enumerate(index) if isinstance(idx, Expression)]
+            if len(expr_dim) == 1: # optimization, only 1 expression, reshape to 1d-element
+                # TODO can we do the same for more than one Expression? Not sure...
```
Wout4, line 451: should be possible to reduce the amount of dimensions by the amount of constants in index, but might be a bit of a braintwister to implement, already found this code hard to understand.

Wout4 (commented):
I will attempt the todo in another pr, I guess it's even more niche to have multidimensional with multiple constants as well

cpmpy/expressions/variables.py
IgnaceBleukx: Yes, we implemented it already but without altering the order of dimensions in the variables.
So `arr[1,x,y]` gets converted to `arr[1][x,y]` but `arr[x,y,1]` does not get changed

IgnaceBleukx:
Updated the example with a real "planning as sat" approach and some more doc tweaks

examples/blocks_world.py
```diff
@@ -0,0 +1,132 @@
...
+    towers = []
+    for next_idx in range(len(state)):
+        if next_idx in state: continue
+        tower = [next_idx + 1]
+        while next_idx >= 0:
+            x = state[next_idx]
+            tower.append(int(x + 1))
+            next_idx = x
+        towers.append(tower[:-1])
+    return sorted(towers, key=lambda x: x[-1])
```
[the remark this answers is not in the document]
Wout4: yeah i guess it's fine like this :)

examples/blocks_world.py
```diff
@@ -0,0 +1,242 @@
...
+
+    This file showcases two ways of solving the problem.
+    The first model is an optimization model with a heuristic upper bound on the number of moves as initial bound.
+    The second appraoch is a "planning as SAT" approach where the model is solved iteratively with an increasing horizon
+
+    Both models are inspired by the MiniZinc model:
+        https://github.com/MiniZinc/mzn-challenge/blob/develop/2022/blocks-world/blocks.mzn
+"""
+
+def get_model(start, goal, n_piles=None):
```
Wout4, line 17: n_piles can't be none right?

Wout4:
Added some code to stop printing after the final state. Fine to be merged for me, unless you still want to add to it?

IgnaceBleukx:
Thanks, ok to leave it open for a bit, I'm generating some instances to check if any bugs occur in the model

PR #534, opened 2024-11-22
Multi dim element optimisation

Wout4:
Building on Ignaces idea, but for multiple constant dimensions.


tias:
Is this related to the following?

```
A = cp.cpm_array([[1,2,3],[7,8,9]])
iv = cp.intvar(0,1)
[A[iv, j] for j in range(A.shape[1])]  # correct: [[1 2 3 7 8 9][(IV267) * 3], [1 2 3 7 8 9][1 + ((IV267) * 3)], [1 2 3 7 8 9][2 + ((IV267) * 3)]]
A[iv,:]  # AttributeError: 'slice' object has no attribute 'name'
A[iv]  # does not seem to check if more dimensions: [1 2 3 7 8 9][IV269]
```

which students stumbled on in car sequencing, e.g.
```[option[s, o] == requires[sequence[s], o] for o in range(n_options)]```
instead of vectorized ```option[s] == requires[sequence[s]]```

**Comment by Wout4** [comment]
the first one we optimise here:
`[A[iv, j] for j in range(A.shape[1])] #[[1 7][IV3135], [2 8][IV3135], [3 9][IV3135]]`
The others aren't really affected, but we can adress them here.
`A[iv]` should just throw an error since the dimensions don't match
`A[iv,:]` means the same in python terms(?) so I don't expect this to work. what would be the expected output of this?

**Comment by IgnaceBleukx** [comment]
Ok so I took the opportunity to also fix #444 in this PR.
In summary: the `Element` global can now take n-ary indices, saved as arguments after the array.
We can **safely** convert this multi-dimensional element to a single-dim element (as we did before) when the bounds of the indices are tight (within the dimension of the arr).

For this I made a helper function, which massively cleans up the code in the __getitem__ of NDVarArray.

Main downside in my opinion is that we now always have to safen Element at the toplevel, not sure what this means performance-wise. I actually think it should be fine, but maybe we should test this.

File: cpmpy/expressions/globalfunctions.py [hunk 1]
```diff
@@ -229,30 +232,51 @@ class Element(GlobalFunction):
...
         a 'numeric' global constraint.
     """
 
-    def __init__(self, arr, idx):
-        if is_boolexpr(idx):
-            raise TypeError("index cannot be a boolean expression: {}".format(idx))
-        if is_any_list(idx):
-            raise TypeError("For using multiple dimensions in the Element constraint, use comma-separated indices")
-        super().__init__("element", [arr, idx])
+    def __init__(self, arr, *idx):
```
tias, line 242: I'm not sure I like idx being a *idx implicit array...

my impression was that numpy is more natural in it, that an index can also be a tuple instead of a single value; and in case of a tuple, the index is the tuple, not that there are multiple indices.

So then here, and in all below, it would be (self, arr, idx) and still super().ini("el", [arr, idx]) etc...?

cpmpy/solvers/ortools.py
```diff
@@ -355,7 +355,7 @@ def transform(self, cpm_expr):
         """
         cpm_cons = toplevel_list(cpm_expr)
         supported = {"min", "max", "abs", "element", "alldifferent", "xor", "table", "negative_table", "cumulative", "circuit", "inverse", "no_overlap"}
-        cpm_cons = no_partial_functions(cpm_cons, safen_toplevel=frozenset({"div", "mod"})) # before decompose, assumes total decomposition for partial functions
+        cpm_cons = no_partial_functions(cpm_cons, safen_toplevel=frozenset({"div", "mod", "element"})) # before decompose, assumes total decomposition for partial functions
```
tias, line 358: right, so this would be for all elements... even the 1d ones, where we know it is often not totally formulated, but intended to be total by the user... I see

cpmpy/solvers/ortools.py
```diff
@@ -464,8 +464,15 @@ def _post_constraint(self, cpm_expr, reifiable=False):
                     return self.ort_model.AddDivisionEquality(ortrhs, *self.solver_vars(lhs.args))
                 elif lhs.name == 'element':
                     # arr[idx]==rvar (arr=arg0,idx=arg1), ort: (idx,arr,target)
-                    return self.ort_model.AddElement(self.solver_var(lhs.args[1]),
-                                                     self.solver_vars(lhs.args[0]), ortrhs)
+                    arr, *idx = lhs.args
+                    if len(idx) > 1: # multi-dim, convert here
```
tias, line 468: can this reuse the `to_1d_element` function, so it detects that it requires higher D, which is not natively supported, so it calls to_1d, and then continues to post this?

tias (commented):
So this is not in the low hanging fruit basket, but still interesting to make progress on afterwards perhaps.

After having gone through the patch, I like it. But it does feel that it is different then normal Element, because of it needing to be total, because of solvers needing to re-detect that its higher dim.

Instead of having both cases in Element, would there be a way at creation time to either create an Element (strictly 1D), or to create an ElementND(). Then this second one can be in the 'no partial' list, and its decomposition would be the 'to 1d function' which would automatically be called for all solvers that dont have a native ND Element?

IgnaceBleukx:
Yes, so indeed it seems like we should make a `ElementMultiD` global constraint indeed. The `__getitem__` can till be used as a dispatcher to automatically select which one to create, after it has removed all dimensions where the user provided a constant anyway.

I'll get to work on that : )

Issue #535, opened 2024-11-25
Min/Max on iterable of length 1


ThomSerg:
When using `cp.max(...)` / `cp.min(...)` on an iterable of length 1 with no decision variables (e.g. `cp.max([1, ])` ), a `TypeError` gets thrown: `'int' object is not iterable`.

CPMpy checks that the list does not have any CPMpy-expressions and if so, forwards the operation to Python's build-in. In the current implementation, a broadcast operator gets used:

```python
if ... :
    builtins.min(*iterable, **kwargs)
```

This causes an issue with the given example: an iterable of length 1. `builtins.min(*[1, ])` becomes `builtins.min(1)`, resulting in the error that an integer is not iterable. This  deviates from Python's standard behaviour and thus causes an inconsistency especially when using `from cpmpy import *` to overload some of the build-ins.

A solution would be to just remove the broadcast operator, since Python's min can directly accept a single iterable as argument, i.e. `builtins.min(*[1, 2]) = builtins.min(1, 2)` is equivalent to `builtins.min([1, 2])`

PR #536, opened 2024-11-26
Remove broadcast in min/max when forwarding to builtins

ThomSerg:
Resolves issue #535 where CPMpy's min/max is inconsistent with Python's built-in min/max on iterators with single (non-cpmpy) element. Old implementation with broadcast operator doesn't take this edge-case into account and causes an exception. The builtin works perfectly fine with handling the iterator directly, no need for broadcasting.


Wout4:
Thanks, good catch!

PR #538, opened 2024-12-03
Linearize power

Dimosts:
Linearizing power.

only `POW(b,n) == IV` supported, with n being an integer, post as b*b*...*b (n times) == IV

### DIFF
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -124,7 +124,7 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
                         raise NotImplementedError("Cannot linearize power without multiplication")
                     # only `POW(b,n) == IV` supported, with n being an integer, post as b*b*...*b (n times) == IV
                     x, n = lhs.args
-                    new_lhs = x
+                    new_lhs = 1
                     for exp in range(n):
                         new_lhs, new_cons = get_or_make_var(x * new_lhs)
                         newlist.extend(new_cons) # TODO: when reified, this will make again new vars - should post directly toplevel instead
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -122,6 +122,8 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
                 elif lhs.name == "pow" and "pow" not in supported:
                     if "mul" not in supported:
                         raise NotImplementedError("Cannot linearize power without multiplication")
+                    if not is_num(lhs.args[1]):
+                        raise NotImplementedError("Cannot linearize power with ")
                     # only `POW(b,n) == IV` supported, with n being an integer, post as b*b*...*b (n times) == IV
                     x, n = lhs.args
                     new_lhs = 1
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -119,6 +119,18 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
                     lhs = Operator("wsum",[[lhs.args[0]], [lhs.args[1]]])
                     cpm_expr = eval_comparison(cpm_expr.name, lhs, rhs)
 
+                elif lhs.name == "pow" and "pow" not in supported:
+                    if "mul" not in supported:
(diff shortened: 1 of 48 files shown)


tias (commented):
Nice.

Maybe add/generate a few testcases for this?

IgnaceBleukx:
Yes, you _should_ be able to just enable the testcase for Exact/Gurobi in `tests/test_constraints.py` using the `EXCLUDE_OPERATORS`  dict

Wout4:
I don't think it's that easy
First of all, this outputs non-flat constraints for n > 2

secondly gurobi supports power for n = 2, but not for n > 2. Right now we just say gurobi supports pow but then throw an exception when n > 2. So we want to linearize it when n > 2. Not sure how to do this with our supported argument (unless we add a quadratic thing).

IgnaceBleukx:
I think it's ok to only support powers of 2 for now as it is the most common case probably. (Still need to check for it of course).

In the future, we can make intermediate result vars:
IV0 == b * b
IV1 == IV0 * b
... 

But that might be tricky as these aux vars have to be assigned toplevel, but linearize has no knowing of toplevel/nestedness...

IgnaceBleukx:
Hmm actually indeed it might be fine as the multiplication is functional - let's just make sure to write some extra tests for it ;) 
So then the question just remains if we support arbitrary powers, or only ^2.

IgnaceBleukx:
Updated this, should be ready for review @Wout4

Wout4:
I would still add somewhere that we only support integers as the exponent, maybe the solver interface or the power constructor? right now ortools will interpret any expression as being an exponent of 2 since we check exp == 2 (which will make a truthy expression if exp is an expression)

cpmpy/transformations/linearize.py
```diff
@@ -119,6 +119,21 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
...
+                    if "mul" not in supported:
+                        raise NotImplementedError("Cannot linearize power without multiplication")
+                    if not is_num(lhs.args[1]):
+                        raise NotImplementedError("Cannot linearize power with ")
+                    # only `POW(b,n) == IV` supported, with n being an integer, post as b*b*...*b (n times) == IV
+                    x, n = lhs.args
+                    new_lhs = 1
+                    for exp in range(n):
+                        new_lhs, new_cons = get_or_make_var(x * new_lhs)
+                        newlist.extend(new_cons) # TODO: when reified, this will make again new vars - should post directly toplevel instead
```
Wout4, line 132: what can of reification can make it to linearization? we already flattened?

Wout4:
I just realised that we are putting constraints in canonical form during the linearisation, so should we do this also for our auxiliary constraints? (x * x = IV0 should become a weighted sum 1* (x*x) -1*(IV0) == 0) since this is the expected output for linearize?

cpmpy/transformations/linearize.py
IgnaceBleukx: Hmm good point, Indeed the TODO will not actually occur in practice as we first do `reify_rewrite` in the interfaces.
I was thinking of the case where `b -> x ** 3` is first linearized, and then the linearized power is put toplevel if the power is not supported in a reifiied context, but that won't be the case in practice :)

IgnaceBleukx:
Nono, we support non-canonical stuff in the linear format (output of linearize):
https://github.com/CPMpy/cpmpy/blob/73d287426238b94b72e3e5cc50435c30493b6161/cpmpy/transformations/linearize.py#L34

Where the GenExpr in this case would be `x * x` which is supported by the solver.

IgnaceBleukx:
I think this is ready to merge now.

Wout4:
looks like in the safening branch we set division in z3 to be real(lhs)/real(rhs) in stead of just lhs/rhs, do you remember why? we want integer division, not real division right

IgnaceBleukx:
I believe it was due to division in Z3 being underspecified:
https://microsoft.github.io/z3guide/docs/theories/Arithmetic/

But indeed, the to-real fix might have been because at first I did not call safening in z3. Fine to revert it indeed.

At some point we should re-check that all solvers are properly doing integer division, with rounding towards zero.

Wout4:
I quickly checked an z3 is also rounding towards zero now, maybe I misremember that they rounded down, or something changed.

PR #539, opened 2024-12-03
Improve exception messages and remove unused imports

Dimosts:
Improve exception messages to be more precise, mostly mentioning what expression they refer to. That makes debugging easier.

Also, remove unused imports.


tias (approved):
that's very thorough, no reason to delay merging

PR #541, opened 2024-12-03
Only implied speedup

tias:
does fail on one test:
FAILED tests/test_solvers.py::TestSolvers::test_incremental_objective - AssertionError: 0 != 5
===================================================== 1 failed, 20130 passed, 8 skipped, 32 warnings in 1261.41s (0:21:01) =====================================================

I hope Wout you can have a look at that, not sure it is from this branch?


cpmpy/transformations/reification.py
```diff
@@ -88,15 +87,14 @@ def only_implies(constraints):
                 newcons.append(cpm_expr)
             else:
                 # BVar1 == BE0 :: ~BVar1 -> ~BE0, BVar1 -> BE0
-                newexprs = ((~a0).implies(recurse_negation(a1)), a0.implies(a1))
-                if isinstance(a1, GlobalConstraint):
-                    newcons.extend(newexprs)
```
Wout4, line 93: is it not worth skipping the retransform in the case of bv -> global ?

tias: it felt dangerous that if the 'original' a1 is global, that the `recurse_negation(a1)` in newexprs would not need further transformations?
it sounds like there must be cases where this is would happen...

or is there some implicit assumption that decompose_globals would have decomposed it? but still... an implication on a negated global?

cpmpy/transformations/reification.py
```diff
@@ -77,7 +77,7 @@ def only_implies(constraints):
         # Comparisons: transform bV == BE
         elif cpm_expr.name == '==' and cpm_expr.args[0].is_bool():
             a0,a1 = cpm_expr.args
-            if isinstance(a0, _BoolVarImpl) and isinstance(a1, _BoolVarImpl):
+            if isinstance(a1, _BoolVarImpl):
```
tias, line 80: for completeness, maybe add  # assumption: a0 is a BoolVar
is it always? or did you derive it from the comparison?

cpmpy/transformations/reification.py
Wout4: push down negation does not decompose, it just returns ~global(args) (after pushing down negation in the args)
So even calling push down negation is not required in the case of a global.

I guess in practise we won't encounter a global here because indeed the global constraint would have to be supported in reified context to make it this far without being decomposed.
Only choco supports reified globals (at the moment), and it does not us only_implies

cpmpy/transformations/reification.py
Wout4: well the prerequisite is only_bv_reifies, so any reification should have a a bv there, will add comment

Wout4:
fine to merge for me, but maybe look at the comments again to see if you agree :)

cpmpy/transformations/flatten_model.py
```diff
@@ -163,7 +163,7 @@ def flatten_constraint(expr):
                 if expr.args[1].name == 'and':
                     a1s = expr.args[1].args
                     a0 = expr.args[0]
-                    newlist.extend(flatten_constraint([a0.implies(a1) for a1 in a1s]))
+                    lst_of_expr.extend([a0.implies(a1) for a1 in a1s])
```
tias, line 166: can you take this 'modify list you are iterating over' one out? its in a separate PR

tias (commented):
agreed (if you remove that the change to flatten_model)

Issue #512, opened 2024-08-09
Overhead in solver interface


IgnaceBleukx:
When doing many "easy" solve calls, the overhead of CPMpy becomes very significant.
We should try to minimize this to some extend as it can slow down many applications relying on repeated solving, such as MUS extraction/enumeration!

The following code snippet gives an idea of the extend of the problem:

```python
import cpmpy as cp
from time import time

n = 100
iterations = 10000

bvars = cp.boolvar(name="bv", shape=n)    

import random
random.seed(0)

solver = cp.SolverLookup.get("pysat")
solver += cp.sum(bvars) >= 0

solve_time = 0
t0 = time()

for i in range(iterations):
    print(f"\rAt iteration {i+1}", end="\t")
    if solver.solve() is False:
        break
    solve_time += solver.status().runtime
    # do something with the assignment
    bvars.value().sum()
    # add new constraint
    solver += cp.any([bv if random.random() >= 0.5 else ~bv for bv in bvars])
print()    
    
print(f"Did {i+1} iterations")
print("Total time spent in solving:", solve_time)
print("Total time:", time()-t0)
```
| Solver | Solve time | Total time |
| -------| -----------| ----------|
| OR-Tools | 990.01s | 1192.569s |
| Pysat |  **0.08s** | 17.94s |
| Gurobi |  38.45s | 126.27s |
| Z3 | 275.06s | 515.65s |

Attached are some of the profiler runs for the above script
[profile.zip](https://github.com/user-attachments/files/16559074/profile.zip)

tias:
Here is a slightly updated version that does the profile and ranks by time (excluding time in subfunctions):
```
import cpmpy as cp
from time import time

n = 100
iterations = 10000

bvars = cp.boolvar(name="bv", shape=n)    

import random
random.seed(0)

solver = cp.SolverLookup.get("pysat")
solver += cp.sum(bvars) >= 0

solve_time = 0
t0 = time()

for i in range(iterations):
    print(f"\rAt iteration {i+1}", end="\t")
    if solver.solve() is False:
        break
    solve_time += solver.status().runtime
    # do something with the assignment
    bvars.value().sum()
    # add new constraint
    solver += cp.any([bv if random.random() >= 0.5 else ~bv for bv in bvars])
print()    
    
print(f"Did {i+1} iterations")
print("Total time spent in solving:", solve_time)
print("Total time:", time()-t0)
```

[...]

PR #542, opened 2024-12-04
remove inline imports

tias:
one example here: inline Element import in var.getitem() caused 1.5 second delay in #512 

we should go over the entire codebase (including solvers, I think I did it there too) and remove all inline imports...


Wout4:
I think @ThomSerg has gone over most of these?

PR #543, opened 2024-12-05
Remove recursion in flatten

Wout4:
in stead just add constraints to the list we are looping over


Wout4:
Extracted this from has_subexpr, since its a separate thing

tias:
I think something went wrong in the merge... it now just shows 2 line changes

tias:
ah no yes, that top line is the line : )

why do that only for this line?
Plus, I think it is dangerous, `for expr in lst_of_expr:`, and changing that lst_of_expr in the loop... that is not a good coding pattern.
If we want to do that, we should do `stack = lst_of_expr; while len(stack) != 0: ...` then you can add stuff to the stack

Wout4:
I guess we are fine with keeping the recursion then, maybe we did it in competition branch because of possibly hitting the recursion limit

Issue #544, opened 2024-12-05
Inconsistent model/solver status


ThomSerg:
Calling `.status()` on a CPMpy `Model` or an implementation of `SolverInterface` (like `CPM_ortools`) returns a `SolverStatus` object with two fields: `runtime` and `exitstatus`. This status object is not consistent across solver backends:

|           | Solve()  |  Solve()  | SolveAll()                                                     |
| -------- | -------- | --------- | -------------------------------------------------------------- |
| **Solver**   | **CSP**      | **COP**       | **CSP/COP**                                                            |
| exact    | FEASIBLE | OPTIMAL   | no state created                                               |
| gcs      | FEASIBLE | FEASIBLE | NOT_RUN                                                        |
| gurobi   | FEASIBLE | OPTIMAL   | no state created                                               |
| ortools  | OPTIMAL  | OPTIMAL   | same as Solve()                                                |
| minizinc | FEASIBLE | OPTIMAL   | same as Solve()                                                |
| Z3       | FEASIBLE | OPTIMAL   | UNSATISFIABLE + 0s runtime (solving with blocking constraints) |
| choco    | FEASIBLE | OPTIMAL   | NOT_RUN                                                        |

For the CSP setting, these are clear bugs to be fixed.

For the COP setting, not sure what the state + runtime should be. For the state, most logical seems:
- UNSATISFIABLE if 0 solutions found and returned within time
- UNKNOWN if 0 solutions found and timeout
- SATISFIABLE if at least 1 solution found
- NOT_RUN as default

For the runtime, the total runtime?

PR #545, opened 2024-12-05
Consistent state reporting

ThomSerg:
Solver are inconsistent when reporting their status after solving.

See issue #544 

There are more fixes to come...


Wout4:
let me know if this is ready for review!

cpmpy/solvers/z3.py
```diff
@@ -161,14 +164,30 @@ def solve(self, time_limit=None, assumptions=[], **kwargs):
 
         # translate exit status
         if my_status == "sat":
-            self.cpm_status.exitstatus = ExitStatus.FEASIBLE
-            if isinstance(self.z3_solver, z3.Optimize):
+            if self.has_objective(): # COP
+                # check if optimal solution found and proven, i.e. bounds are equal
+                lower_bound = self.z3_solver.lower(self.obj_handle)
+                upper_bound = self.z3_solver.upper(self.obj_handle)
+                if lower_bound == upper_bound: # found optimal
```
ThomSerg, line 171: I initially thought that I read somewhere that z3 can return "sat" on a COP when it only found a sub-optimal solution within the time limits (that's why I checked the bounds to see if it is optimal or not). But now I can't seem to find it again. Anyone has any idea if this is even needed, or that "sat" always means a optimal solution has been found for a COP (with suboptimal ending up in the 'unknown' case.

ThomSerg:
Changes / issues fixed
- add exit status instructions to template
- choco
	- broken time_limit in solveAll()
	- status never updated in solveAll()
- exact
	- suboptimal solution status (from exact) changed from "SAT" to "TIMEOUT"
	- keep track of remaining time for solveAll (was never updated)
	- status never created in solveAll()
- gurobi
	- make exitstatus consistent with ortools
	- status never updated in solveAll()
- minizinc
	- make exitstatus consistent with ortools
	- solveAll status was identical to solve -> added missing cases
- solver_interface (e.g. z3)
	- status never created in solveAll()
	- enforce time_limit in solveAll()
- z3
	- never reported on sub-optimal solutions for COP

I'm a bit stuck on gcspy, since untill now I haven't been able to install it anywhere.

IgnaceBleukx:
Updated GCS and added some tests.
I think one tricky thing is when the solution limit is **exactly** the number of feasible solutions.

In our solver-interface superclass, we now report `FEASIBLE`, whereas some solvers (e.g., OR-Tools) report `OPTIMAL`.
One way to solve this is to do one more solve-call, even when the solution limit has reached.
Downside is that this could take a long time for the solver to prove UNSAT....

IgnaceBleukx:
@ThomSerg What should be the next steps on this PR? I think it would be nice if it's part of the next release

ThomSerg:
> @ThomSerg What should be the next steps on this PR? I think it would be nice if it's part of the next release

I believe that with your addition for gcspy, everything is ready for review. The only open question that we still had is how the `runtime` should be measured. But that is part of a larger topic that also affects the meaning of `time_limit`; should it only include the time spend in the solver, or also in CPMpy? But best to leave that for a separate GitHub issue.

I'm still not quite sure if the bounds check for Z3 is needed, but leaving it in place as a safety check can't hurt either.

Don't know if you still would like a fix for checking for optimality when the solution limit has been reached?

IgnaceBleukx:
Ok, I've marked this one ready for review and will do a round of reviewing soon.
Regarding the status after solveAll, maybe @tias can express his opinion on this matter : ) In any case, we should document what we decide to do in the end.
I'm actually fine with solvers having a different status for this edge case. As long as we document it as "OPTIMAL means the solver provenly exhausted the feasible space" or smt like that.

hbierlee:
From offline discussion: the expected behavior (which I believe is implemented) is for solvers which automatically prove all solutions are returned (even if this is exactly the solution_limit, so e.g. or-tools) to return OPTIMAL, and in all other cases where we don't know if they prove this or we know they don't, they should return FEASIBLE.

IMO I wouldn't add the extra call.

ThomSerg:
As for the discussion to use OPTIMAL in a CSP setting, we just replicated the convention that OR-Tools uses. But as seen in the table included in issue #544, it seems to be the only solver that does this. So we would have to decide on which convention to follow.

tias:
I've never seen a solver return 'Optimal' for a satisfaction problem before. I would not have programmed it this way for ortools either, so they might have changed behaviour (or I missed it back then).

OrT docs say "  // More generally, this status represent a success. So we also return OPTIMAL
  // if we find a solution for a pure feasibility problem or if a gap limit has
  // been specified and we return a solution within this limit. In the case
  // where we need to return all the feasible solution, this status will only be
  // returned if we enumerated all of them; If we stopped before, we will return
  // FEASIBLE.
  OPTIMAL = 4;"

MiniZinc docs say "SATISFIED: A solution was found, but possibly not the whole search space was explored." Also OrT: "  // A feasible solution has been found. But the search was stopped before we
  // could prove optimality or before we enumerated all solutions of a
  // feasibility problem (if asked)."

I follow that, but don't want to go as far as concluding that a SAT problem has to return 'optimal'... So it should be 'Feasible' for satisfaction problems, and 'optimal' for solving with obj and solveall if entire search space was explored (e.g. the solvers final state is actually unsatisfiable, but the output given is optimal).

So if it is not an on-purpose choice, then wrt #544 there are only 2 inconsistencies to fix. Do we agree?

IgnaceBleukx:
I think this PR should not be part of the milestone, and rather be part of a `0.10.0` release instead.
We break the user-API here so a major change...

Any code that relies on checking whether status == OPTIMAL will now fail

tias:
I don't know, it would just be 2 changes: ortools to report 'sat' when sat, and the other one...

Maybe go ahead with making the discussed changes @ThomSerg and then we see?

ThomSerg:
A couple of notes after the requested changes:

- pysdd doesn't use `time_limit`
- the meaning of `time_limit` is not consistent (python overhead is not included, especially when solver doesn't natively support solving for multiple solutions (e.g. Exact))
- some code duplication between cpo and solver_interface for solveAll (only enforcement of nogood of current solution differs, maybe extract this to its own method?)

tias:
OK, I can see there is more in this PR then just the changes to these two solver states.

So, I must agree its not low hanging fruit, and we better leave it for the next release (potentially split up)

ThomSerg:
As written in the documentation, we now follow the below convention:

---
The status of solve-call can be the following:
1. `ExitStatus.OPTIMAL`: The solver found a solution to an optimisation problem and proved its optimality.
2. `ExitStatus.FEASIBLE`: The solver found a solution to a satisfaction problem, or a feasible solution to an optimization problem but did not prove optimality
3. `ExitStatus.UNSATIFIABLE`: The solver proved the input problem is unsatisfiable.
4. `ExitStatus.UNKNOWN`: The solver did not find a feasible solution, nor proved the problem is unsatisfiable. Can happen when a time-limit is reached.
5. `ExitStatus.NOT_RUN`: The solver is not run yet (default when initializing a solver)

---

The status of solveAll-call can be the following:
1. `ExitStatus.OPTIMAL`: The solver found all possible solutions to a problem and proved there to be none remaining.
2. `ExitStatus.FEASIBLE`: The solver found a subset of all solutions, or found all solutions but did not prove there to be none remaining.
3. `ExitStatus.UNSATIFIABLE`: The solver proved the input problem is unsatisfiable.
4. `ExitStatus.UNKNOWN`: The solver did not find a feasible solution, nor proved the problem is unsatisfiable. Can happen when a time-limit is reached.
5. `ExitStatus.NOT_RUN`: The solver is not run yet (default when initializing a solver)

---

When looking at the table in #545, only two states had to be updated in the case of `solve()`. For `solveAll()` a lot more changes were needed, since for some solvers state reporting was not properly added. Additionally, some fixes to `time_limit` are included due to it also not functioning as intended. Future improvements to the definition of `time_limit` (to include or not include cpmpy overhead, etc.) are best kept to a separate pull request. This one is focused on fixing what was already there.

It does seem like many of these changes (aside from `solveAll`) will be required for XCSP3.

cpmpy/solvers/choco.py
```diff
@@ -246,6 +257,19 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from
         self.cpm_status = SolverStatus(self.name)
         self.cpm_status.runtime = end - start
 
+        if len(sols): # solutions found
+            if (len(sols) == solution_limit): # matched the set limit (if given)
```
tias, line 261: I guess the simpler one is 'if (solution_limit is None) and ((time_limit is None) or (self...runtime < time_limit)):
  ret optimal
else
  ret feasible
  
?

tias (approved):
great; surprisingly intricate to get the status' right sometimes, esp for solveall...

one optional comment below

cpmpy/solvers/choco.py
ThomSerg: But then we won't get OPTIMAL when `solution_limit` is not None but greater than the available number of solutions. So a solveAll call could still prove to have found all solutions, thus returning OPTIMAL.

tias: ah, very attentive. As said, surprisingly intricate sometimes : )

Issue #546, opened 2024-12-05
Linearization of AllDifferent with constants


ThomSerg:
The linearization of `AllDifferent` makes the assumption that all the `args` are decision variables (calls `.lb` and `.ub` on them for the bipartite matching decomposition). 

The following will thus fail:
```python
import cpmpy as cp
arr = cp.cpm_array([cp.intvar(0, 5), cp.intvar(0, 5), 5, 4]) # combination of decision variables and constants
c = cp.AllDifferent(arr)

from cpmpy import SolverLookup
s = SolverLookup().get("gurobi")
s += c
s.solve()
```

IgnaceBleukx:
Hmm, yes - should be an easy fix by calling `get_bounds(var)` instead

ThomSerg:
That fixes the issue that the 'lb' and 'ub' calculation assumes only decision variables, but the rest of the decomposition is still incompatible. In the same example as above, the channeling constraint (connecting the one-hot boolean encoding to the original integers) will cause trouble:

```python
for arg, row in zip(cpm_expr.args, sigma):
    constraints += [sum(np.arange(lb, ub + 1) * row) + -1*arg == 0]
```
This implements $\sum\limits_{j} B_{i,j}*j = x_i$

For the constants 5 (similarly for 4), we get....
```
(sum([np.int64(0), 1, np.int64(2), np.int64(3), np.int64(4), np.int64(5)] * [BV12, BV13, BV14, BV15, BV16, BV17])) + -5 == 0
```
... which is not a type of expression that that current stage in the transformation expects, resulting in a `NotImplementedError`.

It is also quite wasteful: creating boolean decision variables to encode a constant.

& if the constant is quite large (or small) in can unnecessarily increase the size of the domain that gets represented by the one-hot encoding.

Instead we can just ignore the constants during bounds calculation, and for each constant which falls within the bounds we
completely remove the column in the boolean matrix corresponding to that already taken constant. (! this does require an additional check to see if the constants are "alldiff" amongst themselves, which can be done at construction time).

IgnaceBleukx:
Hmm I can see your point of "wasting" the Boolean variables just to encode a constant, but actually it might sometimes lead to a better linear relaxation, no?

Regarding the summation constraint - I would be fine with just checking whether x is a constant/var and decide on what side of the equality to put it based on that? Overall the case is quite exceptional where a user wants a constant in the alldiff?

IgnaceBleukx:
[...]

PR #547, opened 2024-12-05
AllDifferent linearization with constants

ThomSerg:
See issue #546 

Current decomposition inside the linearization assumes all arguments of `AllDifferent` are decision variables (so no support for constants).

This slightly altered decomposition attempts to fix that. For each of the constants, I remove the "at most once" constraint and added a "at most zero" constraint. Could be made more efficient by simply removing the columns corresponding to the constants. 

**TODO**: still needs a check that the constants are also "alldifff" amongst themselves (can maybe be checked during construction of the `AllDifferent` global constraint?

### DIFF
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -39,6 +39,7 @@ General comparisons or expressions
 """
 import copy
 import numpy as np
+import cpmpy as cp
 from cpmpy.transformations.normalize import toplevel_list
 
 from .flatten_model import flatten_constraint, get_or_make_var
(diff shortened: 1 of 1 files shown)


IgnaceBleukx:
Copied comment from issue #546, lets move the conversation here

Hmm I can see your point of "wasting" the Boolean variables just to encode a constant, but actually it might sometimes lead to a better linear relaxation, no?

Regarding the summation constraint - I would be fine with just checking whether x is a constant/var and decide on what side of the equality to put it based on that? Overall the case is quite exceptional where a user wants a constant in the alldiff?

IgnaceBleukx:
Ok so did a little more thinking about this one: I'm actually fine with encoding the constant with the range of Booleans.

The overhead should be minimal as the Boolean vars will be fixed at rootlevel already, if we do the following:
- replace bound computation of arguments with `get_bounds` (can also be done vectorized now)
- replace the "linking of Boolean variables to integer" as follows:
```python
for arg, row in zip(cpm_expr.args, sigma):
    if is_num(arg):
        constraints += [row[arg-lb] == 1] # constant, encode directly
    else:
        constraints += [sum(np.arange(lb, ub + 1) * row) + -1*arg == 0]

newlist += constraints
```

This handles with the edge-case more explicitly and keeps the main idea of the decomposition more clear in the code

(As a sidenote, we don't need the 'constraints' intermediate var here... We can just append to 'newlist' directly)

IgnaceBleukx:
I updated this PR with some minimal changes compared the the one on master, cfr my comments above.
@ThomSerg and @Wout4 if you could have a look

ThomSerg (commented):
Looks good. I can't approve since I'm the original pull request author.

Wout4:
allright, will merge

PR #550, opened 2024-12-07
ndvar_getitem: move fetch cast to init, saves many isinst+casts

tias:
small efficiency thing


Wout4:
Makes sense, will merge

PR #551, opened 2024-12-07
make has_subexpr internally recursive again

tias:
the rewrite seems to save especially a number of isinstance checks

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -169,15 +169,16 @@ class Expression(object):
         # recursive variant
         def rec_subexpr(lst):
             for el in lst:
-                if isinstance(el, (cp.variables._NumVarImpl, BoolVal)) or \
-                   isinstance(el, cp.variables.NDVarArray) and not el.has_subexpr() or \
-                   is_any_list(el) and not rec_subexpr(el):
-                    pass # check the rest
-                else:
-                    return True
+                if isinstance(el, Expression):
+                    if isinstance(el, (cp.variables._NumVarImpl, BoolVal)) or \
+                       isinstance(el, cp.variables.NDVarArray) and not el.has_subexpr() or \
+                       is_any_list(el) and not rec_subexpr(el):
+                        pass # check the rest
+                    else:
+                        return True
             return False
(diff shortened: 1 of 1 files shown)


cpmpy/expressions/core.py
```diff
@@ -166,26 +166,19 @@ def has_subexpr(self):
         if hasattr(self, '_has_subexpr'):
             return self._has_subexpr
 
-        # Initialize stack with args
-        stack = list(self.args)
-
-        while stack:
-            el = stack.pop()
-            if isinstance(el, Expression):
```
Wout4, line 174: we still need the explicit check that we are dealing with a cpmpy expression, this new interpration will see a constant '3' as a subexpression now since it's not any of our leaf expressions

Wout4:
I added the missing instance check, to see if we are dealing with an expression or a constant. Maybe this will invalidate the idea that we are saving instance checks with the recursive method, not sure how you tested it.

cpmpy/expressions/core.py
```diff
@@ -166,26 +166,21 @@ def has_subexpr(self):
...
-            elif is_any_list(el):
-                # Add list elements to stack for processing
-                stack.extend(el)
-
-        # No subexpressions found
-        self._has_subexpr = False
-        return False
+                if isinstance(el, Expression):
+                    if isinstance(el, (cp.variables._NumVarImpl, BoolVal)) or \
+                       isinstance(el, cp.variables.NDVarArray) and not el.has_subexpr():
```
IgnaceBleukx, line 176: I think that we can off-load some of this to just the object-hierarchy no? Not sure if it faster but if we implement a has_subexpr() for NDVarArray than we'll have a has_subexpr() for each Expression.
So, then this check becomes just:
```
if isinstance(el, Expression):
    if el.has_subexpr():
        return True
    else:
        pass
return False
```

tias:
The current version does not save any isinstance calls; trying Ignace' suggestion...

tias:
Ignace's version saves isinstance calls, but not runtime... I think because it requires additional dispatched function calls...

There don't seem to be clear gains here; I'm just going to close it and keep the code as is.

Issue #554, opened 2024-12-12
cpm_array(m.T) returns m in stead of m.T


Wout4:
```
iv = intvar(-4,4, shape=(5,5))
print(iv)
print(cpm_array(iv.T))
```

not sure what happens, cpm_array calls the ndvararray constructor, (variables.py 
```
    def __init__(self, shape, **kwargs):
        Expression.__init__(self, "NDVarArray", self)
```
but this calls the expression init, without using the kwargs (the transposed matrix is in the buffer argument i think).

but doing this doesnt fix it because we flatten in the expression init:
```
    def __init__(self, shape, **kwargs):
        Expression.__init__(self, "NDVarArray", kwargs['buffer'])
```
```
            if isinstance(arg_list[i], np.ndarray):
                # must flatten
                arg_list[i] = arg_list[i].reshape(-1)

```

this reshape also gets rid of our transposed form somehow

IgnaceBleukx:
Hmm tricky stuff indeed.
[This part of the numpy docs](https://numpy.org/doc/2.1/user/basics.copies.html#how-to-tell-if-the-array-is-a-view-or-a-copy) explains it quite well.

We are currently taking "buffer" as the content of the array, but it is unchanged when acting on a view (such as `ndarray.T`). So, I think we have to check whether we are acting on a view, and if so: either throw an error to the user, or make a proper copy of the view and use that.

Wout4:
i was a bit confused by how our NDVarArray constructor works, but because it inherits from np.ndarray, i guess first numpy's constructor gets called, initialising the 'self' within our own init.
Apperently there is a flag in ndarrays to indicate if you are row-major (C-style) or column-major (Fortran-style). Transposing a matrix just sets this flag, so the data doesn't have to change.
Adding this flag correctly to the call to NDVarArray then fixes our problem (see pull request)

tias:
So we seem to agree that the PR fixes it.

For the specific example, instead of
```
iv = intvar(-4,4, shape=(5,5))
print(iv)
print(cpm_array(iv.T))
```
one should do 
```
iv = intvar(-4,4, shape=(5,5))
print(iv)
print(cpm_array(iv).T)
```
and get desired behaviour? 

but that is not how the bug appeared I guess

Dimosts:
Yes, the issue is that the call to cpm_array can be hidden in e.g. a global constraint. For example, maybe an np array is given somewhere, and in some cases a cpm_array may need to be created.

[...]

PR #555, opened 2024-12-12
fix cpm_array() when order = 'F'

Wout4:
fixes https://github.com/CPMpy/cpmpy/issues/554

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -237,7 +237,8 @@ def cpm_array(arr):
     """
     if not isinstance(arr, np.ndarray):
         arr = np.array(arr)
-    return NDVarArray(shape=arr.shape, dtype=arr.dtype, buffer=arr)
+    order = 'F' if arr.flags['F_CONTIGUOUS'] else 'C'
+    return NDVarArray(shape=arr.shape, dtype=arr.dtype, buffer=arr, order=order)
 
 
 class NullShapeError(Exception):
--- a/tests/test_expressions.py
+++ b/tests/test_expressions.py
@@ -479,6 +479,10 @@ class TestBounds(unittest.TestCase):
         self.assertListEqual([1,[3,5],[6]], ubs)
 
 
+    def test_array(self):
+        m = intvar(-3,3, shape = (3,2), name= [['a','b'],['c','d'],['e','f']])
+        self.assertEqual(str(cpm_array(m)), '[[a b]\n [c d]\n [e f]]')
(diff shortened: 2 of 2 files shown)


Wout4:
accidently gave commit the wrong name..

Wout4:
seems that our NDVarArray constructor is actually setting the base property to the args of the array, so it contains the flatlist of elements..

Wout4:
we can compare the base and the array but that seems a bit excessive

Wout4:
```
data = np.array([_BoolVarImpl(name=n) for n in names])
    # insert into custom ndarray
    r = NDVarArray(shape, dtype=object, buffer=data)
```
in our constructor of boolvar and intvar, we make an array out of data so base is not none anymore

tias:
I remember I tried a couple of things to instantiate a ndarray subclass, and only the above one worked with my very limited knowledge (and patience) at the time...

So, if we know of a better way to initialise NDVarArray... one that is more 'view' compatible, that would be very welcome...

IgnaceBleukx:
Wait, but this is about the `cpm_array` helper method, right?
We don't call the cpm_array method when constructing an array of variables, there we simply wrap the freshly created numpy array into an NDVarArray, which seems correct to me.

I'm proposing to _always_ copy the buffer of the input array when using `cpm_array`, so that it mimics the numpy behaviour.
Secondly, I think we should copy the buffer when initialising `NDVarArray` when the given array is a view. I don't think the latter would really matter for our use-case, but a user may call the constructor directly...

tias:
if you look at the commit in this PR, that is what we already did, and now Wout is just also taking the 'order' into account... so all good for this patch?


(btw our 'friend' noted that to make our custom objects more view-compatible, we can do something like:
```
    def __array_finalize__(self, obj):
        """
        This method is called automatically when a view of the array is created.
        It allows you to propagate custom attributes to the view.
        """
        if obj is None:
            return
        # Copy custom attributes from the source object
        self.description = getattr(obj, 'description', None)
```
e.g. we can do that to set 'name' and '_args' and _has_subexpr instead of what we do inside the __repr__ now; Yes, I'm totally abusing this PR, if its relevant I can open a separate issue?

IgnaceBleukx:
Yes for this patch indeed it's fine, let's move the broader discussion to a new issue :)

Wout4:
so we can merge this? think we need it anyway, wether we make copy or not

IgnaceBleukx:
Jup fine to merge indeed - maybe add a small test for it>

Wout4:
I opened a different issue for the copy/view thing
Maybe you should also open one @tias , I'm not sure how to phrase what you want ;)

PR #560, opened 2024-12-20
Global function in Z3 objective

IgnaceBleukx:
Posting an objective to Z3 was a little buggy as global functions were not decomposed when unsupported.
This PR fixes that by creating an explicit "objective var" when the objective is a global function.


Wout4:
nice easy fix, thanks for updating the tests!

Issue #559, opened 2024-12-20
Objective in GCS does not have any influence on solution quality


IgnaceBleukx:
There seems something wrong in the interface to GCS.
I'm not sure if its CPMpy or the Python API of GCS itself.

```python
import cpmpy as cp

vars = cp.intvar(0,10, shape=3)
model = cp.Model(cp.AllDifferent(vars), maximize=cp.sum(vars))

solver = cp.SolverLookup.get("gcs", model)
assert solver.solve()
print(vars.value())
```
Shows `[0,1,2]` while clearly the optimal value assigns 10,9 and 8 to the variables.

GCS has a handy function `get_api_calls_str()` function, which returns the following log:
```
auto v0 = p.create_integer_variable(0_i, 10_i);
auto v1 = p.create_integer_variable(0_i, 10_i);
auto v2 = p.create_integer_variable(0_i, 10_i);
p.post(AllDifferent{{v0, v1, v2}});
auto v3 = p.create_integer_variable(0_i, 30_i);
p.post(LinearEquality{WeightedSum{} + 1_i * v0 + 1_i * v1 + 1_i * v2 + -1_i * v3, 0_i});
p.minimise(v3); # oh ow...
solve
get_solution_value
get_solution_value
get_solution_value
get_solution_value
get_solution_value
```

I did a little digging already, and found this func in `gcspy-0.1.3/python/gcspy.cpp` which explains the faulty logging, but it seems only part of the problem as the logic is correct.
Moreover, minimizing the negated objective yields the same issue...

```cpp
auto Python::maximise(const string & var_id) -> void
{
#ifdef WRITE_API_CALLS
    api_calls << "p.minimise(v" << var_id << ");" << endl;
#endif
    p.maximise(get_var(var_id));
}
```

@mmcilree maybe you can help out here : )

mmcilree:
Ah... I think I know what's happening. 

If `all_solutions=false`(as is default) gcspy stops as soon as it finds any solution regardless of whether it's optimising or not. But I guess the expected behaviour of solve when maximising should be to keep going until it finds the optimal solution regardless of whether `all_solutions` is set. This was an oversight on my part.

This should be easy to fix. Does GCS get it right in your example if you do `solve(all_solutions=True)`?

IgnaceBleukx:
Hi Matthew,

In the CPMpy interface we do set the `all_solutions` flag to True when we are optimizing:
[...]

PR #561, opened 2024-12-20
558 reporting intermediate solutions

### LINKED ISSUE #558: Reporting intermediate solutions
For several use-cases where you solve hard optimization problems, it might be useful to get the intermediate solutions back from the solver.
I know Gurobi and OR-Tools support this feature through callbacks, but it requires a user to write its own wrapper for each solver he wants to use.

I think it would be nice if we can handle this in a more unified way.
If you agree, let's use this issue on how exactly this feature should be implemented.

My take:
Something similar to the `display` parameter in `solveAll` would be nice I think. 
Here, we can either pass a (list of) expressions whose value is printed after each solution found, or a "real" callback function that gets called on every solution. A user can use the callback himself to track time until this solution or similar stats.

IgnaceBleukx:
Following the discussion in #558.

It adds quite some code in all solver interfaces, but I tried to do it somewhat consistent.
Some solvers are missing still:
- minizinc - this needs a lot of engineering with async functions (similar to solveAll)
- gcs - there is a callback function but as optimization is not working properly (see #559) I could not test it.

For Exact,it is somewhat hacky as we need to go through "runOnce" instead of the nice "toOptimal" function. So there will be some more ping-pong through the interface.


cpmpy/solvers/TEMPLATE.py
```diff
@@ -92,12 +92,14 @@ def __init__(self, cpm_model=None, subsolver=None):
...
 
 
-    def solve(self, time_limit=None, **kwargs):
+    def solve(self, time_limit=None, display=None, **kwargs):
         """
             Call the TEMPLATE solver
 
             Arguments:
             - time_limit:  maximum solve time in seconds (float, optional)
+            - display: display: either a list of CPMpy expressions, OR a callback function, called with the variables after value-mapping
```
ThomSerg, line 101: makes it seem as if the "variables after value-mapping" would get passed to the display callback

cpmpy/solvers/gurobi.py
```diff
@@ -139,9 +158,32 @@ def solve(self, time_limit=None, solution_callback=None, **kwargs):
...
+        if display is not None:
+            assert solution_callback is None, "Cannot have both generic CPMpy callback and specialized gurobi solution callback"
+            if isinstance(display, Expression) or is_any_list(display):
+                _cpm_vars = get_variables(display)
+            else:
+                _cpm_vars = list(self.user_vars)
+            # avoid re-lookup
+            grb_vars = self.solver_vars(_cpm_vars)
+            def solution_callback(model, state, **kwargs):
+                if state == GRB.Callback.MIPSOL: # found new solution
```
ThomSerg, line 173: Yes, that's a much nicer callback than the `.MIP` one

ThomSerg (commented):
Seems a nice quality-of-life improvement. But yeah, the one for exact seems hack-y, especially with the "warning" part. Is there a way to restore the solver afterwards?

cpmpy/solvers/gurobi.py
```diff
@@ -139,9 +158,32 @@ def solve(self, time_limit=None, solution_callback=None, **kwargs):
...
+                        cpm_var._value = val
+
+                    if isinstance(display, Expression):
+                        print(argval(display))
+                    elif is_any_list(display):
+                        print(argvals(display))
+                    else:
+                        display()
+
+        _ = self.grb_model.optimize(callback=solution_callback)
```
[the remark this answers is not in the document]
IgnaceBleukx: Not really, it's the same on master...

cpmpy/solvers/gurobi.py
```diff
@@ -30,11 +30,11 @@
     Module details
     ==============
 """
-
+from examples.csplib.prob050_diamond_free import diamond_free
```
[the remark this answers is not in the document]
IgnaceBleukx: Jup, fixed

cpmpy/solvers/TEMPLATE.py
IgnaceBleukx: Ah hmm I see, I copy-pasted it from the `solveAll` docstring so maybe that one should be updated too? 
Feel free to propose an alternative that would make it more clear : )

IgnaceBleukx: Updated this description to be more clear

ThomSerg: Just noticed that we don't mention that display is only for optimization problems. Add in docstring + add warning?

ThomSerg:
Besides that it is never mentioned + silently enforced that "display" is only for optimization, it seems good to me.

Wout4:
we could add the gcs callback also since that bug is fixed or not?

IgnaceBleukx:
Made quite some changes to this PR after the last review, so maybe another round of revieuwing from @ThomSerg and/or @Wout4 would be appropriate before merging.

I'm wondering if we should really fill the `solver.objective_` argument directly from the underlying solver argument.
Alternatively, we can store the original objective expression and compute the objective value based on that. There were quite some tricky things to get right in this PR...

cpmpy/solvers/minizinc.py
```diff
@@ -240,7 +244,10 @@ def solve(self, time_limit=None, **kwargs):
 
         # make mzn_inst
         (mzn_kwargs, mzn_inst) = self._pre_solve(time_limit=time_limit, **kwargs)
-        
+
+        if display is not None:
+            raise NotImplementedError("TODO: implement MiniZinc callbacks -- requires new async methhod (similar to solveAll())")
```
ThomSerg, line 249: Meant to leave this `TODO`?

cpmpy/solvers/exact.py
```diff
@@ -149,6 +150,8 @@ def solve(self, time_limit=None, assumptions=None, **kwargs):
                            For repeated solving, and/or for use with s.get_core(): if the model is UNSAT,
                            get_core() returns a small subset of assumption variables that are unsat together.
             :type assumptions: list of CPMpy Boolean variables
+            display: either a list of CPMpy expressions, OR a callback function, called with the variables after value-mapping
```
ThomSerg, line 153: why is this documentation of "display" slightly different than that of all the other solvers?

ThomSerg (changes requested):
Went through the entire thing again

cpmpy/solvers/exact.py
```diff
@@ -176,32 +179,69 @@ def solve(self, time_limit=None, assumptions=None, **kwargs):
...
+            self.xct_solver.setAssumptions([(self.solver_var(obj_assump), self.objective_is_min_)])
+
+            if isinstance(display, Expression) or is_any_list(display):
+                _cpm_vars = get_variables(display)
+            else:
+                _cpm_vars = None
+
+            start = time.time()
+            xct_status = self.xct_solver.runOnce(time_limit if time_limit is not None else 0)
+            while xct_status != "INCONSISTENT":
```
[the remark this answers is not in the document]
IgnaceBleukx: Yes, indeed... I realized this the hard way... fixed in commit [ab6a556](https://github.com/CPMpy/cpmpy/pull/561/commits/ab6a556c195abbfa09ac4c87df0f70a6630f4055). Also counts the time spent in translating vars etc. Not sure what the best version is

cpmpy/solvers/minizinc.py
IgnaceBleukx: Choco in Java does have callbacking, so just leaving it here for if and when they implement callbacking directly from Python as well

cpmpy/solvers/choco.py
```diff
@@ -144,6 +144,9 @@ def solve(self, time_limit=None, **kwargs):
         if time_limit is not None:
             self.chc_solver.limit_time(str(time_limit) + "s")
 
+        if display is not None:
+            raise NotImplementedError("Choco does not support solution callbacks, (TODO?)")
```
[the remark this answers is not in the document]
IgnaceBleukx: Choco in Java does have callbacking, so just leaving it here for if and when they implement callbacking directly from Python as well

cpmpy/solvers/exact.py
IgnaceBleukx: Forgot to adapt this one apparently

cpmpy/solvers/choco.py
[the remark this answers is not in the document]
IgnaceBleukx: We _could_ also implement it as in Exact, but there will be a lot more overhead in Choco (needs to pass trough C++ and Java...)

IgnaceBleukx:
Should be fine now @ThomSerg

IgnaceBleukx:
Decided to keep the `solution_callback` argument explicitely in the functions signature, and added proper documentation  for it. Also that it takes precedence over the `display` argument if both are supplied (which I think makes sense)

Should be ready for a re-review @ThomSerg

cpmpy/solvers/choco.py
```diff
@@ -685,7 +685,7 @@ def _get_constraint(self, cpm_expr):
                 raise NotImplementedError(f"Unknown global constraint {cpm_expr}, should be decomposed! If you reach this, please report on github.")
 
         # unlikely base case: Boolean variable
-        elif isinstance(cpm_expr, _BoolVarImpl):    
+        elif isinstance(cpm_expr, _BoolVarImpl):
```
tias: choco only contains space changes maybe take out here?

cpmpy/solvers/exact.py
```diff
@@ -220,7 +220,7 @@ def solve(self, time_limit:Optional[float]=None, assumptions:Optional[List[_Bool
             timeout = time_limit
         else:
             timeout = 0
-            
+
```
tias: also only space changes for exact...

cpmpy/solvers/gcs.py
```diff
@@ -263,6 +271,32 @@ def solve(self, time_limit:Optional[float]=None, prove=False, proof_name:Optiona
             
         return has_sol
 
+    def _get_callback(self, display) -> Callable[[dict[str,int]], None]:
```
tias: is 'display' also typable?

cpmpy/solvers/gurobi.py
```diff
@@ -601,13 +613,45 @@ def solveAll(self, display:Optional[Callback]=None, time_limit:Optional[float]=N
...
                 if grb_status == GRB.TIME_LIMIT: # reached time limit
                     self.cpm_status.exitstatus = ExitStatus.FEASIBLE
-                else: # found all solutions   
+                else: # found all solutions
                     self.cpm_status.exitstatus = ExitStatus.OPTIMAL
         # if unsat or timout with no solution, .solve() will have already set the state accordingly (so nothing to update)
 
         return opt_sol_count
+
+    def _get_callback(self, display, events):
```
tias: can this be typed?

cpmpy/solvers/minizinc.py
```diff
@@ -923,7 +923,7 @@ def solveAll(self, display:Optional[Callback]=None, time_limit:Optional[float]=N
                 call_from_model:    whether the method is called from a CPMpy Model instance or not
                 **kwargs:           any keyword argument, sets parameters of solver object, overwrites construction-time kwargs
 
-            Returns: 
+            Returns:
```
tias: also only space changes for mzn...

you can put these solver files in a separate PR, I don't mind, just not as if its related to intermediate solutions

cpmpy/solvers/z3.py
```diff
@@ -571,4 +578,33 @@ def get_core(self):
         return [self.assumption_dict[z3_var] for z3_var in self.z3_solver.unsat_core()]
 
 
+    def _get_callback(self, display):
```
tias: typing?

cpmpy/solvers/z3.py
```diff
@@ -571,4 +578,33 @@ def get_core(self):
...
+
+        if isinstance(display, Expression) or is_any_list(display):
+            cpm_vars = get_variables(display) # only fill in relevant vars for callback
+        else:
+            cpm_vars = list(self.user_vars) # function can use any variables
+        z3_vars = self.solver_vars(cpm_vars)
+
+        def callback(sol):
```
tias: typing?

cpmpy/solvers/minizinc.py
IgnaceBleukx: These were an artifact of undoing previous changes and merges I think... I've taken them out of this PR

cpmpy/solvers/cpo.py
```diff
@@ -156,14 +156,16 @@ def native_model(self):
...
             Call the CP Optimizer solver
 
             Arguments:
-                time_limit (float, optional):   maximum solve time in seconds 
-                solution_callback (an `docplex.cp.solver.solver_listener.CpoSolverListener` object):   CPMpy includes its own, namely `CpoSolutionCounter`. If you want to count all solutions, 
-                                                                                                        don't forget to also add the keyword argument 'enumerate_all_solutions=True'.
+                time_limit (float, optional):   maximum solve time in seconds
+                solution_callback:              a ``docplex.cp.solver.solver_listener.CpoSolverListener`` object)
+                                                Takes precedence over ``display`` when both are set.
+                display:                        either a list of CPMpy expressions, OR a callback function, called with the variables after value-mapping.
```
ThomSerg: "called after value-mapping the variables" not "with", callable is called without any args.
-> more like docstring of "def solve"'

ThomSerg: otherwise seems like you need to pass a `def display(vars): ...`

cpmpy/solvers/cpo.py
```diff
@@ -156,14 +156,16 @@ def native_model(self):
...
+    def solve(self, time_limit:Optional[float]=None, solution_callback=None, display:Optional[Callback]=None, **kwargs):
         """
             Call the CP Optimizer solver
 
             Arguments:
-                time_limit (float, optional):   maximum solve time in seconds 
-                solution_callback (an `docplex.cp.solver.solver_listener.CpoSolverListener` object):   CPMpy includes its own, namely `CpoSolutionCounter`. If you want to count all solutions, 
-                                                                                                        don't forget to also add the keyword argument 'enumerate_all_solutions=True'.
+                time_limit (float, optional):   maximum solve time in seconds
+                solution_callback:              a ``docplex.cp.solver.solver_listener.CpoSolverListener`` object)
```
ThomSerg: Also "CpoCallback"? What is the difference?

cpmpy/solvers/cpo.py
```diff
@@ -703,8 +713,10 @@ def __init__(self, verbose=False):
             if self.__verbose:
                 self.__start_time = time.time()
 
-        def result_found(self, solver, sres):
-            """Called on each new solution."""
+        def invoke(self, solver: CpoSolver, event:str, sres:CpoSolveResult):
```
ThomSerg: Why do we now use CpoCallback which doesn't just call on solutions, to then filter to only solutions. With HiGHS I found an observable performance difference when doing the event filtering on our side vs the solver's side.

cpmpy/solvers/cpo.py
ThomSerg, line 796: missing documentation of "events". Also what other events are needed than EVENT_SOLUTION for a solution printer? I get that it is useful to support a generic "display" on any event, but why make it part of the CpoSolutionPrinter? It feels like this class has become something more generic / general than just a solution printer. Maybe you need a CpoCallback base class with the event option and then make CpoSolutionCallback subclass it with only EVENT_SOLUTION?

cpmpy/solvers/gcs.py
```diff
@@ -156,32 +156,40 @@ def native_model(self):
...
+            Returns:
                 whether a solution was found.
         """
         # ensure all user vars are known to solver
         self.solver_vars(list(self.user_vars))
-        
+
+        callback = None
+        if display is not None:
+            callback = self._get_callback(display)
```
ThomSerg, line 192: should this approach of a `_get_callback` method become more standard across solvers?

ThomSerg: e.g. also for cpo and ortools, the solvers that have a class-based instead of method based system

cpmpy/solvers/hexaly.py
```diff
@@ -141,13 +141,16 @@ def __init__(self, cpm_model=None, subsolver=None):
...
+    def solve(self, time_limit:Optional[float]=None, solution_callback=None, display:Optional[Callback]=None, **kwargs):
         """
             Call the Hexaly solver
 
             Arguments:
                 time_limit:  maximum solve time in seconds (float, optional)
-                kwargs:      any keyword argument, sets parameters of solver object
+                solution_callback:             Hexaly callback registered on ``TIME_TICKED``.
+                                               Takes precedence over ``display`` when both are set.
+                display:                       generic solution callback; either expressions, a list thereof, or a callable (see ``HexSolutionPrinter``).
```
ThomSerg: not consistent with other solvers, missing some explanation about ... after variable-value assignment ...

tests/test_solvers.py
```diff
@@ -1032,6 +1032,49 @@ def test_vars_not_removed(self, solver):
...
+            kwargs['time_between_ticks'] = 1
+            n = 20 # need a bigger model to see the ticks
+
+        vars = cp.intvar(0,n, shape=n)
+        obj = cp.sum([random.randint(1,n) * (a - b) for a in vars for b in vars])
+        model = cp.Model(cp.AllDifferent(vars), maximize=obj)
+
+        assert model.solve(solver=solver, display=vars, time_limit=3, **kwargs)
+        captured = capsys.readouterr().out
+        assert len(captured) > 0
```
ThomSerg: What if a solver first prints some warning or info message before solving (some do, like info about the license), will this not also make the test succeed?

tests/test_solvers.py
```diff
@@ -1032,6 +1032,49 @@ def test_vars_not_removed(self, solver):
...
+
+        # collect solutions using callback
+        collector = list()
+        assert model.solve(solver=solver, display=lambda :  collector.append(argvals(vars)), time_limit=3, **kwargs)
+        assert len(collector) >= 1
+
+        # print some more information using callback
+        from time import time
+        def display():
```
ThomSerg, line 1050: didn't we used to have tests that wrote something to a global datastructure in "display", so we could more directly test the callback instead of relying on stdout?

cpmpy/solvers/cpo.py
IgnaceBleukx: The solver listener is a more specialized callback, where you have just a few functions that you can implement (conflict reached, new result etc). the CPO Callback is more general, as you can register for any event.

I kept the CPO-specific argument for backwards compatibility, and (hopefully soon) a neater implementation of solveAll

cpmpy/solvers/gcs.py
IgnaceBleukx: I think its ok to have the class-based ones as is? It is nice and explicit

tests/test_solvers.py
IgnaceBleukx: That is the test on the previous line : ) Where  we collect the solutions in a collector

cpmpy/solvers/cpo.py
IgnaceBleukx: Yes, you're right. The events parameter does not make too much sense for a solution printer...
I've removed it now

cpmpy/solvers/cpo.py
IgnaceBleukx: Ok after a lot of trouble I found out the difference between solver listeners and solver callbacks.
If you just use ".solve()", listeners are only invoked for the very last solution, while callbacks are always invoked.
You need to set a particular parameter to make it work:
https://ibmdecisionoptimization.github.io/docplex-doc/cp/docplex.cp.solver.solver.py.html#docplex.cp.solver.solver.CpoSolver.solve

I've updated the code to work with both callbacks and listeners.

IgnaceBleukx:
I think the comments are addressed @ThomSerg

IgnaceBleukx:
I decided to take out the display support for Z3 after all.
It was too unreliable to be tested properly (z3 finds a first solution relatively quick, but then does not improve, even with a large time out).

Alternative is to keep it supported, but skip z3 in the tests?
In any case ready for a round of reviewing.

cpmpy/solvers/cpo.py
```diff
@@ -156,14 +158,18 @@ def native_model(self):
...
         """
             Call the CP Optimizer solver
 
             Arguments:
-                time_limit (float, optional):   maximum solve time in seconds 
-                solution_callback (an `docplex.cp.solver.solver_listener.CpoSolverListener` object):   CPMpy includes its own, namely `CpoSolutionCounter`. If you want to count all solutions, 
-                                                                                                        don't forget to also add the keyword argument 'enumerate_all_solutions=True'.
+                time_limit (float, optional):   maximum solve time in seconds
+                solution_callback:              a ``docplex.cp.solver.solver_listener.CpoSolverListener`` or ``docplex.cp.solver.cpo_callback.CpoCallback`` object, or a list thereof
+                                                Takes precedence over ``display`` when both are set.
```
ThomSerg: Mix of capital and non-capital first letters. In my code editor on a smaller screen makes it more difficult to see which sentence belongs to which argument and where a new argument starts.

cpmpy/solvers/cpo.py
```diff
@@ -773,7 +785,7 @@ def myprint():
                             default/None: nothing displayed
                 solution_limit (default = None): stop after this many solutions 
         """
-        def __init__(self, solver, display=None, solution_limit=None, verbose=False):
+        def __init__(self, solver: CPM_cpo, display=None, solution_limit=None, verbose=False):
```
ThomSerg: Type hints

cpmpy/solvers/gurobi.py
```diff
@@ -155,13 +155,18 @@ def native_model(self):
...
+    def solve(self, time_limit:Optional[float]=None, solution_callback:Optional[Callable]=None, display:Optional[Callback]=None, **kwargs):
         """
             Call the gurobi solver
 
             Arguments:
                 time_limit (float, optional):  maximum solve time in seconds
-                solution_callback:             Gurobi callback function
+                solution_callback:             Gurobi callback function.
+                                               Takes precedence over ``display`` when both are set.
+                display:                       generic solution callback for use during optimization.
```
ThomSerg, line 169: here too, might confuse that the description of "display" already starts the sentence before

cpmpy/solvers/gurobi.py
```diff
@@ -707,13 +719,47 @@ def solveAll(self, display:Optional[Callback]=None, time_limit:Optional[float]=N
...
+
+    def _get_callback(self, display:Callback, events:Iterable) -> Callable:
+        """
+        Get the callback function to use for Gurobi.
+        Arguments:
+            display: either an expression, a list of expressions, or a callback function
+            events: iterable of gurobipy.GRB.Callback event codes
+        """
+
+        self.events=  frozenset(events)
```
ThomSerg: space

tests/test_solvers.py
```diff
@@ -1037,6 +1038,58 @@ def test_vars_not_removed(self, solver):
         #test unique sols, should be same number
         assert len(sols) == 8
 
+    def test_solution_callback(self, solver, capsys):
+        import random
+        random.seed(0)
```
ThomSerg: Why set the seed? Most solvers  (if not all) will ignore this and require you to pass the seed to the solver object no?

ThomSerg (commented):
Very minor styling comments. For the rest looks good!

tests/test_solvers.py
IgnaceBleukx: Ah this was for when I used a randomly generated  problem here, will remove

IgnaceBleukx:
Processed comments and merged master into this one; but its missing a PR "approval" :)

ThomSerg (approved):
Approved

PR #566, opened 2025-01-03
Gurobi separate licence check

ThomSerg:
Separate the supported check of gurobipy in two:
1) check that python package is **installed**
2) check that **license** is ok

Previous implementation did both at once. When the licence failed, it was not easily known that this was the cause and not an issue with the installation of gurobi. Also added explicit printing of the licence error that gurobi throws, as to give a hint to the user what the exact issue is. I had issues with `HostID mismatch`, which was not clear before.

### DIFF
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -71,7 +71,7 @@ class CPM_gurobi(SolverInterface):
 
     @staticmethod
     def supported():
-        return CPM_gurobi.installed() & CPM_gurobi.license_ok()
+        return CPM_gurobi.installed() and CPM_gurobi.license_ok()
 
     @staticmethod
     def installed():
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -71,8 +71,19 @@ class CPM_gurobi(SolverInterface):
 
     @staticmethod
     def supported():
-        # try to import the package
-        try:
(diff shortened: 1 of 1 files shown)


cpmpy/solvers/gurobi.py
```diff
@@ -71,8 +71,19 @@ class CPM_gurobi(SolverInterface):
...
+        try: 
+            import gurobipy as gp
+            return True
+        except ImportError:
+            return False
+        
+    @staticmethod
+    def license_ok():
+        try: 
             import gurobipy as gp
```
Wout4, line 87: so license_ok will also fail if gurobi is not installed? I guess you fixed that by calling them in the right order in the init function, but maybe it makes more sense to check is_installed within license_ok, such that calling license_ok also works separately

cpmpy/solvers/gurobi.py
```diff
@@ -71,8 +71,19 @@ class CPM_gurobi(SolverInterface):
 
     @staticmethod
     def supported():
-        # try to import the package
-        try:
+        return CPM_gurobi.installed() & CPM_gurobi.license_ok()
```
IgnaceBleukx, line 74: use `and` here : )

cpmpy/solvers/gurobi.py
ThomSerg: Thanks, also added a warning to the user to please first check .installed() before .license_check() for good measure.

PR #567, opened 2025-01-03
Z3 auto subsolver

ThomSerg:
A proposal to auto-determine the subsolver type for Z3.

When calling `model.solve(solver='z3')` on a model with an objective function, an exception gets raised since by default the `sat` subsolver gets selected. Instead one has to explicitly write `model.solve(solver='z3:opt')`. For a beginner this is quite confusing, since non of the other solver (as far as I know) require selecting an appropriate subsolver depending on the problem to be solved.

I propose, when no subsolver is provided, to determine the appropriate one depending on the presence of an objective function. This auto-configuration is only possible when going from a model to a CPM_z3 solver object. When directly modeling using CPM_z3, at initialization no objective function will be present to determine subsolver type. So keeping the argument default `"sat"` seems appropriate (instead of `None`).

### DIFF
--- a/cpmpy/model.py
+++ b/cpmpy/model.py
@@ -119,6 +119,9 @@ class Model(object):
         self.objective_ = expr
         self.objective_is_min = minimize
 
+    def has_objective(self):
+        return self.objective_ is not None
+
     def minimize(self, expr):
         """
             Minimize the given objective function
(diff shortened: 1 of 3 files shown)


IgnaceBleukx (approved):
Looks good to me, ideally just change this test to remove the special case for z3.

Issue #564, opened 2024-12-23
No module named 'pkg_resources'


ThomSerg:
In python3.12 `setuptools` no longer comes pre-installed in virtual environments created with venv ([link](https://docs.python.org/3/whatsnew/3.12.html#:~:text=Do%20not%20pre%2Dinstall%20setuptools%20in%20virtual%20environments%20created%20with%20venv.%20This%20means%20that%20distutils%2C%20setuptools%2C%20pkg_resources%2C%20and%20easy_install%20will%20no%20longer%20available%20by%20default%3B%20to%20access%20these%20run%20pip%20install%20setuptools%20in%20the%20activated%20virtual%20environment.)). This in turn makes `pkg_resources` unavailable, which is relied upon by `CPM_exact` to check for version conflicts. `CPM_choco` seems to instead depend on a combination of `importlib.metadata` and `packaging` for handling package metadata. Whilst this currently has an issue of its own (#563), should after its fix probably convert `CPM_exact` to that same approach?

IgnaceBleukx:
Fully agree that we should make the version-checking more coherent across all solvers.
We should probably also add something about it in the documentation and/or in TEMPLATE.py for when solver developers write a new solver interface.

I'm fine with either approach (packaging + importlib) or pkg_resources, alhought the pkg_resources seems to be a little more compact. Indeed, we should indeed add it as a requirement if it's not installed by default...

hbierlee:
By the way, it is also neccesary to install `setuptools` before you can run the test suite (but the developer guide does not contain instructions for this). Same actually goes for `pytest`, perhaps this should be added to the docs.

Issue #563, opened 2024-12-23
`packaging` not in list of dependencies.


ThomSerg:
`CPM_choco` utilises the `packaging` library to check whether the correct version has been installed, but this has not been added as a dependency for CPMpy. Causes `choco` as solver target to always report that `pychoco` is not installed, since the import of `packaging` fails in the `CPM_choco.supported()` method:

```python
def supported():
        # try to import the package
        try:
            import pychoco as chc
            # check it's the correct version
            # CPMPy uses features only available from 0.2.1
            from importlib.metadata import version as get_version
            from packaging import version # <- causes exception and in turn reporting that pychoco is not installed
            pychoco_version = get_version("pychoco")
            if version.parse(pychoco_version) < version.parse("0.2.1"):
                import warnings
                warnings.warn(f"CPMpy uses features only available from Pychoco version 0.2.1, but you have version {pychoco_version}")
                return False
            return True
        except ImportError:
            return False
```

Do we add `packaging` as a dependency (exact will also need it)?

IgnaceBleukx:
Closed by #568

PR #568, opened 2025-01-07
Standardize solver supported (version) check

ThomSerg:
Standardizing the 'supported' check for the solvers, including version checks. Solves both #564 and #563.

Depends on pull request #566 to be merged first.  

The warning message when a version check fails is still a bit unfortunate, since it gets hidden above a more general and less informative exception message:
```
Exception: CPM_exact: Install the python package 'exact' to use this solver interface.
```
and a Traceback:
```
/mnt/c/Users/thoma/Documents/KULeuven/cpmpy/cpmpy/solvers/exact.py:77: UserWarning: CPMpy requires Exact version >=2.1.0 is required but you have version 1.2.1, beware exact>=2.1.0 requires Python 3.10 or higher.
  warnings.warn(f"CPMpy requires Exact version >=2.1.0 is required but you have version "
  File "/mnt/c/Users/thoma/Documents/KULeuven/cpmpy/tst_solver_installed.py", line 20, in <module>
    model.solve(solver=solver)
  File "/mnt/c/Users/thoma/Documents/KULeuven/cpmpy/cpmpy/model.py", line 159, in solve
    s = SolverLookup.get(solver, self)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/thoma/Documents/KULeuven/cpmpy/cpmpy/solvers/utils.py", line 107, in get
    return solver_cls(model, subsolver=subname)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/thoma/Documents/KULeuven/cpmpy/cpmpy/solvers/exact.py", line 102, in __init__
    raise Exception("CPM_exact: Install the python package 'exact' to use this solver interface.")
Exception: CPM_exact: Install the python package 'exact' to use this solver interface.
```

### DIFF
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -78,7 +78,7 @@ class CPM_choco(SolverInterface):
             import pychoco as chc
             # check it's the correct version
             # CPMPy uses features only available from 0.2.1
-            pkg_resources.require("pychoco>=2.1.0")
+            pkg_resources.require("pychoco>=0.2.1")
             return True
         except ModuleNotFoundError:
             return False
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -32,6 +32,10 @@ import time
 
 import numpy as np
 
+import warnings
+import pkg_resources
+from pkg_resources import VersionConflict
+
 from ..transformations.normalize import toplevel_list
 from .solver_interface import SolverInterface, SolverStatus, ExitStatus
 from ..expressions.core import Expression, Comparison, Operator, BoolVal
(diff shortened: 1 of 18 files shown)


cpmpy/solvers/choco.py
```diff
@@ -70,19 +74,20 @@ class CPM_choco(SolverInterface):
...
-                warnings.warn(f"CPMpy uses features only available from Pychoco version 0.2.1, but you have version {pychoco_version}")
-                return False
+            pkg_resources.require("pychoco>=0.2.1")
             return True
-        except ImportError:
+        except ModuleNotFoundError:
+            return False
+        except VersionConflict: # unsupported version of pychoco
+            warnings.warn(f"CPMpy uses features only available from Pychoco version 0.2.1, "
+                          f"but you have version {pkg_resources.get_distribution('exact').version}.")
```
Wout4, line 87: should be choco in stead of exact?

Wout4: pychoco i guess

cpmpy/solvers/gcs.py
```diff
@@ -70,8 +70,10 @@ def supported():
         try:
             import gcspy
             return True
-        except ImportError as e:
+        except ModuleNotFoundError:
             return False
+        except Exception as e:
```
Wout4, line 75: no version conflict here, because we have no version requirements?

cpmpy/solvers/choco.py
ThomSerg: indeed

cpmpy/solvers/gcs.py
ThomSerg: Yes, `pkg_resources.require("...")` is the one that raises `VersionConflict` in the other solvers. For gcspy we currently don't require a specific version, and thus that exception case never occurs.

Wout4:
nice, will merge

PR #569, opened 2025-01-08
Improve normalization of Boolean terms for PySAT

### LINKED ISSUE #552: var comp sum optimisation
bv0 > bv2 + bv3
creates an intvar for the lhs instead of normalising as:
bv0 - (bv2 + bv3) > 0

This is a blocker for #114 which should stay in pseudo-boolean space.

should probably be handled somewhere in flatten's normalize_boolexpr (because the '>' comparison is bool); not sure.

```
import cpmpy as cp
bvs = cp.boolvar(3)
c = bvs[0] > sum(bvs[1:])
s = cp.SolverLookup.get("pysat")
print(c, "==>", s.transform(c))
s += c
s.solve()
```

hbierlee:
Fixes #552. In two cases, normalization of linear constraints does not lead to the supported cardinality constraint for PySAT, crashing with `NotImplemented` exception:

- `a+b-c<0`: negative coefficients leads to a `wsum`, but we need `sum`; by adding a transform function which rewrites such terms, this now leads to `a+b+~c<1`
- `a<b+c`:  the original issue, where this was previously not normalized to `(b+c)<a` (as it is done for `Expr == Var` in `flatten_constraint`). This then introduces a IntVar `x=b+c,a<x`, and integer variables are not supported by PySAT. By also normalizing to `a>(b+c)` (mind the flipped comparator), this is fixed automatically.

Currently draft since the rcpsp example breaks for unknown reason (and perhaps other solvers as well, waiting on CI)

### DIFF
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -78,8 +78,10 @@ class CPM_pysat(SolverInterface):
             from pysat.formula import IDPool
             from pysat.solvers import Solver
             return True
-        except ImportError as e:
+        except ModuleNotFoundError:
             return False
+        except Exception as e:
+            raise e
 
 
     @staticmethod
@@ -113,7 +115,7 @@ class CPM_pysat(SolverInterface):
             see .solvernames() to get the list of available solver(names)
         """
         if not self.supported():
-            raise Exception("CPM_pysat: Install the python 'python-sat' package to use this solver interface "
+            raise Exception("CPM_pysat: Install the python package 'python-sat' to use this solver interface "
                             "(NOT the 'pysat' package!)")
         if cpm_model and cpm_model.objective_ is not None:
             raise NotSupportedError("CPM_pysat: only satisfaction, does not support an objective function")
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -42,66 +42,13 @@ from ..exceptions import NotSupportedError
 from ..expressions.core import Comparison, Operator, BoolVal
 from ..expressions.variables import _BoolVarImpl, NegBoolView, boolvar
 from ..expressions.globalconstraints import DirectConstraint
-from ..transformations.linearize import canonical_comparison
+from ..transformations.linearize import canonical_comparison, only_positive_coefficients
 from ..expressions.utils import is_int, flatlist
 from ..transformations.decompose_global import decompose_in_tree
 from ..transformations.get_variables import get_variables
 from ..transformations.flatten_model import flatten_constraint
 from ..transformations.normalize import toplevel_list, simplify_boolean
 from ..transformations.reification import only_implies, only_bv_reifies
-from ..expressions.utils import eval_comparison
-
-
-# TODO [?] where should we put this transformation function?
-# TODO [?] are zero coefficients removed in linearize? Probably better to do here than there (right now this is actually `only_non_negative_terms`).
-def only_positive_coefficients(lst_of_expr):
-    """
-        Replaces Boolean terms with negative coefficients in linear constraints with terms with positive coefficients by negating its literal.
-        This can simplify a wsum into sum.
-        cpm_expr is expected to be linearized. Only apply after applying linearize_constraint(cpm_expr)
-
-        Resulting expression is linear.
-    """
-    newlist = []
-    for cpm_expr in lst_of_expr:
-        print(f"{cpm_expr=}")
-
-        if isinstance(cpm_expr, Comparison):
-            lhs, rhs = cpm_expr.args
-            new_cons = []
-
-            #    ... -c*b + ... <= k
-            # :: ... -c*(1 - ~b) + ... <= k
-            # :: ... -c + c* ~b + ... <= k
-            # :: ... + c*~b + ... <= k+c
-            if lhs.name == "wsum":
-                weights, args = lhs.args
-                idxes = {i for i, (w, a) in enumerate(zip(weights, args)) if w < 0 and isinstance(a, _BoolVarImpl)}
-                nw, na = zip(*[(-w, ~a) if i in idxes else (w, a) for i, (w, a) in enumerate(zip(weights, args))])
-                rhs += sum(-weights[i] for i in idxes)
-
-                # Simplify wsum to sum if all weights are 1
-                if all(w == 1 for w in nw):
-                    lhs = Operator("sum", [list(na)])
-                else:
-                    lhs = Operator("wsum", [list(nw), list(na)])
-
-            newlist.append(eval_comparison(cpm_expr.name, lhs, rhs)) # TODO [?] why eval_comparison?
-            newlist += new_cons # TODO [?] only_positive_bv linearizes again?
-
-        # reification
-        elif cpm_expr.name == "->":
-            cond, subexpr = cpm_expr.args
-            assert isinstance(cond, _BoolVarImpl), f"{cpm_expr} is not a supported linear expression. Apply " \
-                                                   f"`linearize_constraint` before calling `only_positive_bv` "
-            if isinstance(cond, _BoolVarImpl): # BV -> Expr
-                subexpr = only_positive_coefficients([subexpr])
-                newlist += [cond.implies(expr) for expr in subexpr]
-
-        else:
-            newlist.append(cpm_expr)
-
-    return newlist
 
 
 class CPM_pysat(SolverInterface):
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -39,15 +39,69 @@
 """
 from .solver_interface import SolverInterface, SolverStatus, ExitStatus
 from ..exceptions import NotSupportedError
-from ..expressions.core import Expression, Comparison, Operator, BoolVal
+from ..expressions.core import Comparison, Operator, BoolVal
 from ..expressions.variables import _BoolVarImpl, NegBoolView, boolvar
 from ..expressions.globalconstraints import DirectConstraint
+from ..transformations.linearize import canonical_comparison
 from ..expressions.utils import is_int, flatlist
 from ..transformations.decompose_global import decompose_in_tree
 from ..transformations.get_variables import get_variables
 from ..transformations.flatten_model import flatten_constraint
 from ..transformations.normalize import toplevel_list, simplify_boolean
 from ..transformations.reification import only_implies, only_bv_reifies
+from ..expressions.utils import eval_comparison
+
+
+# TODO [?] where should we put this transformation function?
+# TODO [?] are zero coefficients removed in linearize? Probably better to do here than there (right now this is actually `only_non_negative_terms`).
+def only_positive_coefficients(lst_of_expr):
+    """
+        Replaces Boolean terms with negative coefficients in linear constraints with terms with positive coefficients by negating its literal.
+        This can simplify a wsum into sum.
+        cpm_expr is expected to be linearized. Only apply after applying linearize_constraint(cpm_expr)
+
+        Resulting expression is linear.
+    """
+    newlist = []
+    for cpm_expr in lst_of_expr:
+        print(f"{cpm_expr=}")
+
+        if isinstance(cpm_expr, Comparison):
+            lhs, rhs = cpm_expr.args
+            new_cons = []
+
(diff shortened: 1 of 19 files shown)


tests/test_constraints.py
```diff
@@ -142,7 +143,7 @@ def comp_constraints(solver):
                 if comp_name == "<" and lb >= get_bounds(rhs)[0]:
                     continue
                 yield Comparison(comp_name, numexpr, rhs)
-
+                yield Comparison(comp_name, rhs, numexpr)
```
Wout4, line 146: this is generating testcases with no solutions: 1 < boolexpr has no solutions but boolexpr < 1 does..

hbierlee: I've resolved this now by copying the new normalization code. But this is a bit ugly because the test does exactly what is implemented.

What I think the real solution is: rather than have complicated logic here to ensure satisfiability, is to brute force and cache the solution sets for all these test cases. Brute forcing can be done by generating all assignments for a test case model, and filtering on `all(cons.value() for cons in model)`. Caching can be done via pytest so that it's only done once, and then for new tests.

Upsides:
* Soundness checking is way more comprehensive, because you *also* test for unsatisfiability, and test for missing solutions (if `ALL_SOLS`) rather than only "extra solutions" (which fail constraints).
* Writing tests cases becomes more easy (the above `"if *unsat* : continue"` type statements can be removed)
* (Not sure, but:) Testing may be slightly faster because you don't check constraints, you just check set membership

Downsides:
* takes some engineering to set up properly :)
* brute-forcing may not be possible for bigger test case models (we could fall back on a real solver, but this has some issues if there are bugs in the solver or cpmpy transformation)
* it still relies on the `constraint.value()` functions to be sound, but this is what test rely on anyways (this will also uncover bugs regarding those, so arguably an upside)

hbierlee:
Ready for review (@Wout4  @IgnaceBleukx not sure who)! There are some `TODO [?]` which are basically questions I had, these should be processed and then removed.

cpmpy/solvers/pysat.py
```diff
@@ -39,15 +39,69 @@
...
 from ..transformations.decompose_global import decompose_in_tree
 from ..transformations.get_variables import get_variables
 from ..transformations.flatten_model import flatten_constraint
 from ..transformations.normalize import toplevel_list, simplify_boolean
 from ..transformations.reification import only_implies, only_bv_reifies
+from ..expressions.utils import eval_comparison
+
+
+# TODO [?] where should we put this transformation function?
+# TODO [?] are zero coefficients removed in linearize? Probably better to do here than there (right now this is actually `only_non_negative_terms`).
```
Wout4, line 56: I don't remember why we made everything into wsums in the first place, but it feels a bit like this transformation is just undoing a previous transformation. Maybe we can just not make the wsum when it involves boolvars without coefficients?
In general I think we don't want transformations to be in the solver interface as well, so this one could be in the linearize file imo.

cpmpy/solvers/pysat.py
```diff
@@ -39,15 +39,69 @@
...
+                nw, na = zip(*[(-w, ~a) if i in idxes else (w, a) for i, (w, a) in enumerate(zip(weights, args))])
+                rhs += sum(-weights[i] for i in idxes)
+
+                # Simplify wsum to sum if all weights are 1
+                if all(w == 1 for w in nw):
+                    lhs = Operator("sum", [list(na)])
+                else:
+                    lhs = Operator("wsum", [list(nw), list(na)])
+
+            newlist.append(eval_comparison(cpm_expr.name, lhs, rhs)) # TODO [?] why eval_comparison?
```
Wout4, line 89: It's just a helper function that creates a new comparison expression

cpmpy/solvers/pysat.py
Wout4, line 56: I don't think we handle 0-coefficients at all, I hope we never make them?

cpmpy/solvers/pysat.py
```diff
@@ -39,15 +39,69 @@
...
+from ..expressions.utils import eval_comparison
+
+
+# TODO [?] where should we put this transformation function?
+# TODO [?] are zero coefficients removed in linearize? Probably better to do here than there (right now this is actually `only_non_negative_terms`).
+def only_positive_coefficients(lst_of_expr):
+    """
+        Replaces Boolean terms with negative coefficients in linear constraints with terms with positive coefficients by negating its literal.
+        This can simplify a wsum into sum.
+        cpm_expr is expected to be linearized. Only apply after applying linearize_constraint(cpm_expr)
```
Wout4, line 61: you are not applying it after linearize?

Wout4:
So i think the second case is handled well in flatten, but the for the second case I think we can either fix it be just not turning those expressions into wsums in the first place, or we should maybe handle these cases inside of canonical_comparison, because that's basically what it is supposed to do, put all comparisons in some canonical form. That transformation never got to a nice place still..

cpmpy/solvers/pysat.py
hbierlee: I think linearize makes wsums because it's what a linear solver like gurobi expects. Note I'm not calling linearize in pysat's transform (exactly to avoid the wsum), I just call `canonical_comparison` to get all vars on the LHS. So I don't think this new "positive_coefficients" transform really undoes anything, but I could be wrong.

I agree I should move it to linearize anyway, even if PySAT is the only one that uses this transform for now.

hbierlee: The user might make them, I guess, and then I hope they should be removed at some early stage in the transforms. Let me know if I should add a line of code to do this in flatten constraint or something.

cpmpy/solvers/pysat.py
hbierlee: Ah true. Forgot to update this. Above comment about canonical comparison is related

hbierlee:
Ok, to decide about the second case: I agree it'd be better if we don't turn it into wsum in the first place. What I discussed yesterday with Ignace is that we are stuck because the canonical expression for Gurobi will have negative coefficients, positive literals (BoolVar's) while PySAT wants only cardinality constraints but with negative literals. So there's a split there. Not sure if we can resolve both with one function, so that's why there is now this separate function. But open to ideas:


```
|| only_implies = [((BV11) + (BV12)) < (BV10)]
|| canonical_comparison = [sum([1, 1, -1] * [BV11, BV12, BV10]) < 0]
|| only_positive_coefficients = [sum([BV11, BV12, ~BV10]) < 1]
```

cpmpy/solvers/pysat.py
hbierlee: I guess see discussion below

tests/test_constraints.py
IgnaceBleukx: Let's indeed leave that for another PR ;) I'll update the testsuite here to properly check for always UNSAT constraints, it should not be too hard to do as we have the "get_bounds" helper functions. I'll push it directly here.

hbierlee: I will move this to a feature request (probably I will pick this up when we get to CPMpy auto-grading, or it could be related to our fuzztesting extensions. It's all very adjacent.)

cpmpy/solvers/pysat.py
hbierlee: Since this won't cause bugs, we can resolve 0 coefficients another day.

hbierlee:
So there #[7527f67](https://github.com/CPMpy/cpmpy/pull/569/commits/7527f678684ba5d97345ace7c007e2586ad2acd5) initially failed (`tests/test_solvers.py:tests/test_solvers.py`) with a sub-optimal result for some unknown solver, I guess this was a timeout somehow? Rerunning the job fixed it (but in the meantime I also parametrized the tests in that file, that might also help).

Anyway I think this is ready to finalize and merge. The only remaining question was whether this new transform for only pos. coefficients can somehow be combined with `canonical_comparison`, but the problem is that pysat and gurobi have different ideas of canonical.

IgnaceBleukx:
Hmm I don't think the major changes to `test_solvers` belong in this PR, I would rather cherry-pick those commits and put them in another PR that only updates the tests. Other than that, the code looks fine to me.

IgnaceBleukx:
Hmm, `canonical_comparison` is about the form of the comparison itself (only constants on the rhs), whereas this is more at the level of `only_positive_bv`. So I'm actually fine with keeping it the way it's implemented now.
First putting all variables to the lhs, then swapping polarity of Boolean variables to get rid of negative coefficients.

hbierlee:
Split out parametrizing tests into separate PR

cpmpy/transformations/linearize.py
```diff
@@ -365,3 +365,52 @@ def canonical_comparison(lst_of_expr):
...
+                rhs += sum(-weights[i] for i in idxes)
+
+                # Simplify wsum to sum if all weights are 1
+                if all(w == 1 for w in nw):
+                    lhs = Operator("sum", [list(na)])
+                else:
+                    lhs = Operator("wsum", [list(nw), list(na)])
+
+            newlist.append(eval_comparison(cpm_expr.name, lhs, rhs))
+            newlist += new_cons
```
IgnaceBleukx, line 401: Left over from copy-pasting? Seems unused.

cpmpy/transformations/linearize.py
```diff
@@ -365,3 +365,52 @@ def canonical_comparison(lst_of_expr):
...
+
+        # reification
+        elif cpm_expr.name == "->":
```
IgnaceBleukx, line 404: To be more complete: `if isinstance(cpm_expr, Operator) and cpm_expr.name == "->"`?
Otherwise you might run into the (stupid) edge case where a user makes the variable `cp.boolvar(name="->")`

cpmpy/transformations/linearize.py
```diff
@@ -365,3 +365,52 @@ def canonical_comparison(lst_of_expr):
...
+            cond, subexpr = cpm_expr.args
+            assert isinstance(cond, _BoolVarImpl), f"{cpm_expr} is not a supported linear expression. Apply " \
+                                                   f"`linearize_constraint` before calling `only_positive_bv` "
```
IgnaceBleukx, line 407: Not 'only_positive_bv`, but `only_positive_coefficients`

cpmpy/transformations/linearize.py
```diff
@@ -365,3 +365,52 @@ def canonical_comparison(lst_of_expr):
...
+            if isinstance(cond, _BoolVarImpl): # BV -> Expr
```
IgnaceBleukx, line 408: You just asserted this, so if-statement can be removed : )

tests/test_pysat_cardinality.py
```diff
@@ -125,6 +127,26 @@ def test_pysat_card_implied(self):
...
+    def test_only_positive_coefficients(self):
+        a, b, c = [cp.boolvar(name=n) for n in "abc"]
+        only_pos = only_positive_coefficients([Operator("wsum",[[1,1,-1],[a,b,c]]) > 0])
+        self.assertEqual(str([Operator("sum",[a, b, ~c]) > 1]), str(only_pos))
+
+    def test_pysat_aggregate_sum_sub_expressions(self):
+        bvs = cp.boolvar(3)
+        c = bvs[0] > sum(bvs[1:])
+        s = cp.SolverLookup.get("pysat")
+        print(c, "==>", s.transform(c))
```
IgnaceBleukx, line 147: Stray print

tests/test_pysat_cardinality.py
```diff
@@ -125,6 +127,26 @@ def test_pysat_card_implied(self):
...
+        s += c
+        s.solve()
```
IgnaceBleukx, line 150: Can you also add some tests where the sum contains a mix of integer and boolean variables?

IgnaceBleukx (changes requested):
A few copy-paste artifacts remain, after that: fine to merge.

cpmpy/transformations/linearize.py
hbierlee: Also fixing this and the needless `if` condition after insertion in `only_postive_if` code

hbierlee (commented):
Processed all comments

hbierlee (commented):
Mark review complete (?)

IgnaceBleukx (approved):

PR #571, opened 2025-01-10
Distribute tests over 4 available CPUs (test suite from `21m30s -> 8m45s`)

hbierlee:
Using [`pytest-xdist`](https://pytest-xdist.readthedocs.io/en/stable/), we can run the test suite over multiple CPUs (github provides [4](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners)), reducing the CI time by quite a lot. There is one downside (apart from having the extra dependency in itself), which is that tests need to be deterministic (ie all collected tests need to be the same and in the same order, https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html). So for parametrized tests, that means you need to make sure to sort some test generation code.

### DIFF
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -127,7 +127,7 @@ def comp_constraints(solver):
                            Numexpr != Constant             (CPMpy class 'Comparison')
         - Numeric inequality (>=,>,<,<=): Numexpr >=< Var  (CPMpy class 'Comparison')
     """
-    for comp_name in Comparison.allowed:
+    for comp_name in sorted(Comparison.allowed):
 
         for numexpr in numexprs(solver):
             # numeric vs bool/num var/val (incl global func)
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8 pytest requests pandas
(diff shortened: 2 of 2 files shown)


tests/test_constraints.py
```diff
@@ -127,7 +127,7 @@ def comp_constraints(solver):
                            Numexpr != Constant             (CPMpy class 'Comparison')
         - Numeric inequality (>=,>,<,<=): Numexpr >=< Var  (CPMpy class 'Comparison')
     """
-    for comp_name in Comparison.allowed:
+    for comp_name in sorted(Comparison.allowed):
```
Wout4, line 130: I guess we need to check everywhere that we don't loop over unordered collections?
there is also Operator.allowed for example (probably those are the only 2 cases)

hbierlee: Actually, I rememberd that this works because `Operator.allowed` is `dict` (while `Comparator.allowed` is a `set`), and `dict`'s nowawadays preserve insertion order when iterating (https://docs.python.org/3.7/library/stdtypes.html#mapping-types-dict) since Python 3.7.

What is our minimally supported python version? `setup.py` has `>=3.6`? In that case, I should add `sorted`.

Wout4: I guess this might be a reason to make that >= 3.7 ?

Wout4:
made a pr to update our reqs, will merge this

PR #572, opened 2025-01-10
Parametrize test solvers tests

hbierlee:
A little nicer, and might prevent the optimization test failing for some sub-optimal result (which happened once), or at least tell us which solver was the problem


Wout4:
Nice, I like it this way, also seems to add some testcases so I guess not all of them were done for all solvers before

Wout4:
Will merge

PR #573, opened 2025-01-14
update minimal python version

Wout4:
Don't think we changed this setup file in a long time, just did some logical updates


IgnaceBleukx (approved):
Which features in Python 3.7 do we require that are not in 3.6?
Or is this a package-included issue?

Maybe also mention it somewhere in the docs? Not sure if it's there yet.

Wout4:
running the unit tests in parallel requires 3.7 apparently (because that's when looping over a dictionary became deterministic)

PR #576, opened 2025-01-17
Cp optimizer

Wout4:
New solver

Requires installing the CPLEX Optimization Studio.
Would be nice if someone can check if the executable_installed() function works as expected before and after installing the executable :)

### DIFF
--- a/cpmpy/solvers/cpo.py
+++ b/cpmpy/solvers/cpo.py
@@ -94,18 +94,12 @@ class CPM_cpo(SolverInterface):
             # [GUIDELINE] Add link to documentation of all solver parameters
         """
 
-        # ensure all vars are known to solver
-        # self.solver_vars(list(self.user_vars)) don't think this is needed for cpo
-
-        if time_limit is not None:
-            self.cpo_model.set_parameters(docplex.cp.parameters.CpoParameters(TimeLimit=time_limit))
-
         # [GUIDELINE] if your solver supports solving under assumptions, add `assumptions` as argument in header
         #       e.g., def solve(self, time_limit=None, assumptions=None, **kwargs):
         #       then translate assumptions here; assumptions are a list of Boolean variables or NegBoolViews
 
         # call the solver, with parameters
-        self.cpo_result = self.cpo_model.solve(**kwargs)
+        self.cpo_result = self.cpo_model.solve(LogVerbosity='Quiet', TimeLimit=time_limit, **kwargs)
         # [GUIDELINE] consider saving the status as self.TPL_status so that advanced CPMpy users can access the status object.
         #       This is mainly useful when more elaborate information about the solve-call is saved into the status
 
@@ -157,6 +151,47 @@ class CPM_cpo(SolverInterface):
 
         return has_sol
 
+    def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from_model=False, **kwargs):
+        """
+            A shorthand to (efficiently) compute all (optimal) solutions, map them to CPMpy and optionally display the solutions.
+
+            If the problem is an optimization problem, returns only optimal solutions.
+
+           Arguments:
+                - display: either a list of CPMpy expressions, OR a callback function, called with the variables after value-mapping
+                        default/None: nothing displayed
+                - time_limit: stop after this many seconds (default: None)
+                - solution_limit: stop after this many solutions (default: None)
+                - call_from_model: whether the method is called from a CPMpy Model instance or not
+                - any other keyword argument
+
+            Returns: number of solutions found
+        """
+
+        # check if objective function
+        if self.has_objective():
+            raise NotSupportedError("TEMPLATE does not support finding all optimal solutions")
+
+        # A. Example code if solver supports callbacks
+        if is_any_list(display):
+            callback = lambda: print([var.value() for var in display])
+        else:
+            callback = display
+
+        self.cpo_model.start_search(TimeLimit=time_limit, SolutionLimit=solution_limit, **kwargs)
+        #self.cpo_model.start_search(params=[docplex.cp.parameters.CpoParameters(TimeLimit=time_limit)], **kwargs)
+        # clear user vars if no solution found
+        if self.TPL_solver.SolutionCount() == 0:
+            for var in self.user_vars:
+                var.clear()
+        return self.TPL_solver.SolutionCount()
+
+        # clear user vars if no solution found
+        if solution_count == 0:
+            for var in self.user_vars:
+                var.clear()
+
+        return solution_count
 
     def solver_var(self, cpm_var):
         """
@@ -197,21 +232,13 @@ class CPM_cpo(SolverInterface):
 
             are permanently posted to the solver)
         """
-        # make objective function non-nested
-        #(flat_obj, flat_cons) = flatten_objective(expr)
-        #self += flat_cons # add potentially created constraints
-        #self.user_vars.update(get_variables(flat_obj)) # add objvars to vars
-
-        # make objective function or variable and post
-        #obj = self._make_numexpr(flat_obj)
-        # [GUIDELINE] if the solver interface does not provide a solver native "numeric expression" object,
-        #         _make_numexpr may be removed and an objective can be posted as:
-        #           self.TPL_solver.MinimizeWeightedSum(obj.args[0], self.solver_vars(obj.args[1]) or similar
-
+        if self.has_objective():
+            self.cpo_model.remove(self.cpo_model.get_objective_expression())
+        expr = self._cpo_expr(expr)
         if minimize:
-            self.docplex.cp.modeler.minimize(expr)
+            self.cpo_model.add(docplex.cp.modeler.minimize(expr))
         else:
-            self.docplex.cp.modeler.maximize(expr)
+            self.cpo_model.add(docplex.cp.modeler.maximize(expr))
 
     def has_objective(self):
         return self.cpo_model.get_objective() is not None
@@ -362,8 +389,6 @@ class CPM_cpo(SolverInterface):
                 elif cpm_con.name == "div":
                     return x // y
                 elif cpm_con.name == "pow":
-                    '''if not is_num(cpm_con.args[1]): # TODO is this supported with exponent expression?
-                        raise NotSupportedError(f"CPO only supports power constraint with constant exponent, got {cpm_con}")'''
                     return x ** y
                 elif cpm_con.name == "mod":
                     return x % y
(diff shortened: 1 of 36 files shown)


ThomSerg:
Checked the `executable_installed()`, which seems to function as expected both before and after installing `docplex` / `CPLEX Optimization Studio`

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+from ..expressions.variables import _BoolVarImpl, NegBoolView, _IntVarImpl, _NumVarImpl
+from ..expressions.utils import is_num, is_any_list, eval_comparison, argval, argvals
+from ..transformations.get_variables import get_variables
+from ..transformations.normalize import toplevel_list
+from ..transformations.decompose_global import decompose_in_tree
+
+"""
+    Interface to CP Optimizers API
+
+    <some information on the solver>
```
tias, line 17: yes : )
From their own docs:
CP Optimizer, also a feature of IBM ILOG Optimization Studio, is a software library of constraint programming tools supporting constraint propagation, domain reduction, and highly optimized solution search.

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+        # try to import the package
+        try:
+            import docplex.cp as docp
+            return True
+        except ModuleNotFoundError:
+            return False
+
+    @staticmethod
+    def executable_installed():
+        if shutil.which('cpoptimizer') is None:
```
tias, line 68: I installed it and also the python link, but I did not bother setting up the CLI PATH. So this fails, even though my docplex can model and solve perfectly...

I think this has to be more similar to the gurobi solver interface instaed, where we have a 'license_OK()' function that checks something purely through the python API. If you can't find a way then for all I care it just does docplex.mp.model.Model().solve() or smth...

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+            raise Exception("CPM_cpo: Install the python package 'docplex'")
+
+        if not self.executable_installed():
+            raise Exception("You need to install the CPLEX Optimization Studio to use this solver.")
+
+        import docplex.cp.model as dom
+        assert subsolver is None
+
+        # initialise the native solver object
+        # [GUIDELINE] we commonly use 3-letter abbrivations to refer to native objects:
```
tias, line 91: please remove all [GUIDELINE] comments

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+    def solve(self, time_limit=None, **kwargs):
+        """
+            Call the CP Optimizer solver
+
+            Arguments:
+            - time_limit:  maximum solve time in seconds (float, optional)
+            - kwargs:      any keyword argument, sets parameters of solver object
+
+            Arguments that correspond to solver parameters:
+            # [GUIDELINE] Please document key solver arguments that the user might wish to change
```
tias, line 110: some key solver args would be nice... e.g. log_output is one, maybe smth wrt nr of threads?

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+            #       for example: assumptions=[x,y,z], log_output=True, var_ordering=3, num_cores=8, ...
+            # [GUIDELINE] Add link to documentation of all solver parameters
+        """
+
+        # [GUIDELINE] if your solver supports solving under assumptions, add `assumptions` as argument in header
+        #       e.g., def solve(self, time_limit=None, assumptions=None, **kwargs):
+        #       then translate assumptions here; assumptions are a list of Boolean variables or NegBoolViews
+
+        # call the solver, with parameters
+        self.cpo_result = self.cpo_model.solve(LogVerbosity='Quiet', TimeLimit=time_limit, **kwargs)
```
tias, line 120: by setting LogVerbosity here it can't be overwritten potentially? If you add LogVerbosity='Quiet' to our def solve() then perhaps it would?

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+                - any other keyword argument
+
+            Returns: number of solutions found
+        """
+
+        import docplex.cp as docp
+        #cpo_solver = self.cpo_model.start_search(TimeLimit=time_limit, SolutionLimit=solution_limit, LogVerbosity='Quiet', **kwargs)
+        solution_count = 0
+        if solution_limit is None:
+            solution_limit = 99999
```
tias, line 193: that is not advisable... we can solve nqueens over 100k in a very fast way.
Change the 'while' below to 'while solution_limit is None or solution_count < lmt'?

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+        while solution_count < solution_limit:
+            cpo_result = self.solve(time_limit=time_limit, **kwargs)
+            if not cpo_result:
+                break
+            solution_count += 1
+            '''if cpo_result.is_solution_optimal():
```
tias, line 199: not sure what this means?

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+            self.cpo_model.add(docp.modeler.forbidden_assignments(solvars, [vals]))
+            if display is not None:
+                if isinstance(display, Expression):
+                    print(argval(display))
+                elif isinstance(display, list):
+                    print(argvals(display))
+                else:
+                    display()  # callback
+        return solution_count
+    def solver_var(self, cpm_var):
```
tias, line 228: empty line inbetween missing

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+        """
+            Creates solver variable for cpmpy variable
+            or returns from cache if previously created
+        """
+        import docplex.cp as docp
```
tias, line 233: this might hurt performance.... we are going to call `solver_var()` many many times, and if each requires an import check...

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+        expr = self._cpo_expr(expr)
+        if minimize:
+            self.cpo_model.add(dom.minimize(expr))
+        else:
+            self.cpo_model.add(dom.maximize(expr))
+
+    def has_objective(self):
+        return self.cpo_model.get_objective() is not None
+
+    def _make_numexpr(self, cpm_expr):
```
tias, line 280: given the use of 'TPL' inside, I guess this code is not needed for CPO?

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+        """
+        # apply transformations
+        cpm_cons = toplevel_list(cpm_expr)
+        # count is only supported with a constant to be counted, so we decompose
+        supported = {"alldifferent", 'inverse', 'nvalue', 'element', 'table', 'indomain',
+                     "negative_table", "gcc", 'max', 'min', 'abs'}
+        supported_reified = {"alldifferent", 'nvalue', 'element', 'table', 'indomain', 'max', 'min',
+                     "negative_table", 'abs'}
+        cpm_cons = decompose_in_tree(cpm_cons, supported=supported, supported_reified=supported_reified)
+        '''cpm_cons = flatten_constraint(cpm_cons)  # flat normal form
```
tias, line 337: so, without flattening? you can remove this then, maybe add a comment line that it indeed does not require flattening...

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,491 @@
...
+                return dom.min(self._cpo_expr(cpm_con.args))
+            elif cpm_con.name == "max":
+                return dom.max(self._cpo_expr(cpm_con.args))
+            elif cpm_con.name == "abs":
+                return dom.abs(self._cpo_expr(cpm_con.args)[0])
+            elif cpm_con.name == "nvalue":
+                return dom.count_different(self._cpo_expr(cpm_con.args))
+
+
+
```
tias, line 490: some excessive newlines : )

cpmpy/solvers/gurobi.py
```diff
@@ -34,7 +34,7 @@
 from .solver_interface import SolverInterface, SolverStatus, ExitStatus
 from ..exceptions import NotSupportedError
 from ..expressions.core import *
-from ..expressions.utils import argvals
+from ..expressions.utils import argvals, argval
```
tias, line 37: should not be in this commit? or needed at all I guess?

tias (changes requested):
great!
couldn't test it yet because of the installed check (see comments) but looks sweet

cpmpy/solvers/cpo.py
Wout4: Ah yes, still some comments that i have to clean up it seems

cpmpy/solvers/cpo.py
Wout4: Ah yes, should check if it's in the kwargs first!

cpmpy/solvers/cpo.py
Wout4: gurobi also has an import here

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,470 @@
...
+    Interface to CP Optimizers API
+
+    Requires that the 'docplex' python package is installed:
+    $ pip install docplex
+
+    docplex documentation:
+    https://ibmdecisionoptimization.github.io/docplex-doc/
+
+    You will also need to install CPLEX Optimization Studio from IBM's website.
+    There is a free community version available.
```
tias, line 45: maybe add 'and add the location of the cpoptimizer binary to your path'?

I did not think it would be possible that if you do a local pip install, you would still have to add something to the path... but hey, turns out we do ( so your original installer check was correct, I perhaps just tested the MP one instaed of the CP one...)

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,470 @@
...
+    Creates the following attributes (see parent constructor for more):
+    - cpo_model: object, CP Optimizers model object
+    """
+
+    _docp = None  # Static attribute to hold the docplex.cp module
+
+    @classmethod
+    def get_docp(cls):
+        if cls._docp is None:
+            import docplex.cp as docp  # Import only once
```
tias, line 60: very elegant... perhaps something we should do in the other solvers too? To discuss...

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,470 @@
...
+
+    @staticmethod
+    def supported():
+        return CPM_cpo.installed() and CPM_cpo.license_ok()
+
+    @staticmethod
+    def installed():
+        # try to import the package
+        try:
+            import docplex.cp as docp
```
tias, line 72: this could be get_docp() now? saves us one import

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,470 @@
...
+            return False
+
+    @staticmethod
+    def license_ok():
+        if not CPM_cpo.installed():
+            warnings.warn(f"License check failed, python package 'docplex' is not installed! Please check 'CPM_cpo.installed()' before attempting to check license.")
+            return False
+        else:
+            try:
+                from docplex.cp.model import CpoModel
```
tias, line 84: can be self._docp.model?

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,470 @@
...
+        - cpm_model: Model(), a CPMpy Model() (optional)
+        - subsolver: str, name of a subsolver (optional)
+        """
+        if not self.installed():
+            raise Exception("CPM_cpo: Install the python package 'docplex'")
+
+        if not self.license_ok():
+            raise Exception("You need to install the CPLEX Optimization Studio to use this solver.")
+
+        import docplex.cp.model as dom
```
tias, line 105: this also self._docp...?

cpmpy/solvers/cpo.py
```diff
@@ -0,0 +1,470 @@
...
+            See the 'Adding a new solver' docs on readthedocs for more information.
+
+        :param cpm_expr: CPMpy expression, or list thereof
+        :type cpm_expr: Expression or list of Expression
+
+        :return: list of Expression
+        """
+        # apply transformations
+        cpm_cons = toplevel_list(cpm_expr)
+        # count is only supported with a constant to be counted, so we decompose
```
tias, line 302: what about the new no_partial_functions? or can it handle that?

tias (approved):
Very clean, very nice...

Some minor comments.

All tests passed on my local machine, so ready to merge after resolving, cool!

cpmpy/solvers/cpo.py
Wout4: i tried that but it somehow does not work within the __init__ function..

cpmpy/solvers/cpo.py
Wout4: should I revert to the other check? then we can also mention the path in the error message.
I think it just worked on my machine by installing the cplex studio

cpmpy/solvers/cpo.py
Wout4: I think so, not all solvers need it but gurobi does

cpmpy/solvers/cpo.py
Wout4: it gets called in __init__ before self.cpo_model is defined

cpmpy/solvers/cpo.py
Wout4: well actually these are static methods so i guess it wouldn't matter anyway

cpmpy/solvers/cpo.py
Wout4: ah yes I thought it didn't work within init but it was probably the static methods that didn't work

cpmpy/solvers/cpo.py
Wout4: ah yes definitely should be added!
I guess it would be good to add some testcases to our test_solvers to check for this. Test will also make clear whether we need to safen toplevel for div/mod

cpmpy/solvers/cpo.py
Wout4: Because these installed checks are static methods we cannot access self

Wout4:
Tested no_overlap and cumulative, looks good thanks to Ignace.
I think the only open comment is about how to do the installed check, after that this is ready to merge

PR #577, opened 2025-01-21
complex sudoku examples

OrestisLomis:
Adding 4 new examples of more complex sudokus; 2 with a some interesting to implement rules and 2 with interesting representations..

(new PR because there were issues merging from a fork)

### DIFF
--- a/examples/sudoku_ratrun1.py
+++ b/examples/sudoku_ratrun1.py
@@ -1,0 +1,131 @@
+import cpmpy as cp
+import numpy as np
+
+# This cpmpy example solves a sudoku by marty_sears, which can be found on https://logic-masters.de/Raetselportal/Raetsel/zeigen.php?id=000IFI
+
+# sudoku cells
+cells = cp.intvar(1,6, shape=(6,6))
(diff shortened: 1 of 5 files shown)


examples/sudoku_ratrun1.py
```diff
@@ -0,0 +1,131 @@
...
+
+def path_valid(path):
+    constraints = []
+    for r in range(path.shape[0]):
+        for c in range(path.shape[0]):
+            if (r,c) != (4,4): # cupcake is the end of path
+                neighbours = get_reachable_neighbours(r, c)
+
+                # for any pathcell, the next pathcell must always be reachable
+                constraints.append((path[r,c] != 0).implies(cp.any([path[*neighbour] == path[r,c] + 1 for neighbour in neighbours])))
```
Wout4, line 72: i get a syntax error on this line

OrestisLomis: What is the syntax error and python version? Seems to work fine on python 3.12

Wout4: Ah yes we currently require python >=3.8

Issue #578, opened 2025-01-23
Wrong float processing


rbtsv:
I encountered the following error:

```
File "/Users/rubtsov/miniconda3/envs/APO/lib/python3.12/site-packages/cpmpy/solvers/ortools.py", line 221, in solve
    assert int(ort_obj_val) == ort_obj_val, "Objective value should be integer, please report on github"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Objective value should be integer, please report on github
```

<img width="1335" alt="Image" src="https://github.com/user-attachments/assets/0fa6b127-35ce-4749-ac98-a5b1dd4e4549" />

The problem is that the processing of the float in this assert is wrong. One shall take into account some tolerance for such a comparison.

Wout4:
We already fixed this issue of not allowing float objective values in https://github.com/CPMpy/cpmpy/pull/529.
So updating to the latest CPMpy release should help.

It will however still return a float and not convert it to an integer in this case because they are not exactly the same.

PR #579, opened 2025-01-23
Check if returned obj is int within tolerance

ThomSerg:
A possible fix for #578, where the assert to check whether the returned obj from OR-Tools is an `int` fails due to machine precision?

Could do it manually using something like:
```
TOLERANCE = 0.0000000001 # some tolerance due to machine precision
if abs(ort_obj_val - round(ort_obj_val)) < TOLERANCE: # it is an integer?
    ...
```
Alternatively, `numpy` provides a nice function `np.isclose()` exactly for this purpose which has a better implementation than the above

```
# This is only pseudo code, the exact implementation is much more involved
def isclose(a, b, rtol=1e-05, atol=1e-08):
    absolute(a - b) <= (atol + rtol * absolute(b))
```

### DIFF
--- a/cpmpy/solvers/ortools.py
+++ b/cpmpy/solvers/ortools.py
@@ -223,7 +223,7 @@ class CPM_ortools(SolverInterface):
             # translate objective
             if self.has_objective():
                 ort_obj_val = self.ort_solver.ObjectiveValue()
-                if round(ort_obj_val) == ort_obj_val: # it is an integer?
+                if np.isclose(ort_obj_val, round(ort_obj_val)): # it is an integer?
                     self.objective_value_ = int(ort_obj_val)  # ensure it is an integer
                 else: # can happen when using floats as coeff in objective
                     self.objective_value_ = float(ort_obj_val)


Wout4:
I wonder if ortools specifies there tolerance on objective bounds? if so it would make sense to reuse the same relative/absolute bounds here

ThomSerg:
If I read the protofile correctly, OR-Tools defaults to 1e-7: [link](https://github.com/google/or-tools/blob/6fc19304dfad3f5105da74c898a2aa0f8b450120/ortools/sat/sat_parameters.proto#L1301)

ThomSerg:
The Python interface seems to reference `primal_bound` and `dual_bound` for the objective value [link](https://github.com/google/or-tools/blob/6fc19304dfad3f5105da74c898a2aa0f8b450120/ortools/math_opt/python/result.py#L113)

In the proto file these seem to default to 1e-7 [link](https://github.com/google/or-tools/blob/6fc19304dfad3f5105da74c898a2aa0f8b450120/ortools/sat/sat_parameters.proto#L1301)

Not sure if I'm looking in the correct places.

PR #582, opened 2025-01-28
accept kwargs for solveAll

Wout4:
We didn't accept kwargs for solveAll in our model object, which doesn't make sense since our solverinterface does accept kwargs.

Not sure if I should add a test since any keyword will be solver specific

### DIFF
--- a/tests/test_solveAll.py
+++ b/tests/test_solveAll.py
@@ -49,3 +49,9 @@ class TestSolveAll(unittest.TestCase):
 
             except NotSupportedError as e:
                 pass # solver does not support finding all optimal solutions
+
+    def test_solveAll_keywords(self):
+        a, b = cp.boolvar(shape=2)
+        m = cp.Model(a | b)
+
+        self.assertEqual(3, m.solveAll('ortools', log_search_progress=True))
--- a/cpmpy/model.py
+++ b/cpmpy/model.py
@@ -164,7 +164,7 @@ class Model(object):
         self.cpm_status = s.status()
         return ret
 
-    def solveAll(self, solver=None, display=None, time_limit=None, solution_limit=None):
+    def solveAll(self, solver=None, display=None, time_limit=None, solution_limit=None, **kwargs):
         """
             Compute all solutions and optionally display the solutions.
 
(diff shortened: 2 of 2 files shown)


tests/test_solveAll.py
```diff
@@ -49,3 +49,9 @@ def test_solveall_with_obj(self):
 
             except NotSupportedError as e:
                 pass # solver does not support finding all optimal solutions
+
+    def test_solveAll_keywords(self):
+        a, b = cp.boolvar(shape=2)
+        m = cp.Model(a | b)
+
+        self.assertEqual(3, m.solveAll('ortools', log_search_progress=True))
```
ThomSerg, line 57: What does this even test? It will fail when an invalid kwarg is passed to a certain solver, so if either a different solver than the requested "ortools" was used  as backend (so issue somewhere else, nothing to do with solveall kwargs) or if "log_search_progress" is not an option for "ortools" (so it tests the test, if we pass the correct kwarg). It maybe tests if somehow our code would change the kwarg's key value.

It does not seem to test what we would actually want: whether the kwarg even gets passed to the solver through the solveAll call.

ThomSerg (commented):
From a purely functional point of view, seems good.

But as you suggested, from a more conceptual point of view, not sure if it makes sense.
The idea of model.solve() is that it is solver agnostic. Want to use a different backend? No problem, just change the "solver" argument and everything works as expected. By allowing for kwargs, its usage now becomes solver dependent. It seems more logical to me that model.solve() only has args that are supported by every solver. But that's just a personal opinion.

tests/test_solveAll.py
Wout4: Hmm I guess on main this would fail because the Model.solveAll() doesn't accept any kwargs, so it indeed just tests our function headers (which is what is being fixed in this pr).
To your other point, I think it's more about allowing advanced users to do solver specific things. But if they do it's also their responsibility to use correct kwargs.
maybe this is unnecessary to add to the Model class, and you should just make a solver object to call solveAll on.  But then why do we support it for Model.solve()?

tias: So the solver `solve()` and `solveAll()` should definitely accept kwargs...

As to `model.solve()`... its true that you are not supposed to make assumptions about the default solver, but you can name one with "solver='ortools'" and then it is valid to use kwargs... e.g. I can do
`m.solve('ortools', log_search_progress=True)` without having to do the solverLookup dance separately...

We could make this more clear by adding a check that if kwargs is not empty, then solver can not be None

and then solveAll should have consistent behaviour on kwargs

Wout4:
Allright I added a check that throws an exception if the solver is not specified but there are kwargs given to model.solve or model.solveAll.
Seems like a good way to make things consistent.
The test might be a bit silly, but I guess it's a good habit to add a testcase that reflects the change.

PR #583, opened 2025-02-04
Cplex solver

Wout4:


cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,458 @@
...
+                - call_from_model: whether the method is called from a CPMpy Model instance or not
+                - any other keyword argument
+
+            Returns: number of solutions found
+        """
+
+        if time_limit is not None:
+            self.cplex_model.set_time_limit(time_limit)
+        if not call_from_model:
+            warnings.warn("Adding constraints to solver object to find all solutions, "
```
Wout4, line 427: this is copied from gurobi, but I don't actually know what it means..

Wout4:
@IgnaceBleukx This is ready to review if you want.
I added support for modulo with a constant to linearize, we can do the same with div but that's not in main yet.
The solver status remains untested for timeouts and errors, not sure if there is a good way to test all of them..

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,459 @@
...
+
+    def solve(self, time_limit=None, **kwargs):
+        """
+            Call the cplex solver
+
+            Arguments:
+            - time_limit:  maximum solve time in seconds (float, optional)
+            - kwargs:      any keyword argument, sets parameters of solver object
+
+            Examples of supported arguments include:
```
IgnaceBleukx, line 137: These long-lines should be shortened massively... Just a concise description of what the parameter does is enough I feel. E.g., the `log_output` can be just the `solver logs are written to stdout` or smt. If a user wants to know the details of a parameter there is a nice link ; )

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,459 @@
...
+        """
+        # ensure all vars are known to solver
+        self.solver_vars(list(self.user_vars))
+
+        if time_limit is not None and not np.isinf(time_limit):
+            self.cplex_model.set_time_limit(time_limit)
+
+        cplex_objective = self.cplex_model.get_objective_expr()
+        self.cplex_model.solve(**kwargs)
+        # all available solve details:
```
IgnaceBleukx, line 155: You mean a user can extract these statistics with `self.cplex_model.solve_details`? If so, might be good to make that explicit

cpmpy/solvers/cplex.py
IgnaceBleukx: We often also add the parameter that controls the number of threads the solver uses. I think such a parameter is more releavant to a use than e.g., `context`

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,459 @@
...
+        cplex_status = self.cplex_model.solve_details.status
+        if cplex_status == "Feasible":
+            self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+        elif "infeasible" in cplex_status:
+            self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+        elif cplex_status == "Unknown":
+            # can happen when timeout is reached...
+            self.cpm_status.exitstatus = ExitStatus.UNKNOWN
+        elif "optimal" in cplex_status:
+            self.cpm_status.exitstatus = ExitStatus.OPTIMAL
```
IgnaceBleukx, line 171: Will this also be "optimal" when no objective function is given? (see #545 )

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,459 @@
...
+            return cpm_expr
+
+        # decision variables, check in varmap
+        if isinstance(cpm_expr, _NumVarImpl):  # _BoolVarImpl is subclass of _NumVarImpl
+            return self.solver_var(cpm_expr)
+
+        # sum
+        if cpm_expr.name == "sum":
+            return self.cplex_model.sum_vars(self.solver_vars(cpm_expr.args))
+        if cpm_expr.name == "sub":
```
IgnaceBleukx, line 273: Move `sub` to under `wsum` as it is probably the least common case

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,459 @@
...
+
+            # Thanks to `only_numexpr_equality()` only supported comparisons should remain
+            if cpm_expr.name == '<=':
+                cplexlhs = self._make_numexpr(lhs)
+                self.cplex_model.add_constraint(cplexlhs <= cplexrhs)
+            elif cpm_expr.name == '>=':
+                cplexlhs = self._make_numexpr(lhs)
+                self.cplex_model.add_constraint(cplexlhs >= cplexrhs)
+            elif cpm_expr.name == '==':
+                if isinstance(lhs, _NumVarImpl) \
```
IgnaceBleukx, line 351: `... and lhs.name in {'sum', 'wsum', 'sub}`

IgnaceBleukx (commented):
Nice, indeed fairly similar interface to Gurobi, the logic seems to be all there, some documentation updates would be welcome.

One thing: I'm surprised they do not support multiplication, as they do support (a form of) division? Or is this division between a var and a constant only?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,459 @@
...
+        # a direct constraint, pass to solver
+        elif isinstance(cpm_expr, DirectConstraint):
+            cpm_expr.callSolver(self, self.cplex_model)
+
+        else:
+            raise NotImplementedError(cpm_expr)  # if you reach this... please report on github
+
+      return self
+
+    def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from_model=False, **kwargs):
```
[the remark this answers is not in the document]
Wout4: implemented it before realising that, think it can be removed indeed

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,459 @@
...
+    See detailed installation instructions at:
+    https://www.ibm.com/docs/en/icos/22.1.2?topic=2212-installing-cplex-optimization-studio
+    Creates the following attributes (see parent constructor for more):
+    - cplex_model: object, CPLEX model object
+    """
+
+    _domp = None  # Static attribute to hold the docplex.cp module
+
+    @classmethod
+    def get_domp(cls):
```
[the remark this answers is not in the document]
Wout4: DOcplex.MP (:

cpmpy/solvers/cplex.py
Wout4: moved it to the docs of .solve, i guess that's the place where it makes most sense?

cpmpy/solvers/cplex.py
Wout4: Yea that seems to be how it works, cplex always defines an objective, it's just a constant if it's not an optimisation problem

cpmpy/solvers/cplex.py
Wout4: it's on the next line (this is a var)

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,459 @@
...
+            cond, sub_expr = cpm_expr.args
+            assert isinstance(cond, _BoolVarImpl), f"Implication constraint {cpm_expr} must have BoolVar as lhs"
+            assert isinstance(sub_expr, Comparison), "Implication must have linear constraints on right hand side"
+            if isinstance(cond, NegBoolView):
+                cond, trigger_val = self.solver_var(cond._bv), 0
+            else:
+                cond, trigger_val = self.solver_var(cond), 1
+
+            lhs, rhs = sub_expr.args
+            if isinstance(lhs, _NumVarImpl) or lhs.name == "sum" or lhs.name == "wsum":
```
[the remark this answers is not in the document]
Wout4: My original answer was that it's not supported reified, but I guess it should be. But then linearizing makes it into a wsum anyway. Not sure why there is 'sub' in other places, I think we need it in make_numexpr because it could be in the objective as well.
Will add it to the supported_reified set though

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+        super().__init__(name="cplex", cpm_model=cpm_model)
+
+    @property
+    def native_model(self):
+        """
+            Returns the solver's underlying native model (for direct solver access).
+        """
+        return self.cplex_model
+
+    def solve(self, time_limit=None, nb_threads=1, **kwargs):
```
IgnaceBleukx: `nb_threads` should not be a default argument here imo, it should just be a `kwarg`

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+            - time_limit:  maximum solve time in seconds (float, optional)
+            - nb_threads:  how many threads to use during solve (int, optional)
+            - kwargs:      any keyword argument, sets parameters of solver object
+
+            Examples of supported arguments include:
+                - context (optional) – context to use during solve
+                - cplex_parameters (optional) – A set of CPLEX parameters to use
+                - checker (optional) – a string which controls which type of checking is performed. (type checks etc.)
+                - log_output (optional) – if True, solver logs are output to stdout.
+                - clean_before_solve (optional) – default False (iterative solving)
```
IgnaceBleukx: I think some more appropriate parameters relating to the actual solving would be more appropriate here

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+            Call the cplex solver
+
+            Arguments:
+            - time_limit:  maximum solve time in seconds (float, optional)
+            - nb_threads:  how many threads to use during solve (int, optional)
+            - kwargs:      any keyword argument, sets parameters of solver object
+
+            Examples of supported arguments include:
+                - context (optional) – context to use during solve
+                - cplex_parameters (optional) – A set of CPLEX parameters to use
```
IgnaceBleukx: I don't understand this parameter, is it a `dict` with parameters?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+            and for cplex parameters: https://www.ibm.com/docs/en/icos/22.1.1?topic=cplex-topical-list-parameters
+
+            After solving, all solve details can be accessed through self.cplex_model.solve_details:
+            https://ibmdecisionoptimization.github.io/docplex-doc/mp/docplex.mp.sdetails.html#docplex.mp.sdetails.SolveDetails
+        """
+        # ensure all vars are known to solver
+        self.solver_vars(list(self.user_vars))
+
+        # set time limit
+        if time_limit is not None and not np.isinf(time_limit):
```
IgnaceBleukx: Why do we need the `isinf` here? We don't have it for any other solver

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+            if time_limit <= 0:
+                raise ValueError("Time limit must be positive")
+            self.cplex_model.set_time_limit(time_limit)
+
+        # set nb of threads
```
IgnaceBleukx: What is the default? do you really need to set it?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+            # COP
+            if self.has_objective():
+                self.cpm_status.exitstatus = ExitStatus.OPTIMAL
+            # CSP
+            else:
+                self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+        elif cplex_status == "JobFailed":
+            self.cpm_status.exitstatus = ExitStatus.ERROR
+        elif "aborted" in cplex_status:
+            self.cpm_status.exitstatus = ExitStatus.NOT_RUN
```
IgnaceBleukx, line 231: Wouldn't this also be `UNKNOWN` or `ERROR`?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+        # set time limit
+        if time_limit is not None and not np.isinf(time_limit):
+            if time_limit <= 0:
+                raise ValueError("Time limit must be positive")
+            self.cplex_model.set_time_limit(time_limit)
+
+        # set nb of threads
+        self.cplex_model.context.cplex_parameters.threads = nb_threads
+
+        cplex_objective = self.cplex_model.get_objective_expr()
```
IgnaceBleukx: I guess you could also call this below, where you actually use it?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+                self.cplex_model.add_constraint(cplexlhs >= cplexrhs)
+            elif cpm_expr.name == '==':
+                if isinstance(lhs, _NumVarImpl) \
+                        or (isinstance(lhs, Operator) and (lhs.name in {'sum', 'wsum', 'sub'})):
+                    # a BoundedLinearExpression LHS, special case, like in objective
+                    cplexlhs = self._make_numexpr(lhs)
+                    self.cplex_model.add_constraint(cplexlhs == cplexrhs)
+
+                else:
+                    # General constraints
```
IgnaceBleukx: global functions you mean? ;)

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+        """
+
+        # set time limit
+        if time_limit is not None:
+            if time_limit <= 0:
+                raise ValueError("Time limit must be positive")
+            self.cplex_model.set_time_limit(time_limit)
+
+        if solution_limit is None:
+            raise Exception(
```
IgnaceBleukx, line 544: Is this indeed the case? Or is this a copy-paste leftover from Gurobi?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+        solution_count = len(solutions_pool)
+        opt_sol_count = 0
+
+        # clear user vars if no solution found
+        if solution_count == 0:
+            self.objective_value_ = None
+            for var in self.user_vars:
+                var._value = None
+
+        for i in range(solution_count):
```
IgnaceBleukx: I suppose you can also just itertate over the solution pool directly?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+                if isinstance(display, Expression):
+                    print(argval(display))
+                elif isinstance(display, list):
+                    print(argvals(display))
+                else:
+                    display()  # callback
+
+        # Reset pool search mode to default
+        self.cplex_model.context.cplex_parameters.mip.limits.populate = 1
+        self.cplex_model.context.cplex_parameters.mip.pool.intensity = 0  # (optional) max effort for finding solutions
```
IgnaceBleukx: remove comment

IgnaceBleukx (changes requested):
Starting to look good, logic seems to be all there for the most part but
- tests are failing locally (something related to circuit)
- some questions remain in the interface too

cpmpy/solvers/cplex.py
tias: that is what gurobi calls min/max/abs etc in its documentation.

Cplex does not seem to give a name to this class of constraints:
https://ibmdecisionoptimization.github.io/docplex-doc/mp/docplex.mp.model.html#docplex.mp.model.Model.max

so we should indeed stick to our vocabulary : )

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+##
+## cplex.py
+##
+"""
+    Interface to CPLEX Optimizer using the python 'docplex.mp' package
+
+    Requires that the 'docplex' python package is installed:
```
tias: we now have a new standard for how to start a solver file.

its similar to this though... see e.g. ortools on master

(also, you will need to add a file in docs/api/solvers in docs so that the solver is also generated int he docs: https://cpmpy.readthedocs.io/en/latest/adding_solver.html)

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+
+    @staticmethod
+    def supported():
+        return CPM_cplex.installed() and CPM_cplex.license_ok()
+
+    @staticmethod
+    def installed():
+        # try to import the package
+        try:
+            import docplex.mp as domp
```
tias, line 93: this could call get_domp?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,533 @@
...
+
+    @staticmethod
+    def license_ok():
+        if not CPM_cplex.installed():
+            warnings.warn(
+                f"License check failed, python package 'docplex' is not installed! Please check 'CPM_cplex.installed()' before attempting to check license.")
+            return False
+        else:
+            try:
+                from docplex.mp.model import Model
```
tias, line 112: this could call get_domp?

I feel the get_domp isn't used much... cut it our use it consistently?

cpmpy/transformations/linearize.py
```diff
@@ -181,7 +181,7 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False, c
 
 
                 elif lhs.name == "mod" and "mod" not in supported:
-                    if "mul" not in supported:
+                    if "mul" not in supported and not is_num(lhs.args[1]):
```
tias, line 180: put separately in master

cpmpy/solvers/cplex.py
Dimosts: If we put it as kwarg, then the code will get quite ugly. The solver does not accept it as an argument in its solve method, but it has to be set before. Thus, we will have to extract it from the kwargs first, to do so etc.

cpmpy/solvers/cplex.py
Dimosts: These are actually the only parameters. changed the description a bit

cpmpy/solvers/cplex.py
Dimosts: They explain it as "Accepts either a RootParameterGroup object (obtained by cloning the model’s parameters), or a dict of path-like names and values.". I guess if someone wants to use this one they would have to have quite good understanding of CPLex either way.

cpmpy/solvers/cplex.py
Dimosts: it is indeed the case, described here: https://www.ibm.com/docs/en/icos/22.1.2?topic=solutions-how-enumerate-all

cpmpy/solvers/cplex.py
Dimosts: leftover, not sure why it was needed. removed it and everything is fine

cpmpy/solvers/cplex.py
Dimosts: changed to only set when not none is given

cpmpy/solvers/cplex.py
Dimosts: idd

cpmpy/solvers/cplex.py
Dimosts: idd

cpmpy/solvers/cplex.py
Dimosts: it is actually not run, digging in the solve status of docplex. It is when it was aborted either by the user or the system before solve starts.

cpmpy/transformations/linearize.py
Dimosts: will do

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+            
+        # set time limit
+        if time_limit is not None:
+            if time_limit <= 0:
+                raise ValueError("Time limit must be positive")
+            self.cplex_model.set_time_limit(time_limit)
+
+        # set nb of threads
+        if nb_threads is not None:
+            self.cplex_model.context.cplex_parameters.threads = nb_threads
```
IgnaceBleukx: I would assume you can set all types of parameters like this?
Wouldn't it be nicer if we translated the "kwargs" for the user like this?

e.g., 
```python
for param, val in kwargs.items():
    setattr(eval(f"self.cplex_model.context.cplex_parameters.{param}"), val)
```

IgnaceBleukx: The the line below becomes obsolete.
But maybe I'm too optimistic about CPLEXes API user-friendlyness

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+                self.cpm_status.exitstatus = ExitStatus.OPTIMAL
+            # CSP
+            else:
+                self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+        elif cplex_status == "JobFailed":
+            self.cpm_status.exitstatus = ExitStatus.ERROR
+        elif "aborted" in cplex_status:
+            self.cpm_status.exitstatus = ExitStatus.NOT_RUN
+        else:  # another? This can happen when error during solve. Error message will be in the status.
+            raise NotImplementedError(f"Translation of cplex status {cplex_status} to CPMpy status not implemented")  # if a new status type was introduced, please report on GitHub
```
IgnaceBleukx: remove comment after error

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+                    cpm_var._value = solver_val >= 0.5
+                else:
+                    cpm_var._value = round(solver_val)
+            # set _objective_value
+            if self.has_objective():
+                obj_val = self.cplex_model.get_objective_expr().solution_value
+                if round(obj_val) == obj_val: # it is an integer?:
+                    self.objective_value_ = int(obj_val)
+                else: #  can happen with DirectVar or when using floats as coefficients
+                    self.objective_value_ = float(obj_val)
```
IgnaceBleukx, line 254: do we have a test for this explicitely? I think we should.

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+            Creates solver variable for cpmpy variable
+            or returns from cache if previously created
+        """
+        if is_num(cpm_var):  # shortcut, eases posting constraints
+            return cpm_var
+
+        # special case, negative-bool-view
+        if isinstance(cpm_var, NegBoolView):
+            raise Exception("Negative literals should not be part of any equation. "
+                            "See /transformations/linearize for more details")
```
IgnaceBleukx, line 275: ValueError is better here.
Also, the error message could be somewhat more informative.
Smt like: "negative literals cannot be part of any equation, and should be removed by `cpmpy.transformations.linearize.only_positive_bv`" or smt?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+      for cpm_expr in self.transform(cpm_expr_orig):
+
+        # Comparisons: only numeric ones as 'only_implies()' has removed the '==' reification for Boolean expressions
+        # numexpr `comp` bvar|const
+        if isinstance(cpm_expr, Comparison):
+            lhs, rhs = cpm_expr.args
+            cplexrhs = self.solver_var(rhs)
+
+            # Thanks to `only_numexpr_equality()` only supported comparisons should remain
+            if cpm_expr.name == '<=':
```
IgnaceBleukx, line 406: you can merge these three cases using `eval_comparison` from utils.
Not required tho, as you whish : )

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+
+                elif lhs.name == 'mul':
+                    assert len(lhs.args) == 2, "CPLEX only supports multiplication with 2 variables"
+                    a, b = self.solver_vars(lhs.args)
+                    # CPLEX supports quadratic constraints
+                    self.cplex_model.add_constraint(a * b == cplexrhs)
+
+                else:
+                    # Global functions
+                    if lhs.name == 'min':
```
IgnaceBleukx, line 427: merge this if-trail into the previous elifs?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+                cond, trigger_val = self.solver_var(cond._bv), False
+            else:
+                cond, trigger_val = self.solver_var(cond), True
+
+            lhs, rhs = sub_expr.args
+            if isinstance(lhs, _NumVarImpl) or (lhs.name in {'sum', 'wsum', 'sub'}):
+                lin_expr = self._make_numexpr(lhs)
+            else:
+                raise Exception(f"Unknown linear expression {lhs} on right side of indicator constraint: {cpm_expr}")
+            if sub_expr.name == "<=":
```
IgnaceBleukx: ok here you can definately merge them with eval_comparison ; )

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+                self.cplex_model.add_indicator(cond, lin_expr <= self.solver_var(rhs), trigger_val)
+            elif sub_expr.name == ">=":
+                self.cplex_model.add_indicator(cond, lin_expr >= self.solver_var(rhs), trigger_val)
+            elif sub_expr.name == "==":
+                self.cplex_model.add_indicator(cond, lin_expr == self.solver_var(rhs), trigger_val)
+            else:
+                raise Exception(f"Unknown linear expression {sub_expr} name")
```
IgnaceBleukx: ValueError? (everywhere actually)

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+
+        To learn more about solution hinting in CPLEX, see:
+        https://ibmdecisionoptimization.github.io/docplex-doc/mp/docplex.mp.model.html#docplex.mp.model.Model.add_mip_start
+
+        :param cpm_vars: list of CPMpy variables
+        :param vals: list of (corresponding) values for the variables
+        """
+        # Create a MIP start solution from the provided variables and values
+        mip_start_sol = {}
+        for cpm_var, val in zip(cpm_vars, vals):
```
IgnaceBleukx: dict-comphrehension for bonus points?
```python
mip_start_sol = {self.solver_var(cpm_var) : val for cpm_var, val in zip(cpm_vars, vals)}
```

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+    Documentation of the solver's own Python API:
+    https://ibmdecisionoptimization.github.io/docplex-doc/mp/docplex.mp.model.html
+    """
+
+    @staticmethod
+    def supported():
+        return CPM_cplex.installed() and CPM_cplex.license_ok()
+
+    @staticmethod
+    def installed():
```
IgnaceBleukx, line 90: it seems you need BOTH docplex and cplex Python packages... I was very confused as to why it was not working on my machine...

IgnaceBleukx: So to install: `pip install docplex cplex`

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+        cplex_status = self.cplex_model.solve_details.status
+        if cplex_status == "Feasible":
+            self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+        elif "infeasible" in cplex_status:
+            self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+        elif cplex_status == "Unknown":
+            # can happen when timeout is reached...
+            self.cpm_status.exitstatus = ExitStatus.UNKNOWN
+        elif "optimal" in cplex_status:
+            # COP
```
IgnaceBleukx: style: can you put the #COP  and #CSP to the right of the if-statemet?

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+            
+            .. note::
+                technical side note: any constraints created during conversion of the objective
+                are premanently posted to the solver
+        """
+        # make objective function non-nested
+        (flat_obj, flat_cons) = flatten_objective(expr)
+        flat_obj = only_positive_bv_wsum(flat_obj)  # remove negboolviews
+        get_variables(flat_obj, collect=self.user_vars)  # add potentially created variables
+        self += flat_cons
```
IgnaceBleukx: use self.add()

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+            elif sub_expr.name == ">=":
+                self.cplex_model.add_indicator(cond, lin_expr >= self.solver_var(rhs), trigger_val)
+            elif sub_expr.name == "==":
+                self.cplex_model.add_indicator(cond, lin_expr == self.solver_var(rhs), trigger_val)
+            else:
+                raise Exception(f"Unknown linear expression {sub_expr} name")
+
+        # True or False
+        elif isinstance(cpm_expr, BoolVal):
+            self.cplex_model.add_constraint(cpm_expr.args[0])
```
IgnaceBleukx: So according to the tests, this actually fails?
You can mimic it by adding some dummy constraints instead?
```python
if cpm_expr.args[0] # just true
    pass
else:
    a = cp.boolvar()
    self.add(a - a >= 1) # or smt

```

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,579 @@
...
+            self.cplex_model.set_time_limit(time_limit)
+
+        if solution_limit is None:
+            raise Exception(
+                "CPLEX does not support searching for all solutions. If you really need all solutions, "
+                "try setting solution limit to a large number")
+
+        # Ask for multiple solutions
+        self.cplex_model.context.cplex_parameters.mip.limits.populate = solution_limit
+        self.cplex_model.context.cplex_parameters.mip.pool.intensity = 4
```
IgnaceBleukx: can you add a comment to explain what this means?

setup.py
```diff
@@ -31,6 +31,7 @@ def get_version(rel_path):
     "gcs": ["gcspy"],
     "cpo": ["docplex"],
     "pindakaas": ["pindakaas"],
+    "cplex": ["docplex"],
```
IgnaceBleukx: and "cplex"

IgnaceBleukx (commented):
Nice, a few more open things left. Some are very minor, up to you if you want to do them ;)

cpmpy/solvers/cplex.py
Dimosts: I do not have any test failing on cplex, that is weird. But will change it that either way.

cpmpy/solvers/cplex.py
Dimosts: I actually prefer them separate, like in gurobi too

cpmpy/solvers/cplex.py
IgnaceBleukx: It might depend on the Python/docplex/CPLEX version... I think @ThomSerg also found this for CP-optimizer

IgnaceBleukx: But I think you disabled the tests for CPLEX for this ; )

IgnaceBleukx: ah yes, line 939 in test_solvers.py
https://github.com/CPMpy/cpmpy/pull/583/files#diff-61f7e0ac02a380bd4f76344efacaffbf5546b519ec642c50deca7a418ecb10dcR939

cpmpy/solvers/cplex.py
Dimosts: You are too optimistic indeed. We can do that only with part of the kwargs, i.e., the cplex_parameters part. The rest, need to be given as kwargs to solve().

On the other hand, we could just remove nb_threads, and let users use directly "cplex parameters" kwarg to define it, but I think this is too awkward for such an important parameter. A user would have to write something like s.solve(cplex_parameters={'threads': 8})

IgnaceBleukx: Ok thought about this some more.
I think we should just have **kwargs and filter out those that need to be set as attribute, and set the others as **kwargs in .solve()
E.g.,
```python
def solve(self, time_limit=None, **kwargs):
    
   special_args = ["threads", "context", "clean_before_solve", "checker", "log_output"]
   for arg in special_args:
      if arg in kwargs:
          setattr(eval(f"self.cplex_model.context.cplex_parameters.{arg}"), kwargs[arg])
          del kwargs[arg]
   
   self.cplex_model.solve(cplex_parameters=kwargs) # kwargs only `cplex_parameters` now
        

```

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,581 @@
...
+                return False
+
+    @staticmethod
+    def version() -> Optional[str]:
+        """
+        Returns the installed version of the solver's Python API.
+        """
+        try:
+            import pkg_resources
+            return pkg_resources.get_distribution('docplex').version
```
IgnaceBleukx: Can you return both the docplex and CPLEX version here?
E.g., such as in CPO:
https://github.com/CPMpy/cpmpy/blob/2bdfcc70bf2a854102bcbdcce30d460f8fb82251/cpmpy/solvers/cpo.py#L123

cpmpy/solvers/cplex.py
IgnaceBleukx: What do you think of this proposal?

tests/test_solvers.py
```diff
@@ -883,23 +982,27 @@ def test_incremental_assumptions(self, solver):
         assert s.solve(assumptions=[])
 
     def test_vars_not_removed(self, solver):
-            bvs = cp.boolvar(shape=3)
-            m = cp.Model([cp.any(bvs) <= 2])
+            
+        if solver == 'cplex':
```
IgnaceBleukx: I think you can re-enabel this test now?

tests/test_solvers.py
```diff
@@ -987,7 +1091,7 @@ def test_status_solveall(self, solver):
             assert m.status().exitstatus == ExitStatus.FEASIBLE
 
             num_sols = m.solveAll(solver=solver, solution_limit=10)
-            assert num_sols == 10
+            assert num_sols >= 10 # at least 10 solutions
```
IgnaceBleukx: Well there should be exactly 10 if the limit is 10, no?

IgnaceBleukx (commented):
A few more comments remaing, mainly on testing

cpmpy/solvers/cplex.py
Dimosts: Actually, the "clean_before_solve", "checker", "log_output" need to be given during solve, not put in cplex_parameters. The "context" on the other hand, either needs to be given during solve (replacing the existing context) or be put like self.cplex_model.context = context. Only the threads from these special args uses this syntax (along with many other parameters for cplex).

tests/test_solvers.py
Dimosts: Well, in theory yes ... But for cplex no :) cplex enforces to return at least as many, and sometimes may return a few more.

I changed it to >= for all solvers now, but we can do it conditional. Though I do not find it that big of a problem if more are returned.

cpmpy/solvers/cplex.py
Dimosts: I changed the handling of the parameters in a similar way as your suggestion, but basically handling all other parameters first, except the solve() parameters that are given directly in .solve().

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,608 @@
...
+                raise ValueError("Time limit must be positive")
+            self.cplex_model.set_time_limit(time_limit)
+
+        if solution_limit is None:
+            raise Exception(
+                "CPLEX does not support searching for all solutions. If you really need all solutions, "
+                "try setting solution limit to a large number")
+
+        # Ask for multiple solutions
+        self.cplex_model.context.cplex_parameters.mip.limits.populate = solution_limit
```
IgnaceBleukx, line 549: So from the discussion on the test case, I understand this is a "soft" parameter, and CPLEX does not always adhere to it?
If so, I think that in the for-loop below on 558, we should break if the solution limit is reached, such that it is consistent among all solvers.

tests/test_solvers.py
```diff
@@ -707,6 +708,104 @@ def test_gurobi_element(self):
...
         self.assertTrue(iv.value()[idx.value(), idx2.value()] == 8)
 
+    @pytest.mark.skipif(not CPM_gurobi.supported(),
+                        reason="Gurobi not installed")
+    def test_gurobi_float_objective(self):
+        """Test that Gurobi properly handles float objective values."""
+        # Test case with float coefficients that should result in a float objective value
+        x, y, z = cp.boolvar(shape=3, name=tuple("xyz"))
+        
+        # Create a model with float coefficients - this can happen with DirectVar
```
IgnaceBleukx, line 718: I'm not sure what this comment means, but seems fine : )

tests/test_solvers.py
```diff
@@ -707,6 +708,104 @@ def test_gurobi_element(self):
...
+        self.assertTrue(s.solve())
+
+        m = cp.Model([~bv, ~((iv[0] + abs(iv[1])) == sum(iv))])
+        s = cp.SolverLookup.get("cplex", m)
+        self.assertTrue(s.solve())
+
+
+    @pytest.mark.skipif(not CPM_cplex.supported(),
+                        reason="cplex not installed")
+    def test_cplex_float_objective(self):
```
IgnaceBleukx, line 765: I would extend this (and the Gurobi test) to "test_cplex_floats" and also try to add float coefficients in a constraint?

tests/test_solvers.py
```diff
@@ -882,24 +981,25 @@ def test_incremental_assumptions(self, solver):
...
-            #test unique sols, should be same number
-            assert len(sols) == 8
+        # reset value for vars
+        bvs.clear()
+        assert m.solve(solver=solver)
+        for v in bvs:
+            assert v.value() is not None
+        #test solve_all
+        sols = set()
+        solution_limit = 20 if solver == 'gurobi' else None
```
IgnaceBleukx: so why don't you need the solution limit for CPLEX here?

tests/test_solvers.py
```diff
@@ -971,10 +1072,10 @@ def test_status_solveall(self, solver):
         m = cp.Model(cp.any(bv))
 
         limit = None
-        if solver == "gurobi": limit = 100000
+        if solver in ("gurobi", "cplex"): limit = 100000
 
         num_sols = m.solveAll(solver=solver, solution_limit=limit)
-        assert num_sols == 7
+        assert num_sols >= 7 # at least 7 solutions
```
IgnaceBleukx: Hmm no I don't like this. The doc of solveAll says:
"""
 solution_limit: stop after this many solutions (default: None)
"""

So we should make sure this is indeed the case

cpmpy/solvers/cplex.py
```diff
@@ -0,0 +1,608 @@
...
+        :param cpm_vars: list of CPMpy variables
+        :param vals: list of (corresponding) values for the variables
+        """
+        # Create a MIP start solution from the provided variables and values
+        if len(cpm_vars) != len(vals):
+            raise ValueError("Number of variables and values must match")
+        mip_start_sol = {self.solver_var(cpm_var) : val for cpm_var, val in zip(cpm_vars, vals)}
+        self.cplex_model.add_mip_start(mip_start_sol)
+
+    def solveAll(self, display=None, time_limit=None, solution_limit=None, **kwargs):
```
IgnaceBleukx: you are missing call_from_model here

tests/test_solvers.py
Dimosts: well, this test is failing due to that

tests/test_solvers.py
Dimosts: As discussed in person, we cannot currently handle float coefficients in constraints in CPMpy, and this will require bigger changes in the codebase. Thus this is skipped for now, and will focus on it in different PR.

Issue #570, opened 2025-01-09
Add brute force function for arbitrary models for e.g. testing purposes


hbierlee:
It'd be great to be able to brute-force arbitrary models for testing purposes of solvers and transformations. For instance, we wouldn't need the complicated logic in e.g. `comp_constraints` to only generate satisfiable test cases, instead we just brute force and cache the solution sets for all these test cases. Brute forcing can be done by generating all assignments for a test case model, and filtering on `all(cons.value() for cons in model)` (although there is some complexity there). Caching can be done via pytest so that it's only done once, and then for new tests.

Upsides:
* Soundness checking is way more comprehensive, because you also test for unsatisfiability, and test for missing solutions (if ALL_SOLS) rather than only "extra solutions" (which fail constraints).
* Writing tests cases becomes more easy (the above "if *unsat* : continue" type statements can be removed)
* (Not sure, but:) Testing may be slightly faster because you don't check constraints, you just check set membership

Downsides:
* The big question is whether we can have something like`.value()` for an arbitrary model, with many nested constraints and possibly auxiliary variables which we don't automatically get the assignment to. However for simple test cases it should be straightforward.
* Takes some engineering to set up properly :)
* Brute-forcing may be too expensive for bigger test case models (we could fall back on a real solver, but this has some issues if there are bugs in the solver or cpmpy transformation)
* It still relies on the constraint.value() functions to be sound, but this is what the tests rely on anyways (this will also uncover bugs regarding those, so arguably an upside)

Related: future improvements to fuzz-testing and auto-grading student models

Wout4:
We definitely should have .value() functionality for any arbitrary constraint, so this should be possible. Ofcourse only for smaller models, not sure how the caching works in pytest, but still each run could take a long time I suppose. Or can we cache them over multiple runs (= hardcoding them?)

I suppose we could get unlucky where a bug in our value calculation exactly matches a bug in a solver, but other than that we could find bugs in both with this method.

PR #584, opened 2025-02-05
Check infeasible (Comparison) constraints

hbierlee:
We should also generate and check infeasible constraints. For now, all these constraints are "trivially" unsatisfiable Comparison constraints, but these already can ben an issue for PySAT.


hbierlee:
Some of the new infeasible tests are not truly infeasible, because they of course require a slightly more sophisticated check, but rather than spend the time puzzling that out, I will close this PR in favor of a general brute forcing approach e.g. #570.

PR #586, opened 2025-02-13
Cse custom

Wout4:
Did some timing, and in practise (when used on cpmpy expressions) cpm_dict seems to be about 2-3 times slower than python dict. Cashed hash values make it about 2x faster, but I don't see a way we can reliably do that, because we can't know if the args changed. (We do if the direct args are changed with update_args, but what if it happens one level down?)
We get away with it for has_subexpr, because this implies that there isn't enough nestedness to encounter this problem


tias:
Our 'friend' reasoned the following:

When you override __eq__ for value equality, you also need to override __hash__ so that objects that compare equal have the same hash value. This is essential for using your objects as dictionary keys because Python’s dictionaries first use the hash value to quickly locate a bucket and then use __eq__ to resolve collisions.

For example, if your equality check compares certain attributes, you should compute the hash from those same attributes:

```python
class MyKey:
    def __init__(self, a, b):
        self.a = a
        self.b = b

    def __eq__(self, other):
        if isinstance(other, MyKey):
            return self.a == other.a and self.b == other.b
        return NotImplemented

    def __hash__(self):
        return hash((self.a, self.b))

# Now you can use MyKey objects as keys in a dictionary:
d = {}
key1 = MyKey(1, 2)
key2 = MyKey(1, 2)
d[key1] = "value"
print(d[key2])  # prints "value" because key1 == key2 and they have the same hash
```
This way, you can still take advantage of the highly optimized built-in dictionaries without running into issues with collisions or unhashable types.


So, there is a point here...

We use == in modeling, and the dict uses == on collision.
But when the dict uses it, it will always do it on two exactly the same objects. And so, when its a comparison between two exactly the same objects, we can actually do the comparison on the properties that make it itself (name and args).
And if those match... we can simply return a Boolean 'True'.

This will make the dict happy and correct, and our library should also be able to handle it... (there might be some odd edge case where in a transformation we rewrite with == and it returns a Bool while we assume it should be a BoolVal, but adding a check on that there will surely be cheaper than your observations on the custom dict?)

What do you think?

Wout4:
the problem is the case where two expressions are _not_ exactly the same, but their hash happens to be the same. Otherwise their would be no reason to do any of this. Actually our `__eq__` already gives back truthy values, so it is already implemented as above by default

Wout4:
I guess you could test if the hashes are the same within __eq__, but you still don't know if it is being asked by a dict or if a user happens to create a comparison with a hash collision. Both are equally (un)likely.
Maybe you could check who the caller of the __eq__ function is, and then either create an expression, or return a boolean if the caller was python's dict

Wout4:
So I quickly tried out the idea of splitting the functionality in` __eq__`
We speed up all dictionary operations to the python builtin speed
But the slowdown to __eq__ is totally not worth it, since this function is getting called way more than our cse_dict.
Also this feels a bit too hacky

tias:
thx for trying it!

cpmpy/expressions/core.py
```diff
@@ -153,8 +153,8 @@ def __repr__(self):
         return "{}({})".format(self.name, ",".join(strargs))
 
     def __hash__(self):
-        return hash(self.__repr__())
-
+        args = flatlist(self.args)
```
tias, line 156: I'm a bit hesitant for this... you are first creating a list of list of lists, then turning that into a tuple (single tuple? or tuple of tuples?)

I think we should have something that creates tuples directly? also if its ndarray we can call its custom flatten?

cpmpy/expressions/core.py
```diff
@@ -744,3 +744,196 @@ def _wsum_make(arg):
...
     return [1], [arg]
+
+
+# Custom dictionary and set classes for CPMpy expressions.
+# These do not inherit from python dict or set, because python abstracts from the underlying hash table.
+# Meaning we cannot access it ourselves.
+# This makes you rely on super.add, which is what we are trying to avoid..
+class cpm_dict:
+    def __init__(self):
+        # Maps hash values to lists of (key, value) pairs with that hash
```
tias, line 755: Maps hash -> [(key, value)] list for keys with the same hash

cpmpy/expressions/core.py
```diff
@@ -744,3 +744,196 @@ def _wsum_make(arg):
...
+        self._hash_groups = {}
+
+    def same(self, key1, key2):
+        """
+        Custom key comparison logic (override this as needed).
+        Example: Compare keys as strings to allow "1" and 1 to coexist.
+        """
+        return str(key1) == str(key2)
```
tias, line 763: ideally first check on name being equal, which is faster? and then compare args directly?
return key1.name == key2.name and key1.args == key2.args

or you want to keep it generic for non Expressions too?
Still feels like calling the recursive string is excessive... any equality check can stop as soon as one component is different; while string needs to go all through before doing a comparison

cpmpy/expressions/core.py
```diff
@@ -744,3 +744,196 @@ def _wsum_make(arg):
...
+        # Check for existing key using `same`
+        for i, (existing_key, _) in enumerate(hash_group):
+            if self.same(existing_key, key):
+                hash_group[i] = (key, value)  # Update existing entry
+                return
+        # Add new key-value pair if no collision
+        hash_group.append((key, value))
+
+    def __getitem__(self, key):
+        hash_group = self._get_hash_group(key)
```
tias, line 783: its a bit strange that this becomes a mutable operation: it will create an empty bucket for this hash if the hash has not been stored yet...

```
  h = hash(key)
  if h in self._hash_groups:
    for existing_key, value in self._hash_groups:
       if self.same(existing_key, key):
          return value
   raise...
```
I don't know, I'm not sure you really need the _get_hash_group() in any of the functions, but if you think its cleaner it wont matter too much I guess

cpmpy/expressions/core.py
```diff
@@ -744,3 +744,196 @@ def _wsum_make(arg):
...
+
+    def values(self):
+        return [v for group in self._hash_groups.values() for _, v in group]
+
+    def __repr__(self):
+        items = ", ".join(f"{repr(key)}: {repr(value)}" for key, value in self.items())
+        return f"{{{items}}}"
+
+
+class cpm_set:
```
tias, line 826: any use case for it?

cpmpy/solvers/solver_interface.py
```diff
@@ -76,6 +76,7 @@ def __init__(self, name="dummy", cpm_model=None, subsolver=None):
         # initialise variable handling
         self.user_vars = set()  # variables in the original (non-transformed) model
```
tias, line 77: Iguess this is a use case for cpm_set? a hash collision on the Var would not resolve

cpmpy/solvers/solver_interface.py
```diff
@@ -76,6 +76,7 @@ def __init__(self, name="dummy", cpm_model=None, subsolver=None):
...
         self._varmap = dict()  # maps cpmpy variables to native solver variables
```
tias, line 78: I guess this has to be a cpm_dict too? I don't think we store the varname, but the actual object as key... (or we should switch to the varname as key?)

cpmpy/transformations/flatten_model.py
```diff
@@ -315,23 +316,35 @@ def get_or_make_var(expr):
     if __is_flat_var(expr):
         return (expr, [])
 
+    if expr_dict is None:
+        expr_dict_getrmake_var = cpm_dict()
```
tias, line 320: why do you create a custom-name expr_dict here, and not overwrite the one of the arg?
maybe add a line of documentation about it...

tias (commented):
review

tias:
back to "Cashed hash values make it about 2x faster, but I don't see a way we can reliably do that, because we can't know if the args changed. (We do if the direct args are changed with update_args, but what if it happens one level down?)"

To compute the hash, we need to make the args tuples.

What if we make args a tuple at constructor time? Then every Expression is immutable?
( it might still be that one of the args is an nd(var)array; I don't know if we keep that or change it to a list, but even if we keep that, you would need to know dang well what you are doing to modify a expr.args[1][3] = ...)
all are transformations are copy-on-write, so making the objects immutable should not break anything...

and then we can cache hashes?

cpmpy/expressions/core.py
Wout4: Comparing the strings still feels a bit ugly, and indeed we could fail faster, but in practice this should almost always return true (only will be false if we have a hash collision) so failing faster is probably not better if we slow down the check between identical expressions

cpmpy/solvers/solver_interface.py
Wout4: yes.. I guess we will have to do some benchmarking to see if we really want to make this hash-collision proof

cpmpy/transformations/flatten_model.py
Wout4: Ah this is an artifact from trying to dectect where we called __eq__ from, can be reverted

Wout4:
I think even when doing copy on write we can run into problems with cached hashes;
There is two ways we do the copying:
1. Create a new Operator/Comparison with the same name and the new args -> no problem
2. Copy.copy the expression, and then use update_args. Copy.copy will also copy the cached hash value, so we would always have to manually empty it in our transformations first. (or write our own copy?)
Wouldn't work as is, but seems feasible.

As to making our arguments immutable, I don't think it is relevant to our case, because we always overwrite our args, never add or remove to the arguments list. You can still have mutable objects (our cpmpy-expressions for example) in a tuple, and then change those things that are inside the tuple..

cpmpy/solvers/solver_interface.py
Wout4: Yes, just not using the expression object as a key could also be a workaround, but that also involves computing the str everytime we use a set/dict, so not sure if it would be better

cpmpy/expressions/core.py
Wout4: user vars, and comparing sets of solutions in general, like we do in the mcs tools

Wout4:
I just discovered that in python hash(-1) and hash(-2) are both equal to -2
This kind of throws a monkey wrench into our idea of recursively calling hash on the arguments of our expressions, since this will lead to hash collisions quite often.
using cpm_dict can of course deal with collisions, but hash collisions are always slower

tias:
is that in general for negative numbers (or integers), or only very specifically for '-1'? I would not expect this to be general

Wout4:
only for -1, because in C it's a special error return I think.. We could use some different hash function in stead of the python builtin, but something like SHA256 requires a byte-like input which you can get by doing str(expression).encode(). This would obviously defeat the purpose since then we might as well just go back to hash of string.

-1 and -2 are pretty common numbers so I feel like it might still be a problem. For example decomposing some globals where the arguments have -1 and -2 in the domain will lead to creation of x == -1 or x == -2 type expressions

IgnaceBleukx:
Superseded by #679

PR #587, opened 2025-02-14
Improve DIMACS reading and testing

hbierlee:
This improves the DIMACS reading by:

- Making a certain test which was sometimes failing on my machine more deterministic, and just checking the read models directly rather than relying on solve
- Assuming the DIMACS input is well-formed. While inferring the p-line is possible, this is usually more trouble than it's worth, and bad for performance/debug. The current implementation would make 300 BoolVar's if we encountered an out of order literal (e.g. the DIMACS `1 0 300 0` is supposed to be `[bv1, bv300]`, but makes 300 literals).

### DIFF
--- a/cpmpy/tools/dimacs.py
+++ b/cpmpy/tools/dimacs.py
@@ -79,36 +79,40 @@ def read_dimacs(fname):
     m = cp.Model()
 
     with open(fname, "r") as f:
-        typ = None  # CNF/WCNF
-        # TODO infer p-header values (although generally not a good idea)
-        nr_cls = None
-        bvs = None
         clause = []
+        had_p_line = False
         for line in f.readlines():
-            if line.startswith("p"):
+            if line == "" or line.startswith("c"):
+                continue  # skip empty and comment lines
+            elif line.startswith("p"):
                 try:
-                    typ,typ,nr_vars,nr_cls = line.strip().split(" ")
-                    if typ != "cnf":
-                        raise cp.exceptions.NotSupportedError("WDIMACS (WCNF) files are not supported.")
-                    nr_vars = int(nr_vars)
-                    if nr_vars>0:
-                        bvs = cp.boolvar(shape=nr_vars)
-                    nr_cls = int(nr_cls)
+                    _,typ,nr_vars,nr_cls = line.strip().split(" ")
                 except ValueError:
                     raise cp.exceptions.CPMpyException(f"Invalid DIMACS file p-header: {line}")
-            elif line.startswith("c"):
-                continue
+                if typ != "cnf":
+                    raise ValueError("Expected `cnf` (i.e. DIMACS) as file format, but got {typ} which is not supported.")
+                nr_vars = int(nr_vars)
+                if nr_vars>0:
+                    bvs = cp.boolvar(shape=nr_vars)
+                nr_cls = int(nr_cls)
+                had_p_line = True
             else:
+                assert had_p_line, "Expected p-line before clauses"
                 for token in line.strip().split():
                     if token == "0":
                         m+=cp.any(clause)
                         clause = []
                     else:
                         i = int(token.strip())
-                        bv = bvs[abs(i)-1]
+                        try:
+                            bv = bvs[abs(i)-1]
+                        except IndexError:
+                            raise AssertionError("Expected <={nr_vars} variables (from p-line) but found literal {i} in clause {line}")
+
                         clause.append(bv if i > 0 else ~bv)
 
-        assert(len(m.constraints) == nr_cls, f"Number of clauses was declared in p-line as {nr_cls}, but was {len(m.constraints)}")
+        assert len(m.constraints) == nr_cls, f"Number of clauses was declared in p-line as {nr_cls}, but was {len(m.constraints)}"
+        assert not clause, f"Expected last clause to be terminated by 0, but was not {clause}"
 
     return m
 
--- a/cpmpy/tools/dimacs.py
+++ b/cpmpy/tools/dimacs.py
@@ -71,7 +71,8 @@ def write_dimacs(model, fname=None):
 
 def read_dimacs(fname):
     """
-        Read a CPMpy model from a DIMACS formatted file
+        Read a CPMpy model from a DIMACS formatted file striclty following the specification: https://web.archive.org/web/20190325181937/https://www.satcompetition.org/2009/format-benchmarks2009.html
+        Note: the p-line has to denote the correct number of variables and clauses
         :param: fname: the name of the DIMACS file
         :param: sep: optional, separator used in the DIMACS file, will try to infer if None
     """
--- a/cpmpy/tools/dimacs.py
+++ b/cpmpy/tools/dimacs.py
@@ -72,7 +72,6 @@ def write_dimacs(model, fname=None):
 def read_dimacs(fname):
     """
         Read a CPMpy model from a DIMACS formatted file
-        If the number of variables and constraints is not present in the header, they are inferred.
         :param: fname: the name of the DIMACS file
         :param: sep: optional, separator used in the DIMACS file, will try to infer if None
     """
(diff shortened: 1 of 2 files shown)


cpmpy/tools/dimacs.py
```diff
@@ -72,41 +72,43 @@ def write_dimacs(model, fname=None):
 def read_dimacs(fname):
     """
         Read a CPMpy model from a DIMACS formatted file
```
IgnaceBleukx, line 74: Maybe add a link to the DIMACS specification here to say we adhere to the strict specification and assume the file is well formed (that is with a header line)

cpmpy/tools/dimacs.py
```diff
@@ -72,41 +72,43 @@ def write_dimacs(model, fname=None):
...
+        # TODO infer p-header values (although generally not a good idea)
+        nr_cls = None
+        bvs = None
+        clause = []
+        for line in f.readlines():
+            if line.startswith("p"):
+                try:
+                    typ,typ,nr_vars,nr_cls = line.strip().split(" ")
+                    if typ != "cnf":
+                        raise cp.exceptions.NotSupportedError("WDIMACS (WCNF) files are not supported.")
```
IgnaceBleukx, line 92: This assumes the only other thing one might want to parse is weighted cnf?
I would put a more careful warning here such as: `Expected 'cnf' as format but gotten {typ} instead which is not supported`.

cpmpy/tools/dimacs.py
```diff
@@ -72,41 +72,43 @@ def write_dimacs(model, fname=None):
...
-                break
+        typ = None  # CNF/WCNF
+        # TODO infer p-header values (although generally not a good idea)
+        nr_cls = None
+        bvs = None
+        clause = []
+        for line in f.readlines():
+            if line.startswith("p"):
+                try:
+                    typ,typ,nr_vars,nr_cls = line.strip().split(" ")
```
IgnaceBleukx, line 90: This already assumes the first two arguments are the same. Better to make this a separate check:
```
t1, t2, nr_vars, nc_cls = line.strip().split(" ")
assert t1 == t2 == "cnf", "Header line should be formed as "cnf cnf <#vars> <#clauses>"
```

cpmpy/tools/dimacs.py
IgnaceBleukx: Also: NotSupportedError is supposed to be for solver-interfaces mainly, I think a ValueError or even asserion error is better here

cpmpy/tools/dimacs.py
```diff
@@ -72,41 +72,43 @@ def write_dimacs(model, fname=None):
...
+                    if typ != "cnf":
+                        raise cp.exceptions.NotSupportedError("WDIMACS (WCNF) files are not supported.")
+                    nr_vars = int(nr_vars)
+                    if nr_vars>0:
+                        bvs = cp.boolvar(shape=nr_vars)
+                    nr_cls = int(nr_cls)
+                except ValueError:
```
IgnaceBleukx, line 97: Where would this ValueError be thrown? Is it in the `int` casting? If so I'm fine with simply throwing it through to the user which might make debugging even easiser (you can see exactly which line is failing)

cpmpy/tools/dimacs.py
```diff
@@ -72,41 +72,43 @@ def write_dimacs(model, fname=None):
...
+                    raise cp.exceptions.CPMpyException(f"Invalid DIMACS file p-header: {line}")
+            elif line.startswith("c"):
```
IgnaceBleukx, line 99: Add a comment that this line is a comment ; )

cpmpy/tools/dimacs.py
```diff
@@ -72,41 +72,43 @@ def write_dimacs(model, fname=None):
...
+                for token in line.strip().split():
+                    if token == "0":
+                        m+=cp.any(clause)
+                        clause = []
+                    else:
+                        i = int(token.strip())
+                        bv = bvs[abs(i)-1]
+                        clause.append(bv if i > 0 else ~bv)
+
+        assert(len(m.constraints) == nr_cls, f"Number of clauses was declared in p-line as {nr_cls}, but was {len(m.constraints)}")
```
IgnaceBleukx, line 111: You can remove the paratheses after the assert

tests/test_tool_dimacs.py
```diff
@@ -4,65 +4,39 @@
...
+        model = self.dimacs_to_model("p cnf 3 3\n-2 -3 0\n3 2 1 0\n-1 0\n")
+        bvs = sorted(get_variables_model(model), key=str)
+        self.assertEqual(str(model), str(cp.Model(
+            cp.any([~bvs[1], ~bvs[2]]), cp.any([bvs[2], bvs[1],bvs[0]]), ~bvs[0])
+                         ))
+
+    def test_empty_formula(self):
+        model = self.dimacs_to_model("p cnf 0 0")
+        self.assertTrue(model.solve())
+        self.assertEqual(model.status().exitstatus, cp.solvers.solver_interface.ExitStatus.OPTIMAL)
```
IgnaceBleukx, line 31: Instead of the long referencing, do `from cpmpy.solvers.solver_interface import ExitStatus`

cpmpy/tools/dimacs.py
```diff
@@ -72,41 +72,43 @@ def write_dimacs(model, fname=None):
...
-            m += cp.any(clause)
+                for token in line.strip().split():
+                    if token == "0":
+                        m+=cp.any(clause)
+                        clause = []
+                    else:
+                        i = int(token.strip())
+                        bv = bvs[abs(i)-1]
+                        clause.append(bv if i > 0 else ~bv)
+
```
IgnaceBleukx, line 110: Sanity check: check the clause is empty after the file-reading has finished (means the last clause was terminated properly)

hbierlee:
Great comments @IgnaceBleukx , I've changed everything, but I now have proper testing as well for malformed DIMACs files. It should be more robust now.

cpmpy/tools/dimacs.py
```diff
@@ -71,42 +71,49 @@ def write_dimacs(model, fname=None):
...
+                for token in line.strip().split():
+                    i = int(token.strip())
+                    if i == 0:
+                        m+=cp.any(clause)
+                        clause = []
+                    else:
+                        try:
+                            bv = bvs[abs(i)-1]
+                        except IndexError:
+                            raise AssertionError("Expected <={nr_vars} variables (from p-line) but found literal {i} in clause {line}")
```
IgnaceBleukx, line 111: Hmm ValueError would be better here. AssertErrors are not supposed to be thrown explicilty I feel

cpmpy/tools/dimacs.py
```diff
@@ -71,42 +71,49 @@ def write_dimacs(model, fname=None):
...
-                break
+        clause = []
+        had_p_line = False
+        for line in f.readlines():
+            if line == "" or line.startswith("c"):
+                continue  # skip empty and comment lines
+            elif line.startswith("p"):
+                try:
+                    _,typ,nr_vars,nr_cls = line.strip().split(" ")
+                except ValueError:
```
IgnaceBleukx, line 91: Ah that's where the ValueError can come from.
Then I would parse this line in two parts:
```python
params = line.strip().split(" ")
assert len(params) == 4, f"Expected p-header to be formed as ... but got {line}"
_, typ, nr_vars, nr_cls = params
```
So without an explicit try-catch

cpmpy/tools/dimacs.py
```diff
@@ -71,42 +71,49 @@ def write_dimacs(model, fname=None):
...
-            ints = [int(idx.strip()) for idx in txt_ints.split(" ") if len(idx.strip())]
-
-            for i in ints:
-                if abs(i) >= len(bvs):  # var does not exist yet, create
-                    bvs += [cp.boolvar() for _ in range(abs(i) - len(bvs))]
-                bv = bvs[abs(i) - 1]
-                clause.append(bv if i > 0 else ~bv)
-
-            m += cp.any(clause)
+                assert had_p_line, "Expected p-line before clauses"
```
IgnaceBleukx, line 101: Not strictly necessary as `bvs` will not exist otherwise.
That would mean you can parse smt like this tho, which may not be want you want -- no strong opinion here.
"""
c malformed
0 
0
"""

cpmpy/tools/dimacs.py
```diff
@@ -71,42 +71,49 @@ def write_dimacs(model, fname=None):
 
 def read_dimacs(fname):
     """
-        Read a CPMpy model from a DIMACS formatted file
-        If the number of variables and constraints is not present in the header, they are inferred.
+        Read a CPMpy model from a DIMACS formatted file striclty following the specification: https://web.archive.org/web/20190325181937/https://www.satcompetition.org/2009/format-benchmarks2009.html
```
IgnaceBleukx, line 74: Put this url on a new line

cpmpy/tools/dimacs.py
```diff
@@ -71,42 +71,49 @@ def write_dimacs(model, fname=None):
...
-                    bvs += [cp.boolvar() for _ in range(abs(i) - len(bvs))]
-                bv = bvs[abs(i) - 1]
-                clause.append(bv if i > 0 else ~bv)
-
-            m += cp.any(clause)
+                assert had_p_line, "Expected p-line before clauses"
+                for token in line.strip().split():
+                    i = int(token.strip())
+                    if i == 0:
+                        m+=cp.any(clause)
```
IgnaceBleukx, line 105: Nitpicky: we typically put some spaces (so `m += cp.any(clause)`)

IgnaceBleukx (commented):
Left another round of comments, should be ready to merge after these : )

cpmpy/tools/dimacs.py
```diff
@@ -71,42 +71,49 @@ def write_dimacs(model, fname=None):
...
+                        clause = []
+                    else:
+                        try:
+                            bv = bvs[abs(i)-1]
+                        except IndexError:
+                            raise AssertionError("Expected <={nr_vars} variables (from p-line) but found literal {i} in clause {line}")
+
+                        clause.append(bv if i > 0 else ~bv)
+
+        assert not clause, f"Expected last clause to be terminated by 0, but it was not"
```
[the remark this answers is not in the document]
hbierlee: I agree so I will change it, though apparently (fun fact) it's not the PEP way

cpmpy/tools/dimacs.py
hbierlee: Yes, but then you'd get a NameError; I changed it now though to be more explicit.

PR #589, opened 2025-02-14
fix side conditions for precedence

Wout4:
Fuzztesting bugfixes..


IgnaceBleukx (approved):
Looks good, I would just add one comment that makes some clarification.

PR #592, opened 2025-02-18
Optimize Simplify-bool

IgnaceBleukx:
Port has_subexpr optimization to the simplify_bool transformation.
Reduced time to run the script in #512 from 8s to 5s on my machine.

### DIFF
--- a/cpmpy/transformations/normalize.py
+++ b/cpmpy/transformations/normalize.py
@@ -8,10 +8,10 @@ import numpy as np
 import cpmpy as cp
 
 from ..expressions.core import BoolVal, Expression, Comparison, Operator
-from ..expressions.utils import eval_comparison, is_false_cst, is_true_cst, is_boolexpr, is_num, is_bool, is_int
-from ..expressions.variables import NDVarArray
+from ..expressions.utils import eval_comparison, is_false_cst, is_true_cst, is_boolexpr, is_num, is_bool
+from ..expressions.variables import NDVarArray, _BoolVarImpl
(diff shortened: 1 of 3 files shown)


cpmpy/transformations/normalize.py
```diff
@@ -55,35 +56,41 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
-                    filtered_args = [arg for arg in args if not isinstance(arg, BoolVal)]
-                    if len(filtered_args):
-                        newlist.append(Operator("or", filtered_args))
-                    else:
+                for i,a in enumerate(list(args)):
+                    if is_true_cst(a):
+                        newlist.append(1 if num_context else BoolVal(True))
+                        break
+                    elif is_false_cst(a):
+                        args.pop(i)
```
Wout4, line 71: I think we want to do copy on write, so maybe it's better not to change the args in place here. Haven't really thought through the entire copy on write thing, but we might need to enforce that to be able to cash hashes and stuff ;)

IgnaceBleukx:
This was just some try-outs; indeed needs a lot more to get it properly to work ;)

Wout4:
I do remember that simplify bool can't always be skipped if there is no subexpressions, that's why it's not there at the moment. Not sure what the problem was anymore

tias:
looks good; and time reduction remains: from 9s to 6s on this 'tester_ignace.py' file that I have lying around from another PR.

gonna commit to see test-suite succeed, then merge

tias (approved):
all good, only waiting for tests to finish

Issue #398, opened 2023-08-07
gurobi does not support floordivision


Wout4:
the model below is unsat for gurobi, because it doesn't allow rounding of floordiv (only even numbers are valid for x1 because it gets divided by 2). This is not the expected behavior of floordivision

```
iv = intvar(-1,3)
x1 = intvar(1,4, name="x1")

cons = [x1 // 2 == iv, x1 == 1]
print(Model(cons).solve('gurobi'))
```

IgnaceBleukx:
I expected this one to be closed by #516 but apparently not... We should remove "div" from the supported set of linearized constraints in Gurobi instead...

Dimosts:
We should throw an exception or a warning to the user.

PR #593, opened 2025-02-20
linearize division for gurobi

Wout4:
fixes https://github.com/CPMpy/cpmpy/issues/398

### DIFF
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -314,7 +314,7 @@ class CPM_gurobi(SolverInterface):
         # apply transformations, then post internally
         # expressions have to be linearized to fit in MIP model. See /transformations/linearize
         cpm_cons = toplevel_list(cpm_expr)
-        cpm_cons = no_partial_functions(cpm_cons, safen_toplevel={"mod"})  # linearize expects safe exprs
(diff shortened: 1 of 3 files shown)


IgnaceBleukx (approved):

IgnaceBleukx:
Ideally, also add the test from #398; github won't catch it but at least the local tests should

IgnaceBleukx:
Leaving this here for future reference:

Gurobi supports division of two variables, but using different semantics than other solvers.
I.e., Gurobi rounds towards negative-infinity while other solvers round towards 0. To make the result of all sovlers consistent in CPMpy, we manually decompose the division for Gurobi, as if it did not have support for the division in the first place.

tias:
@IgnaceBleukx I'm confused... there is no code to actually linearize 'div', I guess its already inline, so this is about decomposing div for Gurobi because of the semantics?

In that case, do merge, idd.

IgnaceBleukx:
Yes indeed; the linearization code for `div` (and `mod`) has been added to the transformation in #516, but we did not enable it for Gurobi then, as Gurobi has also supports div natively.
Now, because it's semantics are different from the other solvers, we also enabled linearization in this PR for Gurobi.

PR #594, opened 2025-02-21
Numerical IfThenElse

Wout4:
Numerical version of if then else.

### DIFF
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -61,6 +61,7 @@
         Among
         NValue
         Abs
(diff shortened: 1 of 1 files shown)


IgnaceBleukx:
What solvers support this? I would assume z3 and MiniZinc?

Wout4:
Looks like we didn't add it to any of the solvers for the competition, but should be supported by z3 and minizinc I think

tias:
I don't know...

`ite_num()` is a weird name... would we really use it? or want users to use it?

The code does not include an example of its usefulness

I'm putting it on the waitlist till somebody demonstrates a need for it.

PR #595, opened 2025-02-21
check if argument is negative or positive in abs decomp

Wout4:
from our xcsp3 branch, seems to belong in master


IgnaceBleukx:
Tests fails for some pysat test, but seems unrelated?

Wout4:
Yes will merge. Does indeed make more sense to not decompose into another global..

PR #596, opened 2025-02-21
Table has subexpr

Wout4:
avoids looping over the entire table, that consists of constants anyway

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -372,7 +372,7 @@ class Table(GlobalConstraint):
     """
     def __init__(self, array, table):
         array = flatlist(array)
-        if isinstance(table, np.ndarray): # Ensure it is a list
+        if isinstance(table, np.ndarray):  # Ensure it is a list
             table = table.tolist()
         if not all(isinstance(x, Expression) for x in array):
             raise TypeError(f"the first argument of a Table constraint should only contain variables/expressions: "
@@ -388,6 +388,12 @@ class Table(GlobalConstraint):
         arrval = argvals(arr)
         return arrval in tab
 
+    # specialisation to avoid recursing over big tables
+    def has_subexpr(self):
+        if not hasattr(self, '_has_subexpr'): # if _has_subexpr has not been computed before or has been reset
+            arr, tab = self.args # the table 'tab' can only hold constants, never a nested expression
+            self._has_subexpr = any(a.has_subexpr() for a in arr)
+        return self._has_subexpr
 class ShortTable(GlobalConstraint):
     """
         Extension of the `Table` constraint where the `table` matrix may contain wildcards (STAR), meaning there are
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -424,6 +424,13 @@ class ShortTable(GlobalConstraint):
                 return True
         return False
 
+    # specialisation to avoid recursing over big tables
+    def has_subexpr(self):
+        if not hasattr(self, '_has_subexpr'): # if _has_subexpr has not been computed before or has been reset
+            arr, tab = self.args # the table 'tab' can only hold constants, never a nested expression
+            self._has_subexpr = any(a.has_subexpr() for a in arr)
+        return self._has_subexpr
+
 class NegativeTable(GlobalConstraint):
     """The values of the variables in 'array' do not correspond to any row in 'table'
     """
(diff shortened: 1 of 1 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -388,6 +388,12 @@ def value(self):
         arrval = argvals(arr)
         return arrval in tab
 
+    # specialisation to avoid recursing over big tables
+    def has_subexpr(self):
+        if not hasattr(self, '_has_subexpr'): # if _has_subexpr has not been computed before or has been reset
+            arr, tab = self.args # the table 'tab' can only hold constants, never a nested expression
```
IgnaceBleukx, line 394: I agree, but we should assert this also in the constructor of Table then.

IgnaceBleukx (approved):
Add an assertion/typecheck in the constructor of the Table-constraints such that we are 100% certain the has_subexpr valid.
I think that if a solver does not support Table, and hence it's decomposed, this might go wrong currently (the decomposition also works with variables in the table, solver-API's typically don't).

Wout4:
I'm hesitant to put an assert that loops over the entire table, because that is kind of what we are trying to avoid doing with this optimisation. Maybe it should only be a competition thing where we know the input structure for sure

IgnaceBleukx:
If I remember correctly, @ThomSerg found a nice (and efficient) way of checking all elements in a Table whether they are the correct type?

ThomSerg:
For the XCSP3 competition we didn't really do a 'more efficient check', we just skipped some. We added the attribute `.vars` to all expressions, similar to `.args`, which would return only the decision variables of that expression (no constants). This has a big impact for large tables. Since we know that the entries of the table will always be constants, we only had to do a check on the remaining args.

```python
array = cp.intvar(...)
table = np.array(...) # 2D array of constants
cp.Table(array, table)
```

So `.vars` would only need to check `array`, which is a big saving since the `table` array can be quite large.
This implementation might be too tailored to the competition / table constraint, but some generic way of accessing subsets of the variables (where applicable like for table) would be nice (since now we just store them all together as a flat list)

ThomSerg:
```python
class Table(GlobalConstraint):
    ...
    @property
    def vars(self):
        return self._args[0]
    
    # specialisation to avoid recursing over big tables
    def has_subexpr(self):
        if not hasattr(self, '_has_subexpr'): # if _has_subexpr has not been computed before or has been reset
            arr, tab = self.args # the table 'tab' can only hold constants, never a nested expression
            self._has_subexpr = any(a.has_subexpr() for a in arr)
        return self._has_subexpr
```

IgnaceBleukx:
Hmm ok not sure how to continue then... Should we indeed add this type-check? And if so, what would be the most efficient way of doing so?
It seems weird to me of not testing it to be honest... As we do quite some type-checks in all our global constraints.

tias:
```python3 -c "import timeit, numpy as np; from cpmpy import Table, intvar; n=20000; m=200; t=[[i*m+j for j in range(m)] for i in range(n)]; v=intvar(0, n, shape=m); print('Time:', timeit.timeit(lambda: Table(v, t).has_subexpr(), number=10)/10)"```

is 0.7s on my machine.

but so the current PR implementation is not acceptable (though hyperfast as it ignores the arr); might try my suggestions tonight...

IgnaceBleukx:
I would actually be in favor of doing an elaborate type-check when constructing the table constraint in the first place.
I.e., double for-loop over the nested list and checking `is_num` on all of the elements; or doing more fancy stuff such as exploiting the numpy-type.

But the bottom line being: we don't want to check this in the `has_sub_expr` call (which might happen many times during the transformation stack), but just once at construction time. For me, Tables with variables don't make any sense anyway...
If we want to support such things, we need to go a lot further (e.g., checking in the solver interface if `table` has some expressions, and then decomposing...) but that's off-topic for this PR

tias:
Converting to draft.

So I implemented the is-const check on the table in the constructor, which means we don't have to check it in has_subexpr(), all good.

I also added the special case for when table is an np.array.

Now, the np.array case is much easier to implement, and also a stronger check: it ensures that the dimensions actually match, that it is a square table... it is also much faster. It is in fact so much faster that it is more efficient to create an ndarray and assert those properties, then to check with isinstance...

Now it goes further. If we would store the table as an np.array, then we don't need the .tolist and then it is even much much faster...

So now a further refactor presents itself: **do we always store the 'table' part (for Table, NegTable, ShortTable) as an ndarray?**

Advantages: should also be much faster on XCSP3 side: just read it in as ndarray and keep it that way

DIsadvantages: solvers probably expect a list of lists, so we will still need to do a .list() for those solvers I guess... decompositions will stay the same.

PR #597, opened 2025-02-21
XCSP3 in tools

Wout4:
Should also have an easy benchmarking script.


Wout4:
todo: readme, add python an package requirements somewhere
maybe add a downloader of some xcsp3 instances (think we had that somewhere already)

tias (commented):
Some general comments:
* the has_subexprs should be mainlined in a separate pull request?
* I think we should rename the files a bit to set better expectations:
  - callbacks: parser_callbacks? or pycsp3_callbacks?
  - installer: this is actually not an installer of pycsp3 or the parser... its about downloading instances I think? so, xcsp3_downloader?
  - solution.py: I have no idea... should just be part of run_model?
    I guess we might have 2 runners: a runner that is competition-ready (e.g. it accepts the args and writes in the format that the competition expects)
    and a runner that we can use to benchmark our code? e.g. that logs specific runtimes to a csv file?
  - xglobals... I guess this is included by the callbacks file? maybe just extra_globals.py or xcsp3_globals.py?

tias:
also I'm a bit surprsied by the 'tests' subfolder and run_models being in that...

tias:
cleaned up the printing a bit.

doesn't work:
Found 1 models in folder 'tests/models/'
Will use 7 parallel solves, with solver ortools...
- running model tests/models/Fillomino-mini-5-0_c24.xml
  Error parsing: module 'cpmpy' has no attribute 'ShortTable'
  Error solving: cannot access local variable 't_parse' where it is not associated with a value

Ran models for 0 minutes
Warning: no variables in this model (and so, no generated file)!


it doesnt look like shorttable is already in the globals, though it does expect it?

Wout4:
the has_subexpr for table is indeed in another pr: https://github.com/CPMpy/cpmpy/pull/596
The solution.py was for the xcsp3 competition, I think I accidently added it here, not sure if we want those competition specific things in this pr.
Short table was merged last week so maybe you have to update first?

Will update the file names

tias:
Good, but the downloader is a hack...

I think we should follow Serdar's opinion piece: https://osullivan.ucc.ie/CPML2025/papers/kadioglu.pdf
which already is very much in line with our philosophy, but he makes the point for having dataset access like we have in ML tools.

E.g torch has a very elegant dataset class (I'm not suggesting we import torch, but we could mimic such an abstract class):
https://pytorch.org/tutorials/beginner/basics/data_tutorial.html#creating-a-custom-dataset-for-your-files

where a dataset just needs to implement init, len and getitim...

e.g. we could imagine a class like
```
class XCSP3Comp(AbsDataset):
  def __init__(self, root='.', year='2024', category='opt', transform=None, target_transform=None)
```
which would do the downloading like their MNIST dataset etc do... getitem would return (y, x) where y = whatever metadata xcsp3 has about it (e.g. optimal value, name, category, ...), and 'x' is the xcsp3 string xml

what's nice, is that we could then imagine a function `parse_xcsp3` which would take a xcsp3 string xml as input and returns a cpmpy expression tree or so? then we can do

```
D = XCSP3Comp(year='2024', category='opt', transform=parse_xcsp3)
... some parallel for (_, (model,vars)) in D:
    model.solve()
    ...
```

or something? This is just a brainstorm, esp the parse_xcsp3 as a transform... but instead of the current downloader, creating a generic non-CPMpy specific dataset object following the torch style would be great...

tias:
I think the _dataset, _cpmpy and _benchmark files are now in a stable state.

The branch does seem to need some more care to be mergable, e.g. changes on code throughout; irrelevant or needed? also readme, init etc needs to be revised... I'm also not sure if we need an example XML file in the repo (and definitely not sure about that slow/large/awkward one), nor a models/ directory. I think in the readme we can just instruct to use 'benchmark' and then that they can also run an individual file?

tias:
I was wrong... my simpler version did not manage to add a line to the csv file for the rlimit_cpu killed ones.

The current code is also not ideal because it induces a 2sec (or more) overhead on the runtimes, compared to my run of last night.

Its been a huge timesink, I've tried a range of things but I know very little about subprocesses... why do we even want to kill processes, can't we just trust the timeouts (or pay the price)?
Perhaps your previous solution was more elegant after all... I don't know

cpmpy/tools/xcsp3/xcsp3_natives.py
```diff
@@ -0,0 +1,50 @@
...
+
+# --------------------------------- OR-Tools --------------------------------- #
+
+def ort_nooverlap2d(cpm_ortools, cpm_expr):
+    start_x, dur_x, end_x, start_y, dur_y, end_y = cpm_ortools.solver_vars(cpm_expr.args)
+    intervals_x = [cpm_ortools.ort_model.NewIntervalVar(s,d,e, f"xinterval_{s}-{d}-{d}") for s,d,e in zip(start_x,dur_x,end_x)]
+    intervals_y = [cpm_ortools.ort_model.NewIntervalVar(s,d,e, f"yinterval_{s}-{d}-{d}") for s,d,e in zip(start_y,dur_y,end_y)]
+    return cpm_ortools.ort_model.add_no_overlap_2d(intervals_x, intervals_y)
+
+def ort_subcircuit(cpm_ortools, cpm_expr):
```
tias, line 12: see
https://github.com/CPMpy/cpmpy/blob/master/examples/pctsp_ortools.py

cpmpy/tools/xcsp3/xcsp3_natives.py
```diff
@@ -0,0 +1,50 @@
...
+def ort_subcircuitwithstart(cpm_ortools, cpm_expr):
+    x = cpm_expr.args
+    N = len(x)
+    arcvars = cp.boolvar(shape=(N,N))
+    # post channeling constraints from int to bool
+    cpm_ortools += [b == (x[i] == j) for (i,j),b in np.ndenumerate(arcvars)]
+    # post the global constraint
+    # posting arcs on diagonal (i==j) allows for subcircuits
+    ort_arcs = [(i,j,cpm_ortools.solver_var(b)) for (i,j),b in np.ndenumerate(arcvars) if not ((i == j) and (i == cpm_expr.start_index))] # The start index cannot self loop and thus must be part of the subcircuit.
+    return cpm_ortools.ort_model.AddCircuit(ort_arcs)
```
tias, line 32: similar to above
(not that in each case, you will need to add the variables etc directly to the cpmpy model, and then change the `return cpm_ort.ort_mod.AddCircuit(ort_arcs)` by m += DirectConstraint("AddCircuit", ort_arcs)

cpmpy/tools/xcsp3/xcsp3_natives.py
```diff
@@ -0,0 +1,50 @@
+import cpmpy as cp
+import numpy as np
+
+# --------------------------------- OR-Tools --------------------------------- #
+
+def ort_nooverlap2d(cpm_ortools, cpm_expr):
+    start_x, dur_x, end_x, start_y, dur_y, end_y = cpm_ortools.solver_vars(cpm_expr.args)
+    intervals_x = [cpm_ortools.ort_model.NewIntervalVar(s,d,e, f"xinterval_{s}-{d}-{d}") for s,d,e in zip(start_x,dur_x,end_x)]
+    intervals_y = [cpm_ortools.ort_model.NewIntervalVar(s,d,e, f"yinterval_{s}-{d}-{d}") for s,d,e in zip(start_y,dur_y,end_y)]
+    return cpm_ortools.ort_model.add_no_overlap_2d(intervals_x, intervals_y)
```
tias, line 10: With all the ones below covered (I think), that leaves only this one...

Its special, because calling DirectConstraint("add_no_overlap_2d", intx, inty) requires us to create solver specific variables... and we will not add DirectVar to CPMpy...

However, the kind of interface you are defining here is exactly how DirectConstraint also works (except wrapped in a class). So there might be a way out, e.g. (this is a sketch!)

```
class OrtNoOverlap2D(DirectConstraint):  # subclass DirectConstraint, should still pass all our transformations...
  def __init__(...): ...
  
  def callSolver(self, CPMpy_solver, Native_solver):
    # if you overload this one, then you have access to exactly what you need above...
    # you can call start, ... = CPMpy_solver.solver_vars(self.args)
    # then you can call intv_x = Native_solver.NewIntervalVar(...)
```

I think/hope...

tias (changes requested):
You currently change all solvers because you added 'added_natives' to support native constraints.

However, we already have functionality for this! It is the `DirectConstraint` (bottom of GlobalConstraints.py)

With that, I believe we do not need the `added_natives`, see inline

cpmpy/tools/xcsp3/xcsp3_benchmark.py
```diff
@@ -0,0 +1,289 @@
...
+        process.join(timeout=1)
+        # Kill if still alive
+        if process.is_alive():
+            os.kill(process.pid, signal.SIGKILL)
+            process.join()
+
+    result['time_total'] = time.time() - total_start
+
+    # Collect output
+    output = []
```
tias, line 157: should we do it this way and collect all output in an explicit array?
Can't we directly do 
````while parent_conn.poll(...):
   line = parent_conn.recv()
   if line.startswith('s ')...
```
and then an extra check on the 'exit state missing' that I don't fully get

ThomSerg:
I've created separate pull requests for some of the non-xcsp3-specific changes, can all be found under the milestone `xcsp3`.
Some where then removed from this branch, some were kept as to keep everything functioning before merge, some are only in a separate branch `xcsp3_competition` (will be meant for changes specific to the competition, such that this branch can be merged at some point).

tias:
Right, good idea.

Maybe you could make a pull request on the new xcsp3_competition branch; we won't merge it, but it will be more visible.

Then Ignace can take his small 'positive context' patch there, and then we just use this one to finish the merging of the tools infrastructure, but without xcsp3-specific optimisations we might hack in, sounds good

cpmpy/tools/xcsp3/xcsp3_dataset.py
```diff
@@ -0,0 +1,173 @@
...
+
+    """
+    XCSP3 Dataset in a PyTorch compatible format.
+    
+    Arguments:
+        root (str): Root directory containing the XCSP3 instances (if 'download', instances will be downloaded to this location)
+        year (int): Competition year (2022, 2023 or 2024)
+        track (str, optional): Filter instances by track type (e.g., "COP", "CSP", "MiniCOP")
+        transform (callable, optional): Optional transform to be applied on the instance data
+        target_transform (callable, optional): Optional transform to be applied on the file path
```
IgnaceBleukx: I was a bit confused with the docs on this, looking at the code it seems that:
- `transform` is a function that takes as input the filename of the instance
- `target_transform` is a function that takes as input the `metadata`

but that's not apparent from the docs above.

cpmpy/tools/xcsp3/xcsp3_dataset.py
```diff
@@ -0,0 +1,173 @@
...
+"""
+
+import pathlib
+from typing import Tuple, Any
+import xml.etree.ElementTree as ET
+from urllib.request import urlretrieve
+from urllib.error import HTTPError, URLError
+import zipfile
+
+class XCSP3Dataset(object):  # torch.utils.data.Dataset compatible
```
IgnaceBleukx: Do we want to make a more general `CPMpyDataset` class?

cpmpy/tools/xcsp3/xcsp3_globals.py
```diff
@@ -0,0 +1,996 @@
+"""
+Additional global constraints which are not (yet) part of the standard CPMpy collection.
+
+This file contains all the missing global constraints in order to support XCSP3-core, 
+which is a restricted scope of the complete XCSP3 specification (as used for the competitions).
+
+Currently, version 3.2 is supported.
+"""
```
tias: I think this file could benefit from a classdoc overview of the globals provided in this file...

Also, if some have the same name in main cpmpy, they should maybe get a different name that explains their reason to be (e.g. Table -> NonReifiedTable?)

cpmpy/tools/xcsp3/xcsp3_globals.py
```diff
@@ -0,0 +1,996 @@
...
+        return constraining, defining
+
+    def value(self):
+        start_index = self.args[-1]
+        succ = [argval(a) for a in self.args[:-1]] # Successor variables
+
+        # Check if we have a valid subcircuit and that the start_index is part of it.
+        return SubCircuit(succ).value() and (succ[start_index] != start_index)
+
+class Inverse(GlobalConstraint):
```
tias: this one might be in main too by now?

cpmpy/tools/xcsp3/xcsp3_globals.py
```diff
@@ -0,0 +1,996 @@
...
+    def value(self):
+        b,x,y = self.args
+        if argval(b):
+            return argval(x)
+        else:
+            return argval(y)
+
+class Element(GlobalFunction):
+    """
+        XCSP3 copy for doing Gleb-style ILP friendly decomposition
```
tias: I think we should use the main cpmpy one, this is experimental code...

cpmpy/tools/xcsp3/xcsp3_globals.py
```diff
@@ -0,0 +1,996 @@
...
+        Returns the bounds of the (numerical) global constraint
+        """
+        arr, idx = self.args
+        bnds = [get_bounds(x) for x in arr]
+        return min(lb for lb,ub in bnds), max(ub for lb,ub in bnds)
+
+
+class Cumulative(GlobalConstraint):
+    """
+        Global cumulative constraint. Used for resource aware scheduling.
```
tias: use main cpmpy's or document why another one is provided here (ideally with a different name)

cpmpy/tools/xcsp3/xcsp3_globals.py
```diff
@@ -0,0 +1,996 @@
...
+        # demand doesn't exceed capacity
+        lb, ub = min(start), max(end)
+        for t in range(lb, ub+1):
+            if capacity < sum(demand * ((start <= t) & (t < end))):
+                return False
+
+        return True
+
+
+class GlobalCardinalityCount(GlobalConstraint):
```
tias: same, cut or rename/document

cpmpy/tools/xcsp3/xcsp3_natives.py
```diff
@@ -0,0 +1,90 @@
+"""
+A collection of XCSP3 solver-native global constraints.
```
tias: not sure we need these in a separate file... could be part of XCSP3_globals?

I don't feel very strong about it, just careful with all these similarly named files...

cpmpy/tools/xcsp3/xcsp3_solution.py
```diff
@@ -0,0 +1,51 @@
+"""
+Collection of tools for handeling solutions in XCSP3 format.
```
tias: its only used in xcsp3_cpmpy, so I would just include it there?

a file for one function that is only used in one place...

setup.py
```diff
@@ -55,7 +55,7 @@ def get_version(rel_path):
         # Solvers
         **solver_dependencies,
         # Tools
-        # "xcsp3": ["pycsp3"], <- for when xcsp3 is merged
+        "xcsp3": ["pycsp3", "requests"], # didn't add CLI-specific req since some are not cross-platform
```
tias: ah, I didn't realize that. Could you document that in the README?

cpmpy/tools/xcsp3/xcsp3_natives.py
ThomSerg: I'm in favor of joining them into a single file (since with the new implementation build on `DirectConstraint` they are basically globals, so splitting them into a separate file is more an artifact of the old implementation). But besides that, I'm not too fond of the file naming scheme. Why call them `xcsp3_<something>.py` when they are already in the `/tools/xcsp3` directory? Wouldn't it be nicer to be able to call something like `from tools.xcsp3.dataset import XCSP3Dataset`?  `xcsp3_cpmpy` might need to be kept as, to keep it clear that it is the executable submission of cpmpy to xcsp3.

cpmpy/tools/xcsp3/xcsp3_solution.py
ThomSerg: I put it in a separate file so that users could have access to the XCSP3 solution formatting outside of the competition context (the formatting is part of the XCSP3-core specification). Maybe I should move it to `tools/xcsp3/__init__.py` as done for `read_xcsp3`?

cpmpy/tools/xcsp3/xcsp3_natives.py
tias: yes, I agree the filenames don't need to be prepended with xcsp3_, though for the 'runner' that might still make sense idd, go ahead

cpmpy/tools/xcsp3/xcsp3_solution.py
tias: yes; I'm not actually fond of hiding things in --init--, but at least it would be a consistent place : )

cpmpy/tools/xcsp3/README.md
```diff
@@ -0,0 +1,84 @@
+# CPMpy XCSP3 tools
+
+This directory contains a collection of tools for reading, loading and solving XCSP3 instances in CPMpy.
+
+What is included:
+- utilities for reading and loading XCSP3 instances
```
tias, line 6: and writing

tias: (solutions)

cpmpy/tools/xcsp3/README.md
```diff
@@ -0,0 +1,84 @@
...
+- a PyTorch-compatible dataset class for loading (and downloading) XCSP3 instances
+- cli for solving individual XCSP3 instances and outputting the result in the competition format
```
tias, line 8: well, if we consider that part of this, then that is where the code should go, not in the init : )

cpmpy/tools/xcsp3/README.md
```diff
@@ -0,0 +1,84 @@
...
+- cli for benchmarking CPMpy across a large collection of XCSP3 instances
+- cli for processing and visualizing benchmark results
+
+Only the XCSP3-Core specification version 3.2 is currently supported.
```
tias: maybe drop the 'Only'? its the one of the compettition...

cpmpy/tools/xcsp3/xcsp3_globals.py
```diff
@@ -0,0 +1,1041 @@
...
+    NonReifiedTable
+    ShortTable
+    NegativeShortTable
+    MDD
+    Regular
+    NotInDomain
+    NoOverlap2d
+    IfThenElseNum
+    Element
+    Cumulative
```
tias: I think its called DynamicCumulative now; and I don't think I saw Element.

Plus the solver specific ones to add?

cpmpy/tools/xcsp3/xcsp3_globals.py
```diff
@@ -0,0 +1,1041 @@
...
+        # --> not necessary for comp, because positive context
+        # sink = len(self.nodes)
+        # transitions += [[self.node_map[n], v, sink] for n in self.nodes for v in range(lb, ub + 1) if (n, v) not in self.trans_dict]
+        # transitions += [[sink, v, sink] for v in range(lb, ub + 1)]
+
+        # keep track of current state when traversing the array
+        state_vars = intvar(0, len(self.nodes)-1, shape=len(arr))
+        id_start = self.node_map[start]
+        # optimization: we know the entry node of the automaton, results in smaller table
+        defining = [Table([arr[0], state_vars[0]], [[v,e] for s,v,e in transitions if s == id_start])]
```
IgnaceBleukx: I suppose this could become `NonRefiedTable` instead?

cpmpy/tools/xcsp3/xcsp3_dataset.py
ThomSerg: I have added alternative descriptions. The original ones tried to stay a bit too close to their original meaning in the case of more standard pytorch datasets. Here there is not really a "target", just some metadata with more info on the problem instance. And the data (for transform) is not the instances, but paths to the instances.

cpmpy/tools/xcsp3/xcsp3_dataset.py
ThomSerg: Would thinks so. An abstract class for people to build upon.

cpmpy/tools/xcsp3/xcsp3_globals.py
ThomSerg: In more than one place right? Everywhere that 'Table' occurs in "defining" I suppose.

IgnaceBleukx: Yes indeed, but even if it occurs in "constraining" we could replace it, as we know the globals in xcsp3 are always toplevel

cpmpy/tools/xcsp3/xcsp3_dataset.py
IgnaceBleukx: Yes, that could at least serve as a template for others; it seems weird to have to go to a specific implementation (xcsp3 or #701) to see how it's done.
@tias what do you think?

cpmpy/tools/xcsp3/xcsp3_globals.py
ThomSerg: In the xcsp3 globals decomposition for `AllDifferentListsExceptN` we do create a reified Table ourselves. So that's the only place where I now left the standard `Table`, everything else has been replaced with a `NonReifiedTable`.

IgnaceBleukx: I see, you could actually also get away with a NonRefieidTable there as it is an implication (and thus in postive context)

IgnaceBleukx:
Seems ready to go for me, does @tias want to have a last look as well?

cpmpy/tools/xcsp3/xcsp3_dataset.py
tias: No opinion, not a requirement for this PR.
If we add #701 then yes, the second one should perhaps add it...

tias (approved):
Yes, thanks a lot for all the effort in getting it ready cleanly and still with all the features and robustness : )

tias:
so 'yes' means, I looked at it and its ready to merge : )

PR #599, opened 2025-02-26
Add optional dependencies to `setup.py`

ThomSerg:
Add optional dependencies to the `setup.py`'s `extras_require`.
Allows easy-install of solvers through `pip install cpmpy[choco, z3, pysat, ...]`. (with cpmpy-native solver names)
Also useful for bundling optional dependencies, e.g. those required by one of the tools: `pip install cpmpy[xcsp3]`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ setup(
         "gcs": ["gcspy"],
         "cpo": ["docplex"],
         # Tools
-        "xcsp3": ["pycsp3"],
+        # "xcsp3": ["pycsp3"], <- for when xcsp3 is merged
         # Other
         "test": ["pytest"],
     },
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     #extra dependency, only needed if minizinc is to be used.
     extras_require={
         "FULL":  ["minizinc"],
-        "test": ["pytest"],
+        # Solvers
         "ortools": ["ortools"],
         "z3": ["z3-solver"],
         "choco": ["pychoco>=0.2.1"],
@@ -47,7 +47,12 @@ setup(
         "pysat": ["python-sat"],
         "gurobi": ["gurobipy"],
         "pysdd": ["pysdd"],
-        "xcsp3": ["pycsp3"]
+        "gcs": ["gcspy"],
+        "cpo": ["docplex"],
+        # Tools
+        "xcsp3": ["pycsp3"],
+        # Other
+        "test": ["pytest"],
     },
     classifiers=[
         "Programming Language :: Python :: 3 :: Only",
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,16 @@ setup(
     #extra dependency, only needed if minizinc is to be used.
     extras_require={
         "FULL":  ["minizinc"],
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -38,6 +38,21 @@ def get_version(rel_path):
     #extra dependency, only needed if minizinc is to be used.
     extras_require={
         "FULL":  ["minizinc"],
```
hbierlee, line 40: What is this one still for?

hbierlee:
That's great, I prefer this setup!

I'll add some comments.

One question, shouldn't we remove the `requirements.txt` file as well? To me it is not clear why you would need both this and setup.py (although it is clear to other people https://stackoverflow.com/questions/43658870/requirements-txt-vs-setup-py)

setup.py
hbierlee, line 40: What is this one still for?

setup.py
```diff
@@ -38,6 +38,21 @@ def get_version(rel_path):
...
+        "minizinc": ["minizinc"],
+        "pysat": ["python-sat"],
+        "gurobi": ["gurobipy"],
+        "pysdd": ["pysdd"],
+        "gcs": ["gcspy"],
+        "cpo": ["docplex"],
+        # Tools
+        # "xcsp3": ["pycsp3"], <- for when xcsp3 is merged
+        # Other
+        "test": ["pytest"],
```
hbierlee, line 55: Perhaps add `pytest-xdist[psutil]` for the parallel test runs with `pytest -n auto`?

setup.py
```diff
@@ -38,6 +38,21 @@ def get_version(rel_path):
...
     },
```
hbierlee, line 56: Perhaps add a `docs` for building docs?

setup.py
```diff
@@ -38,6 +38,21 @@ def get_version(rel_path):
     #extra dependency, only needed if minizinc is to be used.
```
hbierlee, line 38: Perhaps the github cli config can also be updated to re-use the optional dependencies

setup.py
ThomSerg: That one seems more something to add as a recommendation in the documentation, rather than something that is strictly required to run our testsuite.

hbierlee:
Maybe also recommend this way of installing solvers in the docs and in the import exceptions?

Unfortunately, we can't check for installed extras in the code, otherwise we could have removed the caught import exceptions.

tias:
bump, ready to merge or still open points @hbierlee ?   Please merge if ready.

hbierlee:
One outstanding comment but it is not a mandatory change. Good to merge for me otherwise

tias:
this is nice.

what is the open comment? about the github CI?

that is out of scope... it updates the code and the docs, so it is ready to merge.

So please merge @hbierlee and if you think the CI should be updated, please file a separate issue or PR.

setup.py
hbierlee: This is quite separate, so doesn't have to be done here. If I have time, I'll do it for this release https://github.com/CPMpy/cpmpy/issues/617

hbierlee:
good to merge for me, but I'll leave that to @ThomSerg

PR #600, opened 2025-02-28
Add `pindakaas` solver

hbierlee:
A draft PR for adding `pindakaas` with basic features expected of a CPMpy solver.

Awaiting finalizing the upstream (`pindakaas`) release with the Python interface (https://github.com/pindakaashq/pindakaas/pull/86) to replace the submodule


cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,313 @@
...
+        literals = None
+        coefficients = None
+        comparator = None
+        k = None
+        if isinstance(cpm_expr, _BoolVarImpl):
+            literals = [cpm_expr]
+        elif isinstance(cpm_expr, Operator) and cpm_expr.name == 'or':
+            literals = cpm_expr.args
+        elif hasattr(cpm_expr, 'decompose'):
+            self += a0.implies(cpm_expr.decompose())
```
tias, line 288: a0 seems undefined, I think you just want to self += expr.decomp()

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,313 @@
...
+from ..transformations.normalize import toplevel_list, simplify_boolean
+from ..transformations.linearize import linearize_constraint
+from ..transformations.reification import only_implies, only_bv_reifies
+
+import time
+
+
+class CPM_pindakaas(SolverInterface):
+    """
+    Interface to TEMPLATE's API
```
tias, line 49: TODO, update (but maybe wait 1-2 days till new layout is in master)

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,313 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+##
+## pindakaas.py
+##
+"""
+    Interface to Pindakaas (`pkl`) API
```
tias, line 7: pkl for me is for Pickle files mentally

Given the name Pindakaas, 'pdk' seems more natural to me?

cpmpy/solvers/pindakaas.py
tias: also, this won't work well if 'conditions' is not empty?

tias (commented):
Clean interface.

I don't think (reified) decompositions work well yet.

Rebase on master and do not include changes like /python/env unrelated to the PR pls thx

cpmpy/solvers/pindakaas.py
hbierlee: I believe this attribute is the responsibility of the transformations, so I've removed it.

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,322 @@
+# /usr/bin/env python
+"""
+Interface to Pindakaas (`pkd`) API.
+
+Requires that the `pkd` library python package is installed:
+
+    $ pip install pindakaas
+
+`pkd` is a library to transform pseudo-Boolean and integer constraints into conjunctive normal form.
```
tias, line 9: please see the new way of documenting the top of solvers on master

(and; pdk?)

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,322 @@
...
+from ..transformations.get_variables import get_variables
+from ..transformations.linearize import linearize_constraint
+from ..transformations.normalize import simplify_boolean, toplevel_list
+from ..transformations.reification import only_bv_reifies, only_implies
+from .solver_interface import ExitStatus, SolverInterface
+
+
+class CPM_pindakaas(SolverInterface):
+    """
+    Interface to TEMPLATE's API.
```
tias, line 44: some TEMPLATE's not yet removed
(also, see the new way of starting a solver file and class doc)

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,322 @@
...
+        if self.unsatisfiable:
+            return self
+
+        # add new user vars to the set
+        get_variables(cpm_expr_orig, collect=self.user_vars)
+
+        # transform and post the constraints
+        try:
+            for cpm_expr in self.transform(cpm_expr_orig):
+                self._add(cpm_expr)
```
tias, line 266: is this worth putting in a separate function, if it is only used here?
(function calls can be expensive in hot paths in Python... no inlineing)

cpmpy/transformations/linearize.py
```diff
@@ -168,9 +168,9 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
 
                 elif lhs.name == "pow" and "pow" not in supported:
                     if "mul" not in supported:
-                        raise NotImplementedError("Cannot linearize power without multiplication")
+                        raise TransformationNotImplementedError("Cannot linearize power without multiplication")
```
tias, line 171: I don't think this should be part of the solver commit...

examples/php.py
```diff
@@ -0,0 +1,9 @@
+import cpmpy as cp
```
tias, line 1: should have at least a little comment on it being the pigenhole problem...

should not be part of solver commit, OK to add to examples

tests/test_examples.py
```diff
@@ -9,10 +9,13 @@
 from glob import glob
```
tias, line 9: should not be part of solver commit...

tests/test_pysat_wsum.py
```diff
@@ -5,9 +5,10 @@
 from cpmpy.solvers.pysat import CPM_pysat
 
 import importlib # can check for modules *without* importing them
+pysat_available = importlib.util.find_spec("pysat") is not None
```
tias, line 8: should be in separate commit

tests/test_trans_linearize.py
```diff
@@ -343,7 +344,7 @@ def test_pow(self):
 
         # this is not supported
         cons = a ** b == 3
-        self.assertRaises(NotImplementedError,
+        self.assertRaises(TransformationNotImplementedError,
```
tias, line 347: should not be part of this commit here

tias (commented):
some cleanup requests

tests/test_examples.py
hbierlee: This is always a bit tricky, because I already made separate PR for some of these unrelated issues you see. So once that PR is merged (#651 ) then this diff should disppear. Plus, I am in part already relying on these changes (in this case, running all examples to ensure correctness of the new solver).

Any preference if I should in the meantime remove these unrelated changes untill they are merged back in from master?

cpmpy/solvers/pindakaas.py
hbierlee: This function is used later to handle the implication, which makes it much cleaner (compare to say how pysat.py handles implications).

For performance, recursion can be worse, but in this case we are guaruantueed the deph is at most one (due to linearize). In other cases with more depth, I would still do it like this and use profiling. I'm very in favour of this style of recursively traversing expression trees, because it makes the code much much cleaner. If your function handles `e : Expr`, and this happens to be `p : BV -> sub : Expr`, then you should call the function again on sub. You will see the same pattern in int2bool.

tests/test_examples.py
hbierlee: I've reverted these changes on this branch

tias (commented):
getting there, still some comments

.github/workflows/python-test.yml
```diff
@@ -16,6 +18,7 @@ jobs:
           pip install ".[test, z3, choco, exact, pysat, pysdd, choco, minizinc]"
           pip install pypblib  # dependency of pysat for PB constraints
           pip install pytest-xdist
+          pip install ./submodules/pindakaas/crates/pyndakaas
```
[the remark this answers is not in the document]
hbierlee: Still in the process of updating this.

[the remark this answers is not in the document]
hbierlee: I've removed the submodule, and added an optional `pindakaas` dependency to `setup.py` which installs a "release" cpmpy branch if you run `pip install cpmpy[pindakaas]` which has everything we need:

`"pindakaas": ["pindakaas @ git+https://github.com/hbierlee/pindakaas.git@feature/cpmpy#subdirectory=crates/pyndakaas"]`

Once we release pindakaas 0.2 (https://github.com/pindakaashq/pindakaas/milestone/1) and publish that on pypi, we can switch this just to `"pindakaas": ["pindakaas"]`. Apart from this being ugly, it's also has some authentication/https/git dependency, so obviously still not ideal. But I wanted to clean up things as much as possible.

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,274 @@
+# /usr/bin/env python
+"""
+Interface to Pindakaas (`pindakaas`) API.
+
+Requires that the `pindakaas` library python package is installed:
```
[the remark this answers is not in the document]
hbierlee: Should look similar now. Some functions do not have docs if there is no different with the documentation inherited from `solver_interface.py`

hbierlee:
Arguably ready for review/merge, with one big caveat that we do not have a release on Pypi yet (pending: https://github.com/pindakaashq/pindakaas/milestone/1), so we pip install from source via a pinned commmit on GitHub.

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,258 @@
...
+    Documentation of the solver's own Python API:
+
+    - https://pypi.org/project/pindakaas/
+    - https://github.com/pindakaashq/pindakaas
+
+    """
+
+    @staticmethod
+    def supported():
+        return importlib.util.find_spec("pindakaas") is not None
```
tias: so we can learn for our other solver interfaces... is there a reason you do it this way, instead of the try/except of the template?

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,258 @@
...
+            if cpm_var.name not in self._varmap:
+                (self._varmap[cpm_var.name],) = self.pdk_solver.new_vars(1)
+            return self._varmap[cpm_var.name]
+        else:
+            raise NotImplementedError(
+                f"{self.name}: unexpected variable {cpm_var} of type {type(cpm_var)} not supported"
+            )
+
+    def transform(self, cpm_expr):
+        cpm_cons = toplevel_list(cpm_expr)
```
tias, line 233: partial's too?
It doesn't supports ints at this point, so maybe not that meanginful, but still...

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,258 @@
...
+        # transform and post the constraints
+        try:
+            for cpm_expr in self.transform(cpm_expr_orig):
+                self._add(cpm_expr)
+        except pdk.Unsatisfiable:
+            self.unsatisfiable = True
+
+        return self
+
+    __add__ = add
```
tias: we have some comment about why we do this, could you add that here too?

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def _add(self, cpm_expr, conditions=[]):
```
tias: I would suggest a better name... esp with --add-- defined just above;  `add_transformed` would fit the documentation provided below...

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,258 @@
...
+            self.pdk_solver.add_clause(conditions + self.solver_vars(cpm_expr.args))
+
+        elif cpm_expr.name == "->":  # implication
+            a0, a1 = cpm_expr.args
+            self._add(a1, conditions=conditions + [~self.solver_var(a0)])
+
+        elif isinstance(cpm_expr, Comparison):  # Bool linear
+            # lhs is a sum/wsum, right hand side a constant int
+            lhs, rhs = cpm_expr.args
+            match lhs.name:
```
tias: this is only allowed in Python Someversion (11 or so?), while the minimal requirement for CPMpy is below that (9?), as far as I know??

tias: (and we are not going to bump the minimal version just for this file ; )

cpmpy/solvers/pindakaas.py
hbierlee: The advantage to me is that this does not import the module if the solver is installed. It's more in line with a `supported` check, let's say you just want to check which solvers are installed using `SolverLookup.print_status`, then you don't want to necessarily load all the solvers.

However, this should be a global change for all solvers, so I'll make a separate ticket and just conform to the other style here.

hbierlee:
Comments resolved, and:

- Using the released pindakaas function
- Implemented `get_core`
- However, Pindakaas' `failed` function has a bug, which will need a patch release (see [PR](https://github.com/pindakaashq/pindakaas/pull/135)).

cpmpy/solvers/pindakaas.py
hbierlee: See https://github.com/CPMpy/cpmpy/issues/702

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+
+============
+Installation
+============
+
+Requires that the 'pindakaas' optional dependency is installed:
+
+.. code-block:: console
+
+    $ pip install cpmpy[pindakaas]
```
IgnaceBleukx: I suppose this should be `pip install pindakaas`?

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+        - `pdk_solver`: The pindakaas back-end which will encode and post constraints for the SAT solver
+        - `unsatisfiable`: If a constraint is found to be unsatisfiable during the encoding phase, this flag is set to `True` to prevent further encoding efforts
+        """
+        name = "pindakaas"
+        if not self.supported():
+            raise ImportError(
+                f"CPM_{name}: Install the Pindakaas python library `pindakaas` (e.g. `pip install cpmpy[pindakaas]`) package to use this solver interface"
+            )
+        if cpm_model and cpm_model.objective_ is not None:
+            raise NotSupportedError(
```
IgnaceBleukx: style: no newlines

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+                f"CPM_{name}: only satisfaction, does not support an objective function"
+            )
+
+        import pindakaas as pdk
+
+        assert (
```
IgnaceBleukx: assert can be 1 line, also no brackets needed

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+            return self._solve_return(self.cpm_status)
+
+        if time_limit is not None and time_limit <= 0:
+            raise ValueError("Time limit must be positive")
+
+        # ensure all vars are known to solver
+        self.solver_vars(list(self.user_vars))
+
+        time_limit = None if time_limit is None else timedelta(seconds=time_limit)
+        solver_assumptions = (
```
IgnaceBleukx: well, if you're going to do it in 3 lines, than why not just an if-statement? I think that's more clear

```python
if assumptions is not None:
    assumptions = self.solver_vars(assumptions)
```

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+            raise NotSupportedError(
+                f"CPM_{name}: only satisfaction, does not support an objective function"
+            )
+
+        import pindakaas as pdk
+
+        assert (
+            subsolver is None
+        ), "Pindakaas does not support any subsolvers for the moment"
+        self.pdk_solver = pdk.solver.CaDiCaL()
```
IgnaceBleukx, line 115: When checking the pindakaas github it seems there is support for KisSAT and IntelSAT?
Ah but maybe not in the released version yet?

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+        time_limit = None if time_limit is None else timedelta(seconds=time_limit)
+        solver_assumptions = (
+            None if assumptions is None else self.solver_vars(assumptions)
+        )
+
+        t = time.time()
+        with self.pdk_solver.solve(
+            time_limit=time_limit,
+            assumptions=solver_assumptions,
+        ) as result:
```
IgnaceBleukx, line 160: what does `result` contain? Only the status and solution? Or also some solving statistics?
If it also includes such information, we normally save it as an attribute (self.pdk_result = self.pdk_solver.solve(...)

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+            elif result.status == pdk.solver.Status.UNSATISFIABLE:
+                self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+            elif result.status == pdk.solver.Status.UNKNOWN:
+                self.cpm_status.exitstatus = ExitStatus.UNKNOWN
+            else:
+                raise NotImplementedError(
+                    f"Pindakaas returned an unkown type of result status: {result}"
+                )
+
+            # # True/False depending on self.cpm_status
```
IgnaceBleukx: double "#"

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+            # translate solution values (of user specified variables only)
+            if has_sol:
+                # fill in variable values
+                for cpm_var in self.user_vars:
+                    if cpm_var.name in self._varmap:
+                        lit = self.solver_var(cpm_var)
+                        cpm_var._value = result.value(lit)
+                        if cpm_var._value is None:
+                            cpm_var._value = True  # dummy value
+                    else:  # if pindakaas does not know the literal, it will error
```
IgnaceBleukx: well, this should not happen anymore, right? Because before you solve, we call `self.solver_vars(user_vars)`. So it should be mapped

hbierlee:
Note to self:

- [ ] Still have to enable CSE, I noticed
- [ ] Bug is now fixed on pindakaas develop, have to check if everything is good for new release

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+                self._add_transformed(cpm_expr)
+        except pdk.Unsatisfiable:
+            self.unsatisfiable = True
+
+        return self
+
+    __add__ = add  # avoid redirect in superclass
+
+    def _add_transformed(self, cpm_expr, conditions=[]):
+        """Add for a single, *transformed* expression, implied by conditions."""
```
IgnaceBleukx: I was confused at first what this function did, or why it was required at all (instead of just doing it in `add`).
But now I get that it is because of the implications.
Might be better to rename this to `_post_constraint`, which is also how it's called in the OR-Tools interface.

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,286 @@
...
+        elif isinstance(cpm_expr, Comparison):  # Bool linear
+            # lhs is a sum/wsum, right hand side a constant int
+            lhs, rhs = cpm_expr.args
+            if lhs.name == "sum":
+                literals = lhs.args
+                coefficients = [1] * len(literals)
+            elif lhs.name == "wsum":
+                coefficients, literals = lhs.args
+            else:
+                raise ValueError(
```
IgnaceBleukx: style: put on 1 line

tests/test_solvers.py
```diff
@@ -822,8 +845,8 @@ def test_incremental_assumptions(self, solver):
         if "assumptions" not in inspect.signature((s.solve)).parameters:
             return # solver does not support solving under assumptions
         
-        if solver == "pysdd":
-            return # not implemented in pysdd
+        if solver == ("pysdd", "pindakaas"):
```
IgnaceBleukx: should be "if solver **in** ..."

tests/test_solvers.py
```diff
@@ -871,7 +894,7 @@ def test_false(self, solver):
         assert not cp.Model([cp.boolvar(), False]).solve(solver=solver)
 
     def test_partial_div_mod(self, solver):
-        if solver in ("pysdd", "pysat", "pumpkin"):  # don't support div or mod with vars
+        if solver in ("pysdd", "pysat", "pindkaas", "pumpkin"):  # don't support div or mod with vars
```
IgnaceBleukx: typo

tests/test_solvers.py
```diff
@@ -975,8 +998,8 @@ def test_hidden_user_vars(self, solver):
         Tests whether decision variables which are part of a constraint that never gets posted to the underlying solver
         still get correctly captured and posted.
         """
-        if solver == 'pysdd' or solver == 'pysat':  # pysat and pysdd don't support integer decision variables
-            return
+        if solver in ('pysdd', 'pysat', 'pindakaas'):
+            pytest.skip(reason="pysat and pysdd don't support integer decision variables")
```
IgnaceBleukx: can update this to `f"{solver} does not support integer decision variables"`

cpmpy/solvers/pindakaas.py
hbierlee: Their Python interface is not yet ready, indeed, because I believe their are some problems building them on all platforms. I will talk with Jip about this, but this will be for the next release.

cpmpy/solvers/pindakaas.py
hbierlee: This would be normally clearer (changed it for the `time_limit`), but note I need to keep both the CPMpy `assumptions` and their solver var mapping `solver_assumptions`, and both need to be either set or `None`. So it'd be the ternary but rewritten:

```py
if assumptions is None:
    solver_assumptions = None
else:
    solver_assumptions = self.solver_vars(assumptions)
```

Here and at other places it's a bit brackety, that's just my autoformatter for which I need to tweak the max linewidth setting.

cpmpy/solvers/pindakaas.py
hbierlee: Interestingly, the result *should* only provide a function to directly access the solver's state regarding status, and literals value (when SAT) or failed (when UNSAT). If there are millions of literals in the solution, copying it and storing it for every solution when not needed causes overheads. So I just extract what I need from it to adhere to the CPMpy specs and then release. Internally, pindakaas does do some copying of this result state, but it's not guaruantueed to stay valid after another solve call. This does have some downsides, namely that I need to copy the unsat core when UNSAT, even if the user has not asked for it yet.

cpmpy/solvers/pindakaas.py
hbierlee: Ah nice one, this made the loop significantly shorter.

hbierlee:
I think should be good to merge, tests should pass. The only thing is that I fixed the UNSAT core's, but this is not yet released, so I gated it behind an `Unsupported` exception. Since the `get_core` is an optional solver feature from cpmpy's perspective, so for me it'd be ok to merge this, but I'll run this by Tias later today.

cpmpy/solvers/pindakaas.py
IgnaceBleukx: Okay, then we keep it as is, I already had some suspicion it would be because of Rust lifetimes ; )

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,308 @@
...
+            if has_sol:
+                # fill in variable values
+                for cpm_var in self.user_vars:
+                    # essentially `.solver_var`, but failing if new vars are added
+                    if isinstance(cpm_var, NegBoolView):
+                        lit = ~self._varmap[cpm_var._bv.name]
+                    elif isinstance(cpm_var, _BoolVarImpl):
+                        lit = self._varmap[cpm_var.name]
+                    else:
+                        raise TypeError
```
IgnaceBleukx: Can you fill in this error with some more information? Also not sure if TypeError is the right class to use here... It would mean there is really something gone wrong during transformation/posting of the constraint... 
In OR-Tools we raise a ValueError:
https://github.com/CPMpy/cpmpy/blob/177686f0871f6a85f6328fd349a4de9ff0acebe2/cpmpy/solvers/ortools.py#L263 

For this an appropriate error message could be something like:
```python
f"Integer variables should be encoded using int2bool transformation, but {cpm_var} is integer, please report on github"
```

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,308 @@
...
+                    cpm_var._value = result.value(lit)
+                    if cpm_var._value is None:
+                        cpm_var._value = True  # dummy value
```
IgnaceBleukx: when can this happen?

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,308 @@
...
+                for enc in self.ivarmap.values():
+                    enc._x._value = enc.decode()
+
+            else:  # clear values of variables
+                for cpm_var in self.user_vars:
+                    cpm_var._value = None
+                # we have to save the unsat core here, as the result object does not live beyond this solve call
+                if assumptions is not None:
+                    self.core = [
+                        x for x, s_x in zip(assumptions, solver_assumptions) if result.failed(s_x)
```
IgnaceBleukx: can you put this one 1 line instead?

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,308 @@
...
+                literals = lhs.args
+                coefficients = [1] * len(literals)
+            elif lhs.name == "wsum":
+                coefficients, literals = lhs.args
+            else:
+                raise ValueError(f"Trying to encode non (Boolean) linear constraint: {cpm_expr}")
+
+            lhs = sum(c * l for c, l in zip(coefficients, self.solver_vars(literals)))
+
+            if cpm_expr.name == "<=":
```
IgnaceBleukx: we actually have a helper function for this `eval_comparison` in `expressions.utils`, it also ensures numpy-stuff is done correctly here (e.g., if rhs is a numpy integer).
You can still add an assertion error like `assert cpm_expr.name in frozenset({"<=", ">=", "=="})` or smt

tests/test_int2bool.py
```diff
@@ -22,8 +22,7 @@
 c = intvar(2, 2, name="c")
 
 SOLVERS = [
```
IgnaceBleukx, line 24: I suppose this can just be removed?

cpmpy/solvers/pindakaas.py
hbierlee: Our doc says that "Note that the [value] function can return None if the model/solution is independent of the given literal.", but I can't make it happen from CPMpy's side.

IgnaceBleukx: I see, so creating the variable at pindakaas level is not enough to ensure it gets a value during the solve call?
That's fine then : )

hbierlee: I can't make it happen though.. so I've replaced it with an assert

tests/test_int2bool.py
hbierlee: The `SOLVERS` var? I think it's just an easy way to add additional solvers to test the `int2bool` on e.g. `pysat`, but now it only has `pindakaas` because that is enough for CI purposes. Actually, I've added `pysat` as well, only takes 1 second to run, and it is nice to just have this check for any solver depending on `int2bool`.

hbierlee:
Thanks for the review! The CI is being a little weird, the previous commit finished in 28 minutes, then the failing commit only removed a comment and suddenly timed out, rerunning it and it finished in 48 minutes.

tests/test_int2bool.py
IgnaceBleukx: Yeah but it's overridden right below? Ok to keep but just wondering

hbierlee: Yes, it is not used elsewhere, so it's overridden. Still a little clearer and easier to change/comment/uncomment which solver you're working with during development than having it in-line in the generator.

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,302 @@
...
+from .solver_interface import ExitStatus, SolverInterface
+
+
+class CPM_pindakaas(SolverInterface):
+    """
+    Interface to Pindakaas' Python API.
+
+    Creates the following attributes (see parent constructor for more):
+
+    - ``pdk_solver``: the `pindakaas` solver or formula object
```
tias: please document all of 'ivarmap', 'encoding', 'unsatisfiable', 'core' here; (and none in the init below)

cpmpy/solvers/pindakaas.py
```diff
@@ -0,0 +1,302 @@
...
+
+    Documentation of the solver's own Python API:
+
+    - https://pypi.org/project/pindakaas/
+    - https://github.com/pindakaashq/pindakaas
+
+    """
+
```
tias, line 79: master now has a version() checking tool for each solver, consider adding it (or we fix it in a separate PR, like #715 )

tias (approved):
Great!

One documentation change request, one `version()` comment that has since landed in master.

PR #604, opened 2025-03-18
Simplify nested wsum boolexpr

ThomSerg:
The variables argument of `wsum` was not recursively send through `simplify_boolean`, causing exceptions in downstream transformations. Bug identified by @ElFosco.

Example expression:
```python
bv = cp.boolvar(name="bv", shape=2)
weights = cp.cpm_array([1, 2])
bool_as_ints = cp.cpm_array([0, 1])
cons = sum( weights * (bv != bool_as_ints) ) == 1
```
this will look something like:
```python
sum([1, 2], [bv[0] != 0, bv[1] != 1]) == 1 # wsum
```
which after `simplify_boolean` should become:
```python
sum([1, 2], [bv[0], ~bv[1]]) == 1 # wsum
```

Otherwise causes `normalized_boolexpr` to fail, since it does not expect a `int` to still be present in the right hand side of a `!=` (assumes a var to be present, see #315)

### DIFF
--- a/cpmpy/transformations/normalize.py
+++ b/cpmpy/transformations/normalize.py
@@ -107,6 +107,9 @@ def simplify_boolean(lst_of_expr, num_context=False):
                 else:
                     newlist.append(~args[0])
 
+            elif expr.name == "wsum":
+                newlist.append(Operator(expr.name, (args[0], simplify_boolean(args[1]))))
+
             else: # numerical expressions
                 newlist.append(Operator(expr.name, args))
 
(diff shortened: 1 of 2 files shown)


IgnaceBleukx (commented):
I don't think this is the correct way of fixing this as we now have another recursive call.
On line 69 we already call simplify_boolean on the args... The issue is that a list of lists is not processed recursively and falls into the default-case all the way at the bottom... I guess that's what we should actually fix ; )

That being said, in PR #592 I did quite a big effort in refactoring this transformation, so maybe we can just put a patch for this issue directly there instead.

ThomSerg:
Indeed, didn't see line 69. Will look into creating that patch.

Issue #608, opened 2025-04-08
Bug in AllDifferentExceptN and AllDifferentExcept0 for multiple solvers


peerjosse:
bugged solvers: minizinc, z3
other tested (correct) solvers: gurobi, ortools, choco

The bug in the following program:
- happens for both AllDifferentExceptN as AllDifferentExcept0
- should give 9 solutions given this domain
- disappears when the constraint (x != 0) is added (shouldn't change anything)

```
import cpmpy as cp

x = cp.intvar(1, 9, shape=1)

m = cp.Model()

m += (cp.AllDifferentExceptN([x], 1))

# # Bug disappears:
# m += x != 0

# # Also works:
# m += (cp.AllDifferentExcept0([x]))

# Minizinc shows 1 solution (incorrect)
print("Amount of Minizinc solutions: ", m.solveAll(solver="minizinc", solution_limit=1000))
# z3 shows 1 solution (incorrect)
print("Amount of z3 solutions as check: ", m.solveAll(solver="z3", solution_limit=1000))
# Gurobi shows 9 solutions (correct)
print("Amount of Gurobi solutions: ", m.solveAll(solver="gurobi", solution_limit=1000))
# ortools shows 9 solutions (correct)
print("Amount of ortools solutions as check: ", m.solveAll(solver="ortools", solution_limit=1000))
# pysat shows 9 solutions (correct)
print("Amount of choco solutions as check: ", m.solveAll(solver="choco", solution_limit=1000))
```

Not sure what else to include in a bug report, so asking for any more information is welcomed.

ThomSerg:
This seems to be an issue due to variables never being added to the model. `cp.AllDifferentExceptN([x], 1)` is a constraint that says nothing, thus after decomposition it results in 'nothing' (no constraint at all):

```python
def decompose(self):
        # equivalent to (var1 == n) | (var2 == n) | (var1 != var2)
        return [(var1 == var2).implies(cp.any(var1 == a for a in self.args[1])) for var1, var2 in all_pairs(self.args[0])], []
```

Since the supplied array of decision variables has length 1, an empty list gets returned resulting in nothing to be posted to the solver. So the backend solver is not aware of `x`, as can be seen by its value after solving:

```python
print(x.value()) # >> None
```

[...]

PR #609, opened 2025-04-08
Missing user vars

ThomSerg:
Not all solvers correctly collected all decision variables across all constraints. 

Z3 does not call `get_variables` upon adding a constraint. Comment mentions that z3 uses `solver_var()` via `_z3_expr` to collect the variables during the transformation process, But this misses variables from constraints which decomposed to nothing (see #608).

Minizinc didn't post the collected user vars on `solveAll()`. Regular `solve()` worked perfectly. 

I added the missing lines to ensure all user decision variables are collected and posted to the solver. 
But current implementation posts decision variables twice for Z3 (not sure if this could be an issue).

### DIFF
--- a/cpmpy/solvers/minizinc.py
+++ b/cpmpy/solvers/minizinc.py
@@ -350,6 +350,10 @@ class CPM_minizinc(SolverInterface):
 
     async def _solveAll(self, display=None, time_limit=None, solution_limit=None, **kwargs):
         """ Special 'async' function because mzn.solutions() is async """
+
+        # ensure all vars are known to solver
+        self.solver_vars(list(self.user_vars))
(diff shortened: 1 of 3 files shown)


tias:
note that it does not 'post' variables twice... solver_var() returns it, and creates it if it did not exist yet; so it will never be created twice.

tias:
when the tests finish running you can merge it, thx for extra testcase too

Dimosts:
2 tests are still failing, regarding not supported int vars in pysat

tias:
I see, pysat and pysdd have to be excluded from these tests then... @ThomSerg could you fix that up so we can merge?

PR #616, opened 2025-04-08
Parameters to exact

### LINKED ISSUE #481: Parameters to Exact
When using the Exact solver, it seems some things go wrong when translating the arguments.
I'm not sure whether this is due to CPMpy or the Python interface of Exact itself, maybe @JoD can help here.

I have the following unsat Sudoku model:
```python
import cpmpy as cp
import numpy as np
e = 0
given = np.array([
    [e, e, e,  2, e, 5,  e, e, e],
    [e, 9, e,  e, e, e,  7, 3, e],
    [e, e, 2,  e, e, 9,  e, 6, e],

    [2, e, 1,  e, e, e,  4, e, 9],
    [e, e, e,  e, 7, e,  e, e, e],
    [6, e, 9,  e, e, e,  e, e, 1],

    [e, 8, e,  4, e, e,  1, e, e],
    [e, 6, 3,  e, e, e,  e, 8, e],
    [e, e, e,  6, e, 8,  e, e, e]])

# make the model
cells = cp.intvar(1,9,shape=(9,9),name="cells")

m = cp.Model()
m += [cp.AllDifferent(row) for row in cells]
m += [cp.AllDifferent(col) for col in cells.T]

for i in range(0,9,3):
    for j in range(0,9,3):
        m += cp.AllDifferent(cells[i:i+3, j:j+3])
m += cells[given != 0] == given[given != 0]

print(m.solve(solver="exact"))
>>False
```
Now, I would like a proof-log as to why this is UNSAT.
I should be able to do:
```python

s = cp.SolverLookup.get("exact", m)
s.solve(verbosity=10, **{"proof-log": "path/to/proof"})
```
But the output printed by Exact is:
```
c #vars 2511 #constraints 1337
c PRESOLVE
```

Also, when trying other arguments like `print-opb` no extra output is given.

IgnaceBleukx:
Add small warning on how to set parameters to exact.
Also added a test that checks if the warning is raised, and if the options are correctly parsed.

Closes #481

### DIFF
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -169,8 +169,8 @@ class CPM_exact(SolverInterface):
         
         # set additional keyword arguments
         if(len(kwargs.items())>0):
-            wrn_txt = "Exact only supports options at initialization. Ignoring additional options {kwargs.items()}\n"
-            wrn_txt += "Use cp.SolverLookup.lookup('exact')(**{parameter-with-hyphen: 42}) to set Exact parameters"
+            wrn_txt = f"Exact only supports options at initialization. Ignoring additional options {kwargs.items()}\n"
+            wrn_txt += "Use cp.SolverLookup.lookup('exact', **{parameter-with-hyphen: 42}) to set Exact parameters"
             warnings.warn(wrn_txt)
 
         # ensure all vars are known to solver
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -166,7 +166,7 @@ class CPM_exact(SolverInterface):
                 - False     if no solution is found
         """
         from exact import Exact as xct
-        
+ 
         # set additional keyword arguments
         if(len(kwargs.items())>0):
             wrn_txt = f"Exact only supports options at initialization. Ignoring additional options {kwargs.items()}\n"
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -112,7 +112,6 @@ class CPM_exact(SolverInterface):
         assert subsolver is None, "Exact does not allow subsolvers."
 
         from exact import Exact as xct
-
         # initialise the native solver object
         options = list(kwargs.items()) # options is a list of string-pairs, e.g. [("verbosity","1")]
         options = [(opt[0], str(opt[1])) for opt in options] # Ensure values are also strings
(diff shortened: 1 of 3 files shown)


Dimosts:
Looks good to me.

tests/test_solvers.py
```diff
@@ -477,6 +478,29 @@ def _trixor_callback():
...
+    def test_parameters_to_exact(self):
+    
+        # php with 5 pigeons, 4 holes
+        p,h = 40,39
+        x = cp.boolvar(shape=(p,h))
+        m = cp.Model(x.sum(axis=1) >= 1, x.sum(axis=0) <= 1)
+
+        # this should raise a warning
+        with self.assertWarns(UserWarning):
+            self.assertFalse(m.solve(solver="exact", verbosity=10))
```
Dimosts, line 492: It feels like for that case, when we select exact through model.solve(), we can actually handle it. The warning should just come when the solver object is already created?

through model.solve we use solverlookup, so we could give the parameters there if exact is used as solver.

Dimosts (changes requested):

tests/test_solvers.py
tias: then we have to hardcode 'exact' into model.solve(); where we would have to have it call SolverLookup.lookup (to get the class) instead of the .get()...
its possible I think (elif solver == "exact": s = SolverLookup.lookup(solver)(**kwargs); kwargs = {}) but maybe a bit messy...

but possible.


to make the 'lookup' itself less messy, we could also have 'SolverLookup.get' accept keyword arguments, that are given to the solver.. then it would be s.SolverLookup.get('exact', verbosity=10) instead of cp.SolverLookup.lookup('exact')(verbosity=10)? One could imagine other solvers also having at-constructor-time arguments

IgnaceBleukx: Hmm, yeah I would be in favor of indeed adding `**init_kwargs` to `SolverLookup.get`.
I think it is not unreasonable to ask from a user to use solverlookup instead of model. 
Using extra solver-arguments already hints at more advanced usage of the solver, compared to just solving

IgnaceBleukx:
Added the **init_kwargs to `SolverLookup.get`, should be fine to merge after tests complete

cpmpy/solvers/exact.py
```diff
@@ -167,10 +166,12 @@ def solve(self, time_limit=None, assumptions=None, **kwargs):
                 - False     if no solution is found
         """
         from exact import Exact as xct
-
+
```
tias, line 169: nitpick: this has unneeded spaces

cpmpy/solvers/exact.py
```diff
@@ -167,10 +166,12 @@ def solve(self, time_limit=None, assumptions=None, **kwargs):
                 - False     if no solution is found
         """
         from exact import Exact as xct
-
+        
         # set additional keyword arguments
         if(len(kwargs.items())>0):
-            warnings.warn(f"Exact only supports options at initialization: {kwargs.items()}")
+            wrn_txt = f"Exact only supports options at initialization. Ignoring additional options {kwargs.items()}\n"
+            wrn_txt += "Use cp.SolverLookup.lookup('exact', **{parameter-with-hyphen: 42}) to set Exact parameters"
```
tias, line 173: seems unusual to do a string addition computation for the printing of a warning...

don't we have a more standard way of doing a multi-line warning? if not OK, but it looks unusual to me

tias (approved):
yes much much cleaner this way, thx

two nitpicks below, non-blocking

cpmpy/solvers/exact.py
IgnaceBleukx: It is because of the curly brackets in the second line of tht warning text that I had to split it; otherwise I cannot format the first part to kwargs.items()

Issue #344, opened 2023-06-05
NullShapeError not a CPMpy exception


363734:
Is it wanted that the NullShapeError (variables.py) inherits from Exception and is located within the variables.py file?
Instead of inheriting from CPMpyException and being in the exceptions.py, with other exceptions

IgnaceBleukx:
I'm pretty sure this stems from before we had our own exceptions.
It would indeed be best to refactor the class to subclass our exceptions and put it in the correct file as well.

363734:
Ok. I also found the use of some generic exceptions and NotImplementedExceptions within the gurobi.py file. I'm assuming we can also refactor them to be CPMpy errors

IgnaceBleukx:
Hmm, I'm not sure we want to replace all exceptions with CPMpy exceptions, if there are python-native exceptions which cover the usecase, I think its fine to use those.
We also raise `ValueError`s throughout the codebase, which are non-CPMpy exceptions either.
However, we should definitely make this more consistent where to raise what exception.
Took a brief look at the gurobi file, we mix `NotImplementedError` and just `Exception` in the `__add__` function, so it would be good if we streamline this somewhat.

363734:
What would help would be to add documentation to the exceptions file, and put the 'rules' on what exceptions to use there within the spec.

hbierlee:
Closed in offline discussion: we indeed throw non-cpmpy errors, which is fine.

However I agree the exception documentation is lacking:

- No doc-strings to say when we raise which exception. I think this is something I can add for this release.
- Move all exceptions to exceptions.py
- Best left as a new issue: I believe we should better categorize the exceptions, like: `SolverException(solver, Exception)`? Now in many cases we always write : `PySAT: error in ...`. I believe there should

PR #622, opened 2025-04-10
Improve exceptions (documentations, some refactor)

hbierlee:
This adds some documentation which was missing as pointed out in #344. I tried to understand the exceptions from the context, let me know if I misinterpreted something.

We could merge this. One refactor I'd propose is to simplify by removing solver specific errors.

- Rename `NotSupportedError` to `SolverLimitation(solver)`: whenever the solver `solver` cannot support something in the model (and is not likely to: e.g. integer out of bounds for Minizinc)
- `SolverConfigurationException(solver)`: whenever `solver` is mis-configured (dependency not installed, minizinc not in PAH, license missing, etc..), the user can usually fix this
- `ResultVerificationException` proof logging failed (eg GCSVerificationException)

If it's not worth the API change of exception, we can merge/close this now.

### DIFF
--- a/cpmpy/exceptions.py
+++ b/cpmpy/exceptions.py
@@ -9,31 +9,37 @@ class CPMpyException(Exception):
 
 
 class MinizincPathException(CPMpyException):
+    '''Raised when `minizinc` is not added to PATH'''
     pass
 
 class MinizincNameException(CPMpyException):
+    '''Raised when a variable is a keyword or otherwise violates Minizinc naming rules'''
     pass
 
 class MinizincBoundsException(CPMpyException):
-    pass
-
-class ChocoTypeException(CPMpyException):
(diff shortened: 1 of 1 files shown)


tias:
thanks for checking and documenting that!

It does look like we have been introducing maybe a bit too many exceptions, e.g. GCSVerificationException should just be a VerificationException that any solver could use... Also BoundsException could be a generic one for both minizinc and choco.

But we shouldn't do API changes just for the sake of it, and I find SolverLimitation(solver) going even a bit too far, its OK to just have the name in the str of the error.

So let's leave it at the doc change, and if we find ourselves adding duplicate solver-specific constraints in the future we can create a shared exception then?

tias (approved):
ready as is

PR #623, opened 2025-04-10
Minizinc installed subsolver

ThomSerg:
Calling `cp.SolverLookup.solvernames()` returns all solver(+subsolver) names currently installed on the system, except for minizinc.
With minizinc, all subsolvers get returned, without a check if they're available. This is an issue when someone wants to easily test a model against all installed solvers:

```python
for solvername in cp.SolverLookup.solvernames():
    m.solve(solver=solvername)
    print(solvername , m.status())
```

This will fail on all minizinc subsolvers which aren't installed. 

The minizinc API does not expose the installation status. The current implementation for getting solver names in CPMpy makes use of the following:
```python
minizinc.default_driver.available_solvers()
``` 
It returns a dictionary with for each 'tag' all possible 'drivers'. It does not check if the executables for each driver are available. Additionally, it includes tags which don't represent solvers ('globalizer', 'findmus', ...) ,  tags which represent a 'family' of solvers (like 'cp', 'ilp', ...) and duplicate tags (like 'coinbc', 'cbc', 'org.minizinc.mip.coin-bc')

This alternative implementation proposes 
1) an new way of getting all minizinc solver names without some of the above issues (by directly accessing the minizinc API in the same way that the minizinc python library does
2) a (optional) filtering of minizinc solvers on those that are installed. 

The detection of installation is currently done by attempting to solve a small model with that solver and seeing whether it fails. This has a performance overhead (2.48s vs 0.19s before on my machine). An alternative would be to check if the solver's minizinc configuration file (or the executable) is available on the machine. But then we would have to check [all possible locations](https://docs.minizinc.dev/en/stable/command_line.html#configuration-files), or parse the last output line of a `minizinc --solvers` call (no guarantee that the format of this output will stay the same).

So there currently doesn't seem to be a great solution to this problem.

### DIFF
--- a/cpmpy/solvers/minizinc.py
+++ b/cpmpy/solvers/minizinc.py
@@ -45,7 +45,10 @@ import re
 import warnings
 import sys
 import os
+import json
+import platform
 from datetime import timedelta  # for mzn's timeout
+from pathlib import Path
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


tias:
I think we can expect solvernames() to always only return those installed (also for minizinc).

I also think we should always filter out non-solver minizinc ones (bit weird they do return them).

As to how to best find out which solvers are actually installed for minizinc... I think this is a desirable option for any python-minizinc user, would you mind asking it on their issue tracker? e.g. that we would like a way to get only the list of truly installed solvers? (you can show them your code sample, which will make them realize a better way should be possible ; )

PR #625, opened 2025-04-10
Numpy ufunc

### LINKED ISSUE #124: Running 0.9.7 model in 0.9.8: AttributeError: 'numpy.bool_' object has no attribute 'is_bool'  (reification related?)
The following old v0.9.7 model throws `AttributeError: 'numpy.bool_' object has no attribute 'is_bool')` when running in v9.7.8.
(It's my model http://hakank.org/cpmpy/picking_teams.py )
```
import sys,math,random
import numpy as np
from cpmpy import *
from cpmpy.solvers import *

def picking_teams(s):
    n = len(s)
    n2 = math.ceil(n / 2)
        
    s_sum = sum(s)
    count = [n2,n2]

    print("s_sum:",s_sum)
    print("s_sum % 2:", s_sum % 2)

    x = intvar(1,2,shape=n,name="x")

    # the difference in strength between the teams
    # to be minimized
    d = intvar(0,s_sum,name="d")

    model = Model([d == abs(sum([s[i]*(x[i] == 1) for i in range(n)]) -
                            sum([s[i]*(x[i] == 2) for i in range(n)])
                            ),
                   ],
                 minimize=d
                  )

    # same size of team
    model += [sum([x[i] == 1  for i in range(n)]) == n2,
              sum([x[i] == 2  for i in range(n)]) == n2]

    # symmetry breaking: assign first person to team 1
    model += (x[0] == 1)

    # divisibility of the sum
    model += ((s_sum % 2) == (d % 2))

    num_solutions = 0
    ss = CPM_ortools(model)

    if ss.solve() is not False:
        num_solutions += 1
        xval = x.value()
        print("x:", xval)        
        print("d:", d.value())
        print("Team 1:", [i for i in range(n) if xval[i] == 1])
        print("Team 2:", [i for i in range(n) if xval[i] == 2])
        print(flush=True)
    print()
    print("number of solutions:", num_solutions)
    print("Num conflicts:", ss.ort_solver.NumConflicts())
    print("NumBranches:", ss.ort_solver.NumBranches())
    print("WallTime:", ss.ort_solver.WallTime())
    print()

s = [35, 52, 17, 26, 90, 55, 57, 54, 41, 9, 75, 24, 17, 23, 62, 74, 100, 67, 40, 48, 7, 6, 44, 19, 16, 14, 2, 66, 70, 2, 43, 45, 76, 53, 90, 12, 88, 96, 30, 30, 36, 93, 74, 1, 52, 45, 38, 7, 24, 96, 17, 21, 12, 12, 23, 90, 77, 64, 37, 79, 67, 62, 24, 11, 74, 82, 51, 17, 72, 18, 37, 94, 43, 44, 32, 86, 94, 33, 97, 27, 38, 38, 29, 92, 35, 82, 22, 66, 80, 8, 62, 72, 25, 13, 94, 42, 51, 31, 69, 66]
picking_teams(s)

# Randomize s
n = 1000
s = [random.randint(1,100) for _ in range(n)]
picking_teams(s)
```

The output running the model:
```
s_sum: 4753
s_sum % 2: 1
  File "picking_teams.py", line 96, in <module>
    picking_teams(s)
  File "picking_teams.py", line 73, in picking_teams
    ss = CPM_ortools(model)
  File "/usr/local/lib/python3.7/site-packages/cpmpy/solvers/ortools.py", line 91, in __init__
    super().__init__(name="ortools", cpm_model=cpm_model)
  File "/usr/local/lib/python3.7/site-packages/cpmpy/solvers/solver_interface.py", line 83, in __init__
    self += cpm_model.constraints
  File "/usr/local/lib/python3.7/site-packages/cpmpy/solvers/ortools.py", line 310, in __add__
    cpm_cons = only_bv_implies(flatten_constraint(cpm_con))
  File "/usr/local/lib/python3.7/site-packages/cpmpy/transformations/flatten_model.py", line 136, in flatten_constraint
    flatcons = [flatten_constraint(e) for e in expr]
  File "/usr/local/lib/python3.7/site-packages/cpmpy/transformations/flatten_model.py", line 136, in <listcomp>
    flatcons = [flatten_constraint(e) for e in expr]
  File "/usr/local/lib/python3.7/site-packages/cpmpy/transformations/flatten_model.py", line 143, in flatten_constraint
    assert expr.is_bool(), f"Boolean expressions only in flatten_constraint, `{expr}` not allowed."
AttributeError: 'numpy.bool_' object has no attribute 'is_bool'
```

Here are some other models that throws the same error:

* http://hakank.org/cpmpy/bus_scheduling_csplib.py
  (data files in http://hakank.org/cpmpy/bus_scheduling_csplib/)

* http://hakank.org/cpmpy/linear_combinations.py

* http://hakank.org/cpmpy/partition_into_subsets_of_equal_values.py

For all these models, the error involves this line
```
...
cpm_cons = only_bv_implies(flatten_constraint(cpm_con))
...
```

so it seems that there is some issue with the reification. Has the syntax/logic for reification been changed?

Dimosts:
Most issues with ufuncs were tackled by other PRs.

However, issue #124 revealed a problem when a user tries to create an expression using an np.array or np.int in the left hand side.

To do this, we have to subclass NDArrayOperatorsMixin in CPMpy expressions and implement our intended behaviour through ufuncs.

This is implemented in this PR, adding also some tests based on issue #124.


cpmpy/expressions/core.py
```diff
@@ -110,6 +111,67 @@ class Expression(object):
...
+            np.logical_and: "and",
+            np.logical_or: "or",
+            np.logical_xor: "xor"
+        }
+        
+        # Handle binary operations
+        if len(inputs) == 2 and ufunc in op_map:
+            x, y = inputs
+            if isinstance(x, Expression):
+                return getattr(x, f"__r{op_map[ufunc]}__")(y)
```
tias, line 152: I think understanding and supporting ufunc's properly in CPMpy is good.

But... this is fishy; using the opmap str and just adding an r before it?

why not just swap the arguments and call the existing op?

note that CPMpy already implements all operators and reverse operators on https://docs.python.org/3.11/reference/datamodel.html#emulating-numeric-types and below;
and there exists no --req-- --rne-- etc... so we should NOT create them (lower in this PR), and we need a more appropriate way to do what you want to do here!

tias (changes requested):

cpmpy/expressions/core.py
Dimosts: Well, if it is other > self, just swapping the arguments and having self > other is not the same expression, we need the right-hand side version of it. And instead of doing a custom mapping inside the function, I preferred to actually have the rhs version separately. But we could have a custom mapping for when swapping the arguments indeed.

For using the strings, the issue is that the np functions are different from ours. So we need to map them either to a str that we will use later, or a function that we are going to use directly after.

Dimosts:
Changed the style for mapping np ufuncs to cpmpy operators.

Still not a big fan of it, but it shows what needs to be done to make it work.

IgnaceBleukx:
Does this also fix the following or is it unrelated?
On current master:
```python
import cpmpy as cp
arr = cp.intvar(0,10,shape=3, name="arr")
x = cp.intvar(0,10,name="x")

arr < x # works fine
>> [(arr[0]) < (x) (arr[1]) < (x) (arr[2]) < (x)]
x  < arr # does not broadcast; fails downstream
>> x < ([arr[0] arr[1] arr[2]])
```

Dimosts:
This is indeed unrelated, as this does not call np ufunc. This has to do with how the expression operators handle it.

`arr < x` works fine because this calls the NDVarArray operators.

tias:
indeed, the NDVarArray functions check for and implement the vectorized version, while _IntVarImpl inherits from Expression which simply creates the comparison over itself and the array; we could check in Expression whether the other is an NDVarArray, and then call its reverse...

It could also be that ufuncs are the right way to solve all of this: that they are always elementwise operations, so should handle all operators/comparisons on all CPMpy objects... requires a deeper study

cpmpy/expressions/core.py
```diff
@@ -110,6 +111,64 @@ class Expression(object):
...
+            np.logical_and: [Expression.__and__, Expression.__rand__],
+            np.logical_or: [Expression.__or__, Expression.__ror__],
+            np.logical_xor: [Expression.__xor__, Expression.__rxor__]
+        }
+        
+        # Handle binary operations
+        if len(inputs) == 2 and ufunc in op_map:
+            x, y = inputs
+
+            if isinstance(x, Expression):
```
tias, line 153: wait, Re: Ignace' comment... that is this case (in case we would translate all expression operators to ufuncs, which is what numpy recommends it seems), then we could check here if broadcasting needs to happen?
that would be a major overhaul, but it could lead to consistent broadcasting instead of only NDVarArray being broadcast-aware...

tias:
Coincidentally ended up on CVXpy, which was one of the inspirations of CPMpy.

They (I think its new) actually are numpy compatible without overloading the ndarray, exactly through a ufunc:
https://github.com/cvxpy/cvxpy/blob/master/cvxpy/expressions/expression.py

They are quite smart about it, because in contrast to us where a decision variable is an ndarray of individual decision variable objects; for them it is an object representing ONE virtual n-dimensional object, and indexing into it creates an index object, not subarrays of the individual decision variables. This is much closer to how numpy does it too; and if you write vectorized code it will also be a much more efficient representation...
https://github.com/cvxpy/cvxpy/blob/master/cvxpy/atoms/affine/index.py

eg, for us, we fill in each decisvarimpl's value, and x.value() loops over that. For their approach, we would fill in the x._value with an nd-array of ints, and x.value() would return it straight (and subarrays would directly call on the x._value's ndarray, pure numpy)

Still some things unclear to me, but this looks to be the right way... both from numpy compatibility and from efficiency...

tias:
Superseded by #1062

Issue #613, opened 2025-04-08
Bug in flattening of Increasing-type constraints


peerjosse:
The following program
```python
import cpmpy as cp

m = cp.Model()

a = cp.boolvar()
b = cp.boolvar()
m += cp.IncreasingStrict(1,  (~a).implies(~b) <= 1)

m.solve()
```
gives the error `AttributeError: 'bool' object has no attribute 'has_subexpr'` in `flatten_constraint` from `transformations\flatten_model.py`

Dimosts:
The actual issue is in simplify_boolean().

cp.IncreasingStrict(1,  (~a).implies(~b) <= 1) is decomposed to [(~BV0) -> (~BV1) <= 1 > 1] and when simplify_boolean processes it, it returns False. 

Specifically, in the recursive call in simplify_boolean, (~BV0) -> (~BV1) <= 1 is converted to 1 (as it has num_context=True). Then, with both sides not being an expression, it reaches `newlist.append(eval_comparison(name, lhs, rhs))`, which returns False, instead of BoolVal(False).

PR #626, opened 2025-04-11
Simplify bool fix

Dimosts:
fix of issue #613 

Specifically, simplify_bool did not handle the case where a comparison ended up being True or False in eval_comparison.

### DIFF
--- a/cpmpy/transformations/normalize.py
+++ b/cpmpy/transformations/normalize.py
@@ -221,7 +221,11 @@ def simplify_boolean(lst_of_expr, num_context=False):
                 elif rhs > 1:
                     newlist.append(BoolVal(name in {"!=", "<", "<="})) # all other operators evaluate to False
             else:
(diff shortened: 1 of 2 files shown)


cpmpy/transformations/normalize.py
```diff
@@ -221,7 +221,11 @@ def simplify_boolean(lst_of_expr, num_context=False):
                 elif rhs > 1:
                     newlist.append(BoolVal(name in {"!=", "<", "<="})) # all other operators evaluate to False
             else:
-                newlist.append(eval_comparison(name, lhs, rhs))
+                res = eval_comparison(name, lhs, rhs)
+                if is_bool(res):
```
tias, line 225: this means 'Boolean constant' right? maybe add  # a Boolean constant comment?

Dimosts: yup exactly, added comment

PR #628, opened 2025-04-11
Solverlookup status

ThomSerg:
Adds a `.status()` and `.print_status()` to SolverLookup. The first one returns a list of solver status reports (tuples of `(solver_name, installed, version)`), whilst the second will print a tabulated status report on CPMpy's supported solvers to stdout.

```python
import cpmpy as cp
cp.SolverLookup.status()
# [('ortools', True, '9.12.4544'), ('pysat', True, '1.8.dev16'), ...]
```
---
```python
import cpmpy as cp
cp.SolverLookup.print_status()
```

```console
Solver               Installed  Version        
--------------------------------------------------
ortools              Yes        9.12.4544
z3                   Yes        4.14.1.0       
minizinc             Yes        0.10.0
gcs                  No         -
gurobi               No         -
pysat                Yes        1.8.dev16      
pysdd                No         -
exact                Yes        2.1.0
choco                No         -
cpo                  No         -
```
---

Would like to add subsolver status reporting in the future (code is in place but commented out) once #623 is resolved.

For creating this table, added a `.version()` classmethod to all solvers.


tias:
I think that's awesome.

Do we need a status() that returns it programatically? 
Maybe status() can just be print_status(), and there is no way to get it programmatically? (other than doing the loop over the installed solvers yourself?)

cpmpy/solvers/cpo.py
```diff
@@ -44,7 +44,9 @@
 
 import shutil
 import time
+from typing import Optional
 import warnings
+import pkg_resources
```
tias, line 49: why don't TEMPLATE and choco require the pkg_resource import, and this one does?

cpmpy/solvers/minizinc.py
```diff
@@ -126,26 +129,91 @@ def outdated():
         else:
             # outdated
             return True
+        
+    @staticmethod
+    def _solver_names_and_version():
```
tias: if the current function is called `solvernames` then `_solver_names_and_version()` is not the most obvious name...

`solvernames_with_version()` could be OK, a public function.

you can even wonder if it should not be merged into solvernames, that if the argument `with_version=True` that it returns a tuple of lists, instead of just a list?

cpmpy/solvers/solver_interface.py
```diff
@@ -49,6 +50,13 @@ def supported():
                 [bool]: Solver support by current system setup.
         """
         return False
+    
+    @classmethod
+    def version(cls) -> Optional[str]:
+        """
+        Returns the installed version of the solver's Python API.
+        """
+        return None
```
tias: I think we should raise notimplementederror as below... in case a solver forgets to implement it (e.g. the one that might be merged just after this PR ; )

cpmpy/solvers/utils.py
```diff
@@ -161,6 +161,71 @@ def lookup(cls, name=None):
                 # found the right solver
                 return CPM_slv
         raise ValueError(f"Unknown solver '{name}', chose from {cls.solvernames()}")
+    
+
+    @classmethod
+    def status(cls):
```
tias: should this be `installed()`? or even `version()`?

version would be in line with the function you added to each interface... which then returns an aggregate of the versions?

cpmpy/solvers/utils.py
```diff
@@ -161,6 +161,71 @@ def lookup(cls, name=None):
...
+                        "name": basename + ":" + subn, 
+                        "status": is_installed, 
+                        "version": CPM_slv.solverversion(subn) if installed else None,
+                    }
+                    result.append(subsolver_status)  # Append subsolver status
+        return result
+
+
+    @classmethod
+    def print_status(cls):
```
tias: print_version()?

docs/modeling.md
```diff
@@ -510,27 +510,80 @@ If that is not sufficient or you want to debug an unexpected (non)solution, have
...
+
+```python
+import cpmpy as cp
+cp.SolverLookup.base_solvers() # returns a list of tuples, 
+                               # where each tuple is a pair of (<solver name>, <cpmpy solver class>)
+# [('ortools', <class 'cpmpy.solvers.ortools.CPM_ortools'>), ('z3', <class 'cpmpy.solvers.z3.CPM_z3'>), ('minizinc', <class 'cpmpy.solvers.minizinc.CPM_minizinc'>), ('gcs', <class 'cpmpy.solvers.gcs.CPM_gcs'>), ('gurobi', <class 'cpmpy.solvers.gurobi.CPM_gurobi'>), ('pysat', <class 'cpmpy.solvers.pysat.CPM_pysat'>), ('pysdd', <class 'cpmpy.solvers.pysdd.CPM_pysdd'>), ('exact', <class 'cpmpy.solvers.exact.CPM_exact'>), ('choco', <class 'cpmpy.solvers.choco.CPM_choco'>), ('cpo', <class 'cpmpy.solvers.cpo.CPM_cpo'>)]
+```
+
+
+Additionally, one can get the status of each of the solvers:
```
tias: its not really the status... we use status to refer to the solver status after a solve.

I think version would fit best "one can get the version of each of the solvers:"

cpmpy/solvers/cpo.py
ThomSerg: They both do (in their `supported()` check)

cpmpy/solvers/utils.py
ThomSerg: I agree, my preference goes to `version()` to match with the solver interfaces.

cpmpy/solvers/cpo.py
```diff
@@ -108,6 +110,16 @@ def license_ok():
...
             except:
                 return False
+            
+    @staticmethod
+    def version() -> Optional[str]:
+        """
+        Returns the installed version of the solver's Python API.
+        """
+        try:
+            return pkg_resources.get_distribution('docplex').version
```
IgnaceBleukx: This returns the version of the `docplex` package, but not actually the version of the solver itself.
Can we make this return a string like "{docplex_version}/{cpo_version)"?
You can query the solver version like this:
https://ibmdecisionoptimization.github.io/docplex-doc/cp/docplex.cp.solver.solver.py.html#docplex.cp.solver.solver.CpoSolver.get_solver_version

cpmpy/solvers/minizinc.py
```diff
@@ -126,26 +129,104 @@ def outdated():
...
+        except ValueError:
+            raise ValueError(f"Subsolver '{subsolver}' not found in the list of available solvers.")
+
+    @staticmethod
+    def version() -> Optional[str]:
+        """
+        Returns the installed version of the solver's Python API.
+        """
+        try:
+            return pkg_resources.get_distribution('minizinc').version
```
IgnaceBleukx: Similar here to the cpo-case, ideally we also get the actual MiniZinc version, apart from just the Python API version

cpmpy/solvers/pysat.py
```diff
@@ -116,19 +120,49 @@ def supported():
...
-                if name not in attr:
-                    name = attr[-1]
-                names.append(name)
-        return names
+        if CPM_pysat.supported():
+            from pysat.solvers import SolverNames
+            names = []
+            for name, attr in vars(SolverNames).items():
+                # issue with cryptosat, so we don't include it in our https://github.com/msoos/cryptominisat/issues/765
+                if not name.startswith('__') and isinstance(attr, tuple) and not name == 'cryptosat':
```
IgnaceBleukx: minor: `and name != 'cryptosat`?

cpmpy/solvers/cpo.py
ThomSerg: Done. Unfortunately one has to create both a cpo model and a solver instances just to get access to these fields.

PR #630, opened 2025-04-11
Discussing whether we support constants for Element decomposition

hbierlee:
So the issue is the Element's decomposition does not work for integer constants as index, only integer variables. So the following test failed for z3, and after my fix still for or-tools:

```python
# constant look-up
self.assertTrue(cp.Model([cp.boolvar() == cp.Element([0], 0)]).solve(solver=s))
# constant out-of-bounds look-up
self.assertFalse(cp.Model([cp.boolvar() == cp.Element([0], 1)]).solve(solver=s))
```

However, if the user uses the overloaded Python array indexing (e.g. `[0][0]`), then we don't get this issue in the first place.

So the question is a) should the decompositions of global constraints support constants for (some) arguments, and b) do we support the use of cp.Element, or only of array indexing (e.g. should the model just have [0][0], which would just work)

Depending on these questions, the PR can just be closed, otherwise I need to still resolve an issue for OR-tools.

### DIFF
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -286,14 +286,15 @@ class Element(GlobalFunction):
 
         """
         arr, idx = self.args
-        # for every  `i` in the intersection of array indices and the bounds of `idx`..
+        # Find where the array indices and the bounds of `idx` intersect
         lb, ub = get_bounds(idx)
         new_lb, new_ub = max(lb, 0), min(ub, len(arr) - 1)
-        return [  # ..we post `(idx = i) -> idx=i -> arr[i] <CMP_OP> cpm_rhs`.
-            implies(idx == i, eval_comparison(cpm_op, arr[i], cpm_rhs)) for i in range(new_lb, new_ub + 1)
-        ] + [  # also enforce the new bounds 
-            idx >= new_lb, idx <= new_ub
-        ], []  # no auxiliary variables
+        cons=[]
+        # For every `i` in that intersection, post `(idx = i) -> idx=i -> arr[i] <CMP_OP> cpm_rhs`.
+        for i in range(new_lb, new_ub+1):
+            cons.append(implies(idx == i, eval_comparison(cpm_op, arr[i], cpm_rhs)))
+        cons+=[idx >= new_lb, idx <= new_ub]  # also enforce the new bounds 
+        return cons, []  # no auxiliary variables
 
     def __repr__(self):
         return "{}[{}]".format(self.args[0], self.args[1])
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -70,7 +70,7 @@ import cpmpy as cp
 from ..exceptions import CPMpyException, IncompleteFunctionError, TypeError
 from .core import Expression, Operator, Comparison
 from .variables import boolvar, intvar, cpm_array
-from .utils import flatlist, argval, is_num, is_int, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals, is_bool, is_true_cst, is_false_cst
+from .utils import flatlist, argval, is_num, is_int, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals, is_bool, is_true_cst, is_false_cst, get_bounds
 
 
 class GlobalFunction(Expression):
@@ -290,8 +290,8 @@ class Element(GlobalFunction):
         cons = []
         # the decomposition posts `idx=i -> arr[i] <CMP_OP> cpm_rhs`
         # for every  `i` in the intersection of array indices and the bounds of `idx`
-        lb, ub = idx.get_bounds()
         new_lb, new_ub = max(lb, 0), min(ub, len(arr))
+        lb, ub = get_bounds(idx)
         for i in range(new_lb, new_ub):
             cond = idx == i
             if is_true_cst(cond):
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -68,9 +68,9 @@ import numpy as np
 import cpmpy as cp
 
 from ..exceptions import CPMpyException, IncompleteFunctionError, TypeError
-from .core import Expression, Operator, Comparison
+from .core import Expression, Operator
 from .variables import boolvar, intvar, cpm_array
-from .utils import flatlist, argval, is_num, is_int, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals, is_bool, is_true_cst, is_false_cst, get_bounds, implies
+from .utils import flatlist, argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals, get_bounds, implies
 
 
 class GlobalFunction(Expression):
@@ -237,6 +237,7 @@ def element(arg_list):
     warnings.warn("Deprecated, use Element(arr,idx) instead, will be removed in stable version", DeprecationWarning)
     assert (len(arg_list) == 2), "Element expression takes 2 arguments: Arr, Idx"
     return Element(arg_list[0], arg_list[1])
+
 class Element(GlobalFunction):
     """
         The 'Element' global constraint enforces that the result equals Arr[Idx]
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -70,7 +70,7 @@ import cpmpy as cp
 from ..exceptions import CPMpyException, IncompleteFunctionError, TypeError
 from .core import Expression, Operator, Comparison
 from .variables import boolvar, intvar, cpm_array
-from .utils import flatlist, argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals
+from .utils import flatlist, argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals, is_bool
 
 
 class GlobalFunction(Expression):
@@ -285,13 +285,13 @@ class Element(GlobalFunction):
 
         """
         arr, idx = self.args
-        if is_num(idx): # index is constant, we only enforce one constraint
-            try:
-                return [eval_comparison(cpm_op, arr[idx], cpm_rhs)], []
-            except IndexError: # out-of-bounds is undefined (TODO check how to handle partial fixed result)
-                return [], [False]
-        return [(idx == i).implies(eval_comparison(cpm_op, arr[i], cpm_rhs)) for i in range(len(arr))] + \
-               [idx >= 0, idx < len(arr)], []
+        cons = []
+        for i,var in enumerate(arr):
+            cond = idx == i
+            if is_bool(cond) and not isinstance(cond, cp.BoolVal): # constant
+                cond = cp.BoolVal(cond) # convert True/np.bool_ into BoolVal to enable .implies(..)
+            cons += [cond.implies(eval_comparison(cpm_op, arr[i], cpm_rhs))]
+        return cons + [idx >= 0, idx < len(arr)], []
 
     def __repr__(self):
         return "{}[{}]".format(self.args[0], self.args[1])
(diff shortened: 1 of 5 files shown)


IgnaceBleukx:
I think we should support constants in the array of Element. 
In CPMPy we give the user freedom to use any constant or Expression as a subexpression in a constraint, as long as the types match.
Only constants is of course somewhat silly as it will just result in `BoolVal(False)` or `BoolVal(True)`, but I can certainly imagine a user mixing constants and variables in an Element array.

One way to fix this is in the decomposition is as follows:
```python
def decompose(self):
    arr, idx = self.args
    cons = []
    for i,var in enumerate(arr):
        cond = idx == i
        if is_bool(cond): # constant
            cond = BoolVal(cond)
        cons += [cond.implies(eval_comparison(cpm_op, var, cpm_rhs)]

    return cons + [idx >= 0, idx < len(arr)], []
```

hbierlee:
As discussed offline, we *should* supports constants whereever we support variables.

I had some fix which handles the special case of `idx` being aconstant integer, but I like Ignace more general approach where we handle the case of `cond` turning out to be Boolean. However it has some overlap (we only need one iteration if `idx` is constant), which I think I can easily fix.

hbierlee:
@IgnaceBleukx you can let me know what you think, it should work all cases. So I tightened the loop for the decomposition, which will lead to smaller models in some cases (when the indexes domain is smaller than the array's indexable range). It also checks whether `i==idx` like your approach, with on `True` an early stop, and on `False` skipping the constraint. The question is if these checks have an avoidable transformation performance overhead.

A separate issue was that without the decomposition, the `Element([0],1)` would get errors (like from OR-tools). I fixed this as best I could in the init by turning 1 into a variable, also ensuring the `get_bounds` works later on... not sure if there is a better approach.

cpmpy/expressions/globalfunctions.py
```diff
@@ -285,8 +287,19 @@ def decompose_comparison(self, cpm_op, cpm_rhs):
 
         """
         arr, idx = self.args
-        return [(idx == i).implies(eval_comparison(cpm_op, arr[i], cpm_rhs)) for i in range(len(arr))] + \
-               [idx >= 0, idx < len(arr)], []
+        cons = []
+        # the decomposition posts `idx=i -> arr[i] <CMP_OP> cpm_rhs`
+        # for every  `i` in the intersection of array indices and the bounds of `idx`
+        lb, ub = idx.get_bounds()
```
IgnaceBleukx, line 293: This will go wrong when idx is not an expression (see prev comment). Better to use `utils.get_bounds(idx)` instead.

cpmpy/expressions/globalfunctions.py
```diff
@@ -256,6 +256,8 @@ def __init__(self, arr, idx):
             raise TypeError("index cannot be a boolean expression: {}".format(idx))
         if is_any_list(idx):
             raise TypeError("For using multiple dimensions in the Element constraint, use comma-separated indices")
+        if is_int(idx): # prevents OR-Tools error
```
IgnaceBleukx, line 259: If it is OR-Tools specific, it should be fixed in the OR-Tools interface before posting the constraint. Now the user will see another expression that he created when printing it.

I.e., print(`Element(arr, 42) <= 1337`) will print `arr[IV0] <= 1337`

cpmpy/expressions/globalfunctions.py
```diff
@@ -285,8 +287,19 @@ def decompose_comparison(self, cpm_op, cpm_rhs):
...
+        cons = []
+        # the decomposition posts `idx=i -> arr[i] <CMP_OP> cpm_rhs`
+        # for every  `i` in the intersection of array indices and the bounds of `idx`
+        lb, ub = idx.get_bounds()
+        new_lb, new_ub = max(lb, 0), min(ub, len(arr))
+        for i in range(new_lb, new_ub):
+            cond = idx == i
+            if is_true_cst(cond):
+                return [eval_comparison(cpm_op, arr[i], cpm_rhs)], []
+            elif is_false_cst(cond):
```
IgnaceBleukx, line 299: Not strictly necesarry, as we also check this in the `implies` function of `BoolVal`, but fine to keep.
Would prefer to add a small comment tho

IgnaceBleukx (changes requested):
The idx-replacement with a variable should not happen at construction time, but rather in the OR-Tools interface.

Other than that 2 small comments to fix

cpmpy/expressions/globalfunctions.py
hbierlee: Ah, definitely the function I was looking for! Thanks

cpmpy/expressions/globalfunctions.py
hbierlee: Right, definitely the better approach (I was unable to come up with).

cpmpy/expressions/globalfunctions.py
hbierlee: Ah true, I only checked implies of `BoolVar`, which only checks if the consequent is fixed. But then I'd rather go for simple code depending on the (good) implementation of implies, forgoing the early quit. So I'll remove these. I did have to add an utils.implies function which is also safe for expressions. I think this is very usefull to make more of the globals safe for constants. You can let me know what you think of the final decomposition

cpmpy/expressions/globalfunctions.py
```diff
@@ -285,8 +285,14 @@ def decompose_comparison(self, cpm_op, cpm_rhs):
 
         """
         arr, idx = self.args
-        return [(idx == i).implies(eval_comparison(cpm_op, arr[i], cpm_rhs)) for i in range(len(arr))] + \
-               [idx >= 0, idx < len(arr)], []
+        # for every  `i` in the intersection of array indices and the bounds of `idx`..
+        lb, ub = get_bounds(idx)
+        new_lb, new_ub = max(lb, 0), min(ub, len(arr) - 1)
+        return [  # ..we post `(idx = i) -> idx=i -> arr[i] <CMP_OP> cpm_rhs`.
```
IgnaceBleukx, line 291: I very very very much prefer the previous code with the explicit for-loop here... This is quite hard to parse and it is very unclear what is happening...

cpmpy/expressions/utils.py
```diff
@@ -194,6 +194,17 @@ def get_bounds(expr):
             return int(expr), int(expr)
         return math.floor(expr), math.ceil(expr)
 
+def implies(expr, other):
+    """ like :func:`~cpmpy.expressions.core.Expression.implies`, but also safe to use for non-expressions """
+    if isinstance(expr, cp.expressions.core.Expression):
+        return expr.implies(other)
+    elif arg is True or arg is np.True_:
```
IgnaceBleukx, line 201: This is the same logic as `is_true_cst` which is already in utils.
That being said, I don't think you need this helper function for this PR, and I prefer just checking whether `idx == i` is Boolean is (not) an expression in the Element decomposition.

We could introduce the helper-function nevertheless, but then we should also change the use of `.implies` in other places in the code base. I suppose we have this logic in multiple places at the moment, thereby explicitly checking for `is_bool` and converting to `BoolVal`.
 Better to open an separate issue and/or PR for that as it is out of scope for this one.

cpmpy/expressions/globalfunctions.py
```diff
@@ -70,7 +70,7 @@ def decompose_comparison(self):
 from ..exceptions import CPMpyException, IncompleteFunctionError, TypeError
 from .core import Expression, Operator, Comparison
 from .variables import boolvar, intvar, cpm_array
-from .utils import flatlist, argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals
+from .utils import flatlist, argval, is_num, is_int, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals, is_bool, is_true_cst, is_false_cst, get_bounds, implies
```
IgnaceBleukx, line 73: Not all of these new imports are used

cpmpy/solvers/ortools.py
```diff
@@ -480,9 +480,15 @@ def _post_constraint(self, cpm_expr, reifiable=False):
                 elif lhs.name == 'div':
                     return self.ort_model.AddDivisionEquality(ortrhs, *self.solver_vars(lhs.args))
                 elif lhs.name == 'element':
-                    # arr[idx]==rvar (arr=arg0,idx=arg1), ort: (idx,arr,target)
-                    return self.ort_model.AddElement(self.solver_var(lhs.args[1]),
-                                                     self.solver_vars(lhs.args[0]), ortrhs)
+                    arr, idx = lhs.args
+                    if is_int(idx) and (idx < 0 or idx >= len(arr)):
```
IgnaceBleukx, line 484: So, OR-Tools is fine with constants as long as it is within bounds?

tests/test_globalconstraints.py
```diff
@@ -1325,3 +1325,22 @@ def test_table(self):
...
+                    self.assertFalse(cp.Model([cp.boolvar() == cp.Element([0], 1)]).solve(solver=s))
+                except (NotImplementedError, NotSupportedError):
+                    pass
+
+    def test_element_index_dom_mismatched(self):
+        """
+            Check transform of `[0,1,2][x in -1..1] == y in 1..5`
+            Note the index variable has a lower bound *outside* the indexable range, and an upper bound inside AND lower than the indexable range upper bound
+        """
+        assert str(cp.Element([0,1,2], cp.intvar(-1,1, name="x")).decompose_comparison("==", cp.intvar(1,5, name="y"))) \
```
IgnaceBleukx, line 1345: Please split this into multiple lines for maintainability, and use `self.assertEquals`

cpmpy/expressions/utils.py
IgnaceBleukx: Also, `arg` is undefined here, you mean `expr`

cpmpy/expressions/globalfunctions.py
hbierlee: I can change it back to that loop style (I'm a big fan of generators/maps/functional style, but it's a choice). Note though that this is how it was written before, just with a tighter loop, saver `implies`, and comments!

IgnaceBleukx: Yes, I agree the generators are great, but not if they get so complicated we need to split them and add a comment to explain what they are doing ;)

cpmpy/expressions/utils.py
hbierlee: the reason I don't use `is_true_cst` is because it would check more types than needed. Not sure if that is worth it though for readability.

Definitly good catch on the `arg`.

I thought this function will be helpfull in other casese, but I indeed didn't want to start looking for them in this PR. Please read the previous comment on the decomp itself, this function is necessary for it, if we change it I can put this logic inline and make a separate PR to fix this in some places.

cpmpy/solvers/ortools.py
hbierlee: yes, this is tested

hbierlee: (although I'd be fine with just removing the 2nd condition for simplicity)

cpmpy/expressions/globalfunctions.py
hbierlee: Maybe I could have structured it better (not concatenate them in place, the comments are a bit creatively placed). I will change it to a loop, I can definitely understand why this is harder to read. Although I like this because it just exactly specifies the decomp as you would mathematically: `idx=i -> arr[i] CMP rhs, for all i in range`.

cpmpy/solvers/ortools.py
IgnaceBleukx: Agreed, no strong opinion on keeping or removing it

hbierlee: I'll go with simplicity

cpmpy/expressions/utils.py
hbierlee: See new refactor of the decomp

hbierlee:
You processed all comments, should be about good to merge I think, just we can decide on this new until function in the decomp comment

cpmpy/expressions/globalfunctions.py
IgnaceBleukx: Ok, sounds good indeed!

PR #634, opened 2025-04-13
rework (start of) transformations docs

tias:
to be more useful/accessible on read the docs


ThomSerg (approved):
Good start. I added some fixes to the formatting and changed "normal users" to "typical users"

PR #636, opened 2025-04-15
Fix #617

### LINKED ISSUE #617: Update GitHub CI config to use new extra_installs feature
Perhaps the github CI config can also be updated to re-use the optional dependencies

_Originally posted by @hbierlee in https://github.com/CPMpy/cpmpy/pull/599#discussion_r1973277808_

hbierlee:
Just updating CI to use the optional dependencies, and some minor refactorings/version improvements.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8 pytest pytest-xdist requests pandas
+          pip install pytest pytest-xdist requests pandas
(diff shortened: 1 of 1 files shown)


hbierlee:
This PR can be merged. One remaining issue we can fix (or not), in a new issue or in this PR is:

- Should we remove requirements.txt? I don't see how it is usefull now that we can just use the optional dependencies
  - Same goes for docs/requirements.txt in favour of a docs dependency? I do see the requirements has numpy, which the extra has not, should test if needed

hbierlee:
Happy to leave it in, I think it's good to nudge users towards using `pip install .[...]` such as in the doc (discussed here https://github.com/CPMpy/cpmpy/pull/633#discussion_r2045542521)

PR #638, opened 2025-04-15
Refactor unit test

hbierlee:
Had a lot of trouble understanding this one because the aux var IV numbers go up between test "cases", so I refactored it while working on another issue. I hope we can adopt a little bit more of a parametrized format for the other transform tests as well.

### DIFF
--- a/tests/test_transf_comp.py
+++ b/tests/test_transf_comp.py
@@ -1,29 +1,33 @@
-import unittest
+import pytest
 import numpy as np
 from cpmpy import *
 from cpmpy.transformations.flatten_model import flatten_constraint
 from cpmpy.transformations.comparison import only_numexpr_equality
 from cpmpy.expressions.variables import _IntVarImpl, _BoolVarImpl # to reset counters
 
-class TestTransfComp(unittest.TestCase):
-    def setUp(self):
-        _IntVarImpl.counter = 0
-        _BoolVarImpl.counter = 0
+def _generate_inputs():
+    ivs = intvar(1,9, shape=3, name="x")
+    for case in [
+            (min(ivs) == 3, "[min(x[0],x[1],x[2]) == 3]"),
+            (min(ivs) > 3, "[(min(x[0],x[1],x[2])) == (IV0), IV0 > 3]"),
+            (min(ivs) <= 3, "[(min(x[0],x[1],x[2])) == (IV0), IV0 <= 3]"),
+            (3 != max(ivs), "[(max(x[0],x[1],x[2])) == (IV0), IV0 != 3]"),
+            (3 > max(ivs), "[(max(x[0],x[1],x[2])) == (IV0), IV0 < 3]"),
+            (3 <= max(ivs), "[(max(x[0],x[1],x[2])) == (IV0), IV0 >= 3]"),
+            ]:
+        yield case
 
-    def test_only_numexpr_eq(self):
-        ivs = intvar(1,9, shape=3, name="ivs")
-        
-        cases = [(min(ivs) == 3, "[min(ivs[0],ivs[1],ivs[2]) == 3]"),
-                 (min(ivs) > 3, "[(min(ivs[0],ivs[1],ivs[2])) == (IV0), IV0 > 3]"),
-                 (min(ivs) <= 3, "[(min(ivs[0],ivs[1],ivs[2])) == (IV2), IV2 <= 3]"),
-                 (3 != max(ivs), "[(max(ivs[0],ivs[1],ivs[2])) == (IV4), IV4 != 3]"),
-                 (3 > max(ivs), "[(max(ivs[0],ivs[1],ivs[2])) == (IV6), IV6 < 3]"),
-                 (3 <= max(ivs), "[(max(ivs[0],ivs[1],ivs[2])) == (IV8), IV8 >= 3]"),
-                ]
+@pytest.fixture()
+def setup():
+    _IntVarImpl.counter = 0
+    _BoolVarImpl.counter = 0
+    yield
 
-        # test transformation
-        for (expr, strexpr) in cases:
-            self.assertEqual( str(only_numexpr_equality(flatten_constraint(expr))), strexpr )
-            self.assertTrue(Model(expr).solve())
+class TestTransform:
+    @pytest.mark.parametrize(("constraint", "expected"),_generate_inputs(), ids=str)
+    def test_transforms(self, constraint, expected, setup):
+        flat=only_numexpr_equality(flatten_constraint(constraint))
+        assert str(flat) == expected 
+        assert Model(constraint).solve()
 
 


tests/test_transf_comp.py
```diff
@@ -1,29 +1,33 @@
...
 from cpmpy import *
 from cpmpy.transformations.flatten_model import flatten_constraint
 from cpmpy.transformations.comparison import only_numexpr_equality
 from cpmpy.expressions.variables import _IntVarImpl, _BoolVarImpl # to reset counters
 
-class TestTransfComp(unittest.TestCase):
-    def setUp(self):
-        _IntVarImpl.counter = 0
-        _BoolVarImpl.counter = 0
+def _generate_inputs():
```
tias, line 8: You here separate the logic of the test from the checking of the test. That doesn't feel right... which transformation you apply determines what the input/output mapping is, so that code should stay together?

tests/test_transf_comp.py
```diff
@@ -1,29 +1,33 @@
...
+    _IntVarImpl.counter = 0
+    _BoolVarImpl.counter = 0
+    yield
 
-        # test transformation
-        for (expr, strexpr) in cases:
-            self.assertEqual( str(only_numexpr_equality(flatten_constraint(expr))), strexpr )
-            self.assertTrue(Model(expr).solve())
+class TestTransform:
+    @pytest.mark.parametrize(("constraint", "expected"),_generate_inputs(), ids=str)
```
tias, line 27: looks like black magic to me... has some hidden assumption that the 2nd argument returns tuples of size 2?

and there is some even more magic 'ids=str' argument?

tests/test_transf_comp.py
```diff
@@ -1,29 +1,33 @@
...
+    def test_transforms(self, constraint, expected, setup):
```
tias, line 28: My concern about separating the logic from the test, this is of course just a simple testfile, with just one class whose purpose is to serve one function, which uses a paramtrize and a setup function; that is still overviewable, but if we would start applying this to all transform tests, I fear it would quickly become unoverseeable and unmaintainable...

tias (commented):
its weirder for me that the counters would not go up in a linear execution of the code, but I guess I'm too used to it. I have other questions about maintainability (see below) though

tias:
if its about the variable counters, then I think the most natural 'fix' would be in the original code

```python
for (expr, strexpr) in cases:
    self.setUp()  # reset counters
    assert...
```

hbierlee:
I'll respond to the specific comments, but I think I didn't put in the big advantage of this approach which is that every test case here (defined as `transform(input) == expected_output`) becomes a separate unit test. They are not run sequentially. This is the same approach (and mostly similar code) as in `test_constraints` (where it's defined by `solve(input, solver) == expected_output`). If a test fails, it's much easier to debug:

```
>       assert str(flat) == expected
E       AssertionError: assert '[(max(x[0],x...V0), IV0 < 3]' == '[(max(x[0],x...V0), IV0 < 4]'
E         
E         - [(max(x[0],x[1],x[2])) == (IV0), IV0 < 4]
E         ?                                        ^
E         + [(max(x[0],x[1],x[2])) == (IV0), IV0 < 3]
E         ?                                        ^
```

So originally, the result is this:
```
        cases = [(min(ivs) == 3, "[min(ivs[0],ivs[1],ivs[2]) == 3]"),
                 (min(ivs) > 3, "[(min(ivs[0],ivs[1],ivs[2])) == (IV0), IV0 > 3]"),
                 (min(ivs) <= 3, "[(min(ivs[0],ivs[1],ivs[2])) == (IV2), IV2 <= 3]"),
                 (3 != max(ivs), "[(max(ivs[0],ivs[1],ivs[2])) == (IV4), IV4 != 3]"),
                 (3 > max(ivs), "[(max(ivs[0],ivs[1],ivs[2])) == (IV6), IV6 < 4]"),
                 (3 <= max(ivs), "[(max(ivs[0],ivs[1],ivs[2])) == (IV8), IV8 >= 3]"),
                ]
    
        # test transformation
        for (expr, strexpr) in cases:
>           self.assertEqual( str(only_numexpr_equality(flatten_constraint(expr))), strexpr )
E           AssertionError: '[(max(ivs[0],ivs[1],ivs[2])) == (IV6), IV6 < 3]' != '[(max(ivs[0],ivs[1],ivs[2])) == (IV6), IV6 < 4]'
E           - [(max(ivs[0],ivs[1],ivs[2])) == (IV6), IV6 < 3]
E           ?                                              ^
E           + [(max(ivs[0],ivs[1],ivs[2])) == (IV6), IV6 < 4]
E           ?                                              ^
```

It's not so clear which case is exactly failing, I don't know if IV6 is the 1st or 2nd aux var of the failing test, and trying to debug this test will run all 6 cases unless I comment them out.

tests/test_transf_comp.py
hbierlee: I agree with this pattern. In this case, the transformation is always the same so I didn't make it part of the parametrization, but that's the idea when you start testing different transforms (like in test_constraints, you have different solvers).

tests/test_transf_comp.py
hbierlee: Absolutely agree, this is all pytest magic (same as in test_constraints). The tuple is jsut because a test case is defined by those two arguments (as in prev. comment, I'd add the transform function if there were different ones tested!). The ids=str just names the unit test case (eg `test_transforms[min(x[0],x[1],x[2]) == 3-[min(x[0],x[1],x[2]) == 3]]`)

tests/test_transf_comp.py
hbierlee: See previous two comments. We'd only need one class to run all tests of type after adding the transform to parametrize, given us test cases that check: "Given transform `f`, check that for input model `m` and expected output model `o`, we have `f(m)==o`. 

I don't think this approach will fit with every possible way you want to check a transform, but in many cases it will fit (and become more maintainable because you're not repeating the checking code over and over).

hbierlee:
> if its about the variable counters, then I think the most natural 'fix' would be in the original code
> 
> ```python
> for (expr, strexpr) in cases:
>     self.setUp()  # reset counters
>     assert...
> ```

This fixture was the recommended way by pytest, but I agree it's magic (you have to add the `setup` arg in every test that requires this setup, but you never have to call it :) magic indeed). I can try if this approach works if it's simpler to read?

hbierlee:
It works if you add `setup` as a function which you call at the start of the test function, but I prefer this pytest fixture way since then I'm sure it runs before parametrization creates the constraints (which doesn't happen currently but it easily can!). I think the way it looks now is very readable, creates a test case for every constraint where all aux vars start at 0, all user variables are clearly

tests/test_transf_comp.py
hbierlee: Actually this `str=ids` is the default, so I've just removed it.

tias:
went for the simple change to accommodate that increasing IV numbers would be unexpected here.

PR #639, opened 2025-04-16
Linearize sub

IgnaceBleukx:
Linearization of constraints such as `2 - x == y` went wrong when explcitely posted to the transformation.
Normally, we rewrite this directly as ` 2 + (-x) == y` so it stayed under the radar for some time.

### DIFF
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -152,7 +152,7 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
 
             if lhs.name == "sub":
                 # convert to wsum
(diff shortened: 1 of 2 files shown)


hbierlee:
Is think this is absolutely fine to merge as-is.

Out of interest, is there something to be said for fixing this generally in `__init__` of Operator, or is that not possible (or does it preclude other optimizations?)

hbierlee:
Actually, just found the doc-string explaining this:

```
def _wsum_should(arg):
    """ Internal helper: should the arg be in a wsum instead of sum

    True if the arg is already a wsum,
    or if it is a product of a constant and an expression 
    (negation '-' does not mean it SHOULD be a wsum, because then
     all substractions are transformed into less readable wsums)
    """
```

Then good to merge :)

hbierlee (approved):

PR #641, opened 2025-04-16
SolverLookup: add installed() and status()

tias:
Inspired by #628 but not going as far, I added two functions to solver lookup. After the release, the docs will then be something like

"""
Accessing solvers always happens through the cp.SolverLookup() static class.

To see which solvers CPMpy supports, and which of those are installed and available on your system, run
`cp.SolverLookup().status()`
which will print a human-readable overview.

To loop over all systems installed and available on your machine, do:
```python
for name in cp.SolverLookup.installed():
    s = cp.SolverLookup.get(name)
    s.add(...)
    s.solve()
```
"""

I've been hesitating about the names...

* For me, `status()` is fine to just print it, rather than the longer print_status(), but I'm open to other opinions

* For `installed()` other alternatives are `list()`, `supported()`, `available()`. I want something short but where the meaning is immediately clear, so both supported and binary/license ready to use... maybe `cp.SolverLookup.available()' then? not sure

I'd like this minimal version in the upcoming release, and then how to actually populate a nice `status()` is for in #628

### DIFF
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -102,8 +102,7 @@ class CPM_gurobi(SolverInterface):
             global GRB_ENV
             if GRB_ENV is None:
                 # initialise the native gurobi model object
-                GRB_ENV = gp.Env()
-                GRB_ENV.setParam("OutputFlag", 0)
+                GRB_ENV = gp.Env(params={"OutputFlag": 0})
                 GRB_ENV.start()
             return True
         except Exception as e:
--- a/cpmpy/solvers/utils.py
+++ b/cpmpy/solvers/utils.py
@@ -97,9 +97,9 @@ class SolverLookup():
         """
             Return the list of names of all solvers (and subsolvers) supported on this system.
 
-            If a solver name is returned, it means that both the Python package is installed and the
-            solver's `.supported()` function returns True and is hence ready for immediate use
-            (e.g. any separate binaries are installed if necessary, and licenses are active if needed).
+            If a solver name is returned, it means that the solver's `.supported()` function returns True
+            and it is hence ready for immediate use
+            (e.g. any separate binaries are also installed if necessary, and licenses are active if needed).
 
             Typical use case is to use these names in `SolverLookup.get(name)`.
         """
--- a/cpmpy/solvers/utils.py
+++ b/cpmpy/solvers/utils.py
@@ -82,7 +82,24 @@ class SolverLookup():
                ]
 
     @classmethod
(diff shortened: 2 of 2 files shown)


hbierlee:
Quick thoughts, I like it (similar to `minizinc --solvers`), I would use `print_solvers()`, because it does tell the user we get a string instead of something more programmable like `solver.status()`, and another suggestion for `installed` is `enabled`.

IgnaceBleukx:
I follow Henk here and prefer `print_xxx()` rather than just `status()` to print the overview of the solvers.
Also in `solver.status()` we return an object after the solve-call, so to be consistent we should not return an object from `SolverLookup.status()` too.

No real preference on the `installed/supported/available` thing

tias:
Thomas' original #545 is `print_status()`, I guess we go for that then; though just `print` could also work but what would somebody expects it prints; I guess the contents of what can be looked up, which is what it is doing, hmm hmm

(the repr is very implicit and wont be clear in the docs, like we now realize __add__ was for model)

hbierlee:
Actually, the most appropiate way would probably be to add an executable #[https://setuptools.pypa.io/en/latest/userguide/entry_point.html](script). After `pip install cpmpy`, the user can just run, say, `cpmpy --solvers`.

tias:
I don't see myself explaining in the docs that they have to run a script... I think a doc entry like in my original post is what we should aim for, and its just the names to decide on now

ThomSerg:
I agree that `print_status()` is more clear to the user that it will print and not return something (like Ignace mentions it will otherwise be confusing with the existing behavior of `Solver.status()` ). I originally separated it into `status()` and `print_status()` so that it wouldn't be as tricky as with minizinc's python API to get programmatic access to a list of all solvers, whether or not they are installed and which version. 

Additionally, for `.status()` I actually like Ignace's idea of returning an object similar to `SolverStatus` (since as described in #628 adding new fields to the status in the future would otherwise be tricky for backwards compatibility). But indeed, we can keep `.status()` for after the upcoming release, especially with all the details involved with subsolvers.

As for the name of the other method I like `installed()` and `available()`, especially the second one since it checks more than just installation status (like available on path, license). But unfortunately the solvers already use `supported()` for this same check (but on a per-solver level).

tias:
I've changed it to `available()` which I also like (and to `print_status()` as discussed)

But writing the docs, I do wonder whether it should be `cp.SolverLookup.supported()` given that we idd have `solver.supported()`. Pro is that it will use the same function name, Contra is that here it returns a list of names; for a solver it returns True/False.

Issue #279, opened 2023-04-18
Time limit semantics


IgnaceBleukx:
Every solver has the optional argument `time_limit` in its solve call.

Right now there are some inconsistencies:

```python
from cpmpy import *
model  = Model(boolvar())
model.solve(solver="ortools", time_limit=-1)
>> error
model.solve(solver="gurobi", time_limit=-1)
>> error
model.solve(solver="pysat", time_limit=-1)
>> True
model.solve(solver="z3", time_limit=-1)
>> True
model.solve(solver="minizinc", time_limit=-1)
>> False
```
I think a value <= 0 as time limit should correspond to "no time limit"

Thoughts?

Wout4:
There is no other viable way to interpret a negative time limit, so no time limit is the best option yes

Dimosts:
On the other hand, I believe a value <=0 should be interpreted as do not solve.

This can happen in processes involving multiple solve calls e.g. in a loop, with a time limit for the entire process. Then, you want to stop and return in case you have already reached the time limit instead of giving infinite time!

IgnaceBleukx:
I think it's most appropriate to raise a TimeOutError when the time-limit is less than or equal to 0; this is easy to understand and catch by the caller too (and built into Python 😉 )
________________________________

hbierlee:
I'd have a slight preference for ValueError because we don't throw the TimeOutError when a regular timeout occurs. This regards -5 as an invalid timeout duration. If you think -5 can be interpreted as a valid timeout, then I agree it should timeout immediately rather than have no timeout.

PR #642, opened 2025-04-17
Handle non-positive time limit consistently

Dimosts:
Issue #279 revealed that a non-positive time limit is handled with different ways in different solvers. This PR makes it consistent, raising a ValueError in all solvers in such a case.

Added also tests for it

### DIFF
--- a/cpmpy/solvers/gcs.py
+++ b/cpmpy/solvers/gcs.py
@@ -163,9 +163,8 @@ class CPM_gcs(SolverInterface):
         self.proof_location = proof_location
      
         # set time limit
-        if time_limit is not None:
-            if time_limit <= 0:
-                raise ValueError("Time limit must be positive")
+        if time_limit is not None and time_limit <= 0:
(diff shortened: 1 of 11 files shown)


cpmpy/solvers/gcs.py
```diff
@@ -162,6 +162,11 @@ def solve(self, time_limit=None, prove=False, proof_name=None, proof_location=".
             self.proof_name = proof_name
         self.proof_location = proof_location
      
+        # set time limit
+        if time_limit is not None:
```
IgnaceBleukx, line 166: Don't need the double if-statement here: 
```python
if time_limit is not None and time_limit <= 0:
    raise ValueError
```

IgnaceBleukx: Same in a few other solvers

IgnaceBleukx (approved):
Nice, ValueError seems to be the most appropriate indeed.

One minor comment regarding style, but non-blocker

Issue #624, opened 2025-04-10
Small bug on NValueExcept constraint


peerjosse:
Small bug I found for the NValueExcept constraint with useless input, so not sure if relevant.
The program should assign x to 0, but gives the following error instead: `TypeError: object of type '_BoolVarImpl' has no len()` in `decompose_comparison`

```python
import cpmpy as cp

m = cp.Model()

x = cp.intvar(-3, 5)
y = cp.intvar(-3, 5)
z = cp.intvar(-3, 5)

m += x == cp.NValueExcept([1], 1)

# Example usage (e.g. after removal of a variable from the array)
# m += cp.Decreasing(z, y, cp.NValueExcept([1], 1))

m.solve()
```

The issue is in the following code segment of `globalfunctions.py`:
```python
# introduce boolvar for each possible value
bvars = boolvar(shape=(ub + 1 - lb))
idx_of_n = n - lb
if 0 <= idx_of_n < len(bvars):
```

Dimosts:
It is more general. It can happen if any of the objects in the first argument array is a constant, or even a var with 1 value in the domain.

It is bad usage of the constraint, but still we should support this as it could happen due to data given for domains in the vars.

PR #645, opened 2025-04-17
Fix in NValueExceptN for 1 value range

Dimosts:
Issue #624 showed that NValueExceptN decomposition crashes when the possible values just include one value. This ofc is bad usage of the constraint, but can happen I guess due to data given in a problem instance.

In the beginning I thought of fixing it by putting a __len__ in expressions, returning just 1, as the len(bv) is the first point that it crashes, and I think that this could happen in other cases too. But this did not fix the issue as then bv is not subscriptable.

So ended up in just a simple check if we end up with just one value (lb == ub) and then return the appropriate constraint, depending on whether that value is n or not.

### DIFF
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -444,6 +444,9 @@ class NValueExcept(GlobalFunction):
         lbs, ubs = get_bounds(arr)
         lb, ub = min(lbs), max(ubs)
 
+        if lb == ub: # if the domain has only one value then check if it is n or not
+            return [eval_comparison(cmp_op, int(lb != n), cpm_rhs)], []
+
         constraints = []
 
         # introduce boolvar for each possible value


cpmpy/expressions/globalfunctions.py
```diff
@@ -444,6 +444,9 @@ def decompose_comparison(self, cmp_op, cpm_rhs):
         lbs, ubs = get_bounds(arr)
         lb, ub = min(lbs), max(ubs)
 
+        if lb == ub: # if the domain has only one value then check if it is n or not
```
IgnaceBleukx, line 447: I think what you can do instead, is to make `bvars = boolvar(shape=(ub+1-lb,))` so it will always be an array.
We do smt similar in `make_assump_model` in tools.explain.utils.
I think that the original decomposition should work as is then.

Issue #632, opened 2025-04-12
choco solver crash using gcc constraint


peerjosse:
crashed solver: choco
correctly working solvers: ortools, minizinc, z3, gurobi

In the following program:
- the error `'int' object has no attribute 'handle'` occurs in `make_intvar_array` of pychoco, so not sure if this is a cpmpy bug.
- the error disappears when the 3 gets changed into a variable (e.g. x)

```python
import cpmpy as cp

m = cp.Model()

x = cp.intvar(-3, 5, shape=1)
y = cp.intvar(-3, 5, shape=1)
z = cp.intvar(-3, 5, shape=1)

m += (cp.GlobalCardinalityCount([3, y, z],[1, 2, 3],[0,0,3])) >= 1

# following works correctly
# m += x == 3
# m += (cp.GlobalCardinalityCount([x, y, z],[1, 2, 3],[0,0,3])) >= 1

print(m.solve(solver='ortools'))   # returns no error
print(m.solve(solver='minizinc'))  # returns no error
print(m.solve(solver='z3'))        # returns no error
print(m.solve(solver='choco'))     # returns error: 'int' object has no attribute 'handle'
print(m.solve(solver='gurobi'))    # returns no error
```

Issue #635, opened 2025-04-13
choco crash in InDomain constraint


peerjosse:
crashed solver: choco
working solvers: ortools, minizinc, z3, gurobi

In the following program:
- the `InDomain` constraint leads to an error for solver choco
- the error shows `in method 'int_array_set', argument 2 of type 'int'`
- this crash does not occur when constants are in the second argument instaed of variables
- this crash does not occur when the regular `... in [...]` is used

```python
import cpmpy as cp

m = cp.Model()

x = cp.intvar(-3, 5, shape=1)
y = cp.intvar(-3, 5, shape=1)
z = cp.intvar(-3, 5, shape=1)

m += cp.InDomain(x, [y, z])
# m += cp.InDomain(x, [1, 3]) # works fine
# m += x in [y,z] # works fine

print(m.solve(solver='ortools'))   # returns no error
print(m.solve(solver='minizinc'))  # returns no error
print(m.solve(solver='z3'))        # returns no error
print(m.solve(solver='choco'))     # returns error: in method 'int_array_set', argument 2 of type 'int'
print(m.solve(solver='gurobi'))    # returns no error
```

Dimosts:
Actually, m += x in [y, z] does not work fine, as it does not actually post a constraint.

The way to do this normally is (x == y) | (x == z).

It breaks in choco because it is the only one that does not decompose.

I think the way to go is just to enforce that the second argument should only contain ints?

Dimosts:
Yes, so this actually should be fixed in the InDomain global constraint to enforce the second argument to not contain expressions (or variables), as it does not make sense to use variables there.

PR #646, opened 2025-04-17
Convert ints to vars in GCC first argument for Choco

Dimosts:
Fixes issue #632 

Initial plan was to fix together with the indomain crash from #635, but that issue will be fixed outside of the choco interface.

I also checked other globals in choco if a similar fix is needed, it seems that we have covered similar cases in other globals already.

very specific bug so not sure if test is needed.

### DIFF
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -589,7 +589,7 @@ class CPM_choco(SolverInterface):
                 return self.chc_model.int_value_precede_chain(self._to_vars(cpm_expr.args[0]), cpm_expr.args[1])
             elif cpm_expr.name == "gcc":
                 vars, vals, occ = cpm_expr.args
-                return self.chc_model.global_cardinality(*self.solver_vars([vars, vals]), self._to_vars(occ), cpm_expr.closed)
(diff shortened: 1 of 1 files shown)


cpmpy/solvers/choco.py
```diff
@@ -592,7 +592,7 @@ def _get_constraint(self, cpm_expr):
                 return self.chc_model.int_value_precede_chain(self._to_vars(cpm_expr.args[0]), cpm_expr.args[1])
             elif cpm_expr.name == "gcc":
                 vars, vals, occ = cpm_expr.args
-                return self.chc_model.global_cardinality(*self.solver_vars([vars, vals]), self._to_vars(occ), cpm_expr.closed)
+                return self.chc_model.global_cardinality(self._to_vars(vars), vals, self._to_vars(occ), cpm_expr.closed)
```
IgnaceBleukx, line 595: The solver_vars also converts numpy to int, so you need to keep that part for `vals` : )

IgnaceBleukx (approved):
We should probably figure out a more structural way for testing our globals on all solvers with differnent argument types, to prevent these edge-cases in the future. WIll open a discussion issue for it.

PR #649, opened 2025-04-18
Undefault `pypblib` dependency for `pysat`

### LINKED ISSUE #643: Potential errors with `pblib` dependency of `pysat`
By adding pblib as a default when the user runs `pip install .[pysat]`, we decrease the chance of unsupported (PB) constraints when solving with pysat, but we increase the chance of install errors due to pypblib. When installing, pypblib depends on some system configuration (related to the GNU gcc compiler), which on Linux is usually alright (`gcc` is installed by default), but on Windows (found by @ThomSerg) we seem to be having some problems.

 Should we still have pblib as default installed, or do we only recommend it to install once the user runs into an unsupported PB constraint?

A potential follow-up issue: should we add more platforms to the CI?

hbierlee:
Fixes #643. To properly test it with/without installing pypblib, had to refactor the import caching a bit too (what we're actually trying to do is class initialization, but can't find the proper docs). Anyway, now it should only be imported once.

### DIFF
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -21,7 +21,13 @@
     Installation
     ============
 
-    Requires that the 'python-sat' package is installed. If you want to also solve pseudo-Boolean and cardinality constraints, you should also install its optional dependency 'pypblib', as follows:
+    Requires that the 'python-sat' package is installed:
+
+    .. code-block:: console
+
+        $ pip install pysat
+
+    If you want to also solve pseudo-Boolean and cardinality constraints, you should also install its optional dependency 'pypblib', as follows:
 
     .. code-block:: console
 
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -114,7 +114,8 @@ class CPM_pysat(SolverInterface):
         except Exception as e:
             raise e
 
-    _pb = None  # holds pysat.pb module if its dependency `pypblib` installed
+    """The `pysat.pb` module if its dependency `pypblib` installed, `None` if we have not checked it yet, or `False` if we checked and it is *not* installed"""
+    _pb = None  # 
 
     @staticmethod
     def solvernames():
@@ -446,7 +447,6 @@ class CPM_pysat(SolverInterface):
 
     def _pysat_cardinality(self, cpm_expr, reified=False):
         """ Convert CPMpy comparison of `sum` (over Boolean variables) into PySAT list of clauses """
-        assert self._card is not None, "Native module which should have been cached in init"
 
         # unpack and transform to PySAT argument
         lhs, rhs = cpm_expr.args
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -21,12 +21,11 @@
     Installation
     ============
 
-    Requires that the 'python-sat' package is installed. If you want to also solve
-    pseudo-Boolean constraints, you should also install its optional dependency 'pblib', as follows:
+    Requires that the 'python-sat' package is installed. If you want to also solve pseudo-Boolean and cardinality constraints, you should also install its optional dependency 'pypblib', as follows:
 
     .. code-block:: console
 
-        $ pip install python-sat[pblib]
+        $ pip install pypblib
 
     See detailed installation instructions at:
     https://pysathq.github.io/installation
@@ -82,7 +81,7 @@ class CPM_pysat(SolverInterface):
 
     """
 
-    IMPORT_PYSAT_PBLIB_ERROR = ImportError("PySAT was installed without optional dependency `pblib`, which is required to encode (weighted) sums. To install PySAT with recommended optional dependencies, run `pip install cpmpy[pysat]`.")
+    IMPORT_PYSAT_PBLIB_ERROR = ImportError("The model contains a PB or cardinality constraint, for which PySAT needs an additional dependency (PBLib). To install it, run `pip install pypblib`.")
 
     @staticmethod
     def supported():
@@ -98,6 +97,8 @@ class CPM_pysat(SolverInterface):
             return False
         except Exception as e:
             raise e
+
+    _pblib = None  # holds pysat.[card,pb] modules if installed
 
     @staticmethod
     def solvernames():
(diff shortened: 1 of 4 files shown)


cpmpy/solvers/pysat.py
```diff
@@ -142,14 +141,16 @@ def __init__(self, cpm_model=None, subsolver=None):
...
-        # try to import the pblib module once
-        try:
-            from pysat import card, pb
-            self._card = card
-            self._pb = pb
-        except NameError:
-            self._card = None
-            self._pb = None
+        # try to import pypblib once (TODO I'd like class initialization, but can't find proper docs for it )
+        if CPM_pysat._pb is None:
```
tias, line 145: is there a reason to do this here? Another logical place would be in the classmethod supported(), which would then set cls._card and cls._pb?

hbierlee: It should be a class initialization function which I've read somewhere but I can't for the life of me find the proper documentation..

tias: check the first line of code of this function (129)...

hbierlee: Oops. fixed now.

hbierlee: Ah I see what you mean. Yes I can put it there.

hbierlee: Done

cpmpy/solvers/pysat.py
```diff
@@ -93,12 +90,25 @@ def supported():
             # while we need the 'python-sat' package, some more checks:
             from pysat.formula import IDPool
             from pysat.solvers import Solver
+
+            # try to import pypblib once
+            if CPM_pysat._pb is None:
+                from pysat import card
```
tias, line 96: this is independent of _pb, should be outside the if

cpmpy/solvers/pysat.py
```diff
@@ -93,12 +90,25 @@ def supported():
...
+                CPM_pysat._card = card  # native
+                try:
+                    from pysat import pb  # require pypblib
+                    CPM_pysat._pb = pb
+                except (ModuleNotFoundError, NameError):  # pysat returns the wrong error type (latter i/o former)
+                    CPM_pysat._pb = False  # not installed, avoid reimporting
```
tias, line 102: you set it to False but the check above online 95 checks for None... so should be None?

Also your check in the pb function below should be consistent with the choice you will make

cpmpy/solvers/pysat.py
```diff
@@ -93,12 +90,25 @@ def supported():
...
+
             return True
         except ModuleNotFoundError:
             return False
         except Exception as e:
             raise e
 
+    _pb = None  # holds pysat.pb module if its dependency `pypblib` installed
```
tias, line 110: I don't think this is needed, just like you don't do it for _card either... I think thats fine
(note how it is None here... just remove this?)

cpmpy/solvers/pysat.py
```diff
@@ -438,8 +439,7 @@ def get_core(self):
 
     def _pysat_cardinality(self, cpm_expr, reified=False):
         """ Convert CPMpy comparison of `sum` (over Boolean variables) into PySAT list of clauses """
-        if self._card is None:
-            raise self.IMPORT_PYSAT_PBLIB_ERROR
+        assert self._card is not None, "Native module which should have been cached in init"
```
tias, line 442: I don't think this check is needed... you can assume it exists

cpmpy/solvers/pysat.py
```diff
@@ -465,9 +465,10 @@ def _pysat_cardinality(self, cpm_expr, reified=False):
             raise ValueError(f"PySAT: Expected Comparison to be either <=, ==, or >=, but was {cpm_expr.name}")
 
     def _pysat_pseudoboolean(self, cpm_expr):
-        """ Convert CPMpy comparison of `wsum` (over Boolean variables) into PySAT list of clauses """
-        if self._pb is None:
-            raise self.IMPORT_PYSAT_PBLIB_ERROR
+        """Convert CPMpy comparison of `wsum` (over Boolean variables) into PySAT list of clauses."""
+        if self._pb is False:
+            raise ImportError("The model contains a PB constraint, for which PySAT needs an additional dependency (PBLib). To install it, run `pip install pypblib`.")
+        assert self._pb is not None, "We should have imported the `pb` module if it was installed."
```
tias, line 471: now I'm just confused... you want it to be eitehr false or none?
Just one is fine..,

cpmpy/solvers/pysat.py
```diff
@@ -21,12 +21,11 @@
     Installation
     ============
 
-    Requires that the 'python-sat' package is installed. If you want to also solve
-    pseudo-Boolean constraints, you should also install its optional dependency 'pblib', as follows:
+    Requires that the 'python-sat' package is installed. If you want to also solve pseudo-Boolean and cardinality constraints, you should also install its optional dependency 'pypblib', as follows:
```
tias, line 24: would be more consistent with other sovlers if

"""
Requires that the 'python-sat' package is installed:

.. code-block:: console
        $ pip install pysat
"""
followed by the 'If you...'
where I think the earlier pip install pysat[pblib] is better because its pysat that chooses the dependent version?

tias (changes requested):

cpmpy/solvers/pysat.py
hbierlee: Right, I forgot we didn't make this the recommendation in the other solvers here. Plus I can give them the alternative of `pysat`  if `pypblib` fails to compile.

cpmpy/solvers/pysat.py
hbierlee: I've added a doc string on `_pb` to clarify, but there are (confusingly!) three states, where `None` is we didn't import yet, the module where we did import succesfully and `False` where we tried but it was not imported (so not installed). This *really* prevents us from attempting imports multiple times whether it is installed or not.

You can let me know if you think is ok. Alternative design was to indeed omit the field (as "no import yet"). I thought I'd then have to use `has_attr(CPM_py, "_pb")` or something to distinguish.. but actually this might be more simple and clear, so I will try that now.

cpmpy/solvers/pysat.py
hbierlee: I'll respond in the other `False`  v `None` comment

cpmpy/solvers/pysat.py
hbierlee: Explained in other comment

cpmpy/solvers/pysat.py
hbierlee: Well, I feel this is a little more clear, let me know. Obviously, all quite overkill for something that is only checked when we initialize a solver, which makes code readability more of a priority.

hbierlee: More generally, we do a lot of engineering to avoid import statements, because (so I heard) they cause significant overhead. When we have more regular benchmarking, we should test if this is still true, or is also true if we use [`importlib.find_spec`](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec).

tias:
yes, great. I had not understood how clever the _pb use was before, sorry for that.

Idd, we should benchmark the import cost; the case we had was an import on every _post_constraint, and that was idd to our surprise causing overhead...

PR #653, opened 2025-04-21
int2bool transformation

tias:
a textbook int2bool direct encoding, as a transformation

prototype implemented such that using different encodings for different variables is possible. Way cleaner then earlier prototypes like #319 or #79 

passes test_constraints, but not much tested 'in the wild'.

### DIFF
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -51,10 +51,12 @@
     Module details
     ==============
 """
+from threading import Timer
+
 from .solver_interface import SolverInterface, SolverStatus, ExitStatus
 from ..exceptions import NotSupportedError
 from ..expressions.core import Comparison, Operator, BoolVal
-from ..expressions.variables import _BoolVarImpl, NegBoolView, boolvar
+from ..expressions.variables import _BoolVarImpl, _IntVarImpl, NegBoolView, boolvar
 from ..expressions.globalconstraints import DirectConstraint
 from ..transformations.linearize import canonical_comparison, only_positive_coefficients
 from ..expressions.utils import is_int, flatlist
@@ -65,6 +67,7 @@ from ..transformations.flatten_model import flatten_constraint
 from ..transformations.linearize import linearize_constraint
 from ..transformations.normalize import toplevel_list, simplify_boolean
 from ..transformations.reification import only_implies, only_bv_reifies, reify_rewrite
+from ..transformations.int2bool import int2bool, int2bool_make
 
 
 class CPM_pysat(SolverInterface):
@@ -161,6 +164,7 @@ class CPM_pysat(SolverInterface):
         # initialise the native solver object
         self.pysat_vpool = IDPool()
         self.pysat_solver = Solver(use_timer=True, name=subsolver)
+        self.ivarmap = dict()  # for the integer to boolean encoders
 
         # initialise everything else and post the constraints/objective
         super().__init__(name="pysat:"+subsolver, cpm_model=cpm_model)
@@ -188,8 +192,13 @@ class CPM_pysat(SolverInterface):
                             Note: the PySAT interface is statefull, so you can incrementally call solve() with assumptions and it will reuse learned clauses
         """
 
-        # ensure all vars are known to solver
-        self.solver_vars(list(self.user_vars))
+        # ensure all Boolean vars are known to solver
+        for v in list(self.user_vars):  # can change during iteration
+            if isinstance(v, _BoolVarImpl):
+                self.solver_vars(v)
+            else:  # intvar
+                if not v in self.ivarmap:
+                    self += int2bool_make(self.ivarmap, v)
 
         if assumptions is None:
             pysat_assum_vars = [] # default if no assumptions
@@ -197,13 +206,11 @@ class CPM_pysat(SolverInterface):
             pysat_assum_vars = self.solver_vars(assumptions)
             self.assumption_vars = assumptions
 
-        import time
         # set time limit
         if time_limit is not None:
             if time_limit <= 0:
                 raise ValueError("Time limit must be positive")
             
-            from threading import Timer
             t = Timer(time_limit, lambda s: s.interrupt(), [self.pysat_solver])
             t.start()
             my_status = self.pysat_solver.solve_limited(assumptions=pysat_assum_vars, expect_interrupt=True)
@@ -237,13 +244,21 @@ class CPM_pysat(SolverInterface):
             sol = frozenset(self.pysat_solver.get_model())  # to speed up lookup
             # fill in variable values
             for cpm_var in self.user_vars:
-                lit = self.solver_var(cpm_var)
-                if lit in sol:
-                    cpm_var._value = True
-                elif -lit in sol:
-                    cpm_var._value = False
-                else: # not specified, dummy val
-                    cpm_var._value = True
+                if isinstance(cpm_var, _BoolVarImpl):
+                    lit = self.solver_var(cpm_var)
+                    if lit in sol:
+                        cpm_var._value = True
+                    else:  # -lit in sol (=False) or not specified (=False)
+                        cpm_var._value = False
+                elif isinstance(cpm_var, _IntVarImpl):
+                    assert cpm_var.name in self.ivarmap, "Integer variable %s not found in ivarenc" % cpm_var.name
+                    varenc = self.ivarmap[cpm_var.name]
+                    lits = self.solver_vars(varenc.vars())
+                    # default value=False
+                    vals = [lit in sol for lit in lits]
+                    cpm_var._value = varenc.decode(vals)
+                else:
+                    raise NotImplementedError(f"CPM_pysat: variable {cpm_var} not supported")
 
         else: # clear values of variables
             for cpm_var in self.user_vars:
@@ -302,6 +317,7 @@ class CPM_pysat(SolverInterface):
         cpm_cons = only_bv_reifies(cpm_cons)
         cpm_cons = only_implies(cpm_cons)
         cpm_cons = linearize_constraint(cpm_cons, supported=frozenset({"sum","wsum", "and", "or"}))  # the core of the MIP-linearization
+        cpm_cons = int2bool(cpm_cons, ivarmap=self.ivarmap)
         cpm_cons = only_positive_coefficients(cpm_cons)
         return cpm_cons
 
--- a/cpmpy/transformations/int2bool.py
+++ b/cpmpy/transformations/int2bool.py
@@ -1,0 +1,283 @@
+"""
+Convert integer linear constraints to pseudo-boolean constraints
+"""
+
+from typing import List
+import cpmpy as cp
+from abc import ABC, abstractmethod
+from ..expressions.variables import _BoolVarImpl, _IntVarImpl
+from ..expressions.core import Comparison, Operator
+from ..transformations.get_variables import get_variables
+from ..expressions.core import Expression
+
+def int2bool(cpm_lst: List[Expression], ivarmap=None, encoding="auto"):
+    """
+    Convert integer linear constraints to pseudo-boolean constraints
+    """
+    assert encoding in ("auto", "direct"), "Only auto or direct encoding is supported"
+    if ivarmap is None:
+        ivarmap = dict()
+
+    cpm_out = []
+    for expr in cpm_lst:
+        vs = get_variables(expr)
+        # skip all Boolean expressions
+        if all(isinstance(v, _BoolVarImpl) for v in vs):
+            cpm_out.append(expr)
+            continue
+
+        # check if all variables are in the ivarmap
+        for v in vs:
+            if type(v) is _IntVarImpl and v.name not in ivarmap: 
+                cons = int2bool_make(ivarmap, v, encoding, cpm_out)
+                cpm_out.extend(cons)
+        
+        # we also need to support b -> subexpr
+        # where subexpr's transformation is identical to non-reified expr
+        # we do this with a special flag
+        is_halfreif = False
+        if expr.name == "->":
+            is_halfreif = True
+            b = expr.args[0]  # PAY ATTENTION: we will overwrite expr by the rhs of the ->
+            expr = expr.args[1]
+        
+        # now replace intvars with their encoding
+        if isinstance(expr, Comparison):
+            # special case: lhs is a single intvar
+            lhs,rhs = expr.args
+            if type(lhs) is _IntVarImpl:
+                cons = ivarmap[lhs.name].encode_comparison(expr.name, rhs)
+                if is_halfreif:
+                    cpm_out.extend([b.implies(c) for c in cons])
+                else:
+                    cpm_out.extend(cons)
+            elif lhs.name == "wsum":
+                # if its a wsum, insert encoding of terms
+                newweights = []
+                newvars = []
+                for w,v in zip(*lhs.args):
+                    if type(v) is _IntVarImpl:
+                        # get list of weights/vars to add
+                        ws,vs = ivarmap[v.name].encode_term(w)
+                        newweights.extend(ws)
+                        newvars.extend(vs)
+                    else:
+                        newweights.append(w)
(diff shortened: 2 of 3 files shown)


hbierlee:
@tias I've added some tests for the `int2bool` transform, which generates some constraints that would come out of linearize, and compares number of solutions of the original and transformed constraint (which alltogether takes ~12 seconds on 4 cores).

Fixed some issues, some remain (of which I think most are relevant). For this PR to be merged, what would we like included and what to be future work? This list is in decreasing order of probability:

- [ ] Fix remaining tests
- [ ] Add the remaining order/binary encodings?
- [ ] Implement some `auto` heuristic?

tias:
Thanks Henk!

We should definitely fix the bugs, correctness first.

Then, regarding performance, I can image this taking ages for problems with variables with large domains; so the number of problems in which this could be used would increase substantially if we would add a textbook log encoding with as 'auto' rule some cap on the size of the domain before swithcing to log or so...?

hbierlee:
Sure, let's just add all of them with some heuristic (which can chose per term)

hbierlee:
@tias this is ready for  review (although I can't officially request it via github, presumably because you opened the PR (: ).

Following decision points:

- I've added a `encoding` arg to pysat to choose the encoding "auto"/"direct"/.., but we should probably remove it and default to "auto". I don't think there is any structural way to influence the transformations, right?
- The inequality constraint encodings for the binary encoding is done using a PB linear. While using a lexicographic encoding is probably slightly more effective, I don't think we have to add this now
- I'd like to do the various boolean operations using `cp.any/all` (such as negating a conjunction, which is now done using lists), but then this PR would change pysat's `add`, and there is also the question whether this is worth the overhead. So probably best leave it for now until we can do more regular profiling.
- There is another smaller refactor I might sneak in before the review, but it is not required

For the rest, code and comments should explain things pretty well

hbierlee:
You will see that similar to pindakaas I use recursion, please see discussion about this on https://github.com/CPMpy/cpmpy/pull/600/files#r2066010097

hbierlee:
One more comment, I removed some of the doc-strings in methods of subclasses, since they should be inheritted from the base class. If the method is different in some way, I do add an individual doc-string for it.

cpmpy/solvers/pysat.py
```diff
@@ -320,49 +355,61 @@ def add(self, cpm_expr_orig):
...
-      """
-      # add new user vars to the set
-      get_variables(cpm_expr_orig, collect=self.user_vars)
+        return self
+
+    __add__ = add  # avoid redirect in superclass
 
-      # transform and post the constraints
-      for cpm_expr in self.transform(cpm_expr_orig):
+    def _add_clauses(self, clauses):
```
tias, line 370: I'm not sure why you create this indirection?
function calls are not free in Python...

if you want the code in `_add_expr()` look cleaner, then there define
add_clauses = self.pysat_solver.append_formula

and that will have the 'alias' effect

cpmpy/transformations/int2bool.py
```diff
@@ -0,0 +1,469 @@
...
+    # skip all Boolean expressions
+    if isinstance(expr, (BoolVal, _BoolVarImpl, DirectConstraint)) or expr.name == "or":
+        return [expr], []
+    elif expr.name == "->":
+        # Encode implication recursively
+        p, consequent = expr.args
+        constraints, domain_constraints = _encode_expr(ivarmap, consequent, encoding)
+        return (
+            [p.implies(constraint) for constraint in constraints],
+            domain_constraints + [p == p],  # keep `p` in model
```
tias, line 51: I don't understand the p==p part, the p is still in the model because of the p.implies(cons) constraints above?

cpmpy/transformations/int2bool.py
```diff
@@ -0,0 +1,469 @@
...
+        # the linear may contain Boolean as well as integer variables
+        if isinstance(x, _BoolVarImpl):
+            terms += [(w, x)]
+        else:
+            x_enc, x_cons = _encode_int_var(
+                ivarmap, x, _decide_encoding(x, cmp, encoding)
+            )
+            domain_constraints += x_cons
+            new_terms, k = x_enc.encode_term(
+                w
```
tias, line 158: I find this multi-line for single-letter argument style surprising... you can just put the comment above the line of code, and then it is one nice compact line of code?

cpmpy/transformations/int2bool.py
```diff
@@ -0,0 +1,469 @@
...
+    constraints = lhs_enc.encode_comparison(cmp, rhs)
+    return constraints, domain_constraints
+
+
+def _decide_encoding(x, cmp=None, encoding="auto"):
+    """Decide encoding of `x` via a simple heuristic based on linear comparator and domain size."""
+    if encoding != "auto":
+        return encoding
+    elif _dom_size(x) >= 100:
+        # even slightly large domain sizes suit the binary encoding, as the PB encoding is not expected to be optimized for integer variables!
```
tias, line 193: I'm not sure what 'not expected to be optimized' means... expected by whom? the user? you as a developer? generally known scientific knowledge?

cpmpy/transformations/int2bool.py
```diff
@@ -0,0 +1,469 @@
...
+
+    def __init__(self, x):
+        """Create direct encoding of integer variable `x`."""
+        name = f"EncDir({x.name}"
+        dom_size = _dom_size(x)
+        if dom_size == 0:
+            raise EMPTY_DOMAIN_ERROR
+        elif dom_size == 1:
+            # TODO we would like to use a boolval(true) here rather than a literal, but this causes problems in linearize
+            xs = cp.cpm_array([boolvar(name=name)])
```
tias, line 281: why wrap it in a cpm_array?
if `boolvar()` itself does not work, try `boolvar(shape=(1,))`

cpmpy/transformations/int2bool.py
```diff
@@ -0,0 +1,469 @@
...
+            xs = boolvar(shape=bits, name=f"EncBin({x.name})")
+        super().__init__(x, xs)
+
+    def encode_domain_constraint(self):
+        """Return binary encoding domain constraint (i.e. upper bound is respected with `self._x<=self._x.ub`. The lower bound is automatically enforced by offset binary which maps `000.. = self._x.lb`)."""
+        # encode directly to avoid bounds check for this seemingly tautological constraint
+        return self.encode_comparison("<=", self._x.ub, check_bounds=False)
+
+    def _bitstring(self, d):
+        """Return offset binary representation of `d` as Booleans in order of increasing signicance (e.g. `4` return `001`)."""
```
tias, line 417: isnt this highly unusual? To have 4 be 001 instead of 100? I find it surprising in any case...

cpmpy/transformations/int2bool.py
```diff
@@ -0,0 +1,469 @@
...
+        # more efficient implementation probably not necessary
+        i = self._offset(d)
+        if (
+            i == 0
```
tias, line 421: this remains odd to read for me, like 2 comments above

cpmpy/transformations/int2bool.py
```diff
@@ -0,0 +1,469 @@
...
+            ), f"{self._x} should have already been encoded, so no domain constraints should be returned"
+            return constraint
+        else:
+            raise UNKNOWN_COMPARATOR_ERROR
+
+    def encode_term(self, w=1):
+        return [(w * (2**i), b) for i, b in enumerate(self._xs)], w * self._x.lb
+
+
+def _unzip(iterable):
```
tias, line 464: that looks odd... and unnecessary?

tias (commented):
amazing, and all 3 and an auto rule...

I just have a bit of questions for clarification, and some style remarks that differ wrt the rest of the codebase.

Looking forward to having this included (and idd also in pysdd while we are at it?)

cpmpy/transformations/int2bool.py
hbierlee: True, that's my autoformatter, and putting the comment above fixes it.

cpmpy/transformations/int2bool.py
hbierlee: I've clarified comment, let me know if it's more clear now

cpmpy/transformations/int2bool.py
hbierlee: Ahh I was looking for this (because `shape=1` does not return a listlike, to my frustration!). That's very good to know, that will clean up a lot of parts.

cpmpy/transformations/int2bool.py
hbierlee: Agree

cpmpy/transformations/int2bool.py
hbierlee: Indeed only used in one place (to go from two weights/lits lists to list of 2-tuples), so I've removed it.

cpmpy/solvers/pysat.py
```diff
@@ -130,7 +133,7 @@ def solvernames():
                 names.append(name)
         return names
 
-    def __init__(self, cpm_model=None, subsolver=None):
+    def __init__(self, cpm_model=None, subsolver=None, encoding="auto"):
```
hbierlee, line 136: Question @tias : I've just added and documented this encoding arg, but I'm not sure how we add "solver" (or rather transformation) specific configuration options. Perhaps it's best to leave out this argument and just always set the encoding to "auto" for now?

Advanced users can still change `PySAT.encoding = "direct"`, but then this does not become an expected part of the interface.

hbierlee: I've removed this arg for now, see commit 4b56f926

hbierlee:
One more thought that came up during sub-group, if we want to use int2bool more for our decompositions (i.e. becoming a more flexible version of MapDomain), there is a little more work in making it more solver-independent. As you can see, there are now some changes in pysat to map bools to integers (which could be user vars which are removed from the model). The question is if it is possible to truly makes this work for any solver in a transparent way. I think I'd rather tackle this in a new PR/issue (i.e. a PR that would replace MapDomain).

tias:
Yes, the use of MapDomain can show that an integer is categorical and may not be needed at all when DirectEncoded/mapped, which would require keeping track in the solver object, like int2bool does, what integers are mapped to what Booleans and that could allow eliminating the integer completely.
I think this would benefit linearisation, which is the step before int2bool, so can also help the translation to PB/SAT solvers.

But also definitely yes, this is not for in this PR : )

(actual code review for later)

cpmpy/solvers/pysat.py
tias: ok to remove

A solver would have to add it as a constructor arg, store it, and add it to this specific transformation constructor itself. We barely use solver constructor args at this moment.

cpmpy/solvers/pysat.py
```diff
@@ -156,11 +159,13 @@ def __init__(self, cpm_model=None, subsolver=None):
             # default solver
             subsolver = "glucose4" # something recent...
         elif subsolver.startswith('pysat:'):
-            subsolver = subsolver[6:] # strip 'pysat:'
+            subsolver = ":".split(subsolver)[1] # strip 'pysat:'
```
tias: euh??? subsolver.split(":")[1] perhaps? or am I super confused now

cpmpy/solvers/pysat.py
```diff
@@ -307,11 +340,12 @@ def transform(self, cpm_expr):
         cpm_cons = only_bv_reifies(cpm_cons, csemap=self._csemap)
         cpm_cons = only_implies(cpm_cons, csemap=self._csemap)
         cpm_cons = linearize_constraint(cpm_cons, supported=frozenset({"sum","wsum", "and", "or"}), csemap=self._csemap)  # the core of the MIP-linearization
+        cpm_cons = int2bool(cpm_cons, self.ivarmap, encoding=self.encoding)
```
tias, line 375: could it benefit from the cse_map? (added after this PR started)

cpmpy/solvers/pysat.py
tias: where/how do you print the clauses? or was this for during development? in the latter, its no longer needed now...

tias: it would actually make the pull request cleaner, as more than half of the lines I see below are just from introducing this indirection...

cpmpy/transformations/int2bool.py
tias: you only use this function in 1 place, so it doesn't really matter...

still, this int2bool is supposed to be pedagogical, and its not a regular bitstring that our CS students would expect, but a reversed bitstring, so how about calling the function `_reversed_bitstring(self, d)` and then all is settled? : )

tias (commented):
great. Some minor comments (also to a thread higher up that is not visible in the code review view)

cpmpy/solvers/pysat.py
hbierlee: Good catch, fixed and tested it.

cpmpy/solvers/pysat.py
hbierlee: I thought about this a bit, and I think.. not at the moment. The constraints handled here are (implied) linears; any usefull CSE on them should have been done in linearize (e.g. adding the constraint `2*x+3*y+5*z<=6` twice is already prevented by CSE currently). The type of CSE you can do here is all at the cardinality/pb encoding level, since most encodings effectively add integers together into new ones (e.g. the generalized totalizer will encode the above constraint as `2*x+3*y <= w, w+5*z<=6`, where `w` is a new integer variable; in this case, another constraint could reuse `w` using CSE, and there is some work on this). But this is not something we can influence right now, unless we start making our own encoders.

Another interesting thing of course is that the `ivarmap` acts like CSE (where integer variables are mapped to their encoding / equivalent PB wsum). So `ivarmap` could be replaced by `csemap`, but for now there is no immediate benefit of that.

cpmpy/transformations/int2bool.py
hbierlee: Yep, clarified this further

cpmpy/solvers/pysat.py
hbierlee: Done. The same thing could be said for `add_expr` for every transformed expression, but I clarified that one a bit more, it's a minor change, and resolves some inconsistent indenting in this file, so I suggest we keep it in.

cpmpy/solvers/pysat.py
```diff
@@ -156,11 +158,13 @@ def __init__(self, cpm_model=None, subsolver=None):
             # default solver
             subsolver = "glucose4" # something recent...
         elif subsolver.startswith('pysat:'):
-            subsolver = subsolver[6:] # strip 'pysat:'
+            _, subsolver = subsolver.split(":") # strip 'pysat:'
```
tias: I'm sorry but you are still making the code less safe...
if subsolver is 'pysat:mysolver:funky' then this will return a list of size3!

I don't know what you don't like about the current line... the elif checks for 'startswith('pysat:'), so its very explicit... if you want it more explicit, you can do:
subsolver = subsolver[len('pysat:'):]   or otherwise add (first 6 characters) in the comment... but not this

cpmpy/solvers/pysat.py
tias: OK, no need to change anything then

cpmpy/solvers/pysdd.py
```diff
@@ -186,7 +186,9 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from
                 var._value = None
             return 0
 
-        sddmodels = [x for x in self.pysdd_root.models()]
+
```
tias: that does not seem appropriate in this pull request?

hbierlee:
@tias I think this can be merged to be included in the next release, or we could have another reviewer.

IgnaceBleukx:
I think we still need to enable the transformation for pysdd (and also update `test_constraints.py` accordingly

hbierlee:
For PySDD, I don't believe they support PB constraints directly. We'd have to encode the PB constraint to a BDD/MDD (or perhaps export the one out of Pindakaas at some point).

PR #662, opened 2025-05-02
fix empty clause in gcs

IgnaceBleukx:
As the title suggests: current implemetion causes a segmentation fault, also opened an issue on the GCS repo directly.
But this way our tests go through : )


hbierlee:
Looks good to me! Maybe add a comment with a link to the gcs repo issue.

I wonder how we can remember to remove upstream bug workaround code. Perhaps the easiest is to add an "upstream" issue around for this one.

PR #665, opened 2025-05-05
Optional dependency 'all' solvers

ThomSerg:
Remove the outdated "FULL" optional dependency and add "all", which results in installing all solvers:
```console
pip install -e .[all]
pip install cpmpy[all] # same as : pip install cpmpy[ortools,z3,choco,exact,minizinc,pysat,gurobi,pysdd,gcs,cpo]
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,21 @@ def get_version(rel_path):
 with open("README.md", "r") as readme_file:
     long_description = readme_file.read()
 
+
+solver_dependencies = {
+    "ortools": ["ortools"],
+    "z3": ["z3-solver"],
+    "choco": ["pychoco>=0.2.1"],
+    "exact": ["exact>=2.1.0"],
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -18,6 +18,21 @@ def get_version(rel_path):
 with open("README.md", "r") as readme_file:
     long_description = readme_file.read()
 
+
+solver_dependencies = {
```
IgnaceBleukx, line 22: Can't we load this dynamically from `Solverlookup`? I guess not seeing we also have the version requirements...
__for the future__ we could add a "get_pip_str" method to the solver-interface or add a "MIN_VERSION" attribute so this can be done automatically

IgnaceBleukx (approved):
Nice, but ideally we can do this dynamically (not required for this PR/release)

We DO need to add this to the "adding a solver" documentation page.
After that, fine to merge.

PR #666, opened 2025-05-05
Changelog recovery

ThomSerg:
Bring the changelog back to a file-centric workflow, i.e. update the `changelog.md` and then call
```console
gh release create vX.Y.Z -F changelog.md
```
(see release wiki for more info)

The last few releases have been done though the GitHub interface, resulting in an outdated changelog file. This pull request brings the file back up-to-date.

PR #667, opened 2025-05-05
Changelog template

ThomSerg:
As mentioned in #666, now that we'll again be using the `changelog.md` file for documenting the changes included in each release, it would be nice if we decided on a consistent template.

We've used many names for the same subsections before:
- What's Changed
- What's New
- What's new?
- New:
- ...

Either we create our own template, use the one from [keepachangelog](https://keepachangelog.com/en/1.1.0/) or make some hybrid. But as seen in release 0.9.24, sometimes we might want to break out of a strict template as to for example better highlight our efforts on performance improvements. The format included in this pull request is in the "keepachangelog" format.

Any thoughts /  opinions?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,58 @@
 # Change log
 
+## 0.9.25
+
+### Added
+* **New solver**: IBM CP Optimizer [#576](https://github.com/CPMpy/cpmpy/pull/576)
+* **New global**: `ShortTable` [#469](https://github.com/CPMpy/cpmpy/pull/469)
+* `model.add()` as new default to add constraints to model (instead of `+=`) [#640](https://github.com/CPMpy/cpmpy/pull/640)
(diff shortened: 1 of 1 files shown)


ThomSerg:
Maybe the dropping of Python version 3.7 should be under the label `Deprecated` / `Removed` (if following the "keepachangelog" convention)

IgnaceBleukx:
Regarding the classes we want in the changelog, I think it makes sense to mimic the labels we have on our PRs?
I.e., "Documentation", "Bug", "Enhancement"(can be user experience or performance) or "New solver"

ThomSerg:
Indeed, I want to add some documentation on what template to use, but this PR was for release 0.9.25. So will close this one and later open another with added documentation + a script (used to extract the release info of an individual version and format it for a github release)

PR #652, opened 2025-04-21
Fix tests when `pysat` is installed but `pypblib` is not

hbierlee:
After a fresh install with `pip install -e .[test]`, running `pytest` still fails because of two issues:

- `setuptools` not being installed (which I thought was fixed in #564; should we be able to run the tests without explicitly running `pip install setuptools`?)
- Some small issue with the changed pysat import

This PR fixes that.

PR #668, opened 2025-05-05
Skip non-pysat-specific tests on missing pblib

ThomSerg:
Continuation of #652. Adds a set of useful decorators to aid in applying other decorators to individual tests or to entire test cases. For example, skipping a test if an expected exception occurs. 

As an application, the decorator `@skip_on_missing_pblib()` is proposed (internally makes use of the other decorators). It allows skipping non-pysat-specific tests that fail due to missing optional `pblib` dependency. The test don't actually get skipped, but their failure gets captured and silenced as if it were skipped. This simple decorator prevents a lot of logic that would otherwise be required to skip these tests before (instead of after) running them.

Requires #652 to be merged first.

### DIFF
--- a/tests/test_pysat_cardinality.py
+++ b/tests/test_pysat_cardinality.py
@@ -1,4 +1,3 @@
-import importlib
 import unittest
 import pytest
 import cpmpy as cp 
--- a/tests/test_pysat_cardinality.py
+++ b/tests/test_pysat_cardinality.py
@@ -1,3 +1,4 @@
+import importlib
 import unittest
 import pytest
 import cpmpy as cp 
@@ -6,6 +7,7 @@ from cpmpy.expressions.core import Operator
 from cpmpy.solvers.pysat import CPM_pysat
 from cpmpy.transformations.linearize import only_positive_coefficients
 
+pblib_available = importlib.util.find_spec("pypblib") is not None
 
 SOLVER = "pysat"
 
@@ -134,6 +136,7 @@ class TestCardinality(unittest.TestCase):
         self.assertTrue(ps.solve())
         self.assertGreaterEqual(sum(self.bvs.value()), 2)
 
+    @pytest.mark.skipif(not pblib_available, reason="`pypblib` not installed")
     def test_pysat_card_implied(self):
         b = cp.boolvar()
         x = cp.boolvar(shape=5)
--- a/tests/test_pysat_cardinality.py
+++ b/tests/test_pysat_cardinality.py
@@ -7,7 +7,7 @@ from cpmpy.expressions.core import Operator
 from cpmpy.solvers.pysat import CPM_pysat
 from cpmpy.transformations.linearize import only_positive_coefficients
 
-pblib_available = importlib.util.find_spec("pypblib") is not None
+from utils import skip_on_missing_pblib
 
 SOLVER = "pysat"
 
@@ -136,7 +136,7 @@ class TestCardinality(unittest.TestCase):
         self.assertTrue(ps.solve())
         self.assertGreaterEqual(sum(self.bvs.value()), 2)
 
-    @pytest.mark.skipif(not pblib_available, reason="`pypblib` not installed")
+    @skip_on_missing_pblib()
     def test_pysat_card_implied(self):
         b = cp.boolvar()
         x = cp.boolvar(shape=5)
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,20 +1,40 @@
+import importlib
 import pytest
 from functools import wraps
+import cpmpy as cp
+
+# ---------------------------------------------------------------------------- #
+#                              Generic Decorators                              #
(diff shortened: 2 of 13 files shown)


tests/utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def smart_decorator(method_decorator):
+    """
+    Wraps a method decorator so it can be applied to either:
+    - a function/method: applies the decorator directly
+    - a class: applies the decorator to all test_* methods
+    """
+    def wrapper(obj):
+        if isinstance(obj, type):
```
IgnaceBleukx, line 44: This is a bit criptic and sneaky, can you add a comment to explain what this does exactly?

tests/test_pysat_cardinality.py
```diff
@@ -134,6 +136,7 @@ def test_pysat_different(self):
         self.assertTrue(ps.solve())
         self.assertGreaterEqual(sum(self.bvs.value()), 2)
 
+    @pytest.mark.skipif(not pblib_available, reason="`pypblib` not installed")
```
IgnaceBleukx, line 139: Shouldn't this one also be "skip_on_missing_pb()"?

tests/utils.py
```diff
@@ -0,0 +1,59 @@
...
+            return apply_decorator_to_tests(method_decorator)(obj)
+        elif callable(obj):
+            return method_decorator(obj)
+        else:
+            raise TypeError("smart_decorator can only be used on classes or callables")
+    return wrapper
+
+def skip_on_missing_pblib():
```
IgnaceBleukx, line 52: Add a docstring here saying what it means

IgnaceBleukx (changes requested):
I'm a bit confused, is this PR undoing some changes from #652? 
It seems they're trying to solve the same thing, but differently.
If we have a `skip_on_missing_pblib` maybe it also makes sense to have a `skip_on_missing_pysat_or_pblib`? Now this PR tries to clean things up, but leaves some room for improvement I feel

ThomSerg:
Both pull requests (this one and #652) indeed try to achieve similar goals, but they tackle different subproblems.

In the case of #652, these are tests specifically tailored to PySAT where we know beforehand a missing pblib will cause an exception (due to the presence of PB constraints). By decorating with
```python
@pytest.mark.skipif(not (pysat_available and not pblib_available), reason="`pysat` is not installed" if not pysat_available else "`pypblib` is installed")
```
these tests don't get run when either pysat of pblib is missing.

In this pull request, we're solving the issue of more generic tests with models that after transformation result in PB constraints. When these tests get run on PySAT and pblib is missing, again an exception will occur. Unlike the above setting, here it is more tricky to know if some generic model anywhere in our test set will lead to PB constraints. So as to not clutter the test code too much, I created the `skip_on_missing_pblib` decorator. It simply wraps a try..catch around the testcase and if it raises an error due to missing pblib, we simply ignore its result. So the biggest difference is that the test still gets run (but their failure is ignored). 

I did some experiments, and it is possible to integrate the decorator of #652 into `skip_on_missing_pblib`, we'll just need some optional argument to describe which behavior of the two you want, i.e. "always skip on missing pblib" or "skip only if pblib causes an issue". E.g:

```python
@skip_on_missing_pblib()
def test_we_know_needs_pblib():
    ...
@skip_on_missing_pblib(defer=True)
def generic_test_that_could_fail_on_pblib():
    ...
``` 
Maybe the `defer` argument should be more descriptive like `skip_on_exception_only`

We can't just manually put `@skip_on_missing_pblib()` on all the test, since a lot of them are generated dynamically.

ThomSerg:
With the current implementation #652 is fully integrated in this pull request and would thus no longer be required.

I looked into `skip_on_missing_pysat_or_pblib`, but is seems like this is currently not needed. Instead it is more like:
```python
@skip_on_missing_pysat() # <- does not exist yet
class TestCardinality(unittest.TestCase):
    ...
    @skip_on_missing_pblib()
    def test_pysat_card_implied(self):
        ...
```

So the checks occur at different levels.

---

Branch [skip_on_missing_solver](https://github.com/CPMpy/cpmpy/blob/skip_on_missing_solver/tests/utils.py) contains a possible followup:
```python
@skip_on_missing_solver("<solver_name>")
```
to replace
```python
@pytest.mark.skipif(not CPM_pysat.supported(),
                    reason="PySAT not installed")
```

e.g. `@skip_on_missing_solver("pysat")` for the above missing `@skip_on_missing_pysat()`.

hbierlee:
I'm happy with this decorator, it will allow us to run more tests with solvers where we are not sure whether they support the model or not. With the generic version, we can also remove some of the `if solver (..) : continue` in the future.

Unfortunate we need so much engineering just because of supporting a solver with- and without optional dependency.

IgnaceBleukx:
Ok I see -- so yes this one is fine to merge for me; let's revisit the engineering challenge for making the decorator more generic in the future.

PR #669, opened 2025-05-06
Adding Pumpkin solver

IgnaceBleukx:
Adding the LCG-Pumpkin solver.
WIP and need to update a bunch of tests still


tests/test_solvers.py
```diff
@@ -804,6 +805,35 @@ def test_incremental(self, solver):
         assert x.value()
         assert y.value() + z.value() == 1
 
+
```
tias, line 808: should directly go into master (no further review needed)

cpmpy/transformations/linearize.py
```diff
@@ -561,9 +561,11 @@ def canonical_comparison(lst_of_expr):
             if isinstance(lhs, Comparison) and (is_bool(rhs) or isinstance(rhs, Expression) and rhs.is_bool()):
                 assert cpm_expr.name == "==", "Expected a reification of a comparison here, but got {}".format(cpm_expr.name)
                 lhs = canonical_comparison(lhs)[0]
-            elif is_num(lhs) or isinstance(lhs, _NumVarImpl) or (isinstance(lhs, Operator) and lhs.name in {"sum", "wsum"}):
-                # Bring all vars from rhs to lhs
-                # 1) collect the variables to bring over
+            elif is_num(lhs) or isinstance(lhs, _NumVarImpl) or (isinstance(lhs, Operator) and lhs.name in {"sum", "wsum", "sub"}):
```
tias, line 564: can go into master directly?

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+
+class CPM_pumpkin(SolverInterface):
+    """
+    Interface to Pumpkin's API
+
+    Creates the following attributes (see parent constructor for more):
+
+    - ``pum_solver``: the pumpkin.Model() object
+    Creates the following attributes (see parent constructor for more):
+    - tpl_model: object, Pumpkin's model object
```
tias, line 63: you probably meant to remove this line

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+
+        # initialise the native solver object
+        self.pum_solver = Model() 
+
+        # for objective
+        self._objective = None
+        self.objective_is_min = True
+
+        # initialise everything else and post the constraints/objective
+        super().__init__(name="Pumpkin", cpm_model=cpm_model)
```
tias, line 99: I don't think we do capitals in our solver names

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+
+    @property
+    def native_model(self):
+        """
+            Returns the solver's underlying native model (for direct solver access).
+        """
+        return self.pum_solver
+
+
+    def solve(self, time_limit=None, proof=None, assumptions=None):
```
tias, line 109: so in GCS we already made some decisions for the 'solve' api to proof logging...
there we decided an attribute 'prove' if it should store its proof, and 'verify' if it should be verified... then we also have proof_name and proof_location; I'm not sure these are the best choices, e,g. 'proof=True, proof_file=...' would also make sense, but I do think we should have both solvers have equally named attributes

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+            elif  isinstance(result, OptimisationResult.Unsatisfiable):
+                self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+            elif isinstance(result, OptimisationResult.Unknown):
+                self.cpm_status.exitstatus = ExitStatus.UNKNOWN
+            else:
+                raise ValueError(f"Unknown Pumpkin-result: {result} of type {type(result)}, please report on github...")
+
+        elif assumptions is not None: # check result under assumptions
+            if isinstance(result, SatisfactionUnderAssumptionsResult.Satisfiable):
+                self.cpm_status.exitstatus = ExitStatus.OPTIMAL
```
tias, line 175: I think we agreed that satisfaction would return 'feasible' not optimal

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+                self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+                self._pum_core = result._0
+            elif isinstance(result, SatisfactionUnderAssumptionsResult.Unknown):
+                self.cpm_status.exitstatus = ExitStatus.UNKNOWN
+            else:
+                raise ValueError(f"Unknown Pumpkin-result: {result} of type {type(result)}, please report on github...")
+
+        else: # satisfaction result without assumptions
+            if isinstance(result, SatisfactionResult.Satisfiable):
+                self.cpm_status.exitstatus = ExitStatus.OPTIMAL
```
tias, line 189: likewise

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+            elif isinstance(result, SatisfactionResult.Unknown):
+                self.cpm_status.exitstatus = ExitStatus.UNKNOWN
+            else:
+                raise ValueError(f"Unknown Pumpkin-result: {result} of type {type(result)}, please report on github...")
+
+        # translate solution (of user vars only)
+        has_sol = self._solve_return(self.cpm_status)
+        if has_sol:
+            solution = result._0
+            self.cpm_status.exitstatus = ExitStatus.OPTIMAL
```
tias, line 201: not sure what this is doing here

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+        :param cpm_expr: CPMpy expression, or list thereof
+        :type cpm_expr: Expression or list of Expression
+
+        :return: list of Expression
+        """
+        # apply transformations
+        cpm_cons = toplevel_list(cpm_expr)
+        supported = {"alldifferent", "cumulative", 
+                     "min", "max", "element"}
+        cpm_cons = decompose_in_tree(cpm_cons, supported=supported) # TODO: Pumpkin supports half-reified globals
```
tias, line 294: I'm not sure I understand the TODO comment

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+            if cpm_expr.name == "<":  comp, rhs = Comparator.LessThanOrEqual, rhs - 1
+            if cpm_expr.name == ">":  comp, rhs = Comparator.GreaterThanOrEqual, rhs + 1
+        else:
+            raise ValueError(f"Cannot convert CPMpy expression {cpm_expr} to a predicate")
+
+        var = self._ivars(lhs)
+        return Predicate(var, comp, rhs)
+
+
+    def _ivars(self, cpm_var):
```
tias, line 345: maybe better in singular name (where you then as a special case also accept a list)? also not sure why it is semi-private with the _ivars?

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+
+        :return: self
+        """
+        from pumpkin_py import constraints
+
+        # add new user vars to the set
+        get_variables(cpm_expr_orig, collect=self.user_vars)
+
+        # transform and post the constraints
+        for orig_expr in toplevel_list(cpm_expr_orig, merge_and=True):
```
tias, line 501: transform itself should do the toplevel list? you should directly for cpm_esxpr in self.transform()? or is there a reason not to?

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,531 @@
...
+            for cpm_expr in self.transform(orig_expr):
+                if isinstance(cpm_expr, Operator) and cpm_expr.name == "->": # found implication
+                    bv, subexpr = cpm_expr.args
+                    for cons in self._get_constraint(subexpr):
+                        if isinstance(cons, constraints.Clause):    
+                            raise ValueError("_get_constraint should not return clauses")
+                        self.pum_solver.add_implication(cons, self.solver_var(bv))
+                else:
+                    solver_constraints = self._get_constraint(cpm_expr)
+                    for cons in solver_constraints:
```
tias, line 511: do we care if it returns a clause?

tias (commented):
looking good, some comments and a few questions

cpmpy/solvers/pumpkin.py
IgnaceBleukx: Currently we decompose any nested globals, in the future we should update the decompose_in_tree transformation to extract the ones in positive context to half-reifications.
Then, we can avoid a lot of the decompositions when we're doing e.g., MUS-extraction with assumptions

cpmpy/solvers/pumpkin.py
IgnaceBleukx: Ah yes, so this entire function is a leftover from getting the explanations from the proofs. I had to keep track of the set of user-constraints... will cleanup this function

cpmpy/solvers/pumpkin.py
IgnaceBleukx: If we want to tag the constraints we do; but not in the main CPMpy branch indeed.

cpmpy/solvers/pumpkin.py
IgnaceBleukx: Hmm yeah having both proof-name and proof-location seems a bit redundant. But we shouldn't change the gcs interface in this PR...
So I'll now mimic the gcs interface, and then we change later on if we want.

(Related: probably we can also enable prooflogging in pysat and exact?)

cpmpy/transformations/linearize.py
IgnaceBleukx: Opened a separate PR for this

IgnaceBleukx:
Processed review.

tests/test_solvers.py
```diff
@@ -1,3 +1,4 @@
+import inspect
```
tias, line 1: I don't think you need inspect

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,550 @@
...
+
+        # for objective
+        self._objective = None
+        self.objective_is_min = True
+
+        # initialise everything else and post the constraints/objective
+        super().__init__(name="pumpkin", cpm_model=cpm_model)
+
+    @property
+    def native_model(self):
```
tias, line 111: do we do that for other solvers? there its just in the docs that it is called self.pum_solver, and we don't provide an indirect/abstraction?

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,550 @@
...
+
+            See the 'Adding a new solver' docs on readthedocs for more information.
+
+        :param cpm_expr: CPMpy expression, or list thereof
+        :type cpm_expr: Expression or list of Expression
+
+        :return: list of Expression
+        """
+        # apply transformations
+        cpm_cons = toplevel_list(cpm_expr)
```
tias, line 316: also partial?
also, csemap has since been merged I think

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,550 @@
...
+            if expr.args[1].is_bool():
+                y = self.pum_solver.boolean_as_integer(y)
+            args = [x.scaled(-1 if negate else 1), y.scaled(1 if negate else -1)]
+        else:
+            raise ValueError(f"Unknown expression to convert in sum-arguments: {expr}")
+        return args
+
+    def _get_constraint(self, cpm_expr):
+        """
+            Get a solver's constraint by a supported CPMpy constraint
```
tias: 'by a'? confusing

tias: also, add to the docs "expects a transformed cpm_expr as input, this is a separate function so we can add an indicator variable to it if needed"

tias (commented):
so does this PR depend on the PR of the linearize change? because if we merge this it will also merge the linearize change...

nice. a few minor comments, will be great to add (even if its not on pip yet...)

cpmpy/solvers/pumpkin.py
IgnaceBleukx: We have it for other solvers too, not sure when it was introduced, but already quite some time I think

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,559 @@
...
+        elif isinstance(cpm_expr, Comparison):
+
+            lhs, rhs = cpm_expr.args
+            assert is_num(rhs), "rhs of comparison must be a constant to be a predicate"
+
+            if isinstance(lhs, Operator): # can be sum with single arg
+                if lhs.name == "sum" and len(lhs.args) == 1:
+                    lhs = lhs.args[0]
+                else:
+                    raise ValueError("Lhs of predicate should be a sum with 1 argument") # TODO: also wsum with 1 arg/mul with const?
```
ThomSerg: still a TODO?

cpmpy/solvers/pumpkin.py
```diff
@@ -0,0 +1,559 @@
...
+                y = self.pum_solver.boolean_as_integer(y)
+            args = [x.scaled(-1 if negate else 1), y.scaled(1 if negate else -1)]
+        else:
+            raise ValueError(f"Unknown expression to convert in sum-arguments: {expr}")
+        return args
+
+    def _get_constraint(self, cpm_expr):
+        """
+            Convert a CPMpy expression into a Pumpkin constraint
+            Expects a transformed CPMpy expression, but implemented as a separate function so we can support reification in `add()`
```
ThomSerg: The logic of the below code is implemented in a separate function? Or is the transformed expression implemented in a different function? Reads as if two separate sentences are incorrectly connected together. What is the "but" trying to contradict?

ThomSerg (commented):
Looks good, small comment in some documentation / todo.

cpmpy/solvers/pumpkin.py
IgnaceBleukx: Ah I see the confusion indeed... the 'but' is trying to explain why the logic is not directly in `def add()` BUT in a seperate function instead. Will update

cpmpy/solvers/pumpkin.py
IgnaceBleukx: I didn't find any cases where this could happen in the testsuite, so I'm assuming our transformations already get rid of those cases. I'll implement it nevertheless, just to be sure

IgnaceBleukx:
Should be ready to go now @ThomSerg.

PR #670, opened 2025-05-07
Release 0.9.25

ThomSerg:
Last couple of changes before release:
- mostly the changelog
- 2 small fixes in the test set

For the changelog, a reformat might occur depending on the conclusions of discussion #667, but we can always change that after the fact. I don't think it should be blocking this release. I can however (for now) make the changelog a bit more similar to previous releases. Any thoughts?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -23,6 +23,8 @@
 * Optional dependencies for solvers and tools (setup.py) [#599](https://github.com/CPMpy/cpmpy/pull/599)
 * Documentation for all CPMpy exceptions [#622](https://github.com/CPMpy/cpmpy/pull/622)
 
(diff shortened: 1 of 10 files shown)


changelog.md
```diff
@@ -1,5 +1,67 @@
 # Change log
 
+## 0.9.25
+
+### Added
+* **New solver**: IBM CP Optimizer [#576](https://github.com/CPMpy/cpmpy/pull/576)
+* **New global**: `ShortTable` [#469](https://github.com/CPMpy/cpmpy/pull/469)
+* `model.add()` as new default to add constraints to model (instead of `+=`) [#640](https://github.com/CPMpy/cpmpy/pull/640)
```
IgnaceBleukx, line 8: it's not instead of, right? The previous method still works. So (same behaviour as `+=`

IgnaceBleukx (approved):
Impressive changelog for low-hanging fruit!

PR #671, opened 2025-05-07
Extract release notes from changelog

ThomSerg:
When working on release 0.9.25, I made a collection of tools to simplify future releases. 
These can be seen as "dev tools", i.e. not part of the cpmpy library but definitely useful towards its development.
With this first example, I propose a new subdirectory `/tools` (not to be confused with `/cpmpy/tools`) to start collecting some of these.

---

The tool included in this pull request is meant for generating release notes. When releasing on GitHub using the cli, one can pass in a file to use as markdown description of that release:

```console
gh release create vX.Y.Z -F <file_with_release_notes>
```

This unfortunately will take the contents of the entire file, without filtering on the section of the targeted release. 

In the "Packaging Releasing Publishing" wiki, the following is currently recommended:
```console
gh release create vX.Y.Z -F changelog.md
```
But this will include all changes of all version into the description of the newly added release.

The proposed tool generates the required version-filtered release notes and saves them to a temporary `release-notes.md` file.

To then release:
```console
gh release create vX.Y.Z -F release-notes.md
```

As a bonus, we can now adapt the formatting of the selected changelog section for the purpose of a release. The script already includes:
- remove main title with version number (since this is already included in the GitHub UI)
- upgrade all titles (### -> ##)
- add new title "Release notes"

### DIFF
--- a/tools/extract_release_notes.py
+++ b/tools/extract_release_notes.py
@@ -1,0 +1,85 @@
+"""
+A small CLI tool for gerenating release notes for GitHub.
+
+To create a new CPMpy release, one must call (!! do look at the wiki for more instructions !!):
+
+```console
+gh release create vX.Y.Z -F <file_with_release_notes>
+```
+
+The entire contents of <file_with_release_notes> will then be used as the markdown description of that release.
+Simply passing `changelog.md` will add the entire changelog to the release notes of the newest release.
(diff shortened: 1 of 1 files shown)


ThomSerg:
Changed the directory to `dev/` and added a small readme to collect descriptions of the different dev scripts (more to be added in the future) and other documentation related to development (more internal information that we use, so currently haven't added it to the RTD docs).

PR #672, opened 2025-05-07
remove unnecessary set operations from deletion-based MUS

OrestisLomis:
Small change to MUS computation where assumptions are removed and added back to the to the core set unnecessarily. Seems to matter quite a bit when testing om some toy problems

### DIFF
--- a/cpmpy/tools/explain/mus.py
+++ b/cpmpy/tools/explain/mus.py
@@ -52,11 +52,9 @@ def mus(soft, hard=[], solver="ortools"):
     for c in sorted(core, key=lambda c : -len(get_variables(dmap[c]))):
         if c not in core:
             continue # already removed
-        core.remove(c)
-        if s.solve(assumptions=list(core)) is True:
-            core.add(c)
-        else: # UNSAT, use new solver core (clause set refinement)
-            core = set(s.get_core())
+
+        if not s.solve(assumptions=list(core)):
+            core = set(s.get_core()) # UNSAT, use new solver core (clause set refinement)
 
     return [dmap[avar] for avar in core]
 


cpmpy/tools/explain/mus.py
```diff
@@ -52,11 +52,9 @@ def mus(soft, hard=[], solver="ortools"):
     for c in sorted(core, key=lambda c : -len(get_variables(dmap[c]))):
         if c not in core:
             continue # already removed
-        core.remove(c)
-        if s.solve(assumptions=list(core)) is True:
-            core.add(c)
-        else: # UNSAT, use new solver core (clause set refinement)
-            core = set(s.get_core())
+
+        if not s.solve(assumptions=list(core)):
```
IgnaceBleukx, line 56: But... now you don't do anything with `c`?
So how is this testing whether core - {c} is SAT? It will just do a bunch of SAT calls with the same assumptions...

IgnaceBleukx (changes requested):
I don't think this changes is actually valid? The tests probably work because the solver core we get in the first place is already a MUS?

cpmpy/tools/explain/mus.py
Dimosts: You mean a bunch of UNSAT calls, but yes indeed, c needs to be removed before solve to see if it is sat and to minimize

IgnaceBleukx: Yeah, with SAT-calls I meand solver calls ;)

OrestisLomis: You are completely right.. This doesn't actually work. It can indeed be used as preprocessing though. MUSer2 actually has an option for this, but it is turned off by default and is generally not beneficial from the small experiments that I've already ran.

PR #658, opened 2025-04-29
Fix all-solution `ortools` edge cases

hbierlee:
Two edge cases to fix for `solveAll` on OR-tools:

```python

self = <test_solvers.TestSolvers testMethod=test_ortools_solve_all>

    def test_ortools_solve_all(self):
        x,y,z = cp.intvar(0, 2, shape=3,name="xyz")
        p = cp.boolvar(name="p")
        user_vars = [x, y, z, p]
        sols = []
        cp.Model(p.implies(
            2 * x + 3 * y + 5 * z == 12
        )).solveAll(display=lambda: sols.append(tuple(argvals(user_vars))))
>       assert len(sols) == len(set(sols)), "There should be no duplicate solutions"
E       AssertionError: There should be no duplicate solutions
E       assert 56 == 29
E        +  where 56 = len([(0, 0, 0, False), (0, 0, 0, False), (1, 0, 0, False), (1, 0, 0, False), (1, 1, 0, False), (1, 2, 0, False), ...])
E        +  and   29 = len({(0, 0, 0, False), (0, 0, 1, False), (0, 0, 2, False), (0, 1, 0, False), (0, 1, 1, False), (0, 1, 2, False), ...})
E        +    where {(0, 0, 0, False), (0, 0, 1, False), (0, 0, 2, False), (0, 1, 0, False), (0, 1, 1, False), (0, 1, 2, False), ...} = set([(0, 0, 0, False), (0, 0, 0, False), (1, 0, 0, False), (1, 0, 0, False), (1, 1, 0, False), (1, 2, 0, False), ...])

tests/test_solvers.py:106: AssertionError
____________________________________________________________________________________________ TestSolvers.test_ortools_solve_empty _____________________________________________________________________________________________
[gw5] linux -- Python 3.12.9 /home/hbierlee/Projects/cpmpy/.direnv/python-3.12.9/bin/python

self = <test_solvers.TestSolvers testMethod=test_ortools_solve_empty>

    def test_ortools_solve_empty(self):
        user_vars = []
        sols = []
        cp.Model().solveAll(display=lambda: sols.append(tuple(argvals(user_vars))))
>       assert len(sols) == 1, "An empty CSP (with 0 user variables) should have one solution (with an empty assignment)"
E       AssertionError: An empty CSP (with 0 user variables) should have one solution (with an empty assignment)
E       assert 0 == 1
E        +  where 0 = len([])

tests/test_solvers.py:95: AssertionError
```

Issue #650, opened 2025-04-21
Structured global testing


IgnaceBleukx:
Following on a variety of edge-case issues where global constraints receive unexpected arguments, we should probably figure out a more general way of testing our globals.
E.g., a possible battery of tests would be:

```python
def proper_check(expr):
  for solver in cp.SolverLookup.solvernames():
    assert cp.Model(expr).solve(solver=solver) # check normal setting
    assert expr.value () # check value function
    
    sols, non_sols = set(), set()
    cp.Model(expr).solveAll(solver=solver,display=sols.add(intvar_args.value()))
    cp.Model(~expr).solveAll(solver=solver,display=non_sols.add(intvar_args.value()))
    assert sols.is_disjoint(non_sols)
    assert len(sols) + len(non_sols) == num_assignments
    
    # check reification
    assert cp.Model(expr | ~expr).solveAll(solver=solver) == num_assignments
```

For each global we would then need to write something like this
```python
intvar_args = cp.intvar(...)
expr = FancyNewGlobal(intvar_args)
proper_check(expr)

# mix of var and const
args = intvar_args.tolist()
args[0] = 42
expr = FancyNewGlobal(args)
proper_check(expr)

# only constants
expr = FancyNewGlobal(list(range(...))
proper_check(expr)

# numpy
expr = FancyNewGlobal(np.arange(...))
proper_check(expr)

# boolvars as int
expr = FancyNewGlobal(cp.boolvar(shape=...))
proper_check(expr)

# single argument
expr = FancyNewGlobal([cp.intvar(...)])
proper_check(expr)

# single constant
expr = FancyNewGlobal([1337])
proper_check(expr)
```

It is hard to fully automate these tings as the set of arguments a global expects can be pretty diverse.
Any thoughts on this, or suggestions for added cases in the above are welcome : )

hbierlee:
Oops, wrong button :)

Definitely on board! On the argument side, @PeerJosse has already expanded the fuzztester to generate the correctly typed expression arguments (leading to a lot of these edge cases issues). So that would be a great starting point.

I think that using type-hints may be the most effective way to cover the global/expression arguments. This also has some big additional benefits:

- Runtime type-checking
- Better maintainability (say an argument of some global changes, you just update its type hint and the tests change automatically)
[...]

PR #674, opened 2025-05-08
solveAll returns 1 solution for empty model

### LINKED ISSUE #661: Enumerating solutions returns no solutions for empty CSP
As discussed in #659, the empty CSP should I believe return one solution containing an empty assignment:

```python
tests/test_solvers.py:106: AssertionError
____________________________________________________________________________________________ TestSolvers.test_ortools_solve_empty _____________________________________________________________________________________________
[gw5] linux -- Python 3.12.9 /home/hbierlee/Projects/cpmpy/.direnv/python-3.12.9/bin/python

self = <test_solvers.TestSolvers testMethod=test_ortools_solve_empty>

    def test_ortools_solve_empty(self):
        user_vars = []
        sols = []
        cp.Model().solveAll(display=lambda: sols.append(tuple(argvals(user_vars))))
>       assert len(sols) == 1, "An empty CSP (with 0 user variables) should have one solution (with an empty assignment)"
E       AssertionError: An empty CSP (with 0 user variables) should have one solution (with an empty assignment)
E       assert 0 == 1
E        +  where 0 = len([])

tests/test_solvers.py:95: AssertionError
```

hbierlee:
Goal to fix #661 , superceding the original PR for this, https://github.com/CPMpy/cpmpy/pull/658, because I can't change the base branch in a PR from my fork to the cpmpy/upstream branch.


hbierlee:
I've made the tests more generic, but I'm having some issue pushing atm. Give me a second.

hbierlee:
Resolved the random push issue. So just checking the number of solutions is not good enough for these bugs, we need more, albeit still fairly basic, checks, from which we can resolve the issue. This can be a template for #650 , so I'll document these checks there.

cpmpy/solvers/cpo.py
```diff
@@ -152,11 +152,16 @@ def solve(self, time_limit=None, **kwargs):
...
+        """ 
 
-        # ensure all vars are known to solver
+        # ensure all vars are known to solver        
         self.solver_vars(list(self.user_vars))
 
+        # edge case, empty model, ensure the solver has something to solve
+        if not len(self.user_vars):
+            iv = intvar(1,1)
+            self += iv == 1
```
IgnaceBleukx: We now have `self.add()` which looks a little more natural in these situations.
Also, no real need to make the iv variable itself. It can just be `self.add(intvar(1,1) == 1)`

cpmpy/solvers/gurobi.py
```diff
@@ -478,6 +483,14 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from
             Returns: number of solutions found
         """
 
+        # ensure all vars are known to solver
+        self.solver_vars(list(self.user_vars))
+
+        # edge case, empty model, ensure the solver has something to solve
+        if not len(self.user_vars):
+            iv = intvar(1,1)
+            self += iv == iv
```
IgnaceBleukx: why not `== 1` here too? as in the ortools case?

cpmpy/solvers/minizinc.py
```diff
@@ -93,7 +93,8 @@ class CPM_minizinc(SolverInterface):
     required_version = (2, 8, 2)
 
     @staticmethod
-    def supported():
+    def supported(): 
+        return False
```
IgnaceBleukx: Remove from PR

IgnaceBleukx (changes requested):
Looks good, I would prefer to use `self.add` instead and then some consistency thingies.

Also, did you check with CP-Optimizer if it works there? I know I had some trouble with it for empty models in the past.

cpmpy/solvers/gurobi.py
hbierlee: Seems to also work.

hbierlee:
I've processed the review, and I can confirm the test works with CP-optimizer, and the code looks good to me. So I think it can be merged @IgnaceBleukx

cpmpy/solvers/pysdd.py
```diff
@@ -175,12 +179,16 @@ def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from
...
+        if not len(self.user_vars):
+            self.add(boolvar() == True)
+
         if time_limit is not None:
             raise NotImplementedError("PySDD.solveAll(), time_limit not (yet?) supported")
         if solution_limit is not None:
             raise NotImplementedError("PySDD.solveAll(), solution_limit not (yet?) supported")
 
-        if self.pysdd_root is None:
+        if self.pysdd_root is None or self.pysdd_root.is_false():
```
IgnaceBleukx, line 191: why does this second part need to be removed? Is it related to this PR or a more general cleanup?

IgnaceBleukx (approved):
Great, tests also pass on all installed solvers, so after clarifying this minor comment in `pysdd`, go ahead and merge

hbierlee:
Removed? There's a disjunct added, right? And it is necesary to check if the root is false, otherwise:

```py
tests/test_solvers.py:1009: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
cpmpy/model.py:226: in solveAll
    ret = s.solveAll(display=display,time_limit=time_limit,solution_limit=solution_limit, call_from_model=True, **kwargs)
cpmpy/solvers/pysdd.py:197: in solveAll
    sddmodels = [x for x in self.pysdd_root.models()]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   ValueError: False has no models

pysdd/sdd.pyx:251: ValueError
```

PR #675, opened 2025-05-08
Remove 3.9 syntax from setup.py

ThomSerg:
One of the changes I made to the `setup.py` to support the "all" optional dependency breaks support for python 3.8. 
`a = dict | dict` is apparently syntax only introduced since 3.9, should have done `a = {**dict, **dict}` instead.

Luckily, our GitHub workflow responsible for the release to pip builds the wheels using python 3.9, so the version on pip is not affected.

Came across this issue whilst cleaning up my "dev-tooling", adding the ability to easily run our testsuite against all supported python versions. Will make a pull request with this tooling so that we can catch such issues in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,15 +51,15 @@ setup(
         'numpy>=1.5',
         'setuptools',
     ],
-    extras_require=(
+    extras_require={
         # Solvers
-        solver_dependencies | {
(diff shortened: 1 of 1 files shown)


IgnaceBleukx (approved):
Good catch, fine to merge

Issue #606, opened 2025-04-05
On interval variables and optional interval variables


tias:
So Interval variables and Optional interval variables are the only features of state-of-the-art CP solvers that our library does not yet support...

We don't want to add them as primitive variables because then they can be used in any expression, which requires working out the full semantics and code-wise this would be a lot of special casing etc.

We created #121 DirectVariable prototype for them, to generically support ortools/CP's interval variables, and we actually use it in some projects with companies. But it is also quite some technical debt should we include it in the main code.

But because we do want them for flexible jobshop (e.g. README) and work we do, I gave it some more thought:

So OrTools has 2 types of IntervalVariables (well, also can create versions with implicit 'end', but lets ignore that for now)
* NewIntervalVar
* NewOptionalIntervalVar

And those are used in just 3 constraints of OrTools:
* AddCumulative
* AddNoOverlap -- special case of Cumulative where demand and capacity=1
* AddNoOverlap2D -- special case of 2D NoOverlap

Now because these are all special cases of each other, this means that they will never appear together in a model.

So an IntervalVar is not really a variable in the expression-tree sense, its just a construct to post either a Cumul, NoOv or NoOv2D over...

So if its only to the purpose of these constraints, we can just create constraints for each combo; that's just 4 combo's if we ignore the binpacking NoOv2D for now:

* Cumulative(start, dur, end, demand, capa)
* NoOverlap(start, dur, end)
* CumulativeOptional(start, dur, end, demand, capa, option)
* NoOverlap(start, dur, end, option)

And of those 4, we already have the first two as globals... (although only Cumulative is in the docs, a different TODO)

Now this does ignore one thing: creating an IntervalVar also enforces it to be an interval, e.g. start+dur=end. For solvers like OrTools that have IntervalVars, this is implicit when using our Cumulative (as we create the IntervalVar behind the scenes), but for a solver that decomposes this is not implicit...

So that means we also either need two more constraints:

* IsInterval(start, dur, end)
* IsOptionalInterval(start, dur, end, option)

or, we need to change the signature of the 4 constraints above to include an `enforceInterval=True` argument...

(the latter will avoid users using a solver that supports IntervalVars to still manually post start+dur=end; the former will mean that solvers with IntervalVars will have twice the interval variables created by our model, once in IsInterval, once in Cumulative, which I hope will be very little overhead in practice)

---

Does this reasoning make sense?

Do we want to add the 2Optional globals and hence abort the DirectVar approach?

[...]

PR #676, opened 2025-05-13
Optional intervals for Cumulative and NoOverlap

IgnaceBleukx:
Implementation for optional intervals for the NoOveralp and Cumulative constraints.

Based on discussion in #606.

Currently only implemented for OR-Tools, the tests work.
Other solvers supporting optional variables are CPO and MininZinc, but the API is somewhat different from the normal scheduling constraints for those solvers.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -112,8 +112,10 @@
         InDomain
         Xor
         Cumulative
+        CumulativeOptional
         Precedence
         NoOverlap
+        NoOverlapOptional
         GlobalCardinalityCount
         Increasing
         Decreasing
@@ -662,6 +664,86 @@ class Cumulative(GlobalConstraint):
 
         return True
 
+class CumulativeOptional(GlobalConstraint):
+    """
+        Generalization of the Cumulative constraint which allows for optional tasks.
+        A task is only scheduled if the corresponing is_present variable is set to True.
+    """
+
+    def __init__(self, start, duration, end, demand, capacity, is_present):
+        assert is_any_list(start), "start should be a list"
+        assert is_any_list(duration), "duration should be a list"
+        assert is_any_list(end), "end should be a list"
+
+        start = flatlist(start)
+        duration = flatlist(duration)
+        end = flatlist(end)
+        is_present = [cp.BoolVal(x) if is_bool(x) else x for x in flatlist(is_present)] # normalize
+        assert len(start) == len(duration) == len(end) == len(is_present), "Start, duration, end and is_present should have equal length"
+        n_jobs = len(start)
+
+        for lb in get_bounds(duration)[0]:
+            if lb < 0:
+                raise TypeError("Durations should be non-negative")
+
+        if is_any_list(demand):
+            demand = flatlist(demand)
+            assert len(demand) == n_jobs, "Demand should be supplied for each task or be single constant"
+        else: # constant demand
+            demand = [demand] * n_jobs
+
+        super().__init__("cumulative_optional", [start, duration, end, demand, capacity, is_present])
+
+    def decompose(self):
+        """
+            Time-resource decomposition from:
+            Schutt, Andreas, et al. "Why cumulative decomposition is not as bad as it sounds."
+            International Conference on Principles and Practice of Constraint Programming. Springer, Berlin, Heidelberg, 2009.
+        """
+
+        arr_args = (cpm_array(arg) if is_any_list(arg) else arg for arg in self.args)
+        start, duration, end, demand, capacity, is_present = arr_args
+        cons = []
+
+        # set duration of tasks
+        for t in range(len(start)):
+            cons += [is_present[t].implies(start[t] + duration[t] == end[t])]
+
+        # demand doesn't exceed capacity
+        lb, ub = min(get_bounds(start)[0]), max(get_bounds(end)[1])
+        for t in range(lb,ub+1):
+            demand_at_t = 0
+            for job in range(len(start)):
+                if is_num(demand):
+                    demand_at_t += demand * ((start[job] <= t) & (t < end[job]) & is_present[job])
+                else:
+                    demand_at_t += demand[job] * ((start[job] <= t) & (t < end[job]) & is_present[job])
+
+            cons += [demand_at_t <= capacity]
+
+        return cons, []
+    
+    def value(self):
+        arg_vals = [np.array(argvals(arg)) if is_any_list(arg)
+                    else argval(arg) for arg in self.args]
+
+        if any(a is None for a in arg_vals):
+            return None
+
+        # start, dur, end are np arrays
+        start, dur, end, demand, capacity, is_present = arg_vals
+        # start and end seperated by duration, if the tasks are present
+        if not (~is_present | (start + dur == end)).all():
+            return False
+
+        # demand of present tasks doesn't exceed capacity
+        lb, ub = min(start), max(end)
+        for t in range(lb, ub+1):
+            if capacity < sum(demand * ((start <= t) & (t < end) & is_present)):
+                return False
+
+        return True
+
 
 class Precedence(GlobalConstraint):
     """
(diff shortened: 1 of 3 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -1042,15 +1044,33 @@ def decompose(self, how:str="auto") -> tuple[Sequence[Expression], Sequence[Expr
         if how not in ["time", "task", "auto"]:
             raise ValueError(f"how can only be time, task, or auto (default), but got {how}")
 
-        start, duration, end, demand, capacity = self.args
+        start, *args = self.args
```
tias: not a good pattern, do `start = self.args[0] instead`

cpmpy/expressions/globalconstraints.py
```diff
@@ -1158,6 +1168,186 @@ def __repr__(self) -> str:
...
+        - demand >= 0
+        - start + duration == end
+
+        If the task is not present, the constraint does not enforce any of the above.
+
+        Equivalent to :class:`~cpmpy.expressions.globalconstraints.NoOverlapOptional` when demand and capacity are equal to 1.
+        Supports both varying demand across tasks or equal demand for all jobs.
+    """
+
+    def __init__(self, start: Sequence[Expression],
```
tias: assuming this lands after #864 is in master, please bring in sync with the typing of then-master

cpmpy/expressions/globalconstraints.py
```diff
@@ -1158,6 +1168,186 @@ def __repr__(self) -> str:
...
+        Support time-based decomposition or task-based decomposition.
+        By default, we heuristically select the best decomposition based on the number of tasks and the horizon.
+
+        Arguments:
+            how (str): how the cumulative constraint should be decomposed, can be "time", "task", or "auto" (default)
+
+        Returns:
+            tuple[Sequence[Expression], Sequence[Expression]]: A tuple containing the constraints representing the constraint value and the defining constraints
+        """
+        return Cumulative.decompose(self, how) # re-use heuristic selection of decomposition
```
tias: this is highly unusual... you are passing yourself as the self to an external (non-superclass) class.

I don't think we should do that...

We should either subclass Cumulative (and call its constructor and then overwrite the args, which is a bit unusual, but it would allows us to share the constructor checks), or we just copy-paste its decompose heuristic here.

cpmpy/expressions/globalconstraints.py
```diff
@@ -1158,6 +1168,186 @@ def __repr__(self) -> str:
...
+        return True
+
+    def __repr__(self) -> str:
+        """
+        Returns:
+            str: String representation of the cumulative constraint
+        """
+        start, dur, end, demand, capacity, is_present= self.args
+        if end is None:
+            return f"CumulativeOptional({start}, {dur}, {demand}, {capacity}, {is_present})"
```
tias: or should we put an explicit None for end? I assume you thought about it and that it is a conscious choice

cpmpy/expressions/globalconstraints.py
```diff
@@ -1211,6 +1401,7 @@ def value(self) -> Optional[bool]:
         Returns:
             Optional[bool]: True if the global constraint is satisfied, False otherwise, or None if any argument is not assigned
         """
+        start, dur, end = self.args
```
tias: this seems redundant, should be cut?

cpmpy/expressions/globalconstraints.py
```diff
@@ -1240,6 +1431,99 @@ def __repr__(self) -> str:
...
+        if len(start) != len(duration):
+            raise ValueError("Start and duration should have equal length")
+        if len(start) != len(is_present):
+            raise ValueError("Start and is_present should have equal length")
+        if end is not None and len(start) != len(end):
+            raise ValueError(f"Start and end should have equal length, but got {len(start)} and {len(end)}")
+        
+        super().__init__("no_overlap_optional", [start, duration, end, is_present])
+
+    def decompose(self):
```
tias: type missitng, should be tuple[list[Expression],list[Expression]] (its not in the others, that is to be fixed... might as well do it immediately well here)

also subsequent function has types missing

cpmpy/expressions/utils.py
```diff
@@ -215,20 +215,27 @@ def implies(expr, other):
 
 # Specific stuff for scheduling constraints
 
-def get_nonneg_args(args):
+def get_nonneg_args(args, condition=None):
```
tias: since you are touching it, please document what it returns, ideally typed

cpmpy/solvers/ortools.py
```diff
@@ -606,6 +606,25 @@ def _post_constraint(self, cpm_expr, reifiable=False):
...
                 return self.ort_model.AddCumulative(intervals, demand, cap)
+            elif cpm_expr.name == "cumulative_optional":
+                start, dur, end, demand, cap, is_present = cpm_expr.args
+                # ensure duration is non-negative
+                dur, dur_cons = get_nonneg_args(dur, is_present)
+                self.add(dur_cons)
+
+                if end is None: # need to make the end-variables ourself
+                    end = [intvar(*get_bounds(s+d)) for s,d in zip(start, dur)]
+                    self.add([implies(p, s + d == e) for s,d,e,p in zip(start, dur, end,is_present)])
```
tias: this is highly unlikely to be needed... when you create an (optional)intervalvar given starts/durs/end, I am pretty sure that it is ortools who will maintain this relation for you in its global!

cpmpy/solvers/ortools.py
```diff
@@ -619,6 +638,20 @@ def _post_constraint(self, cpm_expr, reifiable=False):
...
 
                 return self.ort_model.AddNoOverlap(intervals)
+            elif cpm_expr.name == "no_overlap_optional":
+                start, dur, end, is_present = cpm_expr.args
+                dur, dur_cons = get_nonneg_args(dur, is_present)
+                self.add(dur_cons)
+
+                if end is None: # need to make the end-variables ourself
+                    end = [intvar(*get_bounds(s+d)) for s,d in zip(start, dur)]
+                    self.add([s + d == e for s,d,e in zip(start, dur, end)])
```
tias: same for this s+d==e
not also that you were posting it unconditional here (which should be fine for most cases)

tias (commented):
great, still a few more things to check but will be great to have this

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: I don't have a strong opinion on it; for me the None just looked a bit weird. Fine with either -- I've removed it for now

cpmpy/solvers/ortools.py
IgnaceBleukx: You're right, it is fixed now

cpmpy/solvers/cpo.py
```diff
@@ -545,39 +545,44 @@ def _cpo_expr(self, cpm_con):
...
+
+                # usage constraints
+                height, height_cons = get_nonneg_args(height, is_present)
+                cons += self._cpo_expr(height_cons)
+
                 total_usage = []
-                for s, d, e, h in zip(start, dur, end, height):
-                    task, task_cons = self._make_task(s, d, e)
-                    cons += task_cons
+                for i, (task, height) in enumerate(zip(tasks, height)):
```
tias: that looks like a bad idea, `for height in height`... the one in zip should be 'heights'? I find it hard to infer what height/heights is...

its demand, and we made it a list in the constructor, so I guess its list[ExprLike]? so the one in zip should be 'hights'? or maybe just demand_list actually...

cpmpy/expressions/utils.py
```diff
@@ -215,20 +221,27 @@ def implies(expr, other):
 
 # Specific stuff for scheduling constraints
 
-def get_nonneg_args(args):
+def get_nonneg_args(args:ListLike[ExprLike], condition:Optional[ListLike[ExprLike]]=None) -> tuple[list[ExprLike], list[ExprLike]] :
```
tias: on second thought, I know I asked you originally, but I think we should leave this one untyped... ideally in the near future Inside Expressions everything is stored as a list, so then this function should not take 'ListLike' but just 'list'; and I am unsure if it should take ExprLike or Expression... so yeah, maybe keep the doc but remove the exact types her for now?

PR #678, opened 2025-05-15
Update docs and error in canonical comparsion

IgnaceBleukx:
The canonical comparison transformation implicitly assumed the input is flat.
Updated docs and added more explicit error messages.

### DIFF
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -63,10 +63,10 @@ from .normalize import toplevel_list
 from .. import Abs
 from ..exceptions import TransformationNotImplementedError
 
-from ..expressions.core import Comparison, Operator, BoolVal
+from ..expressions.core import Comparison, Expression, Operator, BoolVal
 from ..expressions.globalconstraints import GlobalConstraint, DirectConstraint
 from ..expressions.globalfunctions import GlobalFunction
-from ..expressions.utils import is_num, eval_comparison, get_bounds, is_true_cst, is_false_cst
+from ..expressions.utils import is_bool, is_num, eval_comparison, get_bounds, is_true_cst, is_false_cst
 
 from ..expressions.variables import _BoolVarImpl, boolvar, NegBoolView, _NumVarImpl, intvar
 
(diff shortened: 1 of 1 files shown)


cpmpy/transformations/linearize.py
```diff
@@ -550,7 +558,8 @@ def canonical_comparison(lst_of_expr):
 
         elif isinstance(cpm_expr, Comparison):
             lhs, rhs = cpm_expr.args
-            if isinstance(lhs, Comparison) and cpm_expr.name == "==":  # reification of comparison
+            if isinstance(lhs, Comparison) and (is_bool(rhs) or isinstance(rhs, Expression) and rhs.is_bool()):
```
hbierlee, line 561: So we need both an `is_bool(rhs)` en `rhs.is_bool()`? There is some subtle difference between the two?

cpmpy/transformations/linearize.py
```diff
@@ -550,7 +558,8 @@ def canonical_comparison(lst_of_expr):
...
+                assert cpm_expr.name == "==", "Expected a reification of a comparison here, but got {}".format(cpm_expr.name)
```
hbierlee, line 562: so if the comparison has bool on both side, we expect this to be a full reification. I'm a bit worried about inputs of linear constraints having PB terms?

- Is `var bool >= var bool` handled by this function?
- What if we give this `var bool = var bool`? We see this as a reified constraint as well, or a linear to be canonicalized? I guess we depend on this having the same outcome?

hbierlee (commented):
some questions more than comments, probably not requiring changes. Maybe we need to add a test for this error this change is fixing?

cpmpy/transformations/linearize.py
IgnaceBleukx: Yes, so `is_bool` checks whether it is a Boolean constant (also non-cpmpy ones). `.is_bool()` checks whether a CPMpy expression is of Boolean type.

hbierlee:
Ok looks good to merge to me then!

PR #679, opened 2025-05-16
CSE with normalized expressions

IgnaceBleukx:
Proposal for a common subexpression elimination implementation.

Still in draft for now, but happy to discuss.
- Each expression is "normalized" in the sense that in order to compare to expressions we don't take into account the order of arguments if not necessary (e.g., AllDiff, sum, or, ...)
- To store keys, I used an ordinary dict, with the __bool__ override from PR #660.
- Only some of the solvers use the `expr_dict` argument for now

Initial testing on XCP3 is somewhat promising for linear solvers with lots of decompositions, for OR-Tools the number of expressions/variables only decreases slightly.

Performance-wise, we could still improve the `__eq__` implementation to return `Boolval(True)` if `self` and `other` have the same memory id. This should be safe to do as the same expression will always equal itself.
(We might get in trouble if a user writes `expr == expr` as the only constraint to a solver tho, as the variables in them will not have a value given by the solver).
So for now I have not implemented that.

Other performance gains could be made by cashing the result of `get_repr()`, but the same issues as hash-cashing discussed in #586 also apply here. (Maybe for XCSP3 we can get away with hashing them anyway, but in generally certainly not).

Thought and comments on this are very welcome.


tias:
For now, just thoughts:

I think your --bool--() realization is excellent. With Wout we brainstormed about ways to avoid needing to implement a custom dict/set, but we thought that the Comparison('==') object itself was evaluated to true. But now you realized that actually --bool--() is called on it before, and that if you do modeling with CPMpy, you never call '--bool--()' on it, so we only need to highjack that. Great!

As to the get_repr()... so this will actually build these objects in memory; and python is slow at building objects in memory. This will be copying a lot of things... So let's think, why do we need the repr?

1) to do the 'true' comparison recursively
2) to compute the hash fastly

Now both don't really need the objects to be materialized; 'true' comparison needs recursive pairwise comparisons (incl over lists); hash needs recursive hashing (incl over lists and np arrays).

Could we do away with get_repr() and instead compute the pairwise comparisons directly in the 'bool' function?

(I'll sleep on it too)

tias:
OK, so I ran on XCSP3 2024 COP, just the time until transforming each of the models:
Grand total for ort_base: 109.77 minutes
Grand total for ort_cse_repr: 84.05 minutes
Grand total for ort_cse: 85.62 minutes

So I think that the engineering maintenance of doing 'repr' in every object (instead of just with str) is not worth it at this point; while CSE is worth getting in asap.

(the next potential target might be getting the hash better/faster, rather than the equality check on (few) hash collisions)

cpmpy/transformations/flatten_model.py
```diff
@@ -322,12 +324,18 @@ def __is_flat_var_or_list(arg):
...
            is_star(arg)
 
-def get_or_make_var(expr):
+def get_or_make_var(expr, expr_dict=None):
     """
         Must return a variable, and list of flat normal constraints
         Determines whether this is a Boolean or Integer variable and returns
         the equivalent of: (var, normalize(expr) == var)
     """
+    assert expr_dict is not None
```
tias, line 333: This is making tests fail I think.

So solvers will always give an expr_dict, but in testing we might not. If there is no expr_dict, I think its fine... then we simply don't check in it, nor write to it, so the expr_dict (maybe better called 'cse_map'?) is optional

that also means you don't have to create the expr_dict in any of the functions, in case it is none...

it might even be safer: if the user is not aware that there is some cse_map created and maintained behind the scenes, and they see some expressions not translated, they might be confused.

(for this reason, once cse is active, we have to make sure we tell people that they should 'test' the transform on a 'new' solver instance, because of you do s = SolverLookup.get(), s.transform(m.cons), s += m.cons, s.solve() you are going to be in trouble, because the s += cons is not going to re-post the cse expressions. So testing transform should be: SolverLookup.get(...).transform(m.cons) and then solving should be s = solverlookup.get(..., model), s.solve()

IgnaceBleukx: Ah yes, this is a leftover from my testing... I wanted to make sure I passed expr_dict everywhere so I put an assert in each transformation that called `intvar` or `boolvar`; I'll remove it.

In the current implementation, we would still post all of the new constraints, it's just that during flattening/decomposing, we won't introduce auxiliary variables which have the same meaning.
We want to check before posting, if `cpm_expr` is in `expr_dict` already I think. (Although that might trigger falst-positives... in case of `bv == cpm_expr` is somewhere in the tree..)

Ah, no, the correct thing to do would be to add `cpm_expr` before posting to `expr_dict`, with the value `Boolval(True)`. That should clean things up.

IgnaceBleukx:
> So I think that the engineering maintenance of doing 'repr' in every object (instead of just with str) is not worth it at this > point; while CSE is worth getting in asap.
>
> (the next potential target might be getting the hash better/faster, rather than the equality check on (few) hash collisions)

I see, did you also test how many expressions/constraints remain after? Would be interesting to see how much the normalization is triggered compared to just using the string-representation (will do some testing today either way, including registering how many and which hash-collisions actually occur).

Regarding faster hashing, the usual suspect is to memorize the hashing function, but we need to have a way of notifying the parent expression, if the args of a subexpression change.
Alternatively, instead of creating the object is memory, we could make the hash-function itself recursive, where each `__hash__` actually returns a hash...

E.g., the hash-computation of `AllDifferent(x,y,z)` would be:
```
hash(("AllDifferent", (hash(x), hash(y), hash(z)))
```
I would expect hashing of tupes containing integers to be very fast.

cpmpy/solvers/TEMPLATE.py
```diff
@@ -397,6 +398,8 @@ def add(self, cpm_expr_orig):
             else:
                 raise NotImplementedError("TEMPLATE: constraint not (yet) supported", cpm_expr)
 
+            self.expr_dict[cpm_expr] = BoolVal(True) # constraint is now always true, no need to post it again
```
tias, line 401: now I'm confused...

the expr_dict is a cse_map that maps a constraint, to a decision variable representing that expression; right?

so you don't need to every overwrite it? it will just each time return that decision variable...

I might have confused you with my remark about testing transform, but the way it was was the way it should be I think... its just a map from cons to vars, so we don't need to create that equality again...

IgnaceBleukx: So here I save the constraint mapping to "True" as we post the constraint to the solver.
Hence, anytime the constraint is used later-on as a subexpression to another constraint, we can safely replace it with True

IgnaceBleukx:
Yes indeed, we can leave out any hashing-optmizations from this PR for now. I'll revert it to comparison of strings before it is review-ready; initial testing does seem to hint this is quite bit faster compared to hashing the string of the full expr tree tho, so something to keep in mind for the follow up PR : )

cpmpy/solvers/TEMPLATE.py
IgnaceBleukx: With this change I'm indeed also putting the top-level constraints into the dict, without that addition in the solver-interfaces, it is indeed a mapping from sub-expressions to auxiliary variables.

In the case of Boolean sub-expressions, it will be a Boolean variable so we introduce a double-reification; in the case of Numerical expressions, we just introduce an equality constraint.

Maybe I'm taking this CSE PR too far, will rework it to be a minimal CSE implementation and add some proper testing which will also make it more clear what types of expressions we are "caching" in the dict in what transformations.

IgnaceBleukx:
Cleaned up the changes compared to master, to "just" be CSE.
Will experiment a bit with different hashes later on

tias (changes requested):
Great, getting close.

2 more things:

1) the name... shouldn't we call it 'cse_map'? yes its a dict of expressions, but that does not reveal its purpose... cse_map (or _csemap, we also have _varmap in the solvers) would be more explicit?

2) there is an additional opportunity for cse, e.g. a user might write the toplevel
var = expr
and yet later might wrote some_constraint(expr); they intended to reuse 'var' but forgot/didnt.

I'm not sure where, but if somwhere (in flatten?) we would see an '==' between a variable and an expression, then we can feed that expression into the csemap; likewise, if the epxression is already in the csemap, we can replace the == to be between two variables.

cpmpy/transformations/linearize.py
```diff
@@ -273,7 +273,8 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False):
 
             elif isinstance(lhs, GlobalFunction) and lhs.name == "abs" and "abs" not in supported:
                 if cpm_expr.name != "==": # TODO: remove this restriction, requires comparison flipping
-                    newvar = intvar(*get_bounds(lhs))
+                    newvar, newcons = get_or_make_var(lhs, expr_dict=expr_dict)
+                    newlist += newcons
                     newlist += linearize_constraint([lhs == newvar])
```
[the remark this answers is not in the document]
IgnaceBleukx: Aha yes, idd... And we should call linearize here with the proper args too...

IgnaceBleukx:
I much prefer `csemap` too, changed it everywhere and made the attribute `self._csemap` as you suggested.

For the `var == expr` optimization: yes, that would indeed be nice. But we cannot just do it in flatten_constraint, because we have to know if the `var == expr` was posted toplevel.
So, we have to do it somewhere where we know we are toplevel.

The only two options for this are either `toplevel_list` or the solver-interface itself.
`toplevel list` is nice because then we don't have to think about it in the solver interface, but dangerous as we call toplevel list from everywhere in the code base

We could also add a new transformation for it, but that seems overkill

tias (approved):
Thanks for the diligent changes, looks ready to go.

As to adding the equalities, I notice that the first transformation that has the csemap is decompose_in_tree, and that this transformation is top-level aware.
So, I think doing it there is the least intrusive way, and still fairly clean imho.

If you have doubts, you can merge this as is without.

IgnaceBleukx:
Yes, I tried adding it there as well,but things get tricky again because we have to keep track which constraint actually was the `var == expr` one.
E.g., if we just add `csemap[expr] = var` in decompose_in_tree, and later on, in flatten, we want to flatten `expr`.
Then it will see "ah but I already have made a variable for `expr`, let's just use that one.

So then the constraint will be simplified to `var == var`... So we need to do some more bookkeeping to get that right.

I will merge this as is; I believe it's a very useful contribution in it's current form already : )

PR #680, opened 2025-05-20
only_implies: avoid retransform if no subexpressions

tias:
gives noticable speed-up on the slower xcsp3 instances, I hope it passes the tests...


IgnaceBleukx (approved):
Great, this should avoid a lot of redundant recursive calls. For some XCSP instances the main time was indeed spent in this transformation

PR #681, opened 2025-05-20
CPO zero-duration task in cumulative 

ThomSerg:
Posting a `cumulative` constraint with a task of zero duration currently defaults to UNSAT for cpo. This doesn't match OR-Tools' behavior and what is expected in the XCSP3 specification.

The culprit is the following constraint:

```python
cpo_h == dom.height_at_start(task, task_height)
```

A zero-duration task will always have a height of 0 at its start, conflicting with the enforced `cpo_h`.

This pull request proposes an alternative set of constraints for this special setting. It reduces to a single constraint:
```python
cpo_s == cpo_e
```

No constraints on the height / capacity needed for this task.


tias:
nice investigation and fix

PR #682, opened 2025-05-20
CPO solution callback

ThomSerg:
Adds `solution_callback` to CPO. Is not yet used for `solveAll` due to its implementation not using cpo's native `start_search`.
To be used for XCSP3.


tias (approved):
Nice. Fixed a typo, otherwise looks good

PR #683, opened 2025-05-21
Fix eval comparison

IgnaceBleukx:
The `__bool__` optimization opened up a new rabbit hole.

The root cause was the following snippet:"
```python
import cpmpy as cp

idx = cp.intvar(0,1, name="idx")
arr = cp.cpm_array([100,200])
y = cp.intvar(0,1000)

lookup = arr[idx]
decomp = lookup.decompose_comparison("==", y)
print(decomp)
>> ([not([idx == 0]), not([idx == 1]), idx >= 0, idx <= 1], [])
```

Interenally, we rely on `eval_comparison` but this goes wrong when comparing with numpy integer.
E.g., `np.int64(100) == x` returns `False`.

I patched it in this PR, but we should have a look at how to do this more properly (e.g., in #625 ?)

Should be merged quite soon (tutorial currently fails to run)

### DIFF
--- a/cpmpy/expressions/utils.py
+++ b/cpmpy/expressions/utils.py
@@ -157,6 +157,11 @@ def eval_comparison(str_op, lhs, rhs):
 
         Especially useful in decomposition and transformation functions that already involve a comparison.
     """
+    if isinstance(lhs, np.integer):
+        lhs = int(lhs)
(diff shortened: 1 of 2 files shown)


tias (approved):
good; as bools are also valid integers, maybe add np.bool as well?

PR #686, opened 2025-05-21
support for sub in canonical_comparison

IgnaceBleukx:
As the title suggests, enables support for canonicalisation of expressions such as `x - y == z` which will now become `1*x + -1*y +-1*z` == 0

Uncovered during testing of Pumpkin solver (cherry picked from that branch)


Dimosts:
I am confused. This should be converted to a wsum in flatten, no? 

I just tested this in choco, and it is converted after flatten, so sub never arrives to canonical_comparison

IgnaceBleukx:
Yes, but sometimes we want to call the transformation without flattening. I had a case in the Pumpkin solver where this was done for example.

Dimosts (approved):
Normally sub should be converted to wsum in flatten. But indeed we may went to use it without flatten, and the change is simple

PR #691, opened 2025-06-16
Solution hinting in Gurobi

IgnaceBleukx:
Gurobi also supports solution hinting, similar to Exact, PySAT and OR-Tools.

This small PR enables this feature in our interface too.
Unfortunately, it is quite tricky to test as the models we currently have in the testsuite are simply too easy to solve...

Should we add more difficult models to further test this?

### DIFF
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -483,9 +483,6 @@ class CPM_gurobi(SolverInterface):
         for cpm_var, val in zip(cpm_vars, vals):
             self.solver_var(cpm_var).setAttr("VarHintVal", val)
 
-        # disable update, model will be updated when solve() is called
-        # self.grb_model.update()
-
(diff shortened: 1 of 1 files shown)


cpmpy/solvers/gurobi.py
```diff
@@ -463,6 +463,29 @@ def add(self, cpm_expr_orig):
...
+
+            solver.solver_var(cpm_var).setAttr("VarHintPri", <priority>)
+
+        :param cpm_vars: list of CPMpy variables
+        :param vals: list of (corresponding) values for the variables
+        """
+        for cpm_var, val in zip(cpm_vars, vals):
+            self.solver_var(cpm_var).setAttr("VarHintVal", val)
+
+        # disable update, model will be updated when solve() is called
```
ThomSerg: Can this be removed?

ThomSerg (commented):
Looks good

PR #694, opened 2025-06-18
Refactor and update Cumulative and NoOverlap constraints

IgnaceBleukx:
Proposal to omit the end variables in scheduling constraints such as Cumulative and NoOverlap.

Makes `end` None by default, and adapts solver behaviour accordingly.
It was not possible to make the "dummy" end variables in the constructor, as then they would be added to `user_vars` in the solver interface, which is incorrect.
Most solvers don't actually need the `end` variables in their interface either, except for OR-Tools.

Added a helper-function `get_end_vars` to create these dummy end variables.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -762,7 +762,7 @@ class Cumulative(GlobalConstraint):
                 if t != j:
                     demand_at_start_of_t += [demand[j] * ((start[j] <= start[t]) & (end[j] > start[t]))]
 
-            cons += [demand[t] + sum(demand_at_start_of_t) <= capacity]
+            cons += [(demand[t] + sum(demand_at_start_of_t)) <= capacity]
 
         return cons, []
 
@@ -788,7 +788,7 @@ class Cumulative(GlobalConstraint):
         lbs, ubs = get_bounds(start)
         lb, ub = min(lbs), max(ubs)
         for t in range(lb,ub+1):
-            cons += [cp.sum(d * ((s <= t) & (e > t)) <= capacity for s,e,d in zip(start, end, demand))]
+            cons += [cp.sum(d * ((s <= t) & (e > t)) for s,e,d in zip(start, end, demand)) <= capacity]
 
         return cons, []
 


cpmpy/expressions/globalconstraints.py
```diff
@@ -704,38 +716,43 @@ def decompose(self):
...
+        lbs, ubs = get_bounds(start)
+        lb, ub = min(lbs), max(ubs)
         for t in range(lb,ub+1):
             demand_at_t = 0
             for job in range(len(start)):
-                if is_num(demand):
-                    demand_at_t += demand * ((start[job] <= t) & (t < end[job]))
-                else:
-                    demand_at_t += demand[job] * ((start[job] <= t) & (t < end[job]))
+                demand_at_t += demand[job] * ((start[job] <= t) & (start[job] + duration[job] > t))
```
Dimosts: Why remove the above? It covers the case that the demand is given as a single number

cpmpy/expressions/globalconstraints.py
```diff
@@ -669,10 +669,16 @@ class Cumulative(GlobalConstraint):
...
+    def __init__(self, start, duration, end=None, demand=None, capacity=None):
+        if end is None:
+            end = [None] * len(start)
+        
         assert is_any_list(start), "start should be a list"
         assert is_any_list(duration), "duration should be a list"
         assert is_any_list(end), "end should be a list"
+        
+        assert demand is not None, "demand should be provided but was None"
+        assert capacity is not None, "capacity should be provided but was None"
```
Dimosts, line 701: I don't know what should be done here, but I am not really a fan of giving them a default value (making them optional) and then not allowing the optional value.

cpmpy/solvers/choco.py
```diff
@@ -621,12 +621,21 @@ def _get_constraint(self, cpm_expr):
...
+                    start, demand, cap = self._to_vars([start, demand, cap])
+                else:
+                    start, end, demand, cap = self._to_vars([start, end, demand, cap])
                 # duration can be var or int
                 dur = self.solver_vars(dur)
                 # Create task variables. Choco can create them only one by one
                 tasks = [self.chc_model.task(s, d, e) for s, d, e in zip(start, dur, end)]
+    
                 return self.chc_model.cumulative(tasks, demand, cap)
+            elif cpm_expr.name == "no_overlap": # post as Cumulative with capacity 1
```
Dimosts, line 655: should also be in supported then?

cpmpy/solvers/ortools.py
```diff
@@ -556,11 +556,17 @@ def _post_constraint(self, cpm_expr, reifiable=False):
...
+                start, dur, end, demand, cap = cpm_expr.args
+                if end[0] is None:
+                    end = cpm_expr.get_end_vars()
+                start, dur, end, demand, cap = self.solver_vars([start, dur, end, demand, cap])
                 intervals = [self.ort_model.NewIntervalVar(s,d,e,f"interval_{s}-{d}-{e}") for s,d,e in zip(start,dur,end)]
                 return self.ort_model.AddCumulative(intervals, demand, cap)
             elif cpm_expr.name == "no_overlap":
-                start, dur, end = self.solver_vars(cpm_expr.args)
+                start, dur, end = cpm_expr.args
+                if end[0] is None:
```
Dimosts: this is also checked in get_end_vars, so we can just call it directly?

Dimosts (changes requested):

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: It is fixed in the constructor (already a while, but decompose was not adapted yet)

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Yeah I also agree it is quite annoying... For now this is the best I could come up with, as you cannot make an argument in the middle optional...

Dimosts: I am still troubled by this. So, someone either has to actually use the end var after all, or has to use kwargs to define demand and capacity, which are optional but not that optional after all.

Both minizinc (https://docs.minizinc.dev/en/stable/predicates.html) and essence (https://arxiv.org/pdf/2201.03472) actually do not accept end variables at all, and create them for solvers that need them. Should we stick to having them after all? 

The question is what if the user needs the end variables for something else I guess; In this case, if the users define themselves the end = start +  duration constraints, it is ok. Common subexpression elimination should capture it during the decomposition if needed.

cpmpy/expressions/globalconstraints.py
```diff
@@ -672,30 +672,46 @@ def __repr__(self):
...
+            end = flatlist(end)
+            assert len(start) == len(end), "Start and end should have equal length"
 
         if is_any_list(demand):
             demand = flatlist(demand)
-            assert len(demand) == n_jobs, "Demand should be supplied for each task or be single constant"
         else: # constant demand
-            demand = [demand] * n_jobs
+            demand = [demand] * len(start)
+        assert len(demand) == len(start), "Demand should be supplied for each task or be single constant"
```
Dimosts: This should be inside the if right?

cpmpy/expressions/globalconstraints.py
```diff
@@ -706,49 +722,65 @@ def decompose(self):
             International Conference on Principles and Practice of Constraint Programming. Springer, Berlin, Heidelberg, 2009.
         """
 
-        arr_args = (cpm_array(arg) if is_any_list(arg) else arg for arg in self.args)
-        start, duration, end, demand, capacity = arr_args
+        start, duration, end, demand, capacity = self.args
 
-        cons = []
+        cons = [d >= 0 for d in duration] # enforce non-negative durations
+        cons += [h >= 0 for h in demand] # enforce non-negative demand
```
Dimosts, line 778: Should maybe be in safening? It is defined as a restriction, not as an enforced constraint.

Dimosts: It feels more like safening code than constraining. It is a restriction of the cumulative constraint (https://sofdem.github.io/gccat/aux/pdf/cumulative.pdf) and not something that is enforced. 

However, after in person discussion I agree that it will create more problems if we do it in safening instead. Note to revisit this if we encounter more such cases.

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Hmm yeah, but removing it all together will break every CPMpy model currently out there that uses a Cumulative constraint...
I don't really have too much of an opinion on whether we want to keep the end variables or not, but if we decide to remove them, we should wait a couple of releases and raise a deprecation warning here

IgnaceBleukx:
@Dimosts and/or @tias could you do a (re-)review on this one? It currently blocks #676 as there is a lot of refactoring in this one, that I believe should go in first before we get to the optional intervals

cpmpy/expressions/globalconstraints.py
```diff
@@ -672,83 +672,162 @@ def __repr__(self):
...
+                how (str): how the cumulative constraint should be decomposed, can be "time", "task", or "auto" (default)
+        """
+
+        assert how in ["time", "task", "auto"], "how can only be time, task, or auto (default), but got {}".format(how)
+
+        start, duration, end, demand, capacity = self.args
+
+        lbs, ubs = get_bounds(start)
+        horizon = max(ubs) - min(lbs)
+        if how == "time" or how == "auto" and len(start) <= horizon:
```
tias: please put brackets on your intended interpretation, e.g. () or (  and  )

cpmpy/expressions/globalconstraints.py
```diff
@@ -672,83 +672,162 @@ def __repr__(self):
...
+            end = [start[i] + duration[i] for i in range(len(start))]
+        else:
+            cons += [start[i] + duration[i] == end[i] for i in range(len(start))]
+
+        # demand doesn't exceed capacity
+        for i in range(len(start)):
+            demand_at_i = demand[i]
+            for j in range(len(start)):
+                if i != j:
+                    demand_at_i += demand[j] * ((start[j] <= start[i]) & (end[j] > start[i]))
```
tias: I find this a hard to read style...
I would prefer

demands_at_i = []
for ...

cons.add( demand[i] + cp.sum(demands_at_i) <= capacity )

which is closer to the textbook formula

tias: I think it needs a bit more explanation, e.g. "only check at the start-time of each task"
and maybe rename 'demand_at_i' to 'demand_with_i' or so

cpmpy/expressions/globalconstraints.py
```diff
@@ -672,83 +672,162 @@ def __repr__(self):
...
+        lbs, ubs = get_bounds(start)
+        lb, ub = min(lbs), max(ubs)
         for t in range(lb,ub+1):
             demand_at_t = 0
             for job in range(len(start)):
-                if is_num(demand):
-                    demand_at_t += demand * ((start[job] <= t) & (t < end[job]))
-                else:
-                    demand_at_t += demand[job] * ((start[job] <= t) & (t < end[job]))
+                demand_at_t += demand[job] * ((start[job] <= t) & (end[job] > t))
```
tias: like above, please do this as a list instead of the less readable accumulating sum, so that it shows cons.add( cp.sum(demands_at_t) <= capacity )

Or you could actually also make a helper function demands_at(t) and then
for t in ...:
   cons.add( cp.sum(demands_at(t)) <= capacity )
   
 not sure what is more readable

cpmpy/expressions/globalconstraints.py
```diff
@@ -798,37 +877,68 @@ def value(self):
 
 class NoOverlap(GlobalConstraint):
     """
-    NoOverlap constraint, enforcing that the intervals defined by start, duration and end do not overlap.
+    Global no-overlap constraint. Used for scheduling problems
+    Ensures no tasks overlap and enforces:
+            duration >= 0
+            demand >= 0
```
tias: no demand

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Rewrote it as a single sum. As this is for each time-step, I think it's more natural to write it functionally

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Updated both the doc and the code to make this more explicit

PR #695, opened 2025-06-20
Ensure Expression value is Python native

IgnaceBleukx:
When some arguments are Numpy constants, our current implementation of `.value()` will return a numpy boolean/integer.

This can cause weird things, e.g., in the OUS algorithm, we check whether `constraint.value() is False` which will fail if it is a Numpy False constant.

This PR fixes that by ensuring `def argval` in utils always returns a Python constant, and adds some extra tests.

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -565,16 +565,22 @@ class Comparison(Expression):
     # return the value of the expression
     # optional, default: None
     def value(self):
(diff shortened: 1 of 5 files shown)


tias (approved):
rigorous implementation and tests.

its going to add milliseconds to each argval, but OK, that's our curse of wanting to be numpy compatible I guess : )

PR #696, opened 2025-06-20
Fix CSE edge-case in only_numexpr_equality

IgnaceBleukx:
Because of CSE, only_numexpr_equality can return a single expression to transform a constraint.

E.g, if `csemap = {max(x,y,z) : IV0}`, then calling `only_numexpr_equality([`max(x,y,z) < 42`]) will return `IV0 < 42`.
The old code assumed that if there was only 1 constraint returned, nothing changed.

This cause some tests to fail for Pumpkin, and probably a failing test in #583 is also linked to this (yet to test).

### DIFF
--- a/cpmpy/transformations/comparison.py
+++ b/cpmpy/transformations/comparison.py
@@ -38,7 +38,7 @@ def only_numexpr_equality(constraints, supported=frozenset(), csemap=None):
             if not isinstance(cond, _BoolVarImpl): # expr -> bv
                 res = only_numexpr_equality([cond], supported, csemap=csemap)
                 if len(res) == 1 and cond is not res[0]: # changed, but no new cons due to CSE
-                    newcons[i] = res[0]
+                    newcons[i] = res[0].implies(subexpr)
                 if len(res) > 1:
                     newcons[i] = res[1].implies(subexpr)
                     newcons.insert(i, res[0])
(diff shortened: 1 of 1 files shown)


cpmpy/transformations/comparison.py
```diff
@@ -38,18 +38,18 @@ def only_numexpr_equality(constraints, supported=frozenset(), csemap=None):
             if not isinstance(cond, _BoolVarImpl): # expr -> bv
                 res = only_numexpr_equality([cond], supported, csemap=csemap)
                 if len(res) == 1 and cond is not res[0]: # changed, but no new cons due to CSE
```
tias: I don't know, this code (and the one before it) seems full of assumptions...

what if len(res) == 1 and cond is res[0]? what happens then, or what should happen then?

is the shortcut 'if len(res) == 1 and res[0] == cond: ...pass/smth`?

and what if len(res)==2, then for some reason you have some strong knowledge on what res[0] will be, and that you only need to have an implication over res[1]? but why?

and also what if len(res) >2?

IgnaceBleukx: Yeah indeed a lot of assumptions are made here which seem quite unintuitive at first sight.
I think we should refactor this to a separate helper function`_comparison_to_eq` or smt, which can then get called from all places in this "dispatcher" transforamtion.

cpmpy/transformations/comparison.py
```diff
@@ -28,68 +28,87 @@ def only_numexpr_equality(constraints, supported=frozenset(), csemap=None):
...
 
-            if cpm_expr.name == "==" and is_boolexpr(lhs) and is_boolexpr(rhs): # reification, check recursively
+    E.g., max(x,y,z) < p is rewritten to:
+        max(x,y,z) == iv & iv < p
+
+    :param cpm_expr: the comparison to rewrite
+    :param csemap: the cse map to use
+    :return: the rewritten comparison and the defining constraint
+    """
+    print(f"Rewriting {cpm_expr}")
```
tias: you probably want to take this print out

PR #697, opened 2025-06-24
Choco shorttable STAR int

ThomSerg:
Ensure that the star symbol used for choco's shorttable is an integer. Some tricks have been used to detect star locations by turning the table data type to float and representing the stars as `np.nan`. The smallest non-nan value then gets found and STAR will take the value of 1 less (min - 1). But when posting to the solver, choco expects a int star-symbol, which is currently still a float. Simply fixed by casting to int (the min - 1 is already is an int but represented with the float datatype).


IgnaceBleukx (approved):
Yes, good to merge.

PR #698, opened 2025-06-26
OCUS in tools

IgnaceBleukx:
Slight modification of the previous `optimal_mus` algorithm to allow for puting a meta-consrtaint on the MUS to be found.
Only thing I'm not sure of, is how we should return "Failure"? I.e., if no unsat subset exists given the meta-consrtaint.

Currently I raise a `ValueError` but not sure if this is appropriate? Maybe a custom CPMpy exception is actually better.


cpmpy/tools/explain/mus.py
```diff
@@ -176,9 +183,22 @@ def optimal_mus(soft, hard=[], weights=None, solver="ortools", hs_solver="ortool
             sat_subset += new_corr_subset # extend sat subset with new corr subset, guaranteed to be disjoint
             hs_solver += cp.sum(new_corr_subset) >= 1 # add new corr subset to hitting set solver
 
+    if hs_solver.status().exitstatus == ExitStatus.UNSATISFIABLE:
+        raise ValueError("No unsatisfiable constrained subset could be found") # TODO: better exception?
```
Dimosts: We can just return the empty list? or None?

It basically is an unsat problem, it does not need an exception.

cpmpy/tools/explain/mus.py
```diff
@@ -176,9 +183,22 @@ def optimal_mus(soft, hard=[], weights=None, solver="ortools", hs_solver="ortool
...
+
     return [dmap[a] for a in hitting_set]
 
+
+def optimal_mus(soft, hard=[], weights=None, solver="ortools", hs_solver="ortools", do_solution_hint=True):
```
Dimosts, line 191: For backwards compatibility right? maybe a comment that it will be deprecated?

cpmpy/tools/explain/mus.py
```diff
@@ -294,8 +317,20 @@ def optimal_mus_naive(soft, hard=[], weights=None, solver="ortools", hs_solver="
...
 
+    if hs_solver.status().exitstatus == ExitStatus.UNSATISFIABLE:
+        raise ValueError("No unsatisfiable constrained subset could be found") # TODO: better exception?
+
     return hitting_set
 
 
+    
+
+def optimal_mus_naive(soft, hard=[], weights=None, solver="ortools", hs_solver="ortools"):
```
Dimosts, line 327: For backwards compatibility right? maybe a comment that it will be deprecated?

cpmpy/tools/explain/mus.py
IgnaceBleukx: Well it's meta-unsatisfiable ; ) 
The trouble is, an empty list will not adhere to the constraint given by the user, and it's also the same result as if the conflict is in the hard constraints...

That's why I was thinking an exception might be more appropriate here. It's also how we do it in the pseudocode of the papers ; )

cpmpy/tools/explain/mus.py
IgnaceBleukx: More like a quality of life feature. Similar to what we have with "smus", it's just a convenient shortcut.

cpmpy/tools/explain/mus.py
Dimosts: Ok, then indeed an appropriate CPMpy exception would be better.

PR #701, opened 2025-07-10
PSP-lib dataset and loader

IgnaceBleukx:
Inspired by the XCSP3 dataloader in #597 , I implemented a similar dataset object to download and parse RCPSP problems from psplib.com

Example for now, but may be useful to make it a bit more structured if we start doing this do this for more datasources.

This version is probably too extensive, but any feedback welcome

### DIFF
--- a/examples/psplib.py
+++ b/examples/psplib.py
@@ -1,0 +1,280 @@
+"""
+PyTorch-style Dataset for RCPSP instances from psplib.com
+
+Simply create a dataset instance (configured for the targeted family such as j30,j60 etc...) and start iterating over its contents:
+The `metadata` contains usefull information about the current problem instance.
+"""
+
+import pathlib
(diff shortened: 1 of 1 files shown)


examples/psplib.py
```diff
@@ -0,0 +1,280 @@
...
+        
+        self.families = dict(
+            rcpsp = ["j30", "j60", "j90", "j120"]
+        )
+        self.family_codes = dict(rcpsp="sm", mrcpsp="mm")
+
+        if variant != "rcpsp":
+            raise ValueError("Only 'rcpsp' variant is supported for now")
+        if family not in self.families[variant]:
+            raise ValueError("Track must be specified, e.g. COP, CSP, MiniCOP, ...")
```
tias: error to be updated

Issue #692, opened 2025-06-16
Negation of global constraints


IgnaceBleukx:
As discussed in 
> Francois Fages, Sylvain Soliman. Reifying Global Constraints. [Research Report] RR-8084, INRIA.
2012, pp.18. 

the negation of some global constraints map to other global constraints. A simple example is `~Table` maps to `NegativeTable`, but also for other constraints there are better ways of handling the negation. E.g., `~AllDifferent(x)` maps to `NValues(x) < len(x)`, and for Regular, we can swap the accepting and non-accepting states.

Currently, we decompose the global constraint and negate that, but I think we can do this better, with fairly minimal changes.

**Proposed changes**
I think it would be great if each global constraint has a function `negate()` which returns a list of expressions that represents the negation of the global constraint.
By default it will return `~self` but for others, we can make this better (e.g., Regular/Table/AllDifferent...)

The `negate()` should then be called in the `push_down_negation` transformation.

Happy to discuss : )

OrestisLomis:
Making this a tracker, feel free to edit this comment to add more.

- [x] Table/NegativeTable
- [x] InDomain  
- [x] Xor 
- [ ] AllDifferent
- [ ] Regular

Issue #620, opened 2025-04-10
Boolean constants not handled in Global constraints


peerjosse:
erroneous solvers: ortools, choco, gurobi
working solvers: minizinc, z3

In the following program:
- 3 of the 5 tested solvers gave an error
- 2 of those 3 gave the same error on the same line of `canonical_comparison`
- ortools gave an error I can't really place

```python
import cpmpy as cp

m = cp.Model()

a = cp.boolvar()
b = cp.boolvar()
c = cp.boolvar()

m += cp.Xor([(cp.Xor([a, b, c])) <= True, ~((cp.Xor([a, b, c])) <= True)])

# m.solve(solver='ortools')     # not supported: model.get_or_make_boolean_index(boolval(True)) (in get_or_make_boolean_index)
# m.solve(solver='minizinc')    # returns no error
# m.solve(solver='z3')          # returns no error
# m.solve(solver='choco')       # lhs cannot be an integer at this point! (in canonical_comparison)
# m.solve(solver='gurobi')      # lhs cannot be an integer at this point! (in canonical_comparison)
```

IgnaceBleukx:
So the root-cause of this problem is the `simplify_bool` transformation that should get rid of any Boolean constants  in the expression tree.

In your case:
```
    Xor([(Xor([a, b, c])) <= True, ~((Xor([a, b, c])) <= True)])
=  Xor([(True False])
```
Which __should__ be further simplifed to just `True`.
However, we currently do not support simplification of global constraints in this transformation, and I'm also not really sure to implement this generically....

The error triggered in Choco and Gurobi is quite funny: we decompose the XOR constraint to `1 + 0 == 1` which remains an expression... this also requires more thinking on how to get this correct. Probably should be part of normalization code as well.

I will change the title to reflect the root-cause of the problem better.

PR #703, opened 2025-07-30
Add missing case for Xor in normalize

ThomSerg:
Xor is not part of "Operator" but instead a special global. It was missing in the normalisation for constant bools.

### DIFF
--- a/cpmpy/transformations/normalize.py
+++ b/cpmpy/transformations/normalize.py
@@ -12,7 +12,7 @@ from ..expressions.globalfunctions import GlobalFunction
 from ..expressions.utils import eval_comparison, is_false_cst, is_true_cst, is_boolexpr, is_num, is_bool
 from ..expressions.variables import NDVarArray, _BoolVarImpl
 from ..exceptions import NotSupportedError
-from ..expressions.globalconstraints import GlobalConstraint
+from ..expressions.globalconstraints import GlobalConstraint, Xor
 
 
 def toplevel_list(cpm_expr, merge_and=True):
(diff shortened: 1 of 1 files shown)


cpmpy/transformations/normalize.py
```diff
@@ -225,7 +225,61 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+                0	1	0	1
+                0	1	1	0
+                1	0	0	1
+                1	0	1	0
+                1	1	0	0
+                1	1	1	1
+            """
+            # Simplify nested expression
+            if expr.has_subexpr():
+                expr_args = simplify_boolean(expr.args, num_context=not expr.is_bool())
```
IgnaceBleukx: I would just put `num_context=False` here; that's more explicit.

cpmpy/transformations/normalize.py
```diff
@@ -225,7 +225,61 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+                elif is_false_cst(a):
+                    if args is expr_args: # will remove this one, need to copy args...
+                        args = args.copy()
+                    args.pop(i)
+                else:
+                    i += 1
+                    continue
+            if len(args) == 0: # only constant bools
+                if nr_true_constants % 2 == 0:
+                    newlist.append(0 if num_context else BoolVal(False))
```
IgnaceBleukx: you could merge these cases:
```python
expr_is_true = BoolVal(nr_true_constants % 2 == 1)
newlist.append(int(expr_is_true) if num_context else expr_is_true)
```

cpmpy/transformations/normalize.py
```diff
@@ -225,7 +225,61 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+                else:
+                    newlist.append(1 if num_context else BoolVal(True))
+            elif len(args) == 1: # Xor with single argument can be simplified to just its argument
+                newlist.append(args[0])
+            elif args is not expr.args: # removed something, or changed due to subexpr
+                newexpr = copy.copy(expr)
+                newexpr.update_args(args)
+                newlist.append(newexpr if (nr_true_constants % 2 == 0) else ~newexpr) # negate expression depending on number of 'True' boolean constants
```
IgnaceBleukx: This goes wrong in the current transformation stack (see failing tests I pushed).
The issues is that simplify_boolean comes AFTER decompose_global. While xor is support toplevel, it is not supported nested, but here you make it nested... so it fails later on when we try to post ~xor.

I'm not really sure how to fix it, other than decomposing the xor here...

IgnaceBleukx (dismissed):
This is indeed an open issue (also related to #620 though not fully fixed by this PR).
One issue remaining: namely the changed context of XOR, based on the number of true constants you removed from it. This needs to be fixed before it can be merged.

tias (changes requested):
This sets a precedent, having an 'if' case for two specific named globals in a transformation, actually in two: in negation and in normalize.

As far as I know, all transformations are designed so that there is no 'special case' logic for individual globals in them, e.g. a user can add a new global with the same ease as a cpmpy developer. And as a starting developer there are no 'surprises' with magic places where additional logic happens. With this, we would add special negation and normalisation cases in the transfo...

So I don't like that.

I'm also not sure what the alternative is. For the 'negation' one could imagine an additional abstract function for globals 'negate_custom()' or so?
But for normalize... not sure.

Why did we need a custom normalize for Xor again?

IgnaceBleukx:
I see this might indeed not be the best place to implement the negation;
In issue #692 I proposed a `.negate` function for global constraints, but I forgot about it while implementing this PR...

Xor and IfThenElse could be a good place to start implementing such a function. The default will just return `~self`.

As for normalization of XOR, we run in the trouble later on if we do not normalize it.
Our `simplify_boolean` tranformation promises to remove any Boolean constants from the expression tree, but this is simply not true when they are used in global constraints.

E.g., XOR(a,b,True) ends up as is in OR-Tools, where there is no support for the Boolval, to convert it to a solver var. Alternatively, we can special case it in each solver interface, but that also seems unwanted.

When it gets decomposed, we catch things correctly, as then we now what to do with the constants.

PR #711, opened 2025-08-04
Boolean constant in wsum

### LINKED ISSUE #688: [OR-Tools] TypeError: LinearExpr::weighted_sum() only accept constants as coefficients: 'bool_'
When running the example [counterfactual_explain.py](https://github.com/CPMpy/cpmpy/blob/master/examples/advanced/counterfactual_explain.py) (with OR-Tools solver), there is the following error:
`TypeError: LinearExpr::weighted_sum() only accept constants as coefficients: 'bool_'`

In this part of the example (line 213):

```
        else:
            # TODO: fix needing to cast them to int for OR-Tools
            master_model += [sum(d * x_d) >= sum(d * x_0.value())]
        i += 1
```

The following is run normally:
```
            x_d_int = x_d.astype(int)
            x_0_val_int = x_0.value().astype(int)
            # Add constraint using integer coefficients
            master_model += [sum(d * x_d_int) >= sum(d * x_0_val_int)]
```

IgnaceBleukx:
Allows to use Boolean constants as arguments in a weighted sum, fixes #688 
Especially useful when using a Boolean array as mask during modeling.

Also includes rewrite of `get_bounds` of weighted sum, avoiding sorting of the arguments, which could go wrong previously... Not sure if it can stay in this PR

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -763,12 +763,15 @@ class Operator(Expression):
             weights, vars = self.args
             bounds = []
             #this may seem like too many lines, but avoiding np.sum avoids overflowing things at int32 bounds
-            for i, varbounds in enumerate([get_bounds(arg) for arg in vars]):
-                sortbounds = (list(weights[i] * x for x in varbounds))
-                sortbounds.sort()
-                bounds += [sortbounds]
-            lbs, ubs = (zip(*bounds))
-            lowerbound, upperbound = sum(lbs), sum(ubs) #this is builtins sum, not numpy sum
+            for w, (lb, ub) in zip(weights, [get_bounds(arg) for arg in vars]):
+                x,y = int(w) * lb, int(w) * ub
+                if x <= y: # x is the lb of this arg
+                    lowerbound += x
+                    upperbound += y
+                else:
+                    lowerbound += y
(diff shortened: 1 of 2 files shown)


cpmpy/expressions/core.py
```diff
@@ -755,13 +762,17 @@ def get_bounds(self):
...
             bounds = []
+            lowerbound, upperbound = 0,0
             #this may seem like too many lines, but avoiding np.sum avoids overflowing things at int32 bounds
-            for i, varbounds in enumerate([get_bounds(arg) for arg in vars]):
-                sortbounds = (list(weights[i] * x for x in varbounds))
-                sortbounds.sort()
-                bounds += [sortbounds]
-            lbs, ubs = (zip(*bounds))
-            lowerbound, upperbound = sum(lbs), sum(ubs) #this is builtins sum, not numpy sum
+            for w, (lb, ub) in zip(weights, [get_bounds(arg) for arg in vars]):
```
ThomSerg, line 767: Was going to leave a comment that the list comprehension could be replaced by a generator expression as to prevent unnecessarily creating a list, but apparently [list comprehension can still be faster](https://stackoverflow.com/questions/47789/generator-expressions-vs-list-comprehensions/62709748#62709748)

PR #712, opened 2025-08-04
Convert assumptions to list

### LINKED ISSUE #689: [OR-Tools] Ocus assumptions: TypeError: unhashable type: 'set'
When running [ocus_explanations.py](https://github.com/CPMpy/cpmpy/blob/master/examples/advanced/ocus_explanations.py), the following error is produced:
```
    if cpm_var not in self._varmap:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unhashable type: 'set'
```

The issue stems from line 212: `if not SAT.solve(assumptions=S):`

It will run normally if cast to list: `if not SAT.solve(assumptions=list(S)):`

IgnaceBleukx:
Allows a user to post a `set` instead of `list` of assumption variables during the solve call.
More of a user-experience improvement (personally also encountered this multiple times already) and fixes #689 

Could be part of a larger discussion on what to do with sets in general, do we want to support them as if they were lists?
E.g., should `get_variables(some_set_of_constraints)` also work?


ThomSerg (approved):
Looks good. But was wondering why we don't check if its an iterable instead? Something similar as is done in `_flatten()`. Not that this would make any noticeable difference. Just a thought, pull request can be merged as is.

IgnaceBleukx:
Hmm good point, from a user-perspective it might actually be nice if we just took care of any collection.
I'll adapt it so it calls `flatlist(assumptions` instead.

IgnaceBleukx (commented):
Merged master into this one again and resolved conflicts.
I can't approve it myself as I am the author of the PR, but looks good to go for me now.

cpmpy/solvers/pysdd.py
```diff
@@ -135,7 +135,7 @@ def native_model(self):
         """
         return self.pysdd_root
 
-    def solve(self, time_limit:Optional[float]=None, assumptions:Optional[List[_BoolVarImpl]]=None):
+    def solve(self, time_limit:Optional[float]=None, assumptions:Optional[Iterable[_BoolVarImpl]]=None):
```
ThomSerg: Not sure why pysdd has `assumptions` in its solve interface, doesn't use it.

ThomSerg (commented):
One comment about pysdd and waiting for tests to complete.

tias (approved):
looks good to merge? please do

PR #714, opened 2025-08-06
Refactor/format int2bool/pysat

hbierlee:
Just stylistic changes to keep things consistent how `int2bool` is integrated.


IgnaceBleukx (approved):
Looks good! Indeed makes a lot of sense to do the encoding in `solver_var`.
I'm wondering if we should also make the encoding part of the template solver interface? I'm leaning towards no as it's quite advanced

hbierlee:
Great! The CI didn't work because I pushed to my fork rather than this repo.

I'm also leaning towards no for the moment, because I'm thinking of ways to make int2bool a bit more transparent so you don't need this extra custom solver interface code.

However, we should take the opportunity to remove the asterisk in the waterfall 😄 Who has the `waterfall.png` code..?

IgnaceBleukx:
Yes indeed! I think @tias has the source for that figure : )

PR #716, opened 2025-08-08
Test and fix `version` method for all solvers

hbierlee:
This fixes `version()` for all solvers (on CI).

A small fix that can be merged quickly. Alternatively, we could try out just implementing this in `SolverInterface` class, because most `version` methods use? Avoiding this duplication for most solvers:

```py
@staticmethod
def version() -> Optional[str]:
    """
    Returns the installed version of the solver's Python API.
    """
    try:
        return pkg_resources.get_distribution('pumpkin').version
    except pkg_resources.DistributionNotFound:
        return None
```

### DIFF
--- a/tests/test_solvers.py
+++ b/tests/test_solvers.py
@@ -1047,3 +1047,5 @@ class TestSupportedSolvers:
         num_sols = cp.Model(cp.BoolVal(False)).solveAll(solver=solver, solution_limit=solution_limit)
         assert num_sols == 0
 
+    def test_version(self, solver):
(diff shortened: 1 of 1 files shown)


tias:
in superclass could make sense, but then the static method needs to know the package name of the subclass?

hbierlee:
just a test now

IgnaceBleukx (changes requested):
The test is great, but the PR includes a lot of irrelevant files.
I think this is due to a miss-commit; please remove and go ahead and merge after : )

PR #718, opened 2025-08-14
Add local solver (hexaly)

IgnaceBleukx:
Adding the commercial hexaly solver (previously called LocalSolver).

Interface is relatively feature complete already, but still need to have a thorough look as to which globals are supported or not.


cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+"""
+
+from typing import Optional
+import pkg_resources
+
+from .solver_interface import SolverInterface, SolverStatus, ExitStatus
+from ..expressions.core import Expression, Comparison, Operator, BoolVal
+from ..expressions.globalconstraints import GlobalConstraint, GlobalFunction
+from ..expressions.variables import _BoolVarImpl, NegBoolView, _IntVarImpl, _NumVarImpl
+from ..expressions.utils import is_num, is_any_list, is_boolexpr, eval_comparison
```
Dimosts: There are some unused imports like is_boolexpr here, or Expression before

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+    Documentation of the solver's own Python API:
+    https://www.hexaly.com/docs/last/pythonapi/index.html
+    """
+
+    @staticmethod
+    def supported():
+        # try to import the package
+        try:
+            import hexaly as hex
+            # optionally enforce a specific version
```
Dimosts: no need for the comment

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+        Arguments:
+        - cpm_model: Model(), a CPMpy Model() (optional)
+        - subsolver: str, name of a subsolver (optional)
+        """
+        if not self.supported():
+            raise Exception("CPM_hexaly: Install the python package 'hexaly' to use this solver interface.")
+
+        from hexaly.optimizer import HexalyOptimizer, HxObjectiveDirection
+
+        assert subsolver is None # unless you support subsolvers, see pysat or minizinc
```
Dimosts: in general remove comments from template

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+from ..transformations.get_variables import get_variables
+from ..transformations.normalize import toplevel_list
+from ..transformations.decompose_global import decompose_in_tree
+
+class CPM_hexaly(SolverInterface):
+    """
+    Interface to Hexaly's API
+
+    Creates the following attributes (see parent constructor for more):
+    - hex_model: object, Hexaly's model object
```
Dimosts, line 60: also hex_solver from what I saw later on?

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+
+        from hexaly.optimizer import HexalyOptimizer, HxObjectiveDirection
+
+        assert subsolver is None # unless you support subsolvers, see pysat or minizinc
+
+        # initialise the native solver object
+        self.hex_solver = HexalyOptimizer()
+        self.hex_solver.param.verbosity = 0
+        self.hex_model = self.hex_solver.model
+        self.hex_model.add_objective(0, HxObjectiveDirection.MINIMIZE)
```
Dimosts: Why is that needed here?

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+
+        # initialise everything else and post the constraints/objective
+        super().__init__(name="hexaly", cpm_model=cpm_model)
+
+    def solve(self, time_limit=None, **kwargs):
```
Dimosts, line 116: In some solvers we have native_model() before solve(), while in others we do not. So the real question is, do we want to have it? I believe yes, and thus we should also include it here

Dimosts: have opened a PR for the rest.

cpmpy/solvers/hexaly.py
Dimosts: I guess for solve() when it is not an opt problem. Shouldn't it then be there?

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+            (technical side note: any constraints created during conversion of the objective
+
+            are permanently posted to the solver)
+        """
+        from hexaly.optimizer import HxObjectiveDirection
+        # make objective function or variable and post
+        # remove previous objectives
+        if self.has_objective(): # remove prev objective
+            self.hex_model.remove_objective(1)
+        assert self.hex_model.nb_objectives == 1 # only the constant objective should be left now
```
Dimosts: I feel this is a bit weird. Instead of always having a constant objective, and put others later, shouldnt we just put a constant objective in satisfacito n problems during solve()?

An alternative is, to always have a constant objective, and then just replace that one in this function, without overcomplicating it by adding another one, checking it they are more than one objective or not etc.

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+        hex_obj = self._hex_expr(expr)
+        if minimize:
+            self.hex_model.add_objective(hex_obj,HxObjectiveDirection.MINIMIZE)
+        else:
+            self.hex_model.add_objective(hex_obj,HxObjectiveDirection.MAXIMIZE)
+
+    def has_objective(self):
+        return self.hex_model.nb_objectives > 1
```
Dimosts: feels weird

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+            if cpm_expr.name == "sub":
+                a,b = self._hex_expr(cpm_expr.args)
+                return a - b
+            if cpm_expr.name == "-":
+                return -self._hex_expr(cpm_expr.args[0])
+            if cpm_expr.name == "mul":
+                a,b = self._hex_expr(cpm_expr.args)
+                return a * b
+            if cpm_expr.name == "div":
+                raise NotImplementedError("hexaly supports division, but result is float, TODO")
```
Dimosts: hexaly has also a floor operator, which could be used to enforce integer division
https://www.hexaly.com/docs/last/pythonapi/optimizer/hxoperator.html

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+
+        :return: list of Expression
+        """
+        # apply transformations
+        cpm_cons = toplevel_list(cpm_expr)
+        # no flattening, so also no safening required
+        cpm_cons = decompose_in_tree(cpm_cons, supported={"min", "max", "abs", "element"})
+        return cpm_cons
+
+    def _hex_expr(self, cpm_expr):
```
Dimosts, line 312: it misses DirectConstraint

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+                return self.hex_model.abs(self._hex_expr(cpm_expr.args[0]))
+            if cpm_expr.name == "min":
+                return self.hex_model.min(*self._hex_expr(cpm_expr.args))
+            if cpm_expr.name == "max":
+                return self.hex_model.max(*self._hex_expr(cpm_expr.args))
+            raise ValueError(f"Global function {cpm_expr} is not supported by hexaly")
+
+        raise NotImplementedError(f"Unexpected expression {cpm_expr}")
+
+    def add(self, cpm_expr_orig):
```
Dimosts: In other solvers that have a separate add() function, we have add() before the solver_constraint() methods, and I think it makes sense

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+##
+## hexaly.py
+##
+"""
+    Interface to Hexaly's API
+
+
+    Hexaly is a local search solver with suppport for  global constraints.
```
Dimosts: support

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,387 @@
...
+from .solver_interface import SolverInterface, SolverStatus, ExitStatus
+from ..expressions.core import Expression, Comparison, Operator, BoolVal
+from ..expressions.globalconstraints import GlobalConstraint, GlobalFunction
+from ..expressions.variables import _BoolVarImpl, NegBoolView, _IntVarImpl, _NumVarImpl
+from ..expressions.utils import is_num, is_any_list, is_boolexpr, eval_comparison
+from ..transformations.get_variables import get_variables
+from ..transformations.normalize import toplevel_list
+from ..transformations.decompose_global import decompose_in_tree
+
+class CPM_hexaly(SolverInterface):
```
Dimosts, line 55: solution_hint() is missing, although supported: https://www.hexaly.com/docs/last/technicalfeatures/initialsolution.html

tests/test_constraints.py
```diff
@@ -51,6 +51,7 @@
                      "pindakaas": {"mul", "div", "pow", "mod"},
                      "exact": {},
                      "pumpkin": {"pow", "mod"},
+                     "hexaly": {"div"} # supported but converts result to float, TODO
```
Dimosts: I think it can be handled easily, see my comment in that part

Dimosts (changes requested):
Looks nice and clean!

Mostly minor comments, except a couple, like solution_hint missing, directconstraint not handled, and a suggestion to handle div.

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,418 @@
...
+            return cpm_expr.callSolver(self, self.hex_model)
+
+        raise NotImplementedError(f"Unexpected expression {cpm_expr}")
+
+    
+    def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from_model=False, **kwargs):
+        # time_limit = 10
+        if time_limit is None:
+            raise ValueError("Hexaly does not support exhausitve search to find all solutions. "
+                             "Set time limit or solution limit to do a limited search")
```
Dimosts: We say "set time limit or solution limit", but we only check for time_limit. I guess the error message has to change?

cpmpy/solvers/hexaly.py
```diff
@@ -0,0 +1,418 @@
...
+            raise ValueError(f"Global function {cpm_expr} is not supported by hexaly")
+
+        elif isinstance(cpm_expr, DirectConstraint):
+            return cpm_expr.callSolver(self, self.hex_model)
+
+        raise NotImplementedError(f"Unexpected expression {cpm_expr}")
+
+    
+    def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from_model=False, **kwargs):
+        # time_limit = 10
```
Dimosts: remove comment

Dimosts:
So, there is the issue we discussed in person about the tests:
- we cannot run the tests in parallel on hexaly. Not sure if we can do something for it, but we should check if we can specify that these tests should not run in parallel while the rest can.
- Even when not running in parallel, some tests fail with no tokens available, because the next tests starts too fast compared to the previous ones. And the token is not released yet.

For the second part, we only need to add the following to work:

    def __del__(self):
        self.hex_solver.delete()

Dimosts (approved):
Yes, seems ready. I reran all tests and seem good! Ready to merge for me

tias:
I had another quick look, still looks good.

One thing that jumped a bit into my eye: implication constraints (b -> Cons()) are translated into disjunction (~b | Cons()), but I could imagine that the solver has no really good native implementation for this.

I wonder if we could not use the 'iff' construct, though that one requires a third argument and is stronger...

PR #719, opened 2025-08-18
Pin Pumpkin git commit

hbierlee:
In this tag, the `constraint_tag` argument is not yet required as it is on main (which is break CI atm). Not sure if this fixes it, as locally `maturin build` is still building the latest source for some reason.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -20,7 +20,7 @@ jobs:
           
           # install pumpkin from source
           pip install maturin
-          git clone https://github.com/consol-lab/pumpkin pumpkin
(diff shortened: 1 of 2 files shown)


IgnaceBleukx:
Let's merge this already, and wait for mergeing #720 until Pumpkin is on pip.

hbierlee:
Ok, I think that's probably the quickest fix. The open questions are:

- Pin the commit or the tag? I think the tag as current is fine
- Should we also ensure users install the right source version somehow?

If the answers are respectively yes and no, then it's good to merge for me.

IgnaceBleukx:
1. yes, fine like this I think
2. hmm, maybe just add it in the installation instructions? Now it just says "clone repo https:// .. " you can add the tagged branch there as well.

Ready to go after that

PR #720, opened 2025-08-18
Implement constraint tagging for Pumpkin

IgnaceBleukx:
The Python interface of Pumpkin has changed, and now requires to add a consrtaint tag to each constraint while posting.
This PR updates our interface to do exactly that.

The purpose of the tags is to, while reading the proof-log, figure out which constraints were used to derive a new clause.

In this PR, the tags are chosen such that they correspond to a "transformed" constraint as supported by the solver.
Alternatively, we can also chose to make the tags correspond to the untransformed, user-level consrtaints instead.
In the MiniZinc interface, they correspond to each flatzinc constraint if I remember correcly; so tried to minic this here as well.

### DIFF
--- a/cpmpy/solvers/pumpkin.py
+++ b/cpmpy/solvers/pumpkin.py
@@ -166,7 +166,7 @@ class CPM_pumpkin(SolverInterface):
 
         elif assumptions is not None:
             assert not prove, "Proof-logging under assumptions is not supported"
-            pum_assumptions = [self.to_predicate(a) for a in assumptions]
+            pum_assumptions = [self.to_predicate(a, tag=None) for a in assumptions]
             self.assump_map = dict(zip(pum_assumptions, assumptions))
             solve_func = self.pum_solver.satisfy_under_assumptions
             kwargs.update(assumptions=pum_assumptions)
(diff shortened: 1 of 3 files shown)


IgnaceBleukx:
1. yes, fine like this I think
2. hmm, maybe just add it in the installation instructions? Now it just says "clone repo https:// .. " you can add the tagged branch there as well.

Ready to go after that

IgnaceBleukx:
So Pumpkin is on pipy now, updated the documentation and gh actions; can you have a quick look @hbierlee?

cpmpy/solvers/pumpkin.py
```diff
@@ -381,54 +383,57 @@ def to_predicate(self, cpm_expr):
...
-    def _sum_args(self, expr, negate=False):
+    def _sum_args(self, expr, negate=False, tag=None):
         """
             Helper function to convert CPMpy sum-like operators into pumpkin-compatible arguments.
             expr is expected to be a `sum`, `wsum` or `sub` operator.
 
             :return: Returns a list of Pumpkin integer expressions
         """
         args = []
+        if tag is None: raise ValueError("Expected tag to be provided but got None")
```
hbierlee, line 411: if `tag` is a mandatory argument, maybe just make it positional rather than keyword?

IgnaceBleukx: I had it like that at first, but it can result in quite confusing error messages (either from the Rust level of from deep recursions in our code)

IgnaceBleukx: Ah and also it would require to put `negate' before the tag which does not match the rest of the interface : )

PR #721, opened 2025-08-20
Linearize Improvements

Dimosts:
Moved an improvement initially made in #583 here, and then decided to make a full pass to maybe catch more errors. I did found some, a few more minor like types, and a few more important ones.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -20,7 +20,7 @@ jobs:
           
           # install pumpkin from source
           pip install maturin
-          git clone https://github.com/consol-lab/pumpkin pumpkin
+          git clone https://github.com/consol-lab/pumpkin pumpkin --branch pumpkin-core-v0.2.1
           cd pumpkin/pumpkin-solver-py
           maturin build
           pip install ../target/wheels/*.whl # name can change based on version
--- a/cpmpy/solvers/pumpkin.py
+++ b/cpmpy/solvers/pumpkin.py
@@ -18,7 +18,7 @@
 
     The `pumpkin_solver_py` python package is currently not available on PyPI.
     It can be installed from source using the following steps:
-     1. clone the repository from github: https://github.com/consol-lab/pumpkin
+     1. clone the release `v0.2.1` repository from github, e.g.: `git clone https://github.com/consol-lab/pumpkin pumpkin --branch pumpkin-core-v0.2.1`
      2. install the "maturin" package to build the python bindings: :code:`pip install maturin`
      3. build and install the package: :code:`cd pumpkin/pumpkin-solver-py && maturin develop`
 


cpmpy/transformations/linearize.py
```diff
@@ -106,10 +106,6 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False, c
             elif cpm_expr.name == "or" and cpm_expr.name not in supported:
                 newlist.append(sum(cpm_expr.args) >= 1)
 
-            # xor
-            elif cpm_expr.name == "xor" and len(cpm_expr.args) == 2:
```
IgnaceBleukx, line 110: this indeed seems like ancient relic from where we did not have a good xor decomposition... fine to remove

cpmpy/transformations/linearize.py
```diff
@@ -181,9 +177,10 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False, c
 
 
                 elif lhs.name == "mod" and "mod" not in supported:
-                    if "mul" not in supported:
+                    if "mul" not in supported and not is_num(lhs.args[1]):
```
IgnaceBleukx, line 180: this `and` should be an `or` I think?

cpmpy/transformations/linearize.py
```diff
@@ -582,7 +579,7 @@ def canonical_comparison(lst_of_expr):
                 elif isinstance(rhs, Operator) and rhs.name == "-":
                     lhs2, rhs = [rhs.args[0]], 0
                 elif isinstance(rhs, Operator) and rhs.name == "sum":
-                    lhs2, rhs = [-1 * b if isinstance(b, _NumVarImpl) else 1 * b.args[0] for b in rhs.args
+                    lhs2, rhs = [-1 * b if isinstance(b, _NumVarImpl) else -1 * b.args[0] for b in rhs.args
```
IgnaceBleukx, line 582: wow... luckily this is just some fail-safe code as normally flatten wil already put sum operators to the left of a comparison

cpmpy/transformations/linearize.py
Dimosts: yea, I actually traced to see what happens with XOR and never goes here. Either way, it is kinda wrong, i.e. does not deal with higher arity XOR.

cpmpy/transformations/linearize.py
Dimosts: Actually no, I also initially went with or, and tests started failing.. 

Then I discovered that it is indeed an and that is needed here. if it is num, it is ok to not support mul, cause support mul means support mul with 2 variables.

cpmpy/transformations/linearize.py
Dimosts: yea, could not manage to actually create a test that it fails here, as it is handled before, but still yea.

PR #725, opened 2025-09-04
Remove simplification of nullifying arguments from constructors

### LINKED ISSUE #713: Unconstrained variables not added to solver vars leading to incorrect `solveAll` count
The following test shows that a constraint `x | True` is never added to the solver, and thus `x` does not become a user variable, and thus we only get one (empty) solution with `x = None`, rather than `x = False`/`True`.

```py
def test_solve_all_user_vars(self, solver):
    x = cp.boolvar()
    m = cp.Model(x | True)
>       assert m.solveAll(solver) == 2
E       AssertionError: assert 1 == 2
E        +  where 1 = solveAll('ortools')
E        +    where solveAll = Constraints:\n    boolval(True)\nObjective: None.solveAll
```

Even if we accept `x` as `None` meaning it could be assigned either (as some solvers do), the solution count would quite arguably still be wrong?

IgnaceBleukx:
Previously, we would simplify things such as `x & False` to `False`. 
However, this raises the problem that the solver does not know `x`, and hence `x` will remain unassigned in the solution.
Additionally, solution counts can be wrong when using solveAll (issue #713).

This PR closes #713

### DIFF
--- a/tests/test_expressions.py
+++ b/tests/test_expressions.py
@@ -628,6 +628,40 @@ class TestBuildIns(unittest.TestCase):
         self.assertEqual(str(gt), str(cp.max(v for v in self.x)))
         self.assertEqual(str(gt), str(cp.max(self.x[0], self.x[1], self.x[2])))
 
+from cpmpy.transformations.get_variables import get_variables
+class TestNullifyingArguments(unittest.TestCase):
+
+    def setUp(self):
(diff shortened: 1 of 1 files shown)


Dimosts:
It would be nice though if this simplification happens while transforming instead

Dimosts (approved):

PR #726, opened 2025-09-04
SolveInterface improvements and tests

Dimosts:
Some of our solvers have a native_model() function to return the native model, while some does not. Also, while the SolverInterface has this function, the template.py does not.

I do believe that it is useful to have it, and thus implemented in all solvers and in template.

### DIFF
--- a/tests/test_solverinterface.py
+++ b/tests/test_solverinterface.py
@@ -272,7 +272,7 @@ def test_time_limit(solver_name):
     
     # Test with positive time limit
     assert solver.solve(time_limit=1.0)
-    assert solver.status().exitstatus == ExitStatus.FEASIBLE or solver.status().exitstatus == ExitStatus.OPTIMAL
+    assert solver.status().exitstatus == ExitStatus.FEASIBLE
     
     # Test with negative time limit should raise ValueError
     try:


cpmpy/solvers/TEMPLATE.py
```diff
@@ -159,13 +159,21 @@ def __init__(self, cpm_model=None, subsolver=None):
...
         #           internally, the constructor relies on `add()` which uses the above solver native object(s)
         super().__init__(name="TEMPLATE", cpm_model=cpm_model)
 
 
+    @property
+    def native_model(self):
+        """
+            Returns the solver's underlying native model (for direct solver access).
+        """
+        return self.TPL_model
```
IgnaceBleukx, line 175: Hmm I would return the solver here. Having both Model and Solver is rare (only choco and OR-Tools I think)

cpmpy/solvers/gcs.py
```diff
@@ -138,6 +138,13 @@ def __init__(self, cpm_model=None, subsolver=None):
         # initialise everything else and post the constraints/objective
         super().__init__(name="Glasgow Constraint Solver", cpm_model=cpm_model)
 
+    @property
+    def native_model(self):
+        """
+            Returns the solver's underlying native model (for direct solver access).
+        """
+        return self.gcs
```
IgnaceBleukx, line 146: Hmm we should rename this to gcs_model, but not for this PR...

cpmpy/solvers/TEMPLATE.py
Dimosts: and hexaly. But indeed only a few have.  But if they do, we do want to get the model one

hbierlee:
I think we should add tests for these solver level interfaces, even if it's just `assert slv.native_model() is not None`

Dimosts:
Maybe then we should add similar tests for all mandatory parts of the solver interfaces. I will add some

Dimosts:
Changed plan. Just altered the scope of this PR to include the changes to solverinterface.py and the relevant tests all together. Need to re-review though.

tests/test_solverinterface.py
```diff
@@ -5,85 +5,320 @@
...
+    # Test with positive time limit
+    assert solver.solve(time_limit=1.0)
+    assert solver.status().exitstatus == ExitStatus.FEASIBLE or solver.status().exitstatus == ExitStatus.OPTIMAL
+    
+    # Test with negative time limit should raise ValueError
+    try:
+        solver.solve(time_limit=-1)
+        assert False, f"Solver {solver_name} should raise ValueError for negative time limit"
+    except ValueError:
+        pass  # Expected behavior
```
hbierlee, line 282: here it'd be good to have a problem which times out, say a medium difficult problem to solve with TO of 1 sec

tests/test_solverinterface.py
```diff
@@ -5,85 +5,320 @@
...
+    solver += [x | y]
+    
+    # Before solving, runtime should be None
+    assert solver.status().runtime is None
+    
+    # After solving, runtime should be recorded
+    solver.solve()
+    status = solver.status()
+    assert status.runtime is not None
+    assert status.runtime >= 0  # Should be non-negative
```
hbierlee, line 326: what if we resolve? Should the runtime increase further, or reset?

tests/test_solverinterface.py
```diff
@@ -5,85 +5,320 @@
...
+        solution_count = 0
+        def count_solution():
+            nonlocal solution_count
+            solution_count += 1
+            
+        if solver_name == "pysdd":
+            # pysdd doesn't support solution_limit
+            total = solver.solveAll(display=count_solution)
+        else:
+            total = solver.solveAll(display=count_solution, solution_limit=10)
```
hbierlee, line 353: another test: solve the same problem with `assert solver.solveAll(display=count_solution, sol_limit=2) == 2

tests/test_solverinterface.py
```diff
@@ -5,85 +5,320 @@
...
+    solver = solver_class()
+    
+    # Initially should have no objective
+    assert not solver.has_objective()
+    
+    # Add an objective if supported
+    try:
+        ivar = intvar(1, 10)
+        solver.minimize(ivar)
+        assert solver.has_objective()
```
hbierlee, line 299: assert ivar.value() == 1?

hbierlee: and solver.obj_value() == 1?

tests/test_solverinterface.py
```diff
@@ -5,85 +5,320 @@
...
+    
+    try:
+        solver_bool = solver.solver_var(bool_var)
+        solver_neg_bool = solver.solver_var(neg_bool_var)
+        
+        # Both should return something
+        assert solver_bool is not None
+        assert solver_neg_bool is not None
+        
+    except Exception as e:
```
hbierlee: it might be good to make this a specific exception, and ensure all solver who do not support neg_bools to raise that one. Catching `Exception` types means that real errors are hidden

tests/test_solverinterface.py
```diff
@@ -5,85 +5,320 @@
...
+@skip_on_missing_pblib(skip_on_exception_only=True)
+def test_solve_infeasible(solver_name):
+    solver_class = SolverLookup.lookup(solver_name)
+    solver = solver_class()
+
+    bvar = boolvar(shape=3)
+    x, y, z = bvar
+
+    solver += x.implies(y & z)
+    solver += ~ z
```
hbierlee, line 129: here (or in a separate test), we could do an additional solve call and status assertion to see whether the solver can transition from SAT status to unsat

tests/test_solverinterface.py
hbierlee: and perhaps test another `solver.maximize(ivar)` on the following lines

tests/test_solverinterface.py
Dimosts: Tried it a bit, issue is it slows down the tests a bit for all solvers, while also sometimes it completely crashes the tests due to an upstream fatal error from exact

hbierlee: Up to you :) Can also be a new issue/PR.

I think 1 second per solver (+transformation time) is an ok concession to ensure the timeout works. Failing timeouts are very destructive, hard to discover type of bugs. For exact, I'd skip the test and make an upstream issue.

tests/test_solverinterface.py
Dimosts: maximize indeed, for the values, we have separate test later on the test_objective (i will also add maximize there too).

tests/test_solverinterface.py
Dimosts: i tested locally, it resets everywhere, i.e. it is the last runtime

tests/test_solverinterface.py
Dimosts: We do have some specific tests for solveall in test_solveall.py, that cover this case. No need to do it also here I think, more to focus on the interface workings.

tests/test_solverinterface.py
Dimosts: Tried to make models that need more than 1 sec to find a solution, but to achieve that (without being trivially satisfied or unsatisfied) i would need to make advanced models or huge models that also need quite some time for the transformations for some solvers.

Also, some problems that are difficult for some solvers, are solved directly on others.

I think this goes too far to check for all solvers

Dimosts: From my tests though I think the time limit is maybe not enforced on pindakaas

hbierlee: yes.. definitely not trivial to test.

Similar problem in pindakaas, but we don't time out during the "transformation" process inside cpmpy or pindakaas, only once the solver starts. Definitely out-of-scope for this PR :) thanks for the attempt

PR #727, opened 2025-09-04
Skip psplib test

Dimosts:
The test randomly fails on github often, while also slows down a lot the testing. Also, it creates around 480 files each time. It does not seem needed so excluding it.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -19,7 +19,10 @@ import itertools
 EXAMPLES = glob(join("examples", "*.py")) + glob(join("examples", "csplib", "*.py"))
 ADVANCED_EXAMPLES = glob(join("examples", "advanced", "*.py"))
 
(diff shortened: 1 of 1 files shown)


tests/test_examples.py
```diff
@@ -19,7 +19,7 @@
 EXAMPLES = glob(join("examples", "*.py")) + glob(join("examples", "csplib", "*.py"))
 ADVANCED_EXAMPLES = glob(join("examples", "advanced", "*.py"))
 
-SKIPPED_EXAMPLES = ["ocus_explanations.py"]  # waiting for issues to be resolved
+SKIPPED_EXAMPLES = ["ocus_explanations.py", "psplib.py"]  # waiting for issues to be resolved
```
IgnaceBleukx: Maybe add the actual reason in comment here? Could put each of the excluding files on a separate line and put the comment next to it

PR #729, opened 2025-09-10
Rc2 maxsat solver

tias:
this needs more extensive testing...

(but its small test-case works)

also, there was a tricky aspect of using direct encoding (with weight 0 literals) in the objective... henk could you have a look?

### DIFF
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -268,12 +268,7 @@ class CPM_pysat(SolverInterface):
 
         # translate exit status
         if my_status is True:
-            # COP
-            if self.has_objective():
-                self.cpm_status.exitstatus = ExitStatus.OPTIMAL
-            # CSP
-            else:
-                self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+            self.cpm_status.exitstatus = ExitStatus.FEASIBLE
         elif my_status is False:
             self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
         elif my_status is None:
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -304,6 +304,8 @@ class CPM_pysat(SolverInterface):
         else: # clear values of variables
             for cpm_var in self.user_vars:
                 cpm_var._value = None
+            for enc in self.ivarmap.values():
+                enc._x._value = None
 
 
         return has_sol
--- a/cpmpy/solvers/rc2.py
+++ b/cpmpy/solvers/rc2.py
@@ -306,10 +306,12 @@ class CPM_rc2(CPM_pysat):
                 if x.name not in self.ivarmap:
                     self.solver_var(expr)  # make sure its known
                 # replace the intvar with its linear encoding
-                enc = self.ivarmap[expr.name]
-                assert enc.name == "wsum", "CPM_rc2: expected encoding to be wsum, got {enc}"
-                new_weights.extend(enc.args[0])
-                new_xs.extend(enc.args[1])
+                enc = self.ivarmap[x.name]
+                terms, const = enc.encode_term(w)
+                self.objective_const_ += const
+                encw, encx = zip(*terms)
+                new_weights.extend(encw)
+                new_xs.extend(encx)
             else:
                 raise NotImplementedError(f"CPM_rc2: Non supported objective {expr} (yet?)")
 
--- a/cpmpy/solvers/rc2.py
+++ b/cpmpy/solvers/rc2.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+##
+## pysat.py
+##
+"""
+    Interface to PySAT's RC2 MaxSAT solver API
+
+    PySAT is a Python (2.7, 3.4+) toolkit, which aims at providing a simple and unified
+    interface to a number of state-of-art Boolean satisfiability (SAT) solvers. It also
+    includes the RC2 MaxSAT solver.
+    (see https://pysathq.github.io/)
+
+    .. warning::
+        It does not support satisfaction, only optimization.
+    
+    Always use :func:`cp.SolverLookup.get("rc2") <cpmpy.solvers.utils.SolverLookup.get>` to instantiate the solver object.
+
+    ============
+    Installation
+    ============
+
+    Requires that the 'python-sat' package is installed:
+
+    .. code-block:: console
+
+        $ pip install pysat
+
+    If you want to also solve pseudo-Boolean constraints, you should also install its optional dependency 'pypblib', as follows:
+
+    .. code-block:: console
+
+        $ pip install pypblib
+
+    See detailed installation instructions at:
+    https://pysathq.github.io/installation
+
+    The rest of this documentation is for advanced users.
+
+    ===============
+    List of classes
+    ===============
+
+    .. autosummary::
+        :nosignatures:
+
+        CPM_rc2
+
+    ==============
+    Module details
+    ==============
+"""
+from threading import Timer
+from typing import Optional
+import warnings
+import pkg_resources
+
+from .solver_interface import SolverInterface, SolverStatus, ExitStatus
+from .pysat import CPM_pysat
+from ..exceptions import NotSupportedError
+from ..expressions.core import Comparison, Operator, BoolVal
+from ..expressions.variables import _BoolVarImpl, _IntVarImpl, NegBoolView
+from ..expressions.globalconstraints import DirectConstraint
+from ..transformations.linearize import canonical_comparison, only_positive_coefficients
+from ..expressions.utils import is_int, flatlist
+from ..transformations.comparison import only_numexpr_equality
+from ..transformations.decompose_global import decompose_in_tree
+from ..transformations.get_variables import get_variables
+from ..transformations.flatten_model import flatten_constraint, flatten_objective
+from ..transformations.linearize import linearize_constraint
+from ..transformations.normalize import toplevel_list, simplify_boolean
+from ..transformations.reification import only_implies, only_bv_reifies, reify_rewrite
+from ..transformations.int2bool import int2bool, _encode_int_var, _decide_encoding
+
+
+class CPM_rc2(CPM_pysat):
+    """
+    Interface to PySAT's RC2 MaxSAT solver API.
+
+    Creates the following attributes (see parent constructor for more):
+
+    - ``pysat_vpool``: a pysat.formula.IDPool for the variable mapping
+    - ``pysat_solver``: a pysat.examples.rc2.RC2() (or .RC2Stratified())
+    - ``ivarmap``: a mapping from integer variables to their encoding for `int2bool`
+    - ``encoding``: the encoding used for `int2bool`, choose from ("auto", "direct", "order", or "binary"). Set to "auto" but can be changed in the solver object.
+
+    The :class:`~cpmpy.expressions.globalconstraints.DirectConstraint`, when used, calls a function on the ``pysat_solver`` object.
+
+    Documentation of the solver's own Python API:
+    https://pysathq.github.io/docs/html/api/examples/rc2.html
+
+    .. note::
+        CPMpy uses 'model' to refer to a constraint specification,
+        the PySAT docs use 'model' to refer to a solution.
+
+    """
+
+    @staticmethod
+    def supported():
+        # try to import the package
+        try:
+            import pysat
+            # there is actually a non-related 'pysat' package
+            # while we need the 'python-sat' package, some more checks:
+            from pysat.formula import IDPool
+            from pysat.solvers import Solver
+            from pysat.examples import rc2
+
+            from pysat import card
+            CPM_rc2._card = card  # native
+
+            # try to import pypblib and avoid ever re-import by setting `_pb`
+            if not hasattr(CPM_rc2, ("_pb")):
+                try:
+                    from pysat import pb  # require pypblib
+                    """The `pysat.pb` module if its dependency `pypblib` installed, `None` if we have not checked it yet, or `False` if we checked and it is *not* installed"""
+                    CPM_rc2._pb = pb
+                except (ModuleNotFoundError, NameError):  # pysat returns the wrong error type (latter i/o former)
+                    CPM_rc2._pb = None  # not installed, avoid reimporting
+
+            return True
+        except ModuleNotFoundError:
+            return False
+        except Exception as e:
+            raise e
+
+    def __init__(self, cpm_model=None, subsolver=None, stratified=False):
+        """
+        Constructor of the native solver object
+
+        Requires a CPMpy model as input, and will create the corresponding
+        PySAT clauses and solver object
+
+        Only supports optimisation problems (MaxSAT)
+
+        Arguments:
+            cpm_model (Model(), optional): a CPMpy Model()
+            subsolver (None): ignored
+            stratified (bool, optional): use the stratified solver (default: False)
+        """
+        if not self.supported():
+            raise ImportError("PySAT is not installed. The recommended way to install PySAT is with `pip install cpmpy[pysat]`, or `pip install python-sat` if you do not require `pblib` to encode (weighted) sums.")
+        if cpm_model and cpm_model.objective_ is None:
+            raise NotSupportedError("CPM_rc2: only optimisation, does not support satisfaction")
+
+        from pysat.formula import IDPool, WCNF
+        from pysat.examples import rc2
+
+        # determine subsolver
+        if stratified:
+            self.pysat_solver = rc2.RC2Stratified(WCNF())
+        else:
+            self.pysat_solver = rc2.RC2(WCNF())
+        # fix an inconsistent API
+        self.pysat_solver.append_formula = lambda lst: [self.pysat_solver.add_clause(cl) for cl in lst]
+        self.pysat_solver.supports_atmost = lambda: False
+        # TODO: accepts native cardinality constraints, not sure how to make clear...
+
+        # objective value related
+        self.objective_value_ = None
+        self.objective_const_ = 0  # constant to add to the objective of the solver
+
+        # initialise the native solver object
+        self.pysat_vpool = IDPool()
+        self.ivarmap = dict()  # for the integer to boolean encoders
+        self.encoding = "auto"
+
+        # initialise everything else and post the constraints/objective (skip PySAT itself)
+        super(CPM_pysat, self).__init__(name="rc2", cpm_model=cpm_model)
+
+    @property
+    def native_model(self):
+        """
+            Returns the solver's underlying native model (for direct solver access).
+        """
+        return self.pysat_solver
+
+    def solve(self, time_limit=None):
+        """
+            Call the RC2 MaxSAT solver
+
+            Arguments:
+                time_limit (float, optional):   Maximum solve time in seconds. Auto-interrups in case the
+                                                runtime exceeds given time_limit.
+                                                
+                                                .. warning::
+                                                    Warning: the time_limit is not very accurate at subsecond level
+        """
+
+        # ensure all vars are known to solver
+        self.solver_vars(list(self.user_vars))
+
+        # the user vars are only the Booleans (e.g. to ensure solveAll behaves consistently)
+        user_vars = set()
+        for x in self.user_vars:
+            if isinstance(x, _BoolVarImpl):
+                user_vars.add(x)
+            else:
+                # extends set with encoding variables of `x`
+                user_vars.update(self.ivarmap[x.name].vars())
+
+        self.user_vars = user_vars
+
+        sol = None
+        # set time limit
+        if time_limit is not None:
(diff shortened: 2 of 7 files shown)


cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,351 @@
...
+                for encw, encx in tlst:
+                    if encw != 0:
+                        new_weights.append(encw)
+                        new_xs.append(encx)
+            elif isinstance(x, int):
+                const += w*x
+            else:
+                raise NotImplementedError(f"CPM_rc2: Non supported term {w,x} in objective {flat_obj} (yet?)")
+
+        # positive weights only, flip negative
```
hbierlee: don't we have `only_positive_coefficients` for this? Or are you not using that because it's a `Comparison`?

tests/test_constraints.py
```diff
@@ -14,6 +14,7 @@
 #   make sure that `SolverLookup.get(solver)` works
 # also add exclusions to the 3 EXCLUDE_* below as needed
 SOLVERNAMES = [name for name, solver in SolverLookup.base_solvers() if solver.supported()]
+SOLVERNAMES = [n for n in SOLVERNAMES if n not in ["rc2"]]  # only supports optimisation
```
hbierlee: on or the other line should be removed?

tests/test_rc2_obj.py
```diff
@@ -0,0 +1,231 @@
...
+        model += x[1].implies(x[2])  # if x[1] is true, then x[2] must be true
+        
+        # Solve with RC2
+        solver = CPM_rc2(model)
+        solved = solver.solve()
+        
+        self.assertTrue(solved)
+        self.assertIsNotNone(solver.objective_value())
+        # The optimal solution should have x[0]=True, x[1]=True, x[2]=True for objective value 3
+        # But RC2 might find a different solution due to the constraints
```
hbierlee: sorry, what? Why won't it just find the optimal solution where all vars are assigned true?

tests/test_solvers.py
```diff
@@ -838,6 +842,9 @@ def test_incremental_objective(self, solver):
         assert s.objective_value() == 25
 
     def test_incremental(self, solver):
+        if solver == "rc2":
+            pytest.skip("not incremental")
```
hbierlee: I thought rc2 was incremental?

tests/test_rc2_obj.py
```diff
@@ -0,0 +1,231 @@
...
+        m.maximize(-4*x[0] + 3*x[1])
+        
+        # Solve with RC2
+        solver = CPM_rc2(m)
+        solved = solver.solve()
+        
+        self.assertTrue(solved)
+        self.assertEqual(solver.objective_value(), 3)
+        self.assertEqual(list(x.value()), [False, True])
+
```
hbierlee, line 247: perhaps we can have a test for incrementally updating the objective (ie adding additional soft clauses)? Or do we not expose this yet?

tests/test_rc2_obj.py
hbierlee: @tias this I still don't understand, and in my case it just always finds the optimal solution with objective value 3

tests/test_rc2_obj.py
hbierlee: not exposed

tests/test_solvers.py
```diff
@@ -1008,10 +1010,13 @@ def test_vars_not_removed(self, solver):
     # minizinc: ignore inconsistency warning when deliberately testing unsatisfiable model
     @pytest.mark.filterwarnings("ignore:model inconsistency detected")
     def test_false(self, solver):
-        assert not cp.Model([cp.boolvar(), False]).solve(solver=solver)
+        try:
```
hbierlee: note to self ; this might be a mistake when merging, I'll remove this change

cpmpy/solvers/pysat.py
```diff
@@ -229,17 +229,7 @@ def solve(self, time_limit=None, assumptions=None):
...
-        user_vars = set()
-        for x in self.user_vars:
-            if isinstance(x, _BoolVarImpl):
-                user_vars.add(x)
-            else:
-                # extends set with encoding variables of `x`
-                user_vars.update(self.ivarmap[x.name].vars())
-
-        self.user_vars = user_vars
+        self.user_vars = get_user_vars(self.user_vars, self.ivarmap)
```
tias: get_user_vars is a too generic name... I think `bool_user_vars` would be better aligned with its goal

cpmpy/solvers/pysat.py
```diff
@@ -254,40 +244,38 @@ def solve(self, time_limit=None, assumptions=None):
...
+        elif feasible is None:
             # can happen when timeout is reached...
             self.cpm_status.exitstatus = ExitStatus.UNKNOWN
         else:  # another?
-            raise NotImplementedError(my_status)  # a new status type was introduced, please report on github
+            raise NotImplementedError(feasible)  # a new status type was introduced, please report on github
+
+        return self._process_solution(self.pysat_solver.get_model())
 
+    def _process_solution(self, sol):
```
tias, line 278: document why we want this in a separate function:
This function can be re-used in subclasses (e.g. rc2)

cpmpy/solvers/pysat.py
```diff
@@ -436,9 +426,13 @@ def _post_constraint(self, cpm_expr):
 
         elif isinstance(cpm_expr, Comparison): # root-level comparisons have been linearized
             if isinstance(cpm_expr.args[0], Operator) and cpm_expr.args[0].name == "sum":
-                self.pysat_solver.append_formula(self._pysat_cardinality(cpm_expr))
+                c = self._pysat_cardinality(cpm_expr)
+                print("C", c)
```
tias: is this debug leftover?

cpmpy/solvers/pysat.py
```diff
@@ -436,9 +426,13 @@ def _post_constraint(self, cpm_expr):
...
+                self.pysat_solver.append_formula(c)
             elif isinstance(cpm_expr.args[0], Operator) and cpm_expr.args[0].name == "wsum":
-                self.pysat_solver.append_formula(self._pysat_pseudoboolean(cpm_expr))
+                c = self._pysat_pseudoboolean(cpm_expr)
+                print("PB", c)
```
tias: is this debug leftover?

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,298 @@
...
+        # In order for all features to work (e.g. `process` option), RC2 should be bootstrapped from a formula
+        self.pysat_solver = WCNFPlus()
+
+        def append_formula(cnf, weight=None):
+            """Some RC2 oracles/solvers support AtMostK constraints in the form of CNFPlus objects, which can be passed through this function in addition to regular clauses."""
+            clauses = []
+            atmosts = []
+            if isinstance(cnf, list):
+                clauses = cnf
+            elif isinstance(cnf, tuple):
```
tias: euh... what?
so somebody will use 'append_formula' somewhere, and it will give it a tuple, and they will know what this tuple should be, and elsewhere we should know that any tuples is an atmost?

this is is too implicit...

in pysat itself we do not support native atmostk? then I find it weird that we hack it in here, and especially in this way...

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,298 @@
...
+                atmosts = cnf
+            else:  # CNFPlus
```
tias: is this expected? can people given an entire CNFPlus to us? or to pysat for that manner?
how would you typehint this `add_formula`... and is the typehint the same as the original `add_formula` function? If not, this is imho not right...

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,298 @@
...
+                self.pysat_solver.append(c, weight=weight)
+            for c in atmosts:
+                self.pysat_solver.append(c, weight=weight, is_atmost=True)
+
+        self.pysat_solver.append_formula = append_formula
+        self.pysat_solver.add_clause = lambda c, weight=None: self.pysat_solver.append(c, weight=weight)
+        # Note: in case it becomes neccessary in the future to support CNFPlus objects for `add_clause`, this function should be replaced by: `lambda c, weight=None: append_formula([c], weight=weight)`
+
+        # We collect all AtMostK constraints in the WCNFPlus object, then encode them if the solver (set at `solve`) does not accept them
+        self.pysat_solver.supports_atmost = lambda: True
```
tias: weird that this is with a lambda

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,298 @@
...
+        self.solver_vars(list(self.user_vars))
+
+        # the user vars are only the Booleans (e.g. to ensure solveAll behaves consistently)
+        self.user_vars = get_user_vars(self.user_vars, self.ivarmap)
+
+        # TODO: set time limit (awaiting upstream PR https://github.com/pysathq/pysat/pull/211)
+        if time_limit is not None:
+            raise NotImplementedError("CPM_rc2: time limit not yet supported")
+
+        # hack to support decision problems
```
tias: I don't think we should? we can just raise that the solver is not for that?

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,298 @@
...
+        if not self.has_objective():
+            self.pysat_solver.add_clause([self.pysat_solver.nv + 1], weight=1)
+
+        # determine subsolver
+        default_kwargs = {"solver": "glucose3", "adapt": True, "exhaust": True, "minz": True}
+        stratified = kwargs.get("stratified", True)
```
tias, line 172: why handle stratified differently? Can just be in default_kwargs as true too?

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,298 @@
...
+        slv_kwargs = kwargs if kwargs else default_kwargs
+
+        # get subsolver kwarg or default, then check if it supports AtMostK constrains
+        sub_solver = slv_kwargs.get("solver", default_kwargs["solver"])
+        if sub_solver and not Solver(name=sub_solver).supports_atmost():
+            # encode AtMostK constraints since they are unsupported by sub-solver oracle
+            atmosts = self.pysat_solver.atms
+            self.pysat_solver.__class__ = WCNF
```
tias: a line like this is black magic and to be avoided

cpmpy/transformations/linearize.py
```diff
@@ -673,6 +673,13 @@ def canonical_comparison(lst_of_expr):
 
     return newlist
 
+def only_positive_coefficients_(ws, xs):
+    """ Helper function which weight """
```
tias: with weight?

tests/test_rc2_obj.py
tias: might be from before I got the objective right... yes, it should just check for 3 and that should work : )

cpmpy/solvers/rc2.py
hbierlee: - the reason that it's more complicated here with the overriding of `append_formula` is that we collect the encoding in WCNFPlus and then instantiate the solver in `solve` from the WCNFPlus. This is so that we can give the solver args there as in most solvers (Exact being the only exception). The implementation may be simplified if we do require the RC2 solve args to be given to __init__ so we don't have this `WCNFPlus` in-between.
- in pysat, we do support native atmost k's (https://github.com/CPMpy/cpmpy/blob/master/cpmpy/solvers/pysat.py#L547)
- Side note: nobody is supposed to use `append_formula`, it's a method of a "private" field (although I suppose they could get it through the `native_model`).

cpmpy/solvers/rc2.py
hbierlee: Mostly this just fits all the possible outputs of pysat's encoding features (`Card`, `PB_enc`) onto the inputs of WCNFPlus.

cpmpy/solvers/rc2.py
hbierlee: Internally, this is how PySAT checks for this

cpmpy/solvers/rc2.py
hbierlee: I'm also fine with that, it just seemed a very easy extension + allows us to more easily test RC2

cpmpy/solvers/rc2.py
hbierlee: Because `stratified` is not an RC2 argument, you use it by using the different class. I can add it to `default_kwargs`, but then I need to remove it again when I pass the args to `compute`

cpmpy/solvers/rc2.py
hbierlee: yeah.. there's no official way to convert WCNFPlus to WCNF, and this avoids copying. I've made a change which avoids the downclass though.

tias:
> Small comments are resolved, I think there are two knots to cut:
> 
> * Simple one: should we keep in the slight hack to enable solving decision problems with RC2? To me it seems useful, and the hack is theoretically sound
> * See comment near WCNFPlus/AtMostK stuff about the implementation, but I believe the main reason it's made complicated is because of the in-between WCNFPlus object required to keep solve arguments in `solve` rather than in `__init__`. I see 3 options:
>   
>   1. Keep AtMostK stuff as implemented, so that we support all features and sub-solvers for RC2 (note that pysat does support atmostk).
>   2. Pass encoded constraints directly to RC2 without the WCNF[Plus] in-between, this may simplify the implementation, and keeps the AtMostK native feature, however, we have to move (some) solver args to `__init__` (like in Exact), and other features like `process` will be lost as they only work if RC2 is bootstrapped from a formula.
>   3. Omit AtMostK support, just always encode everything into a WCNF object, passed to RC2 at solve time (which keeps solve arguments there)

special cases become maintenance burdens, so
* no, only optimisation (do you want to spend time on a bugreport where a user complains that solving a SAT Problem with a MaxSAT solver is slow or has some unimaginable behaviour because Alexey did not expect a constant objective? if somebody has a good enough reason to want to do this, they can add the constant objective themselves)
* 3: omit atmostk support and keep the original design decision of having a wcnf object that is passed to rc2 at solve time

cpmpy/solvers/pysat.py
```diff
@@ -233,17 +233,7 @@ def solve(self, time_limit:Optional[float]=None, assumptions:Optional[List[_Bool
...
-        user_vars = set()
-        for x in self.user_vars:
-            if isinstance(x, _BoolVarImpl):
-                user_vars.add(x)
-            else:
-                # extends set with encoding variables of `x`
-                user_vars.update(self.ivarmap[x.name].vars())
-
-        self.user_vars = user_vars
+        self.user_vars = encode_user_vars(self.user_vars, self.ivarmap)
```
tias: I think the function is better called 'only_bool_user_vars', it is now confusing that it takes user vars and instead of manipulating the dict, what it does is return a new dict...

An alternative is to not have it return anything, and change the code so that for each non BoolVarImpl, it first removes the int and then updates with the bools

tias: I think the function is better called 'only_bool_user_vars', it is now confusing that it takes user vars and instead of manipulating the dict, what it does is return a new dict...

An alternative is to not have it return anything, and change the code so that for each non BoolVarImpl, it first removes the int and then updates with the bools

cpmpy/solvers/pysat.py
```diff
@@ -258,40 +248,39 @@ def solve(self, time_limit:Optional[float]=None, assumptions:Optional[List[_Bool
...
 
+    def _process_solution(self, sol):
+        """Process solution `sol` from PySAT, and handles post-processing int2bool. Used in subclass e.g. RC2."""
         # True/False depending on self.cpm_status
         has_sol = self._solve_return(self.cpm_status)
 
         # translate solution values (of user specified variables only)
         if has_sol:
-            sol = frozenset(self.pysat_solver.get_model())  # to speed up lookup
+            sol = frozenset(sol)
```
tias: please keep comment '# to speed up lookup'

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,287 @@
...
+        self.cpm_status = SolverStatus(self.name)
+        self.cpm_status.runtime = solver.oracle_time()
+
+        # translate exit status
+        if solution is None:
+            # `None` for either unsat or unknown!
+            self.cpm_status.exitstatus = ExitStatus.UNKNOWN if solver.interrupted else ExitStatus.UNSATISFIABLE
+        elif self.has_objective():
+            self.cpm_status.exitstatus = ExitStatus.OPTIMAL
+        else:
```
tias, line 200: else case no longer exists, will always have objective

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,287 @@
...
+            self.cpm_status.exitstatus = ExitStatus.FEASIBLE
+
+        return self._process_solution(solution)
+
+    def _process_solution(self, sol):
```
tias, line 205: not needed as separate function? can just call has_sol = self._process and then set the objective and only then return?

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,287 @@
...
+        from pysat.examples import rc2
+
+        # ensure all vars are known to solver
+        self.solver_vars(list(self.user_vars))
+
+        # the user vars are only the Booleans (e.g. to ensure solveAll behaves consistently)
+        self.user_vars = encode_user_vars(self.user_vars, self.ivarmap)
+
+        if not self.has_objective():
+            if "PYTEST_CURRENT_TEST" in os.environ:  # support decision problems for testing purposes
```
tias: I don't think this embraces "special cases become maintenance burdens".

if a pytest wants to call it without an objective, its the pytest that has to find a workaround, not this solver object. Please just raise notsupported.

cpmpy/solvers/rc2.py
```diff
@@ -0,0 +1,287 @@
...
+            raise NotImplementedError(f"CPM_rc2: Non supported objective {obj} (yet?)")
+
+        terms, cons, k = _encode_lin_expr(self.ivarmap, xs, weights, self.encoding)
+
+        self += cons
+        const += k
+
+        terms = [(w, x) for w,x in terms if w != 0]  # positive coefficients only
+        ws, xs = zip(*terms)  # unzip
+        new_weights, new_xs, k = only_positive_coefficients_(ws, xs) # this is actually only_non_negative_coefficients
```
tias: since you introduce that function, feel free to rename it!
(also the comment 2 lines about about positive only should be non-zero only...)

cpmpy/transformations/int2bool.py
```diff
@@ -95,6 +96,28 @@ def _encode_int_var(ivarmap, x, encoding):
         return (ivarmap[x.name], ivarmap[x.name].encode_domain_constraint())
 
 
+def _encode_lin_expr(ivarmap, xs, weights, encoding, cmp=None):
```
tias, line 99: please add minimal documentation

cpmpy/transformations/linearize.py
```diff
@@ -546,6 +546,13 @@ def canonical_comparison(lst_of_expr):
 
     return newlist
 
+def only_positive_coefficients_(ws, xs):
+    """Helper function to make coefficients positive for Boolean terms given as list of coefficients `ws` and list of Boolean variables `xs`."""
+    indices = {i for i, (w, x) in enumerate(zip(ws, xs)) if w < 0 and isinstance(x, _BoolVarImpl)}
+    nw, na = zip(*[(-w, ~x) if i in indices else (w, x) for i, (w, x) in enumerate(zip(ws, xs))])
+    cons = sum(ws[i] for i in indices)
```
tias: 'cons' surely is highly confusing with 'constraints', rather than 'constant'? especially because we use 'cons' for constrains elsewhere, see rc2

cpmpy/solvers/rc2.py
hbierlee: It overrides `super()._process_solution` (from `CPM_pysat`). What you say is possible, but then `CPM_rc2` will still have that function, but incorrect, since it needs to do a little more than pysat. This feels more natural?

I've added a docstring in the subclass too, now

cpmpy/solvers/rc2.py
hbierlee: Yes, but the already-existing `only_positive_coefficients` then has to be renamed as well (in a lot of code).

So, alternatives:

1. rename only `only_positive_coefficients_` to `only_non_negative_coefficients_`?
2. rename both like that?
3. or did you mean you don't like the end `_`, so it should be e.g. `only_positive_coefficients_from_terms`?

cpmpy/solvers/pysat.py
hbierlee: All the encoding functions in int2bool follow this pattern: `encode_type_of_thing`, and they are all functional and immutable (just return the encoding of what is passed in). Should I refactor that in a separate PR perhaps (make them all `only_bool_type_of_thing`? I agree it's more consistent with the existing naming.

hbierlee:
I excluded it the same way as we do with pysdd, in conftest, which will also fail tests if you do `--solver pysdd`.

As discussed internally, I will now change this to remove RC2 explicitly from non-optimization tests.

Issue #737, opened 2025-09-17
NValue decompose issue with some inputs


Dimosts:
NValue decompose has issues, even in simple cases where we want to post the globalfunction on a transposed array of variables.

The issue is that it is not contiguous anymore.

Example:

```
  ivar = intvar(1, 10, shape=(3,3))
  solver += [NValue(i) <= 3 for i in ivar.T]
```

This fails with any solver needing to decompose the constraint.

IgnaceBleukx:
Well... why do we even need the call to `cpm_array` in the decomp? Quick fix would be to just remove that, and/or iterate over the array ourselves in the `cp.any` call

PR #738, opened 2025-09-17
Not contiguous array handling in cpm_array

Dimosts:
Issue #737 revealed that our cpm_array function does not handle non-contiguous arrays, which can happen easily even with just a transpose.

In this PR, I just add a check for that, and if it is not contiguous, we make it.

Also added tests both with the nvalue constraint, where it was originally found, and a separate one for cpm_array alone.

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -256,7 +256,7 @@ def cpm_array(arr):
     """
     if not isinstance(arr, np.ndarray):
         arr = np.array(arr)
-    elif not arr.flags['C_CONTIGUOUS'] and not arr.flags['F_CONTIGUOUS']:   # Ensure the array is contiguous before creating NDVarArray
+    elif not arr.flags['C_CONTIGUOUS'] and not arr.flags['F_CONTIGUOUS']:   # Ensure the array is contiguous
         arr = np.ascontiguousarray(arr)
     
     order = 'F' if arr.flags['F_CONTIGUOUS'] else 'C'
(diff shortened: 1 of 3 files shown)


cpmpy/expressions/variables.py
```diff
@@ -256,6 +256,9 @@ def cpm_array(arr):
     """
     if not isinstance(arr, np.ndarray):
         arr = np.array(arr)
+    elif not arr.flags['C_CONTIGUOUS'] and not arr.flags['F_CONTIGUOUS']:   # Ensure the array is contiguous
```
IgnaceBleukx: Numpy actually provides a test for this: `arr.flags['FORC']`
https://numpy.org/doc/2.0/reference/generated/numpy.ndarray.flags.html

tests/test_expressions.py
```diff
@@ -717,6 +717,14 @@ def test_dict(self):
         
 class TestUtils(unittest.TestCase):
 
+    def test_cpm_array(self):
+        x = cp.intvar(0,10, shape=(3, 3))
+        self.assertIsInstance(cpm_array(x), NDVarArray)
+        self.assertEqual(cpm_array(x).shape, (3, 3))
+
+        self.assertIsInstance(cpm_array(x.T), NDVarArray)
+        self.assertEqual(cpm_array(x.T).shape, (3, 3))
```
IgnaceBleukx: well, (3,3) is quite uninteresting ;) Can you change it to (3,4)?

IgnaceBleukx (commented):
Two small comments

tests/test_expressions.py
Dimosts: that is so true

PR #739, opened 2025-09-18
Fix minor bugs in existing examples

kostis-init:


examples/csplib/prob026_sport_scheduling.py
```diff
@@ -38,7 +38,7 @@ def sport_scheduling(n_teams):
         # can be written cleaner, see issue #117
         # sum((home == t) | (away == t), axis=1) <= 2
         for p in range(n_periods):
-            model += sum((home[p] == t) | (away[p] == t)) <= 2
+            model += sum((home[:, p] == t) | (away[:, p] == t)) <= 2
```
hbierlee, line 41: if the model ran in the test, wouldn't this have thrown an exception?

hbierlee (approved):

examples/csplib/prob026_sport_scheduling.py
kostis-init: The code was already syntactically correct, but there was a logical/modelling error (adding the constraint through the wrong axis)

PR #742, opened 2025-09-19
Fix non-iterable + index range and add constraint nr 1

ThomSerg:
The fuzz-tester detected an issue with the decomposition for the `Precedence` global.

When j is 0 or 1, `args[:j]) == s` can result in a single bool or boolvar, which is not iterable. Wrapping it with `cpm_array` fixes it. 

Two other issues I noticed when going through the code:
1) j should start from 1, as `args[:j]` will otherwise be empty
2) constraint nr 1 from the paper was missing? `constraints.append(args[0] != t)`

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -774,11 +774,14 @@ class Precedence(GlobalConstraint):
         args, precedence = self.args
         constraints = []
         for s,t in zip(precedence[:-1], precedence[1:]):
-            for j in range(len(args)):
+            # constraint 1 from paper
+            constraints.append(args[0] != t) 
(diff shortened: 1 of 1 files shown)


Dimosts (changes requested):
Looks good. 

Needs one more thing: 2 tests, to showcase the 2 cases that fail without it (i.e., one where without the cpm_array it crashes, and 1 simple where the first constraint is violated)

ThomSerg:
Removed the `cpm_array` part, can't seem to replicate it. Probably more caused by the fuzztester itself. Will open a new issue/pull request if I encounter it again (now that the fuzzer has been improved).

Dimosts (approved):

PR #746, opened 2025-09-22
Skip slow examples

IgnaceBleukx:
Some of the examples we have are very slow for certain solvers.
This PR ensures to exclude any examples which ran for more than 5s on my machine.

The pytest-timeout library apparently does not kill a sub-process (i.e., the solver) after the time-limit has been reached.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -19,8 +19,8 @@ import itertools
 EXAMPLES = glob(join("examples", "*.py")) + glob(join("examples", "csplib", "*.py"))
 ADVANCED_EXAMPLES = glob(join("examples", "advanced", "*.py"))
 
-EXAMPLES = sorted(EXAMPLES, key=str)
-ADVANCED_EXAMPLES = sorted(ADVANCED_EXAMPLES, key=str)
+EXAMPLES = sorted(EXAMPLES)
+ADVANCED_EXAMPLES = sorted(ADVANCED_EXAMPLES)
 
 SKIPPED_EXAMPLES = [
                     "ocus_explanations.py", # waiting for issues to be resolved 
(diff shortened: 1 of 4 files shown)


tests/test_examples.py
```diff
@@ -19,10 +19,21 @@
...
                     "psplib.py" # randomly fails on github due to file creation
-                    ]  
+                    ]
+
+SKIP_MIP = ['npuzzle.py', 'tst_likevrp.py', 'sudoku_', 'pareto_optimal.py',
+            'prob009_perfect_squares.py', 'blocks_world.py', 'flexible_jobshop.py',
+            'mario', 'pareto_optimal','prob006_golomb.py', 'tsp.py', 'prob028_bibd.py', 'prob001_car_sequence.py'
+            ]
+
+SKIP_MZN = ['blocks_world.py', 'flexible_jobshop.py', 'pareto_optimal.py', 'npuzzle.py', 'sudoku_']
```
hbierlee, line 35: why sudoku_? intended to be used with `str.start_with`?

tests/test_examples.py
```diff
@@ -19,10 +19,21 @@
 EXAMPLES = glob(join("examples", "*.py")) + glob(join("examples", "csplib", "*.py"))
 ADVANCED_EXAMPLES = glob(join("examples", "advanced", "*.py"))
 
+EXAMPLES = sorted(EXAMPLES, key=str)
```
hbierlee: why `key=`?

hbierlee (approved):
seems good, we can try to fix the timeout in another PR.

I assume giving all these problems an even smaller default instance is not a good option?

Otherwise ok to merge.

tests/test_examples.py
IgnaceBleukx: It's used as "if name in filename", but adding the `_` runs the "normal" sudoku example still

tests/test_examples.py
IgnaceBleukx: You're right, that is not required here

IgnaceBleukx:
I tried putting a smaller default example in the csplib ones, but was still too slow indeed.

hbierlee:
Every additional example that runs is a plus (both from testing, and user perspective), but definitely no need to do to much trouble. LGTM

PR #750, opened 2025-09-25
Upgrade to pindakaas 0.2

hbierlee:
This time from right remote. This PR upgrade pindakaas to 0.2

TODO:

- [x] Enable `get_core`
- [x] Workaround `get_vars(1)` bug
- [x] Pin version?

### DIFF
--- a/cpmpy/solvers/pindakaas.py
+++ b/cpmpy/solvers/pindakaas.py
@@ -216,7 +216,7 @@ class CPM_pindakaas(SolverInterface):
             # insert if new
             if cpm_var.name not in self._varmap:
                 self._varmap[cpm_var.name] = self.pdk_solver.new_vars(1)[0]  # TODO workaround https://github.com/pindakaashq/pindakaas/issues/159
-                return self._varmap[cpm_var.name]
+            return self._varmap[cpm_var.name]
         elif isinstance(cpm_var, _IntVarImpl):  # intvar
             if cpm_var.name not in self.ivarmap:
                 enc, cons = _encode_int_var(
--- a/cpmpy/solvers/pindakaas.py
+++ b/cpmpy/solvers/pindakaas.py
@@ -215,7 +215,7 @@ class CPM_pindakaas(SolverInterface):
         elif isinstance(cpm_var, _BoolVarImpl):  # positive literal
             # insert if new
             if cpm_var.name not in self._varmap:
-                (self._varmap[cpm_var.name],) = self.pdk_solver.new_vars(1)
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -30,8 +30,8 @@ def get_version(rel_path):
     "pysdd": ["pysdd"],
     "gcs": ["gcspy"],
     "cpo": ["docplex"],
-    "pindakaas": ["pindakaas"],
-    "pumpkin": ["pumpkin-solver>=0.2.1"]
+    "pumpkin": ["pumpkin-solver>=0.2.1"],
+    "pindakaas": ["pindakaas==0.2.0"],
```
IgnaceBleukx: I'm not actually a fan of pinning the exact version of the solver. I feel we should always be supporting the lastest version of any solver, even if the tests fail on the CI because of it.
It will at least warn us of the new version being out, and typically the fix is quite easy to do as well.

For I would change this to `pindakaas>=0.2.0`

cpmpy/solvers/pindakaas.py
```diff
@@ -215,7 +215,7 @@ def solver_var(self, cpm_var):
         elif isinstance(cpm_var, _BoolVarImpl):  # positive literal
             # insert if new
             if cpm_var.name not in self._varmap:
-                (self._varmap[cpm_var.name],) = self.pdk_solver.new_vars(1)
+                self._varmap[cpm_var.name] = self.pdk_solver.new_vars(1)[0]  # TODO workaround current issue
```
IgnaceBleukx: What issue?

setup.py
hbierlee: Not sure I agree, but I'll go with that then :)

cpmpy/solvers/pindakaas.py
```diff
@@ -215,8 +215,8 @@ def solver_var(self, cpm_var):
         elif isinstance(cpm_var, _BoolVarImpl):  # positive literal
             # insert if new
             if cpm_var.name not in self._varmap:
-                (self._varmap[cpm_var.name],) = self.pdk_solver.new_vars(1)
-            return self._varmap[cpm_var.name]
+                self._varmap[cpm_var.name] = self.pdk_solver.new_vars(1)[0]  # TODO workaround https://github.com/pindakaashq/pindakaas/issues/159
+                return self._varmap[cpm_var.name]
```
IgnaceBleukx: This indent is the problem of failing tests. It now only returns something when the var did not exist yet

IgnaceBleukx (changes requested):
Thank you Dimos for noticing this, I found the problem is the extra indent in the solver_var function

hbierlee:
Thanks. If the rest is ok, then I will merge if the tests succeed, then.

PR #756, opened 2025-09-30
Change `cp.sum(*iterable, **kwargs)` to `cp.sum(iterable, **kwargs)`

hbierlee:
This is in line with the Python's `builtins.sum`

### DIFF
--- a/cpmpy/expressions/python_builtins.py
+++ b/cpmpy/expressions/python_builtins.py
@@ -142,15 +142,14 @@ def min(*iterable, **kwargs):
     return Minimum(iterable)
 
 
-def sum(*iterable, **kwargs):
+def sum(iterable, **kwargs):
     """
         sum() overwrites the python built-in to support decision variables.
 
         if iterable does not contain CPMpy expressions, the built-in is called
         checks if all constants and uses built-in sum() in that case
     """
-    if len(iterable) == 1:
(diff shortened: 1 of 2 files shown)


hbierlee:
I checked the behaviour again for all operators, they seem consistent with Python 3.12. No examples needed updating. So I think this is good to merge.

One more change I propose:

The straightforward way to check that our cpmpy override behaviour is the same as the python built-in is by comparing the signatures. Currently, they are the same, except for `max` and `min`, which for Python is:

```py
max(arg1: SupportsLessThanT, arg2: SupportsLessThanT, /, *_args: SupportsLessThanT, key: None=...) -> SupportsLessThanT
```

but we have:

```py
max(*iterable, **kwargs)
```

The behaviour is almost the same, except that we'll get a slightly different error. I will align that as well, then it's ready for review/merge.

hbierlee:
Ok, I looked into getting the same type signature for max/min, but it seems overkill for our case. It uses some pretty extensive typing to accomplish the argument overloading with `@overload` (https://typing.python.org/en/latest/spec/overload.html), plus there are not 2 but 6 `max` overloaded functions. I think our current solutions suffices, you almost get the same error as in Python, just from within the function rather than from the top.

To maintain this kind of complexity, we'd need some way to dynamically get the same signatures as the Python built-ins, but this is also out-of-scope.

Issue #751, opened 2025-09-25
Timeout on Tuner not properly enforced


ElFosco:
The timeout is not correctly enforced in both **ParameterTuner** and **GridSearchTuner**, which can lead to errors or unexpected behaviour.  

---

#### Problem 1: Init solver call not constrained by timeout
Both tuners perform an **Init solver** call without applying the timeout:  

```python
# Init solver
solver = SolverLookup.get(self.solvername, self.model)
solver.solve(**self.best_params)
```
#### Problem 2: Negative timeout values

When checking if the algorithm has exceeded the time limit, the computed `time_limit - (time.time() - start_time)` can become negative, which is not supported by OR-Tools:  

```python
timeout = self.best_runtime
# set timeout depending on time budget
if time_limit is not None:
    timeout = min(timeout, time_limit - (time.time() - start_time))
# run solver
solver.solve(**params_dict, time_limit=timeout)
```

OrestisLomis:
Fixed by #970

PR #757, opened 2025-09-30
adding multi-instance-tuner

ElFosco:
Adding multi-instance function for both ParameterTuner and GridSearchTuner.
This pull request fixes Issue #751.


ElFosco:
Sorry, apparently there is a bug that I just noticed now. I am going to fix it and should be ready

PR #758, opened 2025-10-01
JSPLib data loader

IgnaceBleukx:
Similar spirit as the XCSP3 competition and PSPlib data loaders, but for jobshop instances from the JSPLib github repo.

It is a bit hacky because github does not allow you to download an arbitrary directory from a repo, so had to download the full repo instead and extract whatever is relevant.


IgnaceBleukx:
Closing as this is part of #1037

PR #762, opened 2025-10-03
CPLEX solveAll time limit reached status

### LINKED ISSUE #761: SolveAll test sensitive to system performance
`test_status_solveall` in `test_solvers.py` adds a bunch of decision variables to a model and then assumes that the solver would not be able to enumerate all solutions within a time limit of 1 second. This assumption is too sensitive to system performance / solver performance. On `cplex` on a more powerful machine it is able to return OPTIMAL within the timelimit, causing the test to fail.

ThomSerg:
Fixes #761, missing exit status case for `solveAll` when time limit is reached.

### DIFF
--- a/cpmpy/solvers/cplex.py
+++ b/cpmpy/solvers/cplex.py
@@ -623,7 +623,7 @@ class CPM_cplex(SolverInterface):
             if opt_sol_count == solution_limit:
                 self.cpm_status.exitstatus = ExitStatus.FEASIBLE 
             else:
-                if cplex_status == "Unknown": # reached time limit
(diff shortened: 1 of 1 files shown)


ThomSerg:
Maybe defaulting to `OPTIMAL` for all other exit statuses is not that safe. If something ever changes upstream, suddenly we could incorrectly return a bunch of optimal.

The correct exit status to match against seems to be `"all reachable solutions enumerated, integer optimal"`

IgnaceBleukx (commented):
I would prefer to have an exhaustive elif-trail to check each status, where the default is to raise an error if the status is unexpected.

In the `solve()` function there are a few cases which are not here (e.g., "JobFailed")

Issue #760, opened 2025-10-02
CPLEX Non-convex QCP in constraints


ThomSerg:
A transformation seems to be missing in the waterfall for CPLEX. Non-convex expressions, which are only supported in the objective function, can end up in the posted constraints.

Example model:
```python
import cpmpy as cp

p = cp.boolvar()
x,y = cp.intvar(0, 10, shape=2)
m = cp.Model(p.implies(x*y != 1))

m.solve(solver="cplex")
```
```console
>> NotImplementedError: Translation of cplex status Non-convex QCP to CPMpy status not implemented
```
The above model transforms to:
```
((IV0) * (IV1)) == (IV2)   <-- non-convex
(BV0) -> (sum([1, -100] * [IV2, BV1]) <= 0)
(BV0) -> (sum([1, -2] * [IV2, BV1]) >= 0)
sum([-1, 1] * [BV0, BV1]) <= 0
```

tias:
fixed, now throws nonsupportd error

PR #763, opened 2025-10-03
CPLEX unsupported quadratic equality

ThomSerg:
Raises proper error for #760. Is still not perfect, since non-convex equality can appear after transformation (e.g. decomposition), making the error message confusing for the user.

Proper fix would be to look into the linearisation of the product of two integer decision variables.


hbierlee (approved):
This is fine if there is no linearization of multiplication. Two thoughts:

- We're sure that a, b are both integer variables? 
- Maybe add an issue for encoding `mul`?

Good to merge if CI passes

PR #767, opened 2025-10-06
add hexaly to readme, name it a global opt solver

tias:
hexaly calls itself a global optimizer, we should do the same (so on README, new category: GO, and in docs reademe, global opt.)

Is that OK? e.g. @IgnaceBleukx 

Rest looks fine.

@hbierlee do you want to use this PR to discuss how we should categorize Pindakaas?

It accepts PB-level input, so I would consider it a PB solver... And although it uses a SAT solver, could we call it a SAT solver?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ Install simply with `pip install cpmpy`
 CPMpy can translate to a wide variety of constraint solving paradigms, including both commercial and open-source solvers.
 
 * **CP Solvers**: OR-Tools (default), IBM CP Optimizer (license required), Choco, Glasgow GCS, Pumpkin, MiniZinc+solvers
-* **ILP Solvers**: Gurobi (license required)
+* **ILP Solvers**: Gurobi (license required), CPLEX (license required)
+* **GO Solvers**: Hexaly (license required)
 * **SMT Solvers**: Z3
 * **PB Solvers**: Exact, Pindakaas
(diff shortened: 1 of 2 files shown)


hbierlee:
Short answer: SAT category (meaning, SAT-based).

Refining my comment in #728, there is the question of _modeling paradigm_ ('what input does it support?') and _solving paradigm_ ('what core solver tech does it us'?). If we're being precise, neither question is actually straightforward to answer.

For us, I think the most sensible way to do this categorization is to answer the solving paradigm question (for pindakaas, and for other solvers!). Quite clearly, Pindakaas is a SAT-based solver, since SAT solvers are its core, so it should go in the SAT category (along with PySAT). Note: Pindakaas is not a SAT-solver, the category should only say it's SAT-based (with SAT solvers at its core).

Using the modeling paradigm could definitely also make sense (especially if we're trying to map out the waterfall), but it is murkier for many solvers. (e.g. gurobi acceps non-linear constraints, pindakaas accepts Boolean logic whereas traditional PB solver do not..). Plus, I think that the modeling paradigm matters less to the users of CPMpy, since this is what CPMpy abstracts away.

tias:
@IgnaceBleukx I also updated the top of hexaly's file. THey don't like being called a local solver anymore, in general we should use a similar description tagline as that the solver uses on their own website, which I've adapted now. ALso slight tweak to install info, explicit on binary install.

PR #768, opened 2025-10-13
Typehints in core.py

IgnaceBleukx:
Start of typehinting implementation.

Currently stuck on defining the "is_any_list" type, `mypy` is complaining about a lot it seems...
Something that's quite annoying is that we have explicit checks for which type a certain argument has (e.g., is_bool or is_num), but mypy of course has no idea these functions check what they say they check... so it complains even tho we know it is correct (there are "ignore" tags but they seem cumbersome)

### DIFF
--- a/cpmpy/solvers/TEMPLATE.py
+++ b/cpmpy/solvers/TEMPLATE.py
@@ -53,6 +53,7 @@ import warnings
 from packaging.version import Version
 
 from .solver_interface import SolverInterface, SolverStatus, ExitStatus, Callback
+from ..exceptions import NotSupportedError
 from ..expressions.core import Expression, Comparison, Operator
 from ..expressions.variables import _BoolVarImpl, NegBoolView, _IntVarImpl, _NumVarImpl
 from ..expressions.utils import is_num, is_any_list, is_boolexpr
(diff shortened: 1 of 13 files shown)


ThomSerg:
Using `typing.cast` we can give the required info to the static type checker. I included some small examples.
https://peps.python.org/pep-0484/#casts
It is an "empty function", so it will have a very very small performance impact.

Additionally, a lot of the code requires the args_list to be an "indexable", which `Collection` is not. The `Sequence` type would fix this (e.g. `NestedList = Sequence[Union[T, "NestedList[T]"]]`), but I believe this wouldn't allow for passing iterators. But doesn't seem to be a problem here, since we consume the iterator during construction (in the `Expression.__init__`).

In general I quite like this start.

tias:
OK so I went down the rabbit hole and now core is type hinted and everything is updated so that mypy runs without errors on the whole repo (except for a duplicate file, to be excluded?)

I refound my appreciation for typing, I think it found some bugs along the way (e.g. the xor with consts). Also peculiarities from design choices from the earliest days; e.g. what is stored in the args of an expression.

For the latter, I was going to write that its not worth trying to fix any of that, but then the issue on the runtime came in and so now I am considering to clean up Expression, e.g. that it would be a simple container where args is tuple[Any, ...] and its constructor does nothing, the subclasses have to; and then we can store numpy arrays in it directly, which should also easy type checking some arguments (e.g. for table, maybe even wsum).

None of that should be in this already mega PR though.

It would be good to have it in as one of the next ones, as it might be tedious to rebase every time.

Issue #764, opened 2025-10-03
Linearisation of integer multiplication


ThomSerg:
As became clear in #760, and temporarily fixed in #763, we need a linearisation for the multiplication of two integer decision variables. 

The following model is currently not supported for cplex:
```python
import cpmpy as cp

x,y = cp.intvar(0, 10, shape=2)
m = cp.Model(x*y == 1)

m.solve(solver="cplex")
```

PR #769, opened 2025-10-15
Linearize multiplication with Boolean and Integer

IgnaceBleukx:
Implementation of multiplication constraints of the form `bv * iv <comp> rhs`

These are now translated to `bv -> iv <comp> rhs & (~bv) -> 0 <comp> rhs`

The only solver affected is CPLEX, which does not really support multiplication anyway... Our interface kept `mul` in the model previously, but always raised an error when it was about to be posted.
Additionally, the Cumulative decomposition relies on Boolean-Integer multiplication, which does not work on the current master.

Related to #764

### DIFF
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -86,15 +86,15 @@ def numexprs(solver):
             yield Operator("wsum", [list(range(len(NUM_ARGS))), NUM_ARGS])
             yield Operator("wsum", [[True, BoolVal(False), np.True_], NUM_ARGS]) # bit of everything
             continue
-        elif name == "mul":
-            yield Operator(name, [3,NUM_ARGS[0]])
-            yield Operator(name, NUM_ARGS[:2])
-            if solver != "minizinc": # bug in minizinc, see https://github.com/MiniZinc/libminizinc/issues/962
-                yield Operator(name, [3,BOOL_ARGS[0]])
         elif name == "div" or name == "pow":
             yield Operator(name, [NN_VAR,3])
         elif name == "mul" and "mul-int" not in EXCLUDE_OPERATORS.get(solver, {}):
+            yield Operator(name, [3, NUM_ARGS[0]])
             yield Operator(name, NUM_ARGS[:arity])
+            yield Operator(name, NUM_ARGS[:2])
+            if solver != "minizinc":  # bug in minizinc, see https://github.com/MiniZinc/libminizinc/issues/962
+                yield Operator(name, [3, BOOL_ARGS[0]])
+
         elif name == "mul" and "mul-bool" not in EXCLUDE_OPERATORS.get(solver, {}):
             yield Operator(name, BOOL_ARGS[:arity])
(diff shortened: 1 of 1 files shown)


IgnaceBleukx:
Some tests are still failing, converting to draft for now

cpmpy/transformations/linearize.py
```diff
@@ -128,14 +128,39 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False, c
                             continue
                         elif is_false_cst(lin):
                             indicator_constraints=[] # do not add any constraints
-                            newlist+=linearize_constraint([~cond], supported=supported, csemap=csemap) # post linear version of unary constraint
+                            newlist+=linearize_constraint([~cond], supported=supported, csemap=csemap, reified=reified) # post linear version of unary constraint
                             break # do not need to add other
-                        else:
+                        elif "->" in supported and not reified:
                             indicator_constraints.append(cond.implies(lin)) # Add indicator constraint
+                        else: # Nested -> or implication constraints are not supported
```
hbierlee: very minor thing, but this comment threw me off (I thought it'd say something about the case we're in). I'd prefer if the comments was moved to the block that checks this support, or if the asserts themselves just tell us this in their message.

hbierlee: on the same thread:

- so the assert below should have a message why we expect this comparison, just like the on further below.
- shouldn't we raise the specific excpetion about this, like e.g. TransfNotImplemented ?

cpmpy/transformations/linearize.py
```diff
@@ -158,9 +183,26 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False, c
...
+                    elif isinstance(lhs.args[0], _BoolVarImpl):
+                        bv_idx = 0
+                    elif isinstance(lhs.args[1], _BoolVarImpl):
+                        bv_idx = 1
+
+                    if bv_idx is not None:
+                        # bv * iv <comp> rhs, rewrite to (bv -> iv <comp> rhs) & (~bv -> 0 <comp> rhs)
+                        bv, iv = lhs.args[bv_idx], lhs.args[1-bv_idx]
+                        bv_true = bv.implies(eval_comparison(cpm_expr.name, iv, rhs))
+                        bv_false = (~bv).implies(eval_comparison(cpm_expr.name, 0, rhs))
```
hbierlee, line 202: is this case tested?

IgnaceBleukx: The extra tests in `test_trans_linearize` use solveAll, so all cases should be tested by that

cpmpy/transformations/linearize.py
IgnaceBleukx: I'll update the comment to make it a bit more clear indeed.
Regarding the asserts: to me this is just a sanity check, and it should always go through if the rest of **our** code is correct. I.e., the user normally can put in any expression, without triggering this assert

PR #774, opened 2025-10-17
Skip solver specific example test when solver not installed

ThomSerg:
Some of the examples which are included in our test set target very specific solver backends, different from the default OR-Tools. Whilst not an issue on the GitHub runner (since these solvers get installed in the runner), these tests will fail on a brand new install. I'm of the opinion that all tests should pass with the default minimal install. So I made these tests optional, adding the mechanism to define which examples require which specific solvers and skip those examples when that solver is not available.

P.S. Apologies for the typo in the branch, might cause some confusion when trying to checkout the code and the branch does not seem to exist.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
     zip_safe=False,
     install_requires=[
         'ortools>=9.9',
-        'numpy>=1.5',
+        'numpy>=1.5,<2',
         'setuptools',
     ],
     extras_require={


tests/test_examples.py
```diff
@@ -59,6 +65,13 @@ def test_example(solver, example):
         return pytest.skip(reason=f"exclude {example} for {solver}, too slow or solver-specific")
     if solver == 'minizinc' and any(x in example for x in SKIP_MZN):
         return pytest.skip(reason=f"exclude {example} for {solver}, too slow or solver-specific")
+    # Skip solver specific tests for which the solver is not installed
+    for skip_name in _SOLVER_SPECIFIC_EXAMPLES_REVERSE.keys():
```
IgnaceBleukx, line 69: Hmm I find `skip_name` unintuitive, maybe just `example_name`?

PR #775, opened 2025-10-19
Update solve and solveAll with typehints

### LINKED ISSUE #745: GCS float time limit
Our `.solve()` interface across all different backend solvers promises support for float `time_limit`. GCS only supports `int`s.

Captured by `tests/test_solverinterface.py:test_time_limit`

```console
E       TypeError: solve(): incompatible function arguments. The following argument types are supported:
E           1. (self: gcspy.GCS, all_solutions: bool = True, timeout: Optional[int] = None, solution_limit: Optional[int] = None, callback: Optional[Callable[[dict[str, int]], None]] = None, prove: bool = False, proof_name: Optional[str] = None, proof_location: Optional[str] = None) -> dict[str, int]
E       
E       Invoked with: <gcspy.GCS object at 0x7f5b2711ffb0>; kwargs: all_solutions=False, timeout=1.0, callback=None, prove=False, proof_name=None, proof_location='.'
```

IgnaceBleukx:
Update the function headers of `solve` and `solveAll` with type hints for the time_limit argument.

Something weird was the `def solve()` function of the superclass, it has some weird `model` argument which did not make any sense to me; according to git blame it has been there since 2021.

Also fixes #745 (new version of gcs)

### DIFF
--- a/cpmpy/solvers/hexaly.py
+++ b/cpmpy/solvers/hexaly.py
@@ -40,12 +40,13 @@
 """
 
 from typing import Optional, List
+import time
 
 from .solver_interface import SolverInterface, SolverStatus, ExitStatus, Callback
 from ..expressions.core import Expression, Comparison, Operator, BoolVal
 from ..expressions.globalconstraints import GlobalConstraint, GlobalFunction, DirectConstraint
 from ..expressions.variables import _BoolVarImpl, NegBoolView, _IntVarImpl, _NumVarImpl
-from ..expressions.utils import is_num, is_any_list, eval_comparison, flatlist
+from ..expressions.utils import argval, argvals, is_num, is_any_list, eval_comparison, flatlist
 from ..transformations.get_variables import get_variables
 from ..transformations.normalize import toplevel_list
 from ..transformations.decompose_global import decompose_in_tree, decompose_objective
@@ -118,7 +119,7 @@ class CPM_hexaly(SolverInterface):
     def native_model(self):
         return self.hex_model
 
-    def solve(self, time_limit:Optional[float]=None, **kwargs):
+    def solve(self, time_limit:Optional[float]=None, solution_callback=None, **kwargs):
         """
             Call the Hexaly solver
 
@@ -147,17 +148,26 @@ class CPM_hexaly(SolverInterface):
 
         # set solver parameters
         for arg, val in kwargs.items():
-            setattr(self.hex_solver, arg, val)
+            setattr(self.hex_solver.param, arg, val)
 
         if self.is_satisfaction: # set dummy objective for satisfaction problems
             self.hex_model.add_objective(0, HxObjectiveDirection.MINIMIZE)
 
+        # register solution callback
+        if solution_callback is not None:
+            from hexaly.optimizer import HxCallbackType
+            self.hex_solver.add_callback(HxCallbackType.TIME_TICKED, solution_callback)
+        
         # new status, translate runtime
         self.hex_model.close() # model must be closed
         self.hex_solver.solve()
         self.hex_sol = self.hex_solver.get_solution()
         self.cpm_status = SolverStatus(self.name)
         self.cpm_status.runtime = self.hex_solver.statistics.running_time # wallclock time in (float) seconds
+
+        # unregister solution callback
+        if solution_callback is not None:
+            self.hex_solver.remove_callback(HxCallbackType.TIME_TICKED, solution_callback)
 
         # Translate solver exit status to CPMpy exit status
         # CSP:                         COP:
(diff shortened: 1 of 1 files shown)


Dimosts (changes requested):
Why only time_limit and not e.g. also solution_limit? I suggest that we should do all that w can in the PR.

cpmpy/solvers/choco.py
```diff
@@ -230,7 +230,7 @@ def solve(self, time_limit=None, **kwargs):
 
         return has_sol
 
-    def solveAll(self, display=None, time_limit=None, solution_limit=None, call_from_model=False, **kwargs):
+    def solveAll(self, display=None, time_limit:Optional[float]=None, solution_limit:Optional[int]=None, call_from_model=False, **kwargs):
```
ThomSerg: The type hint for `display` can be `Optional[Expression|list[Expression]|callable]`

cpmpy/solvers/gcs.py
```diff
@@ -95,6 +95,7 @@ def supported():
         # try to import the package
         try:
             import gcspy
+            pkg_resources.require("gcspy>=0.1.8")
```
ThomSerg: Probably to a separate PR? I will steal this one for the "solver version pinning" commit.

cpmpy/solvers/gcs.py
```diff
@@ -250,7 +251,7 @@ def solve(self, time_limit=None, prove=False, proof_name=None, proof_location=".
             
         return has_sol
 
-    def solveAll(self, time_limit=None, display=None, solution_limit=None, call_from_model=False, 
+    def solveAll(self, time_limit:Optional[float]=None, display=None, solution_limit:Optional[int]=None, call_from_model=False,
                  prove=False, proof_name=None, proof_location=".", verify=False, verify_time_limit=None, veripb_args = [], 
                  display_verifier_output=True, **kwargs):
```
ThomSerg:
```python
def solve(self, time_limit:Optional[float]=None, prove:bool=False, proof_name:Optional[str]=None, proof_location:str=".",
              verify:bool=False, verify_time_limit:Optional[float]=None, veripb_args:List[str]=[], display_verifier_output:bool=True, **kwargs):
```

Especially for the optional strings add some hints? Like `proof_name`.

cpmpy/solvers/gurobi.py
```diff
@@ -151,7 +151,7 @@ def native_model(self):
         return self.grb_model
 
 
-    def solve(self, time_limit=None, solution_callback=None, **kwargs):
+    def solve(self, time_limit:Optional[float]=None, solution_callback=None, **kwargs):
```
ThomSerg, line 157: Hmm, `solution_callback` is not even mentioned in the docstring.

cpmpy/solvers/pysat.py
```diff
@@ -491,6 +491,7 @@ def solution_hint(self, cpm_vars, vals):
         cpm_vars = flatlist(cpm_vars)
         vals = flatlist(vals)
         assert (len(cpm_vars) == len(vals)), "Variables and values must have the same size for hinting"
+        assert all(var.is_bool() for var in cpm_vars), "PySAT interface currently only supports Boolean variables in solution hint"
```
ThomSerg, line 504: Maybe temporarily add this to the docstring so it ends up in the documentation?

cpmpy/solvers/solver_interface.py
```diff
@@ -138,19 +139,12 @@ def objective(self, expr, minimize):
...
-            Build the CPMpy model into solver-supported model ready for solving
-            and returns the answer (True/False/objective.value())
+            Call the underlying solver.
 
             Overwrites self.cpm_status
 
-            :param model: CPMpy model to be parsed.
-            :type model: Model
-
-            :param time_limit: optional, time limit in seconds
```
ThomSerg, line 153: Why also remove this one?

cpmpy/solvers/gcs.py
ThomSerg: Also is this the version change where GCS supports float time limits?

cpmpy/solvers/choco.py
ThomSerg: Not sure if we want such verbose type hints

cpmpy/solvers/gurobi.py
ThomSerg: Not that it should be fixed here, just noticed it to be missing.

ThomSerg (commented):
Some small comments

cpmpy/solvers/gcs.py
IgnaceBleukx: Yes that's indeed the version where GCS started to allow for float time limits.
It was still in the "old" style of checking the version, I've updated it now. To me this is not version "pinning" yet, as it just requires a minimum version (similar to exact and pumpkin)

cpmpy/solvers/gurobi.py
IgnaceBleukx: It's a minor change, I'll just add it here

cpmpy/solvers/solver_interface.py
IgnaceBleukx: Woops that was too fast indeed, updated it

cpmpy/solvers/choco.py
IgnaceBleukx: I think it's a bit too verbose if we add this too... It's clear from the docstring imho

hbierlee: I believe the type hints are actually inheritted, so we (thankfully) do not have to duplicate them across all the solvers:

```py
class A:
    def x(self, y: int):
        pass


class B(A):
    pass


import typing

print(typing.get_type_hints(A().x))
print(typing.get_type_hints(B().x))
```

Prints

```
{'y': <class 'int'>}
{'y': <class 'int'>}
```

hbierlee: As discussed internally, complex types can be captured in type aliases, and I think the more complex the argument, the more useful the typing becomes. So I'd be in favour of adding this type hint for display (plus, I've had a lot of trouble with figuring out what exactly to pass to display..)

cpmpy/solvers/choco.py
```diff
@@ -152,7 +152,7 @@ def native_model(self):
         """
         return self.chc_model
 
-    def solve(self, time_limit=None, **kwargs):
+    def solve(self, time_limit: Optional[float]=None, **kwargs):
```
hbierlee, line 164: Apparently, modern (3.10) python type hints for optional use: `time_limit: float | None = None` (see https://stackoverflow.com/a/51710151)

cpmpy/solvers/choco.py
hbierlee: (for me, the point of type hints is not to have them everywhere in the code, it's for the type checkers/documentation/tests to pick them up; you may want to double-check that the docs pick this up, though, there is some confusion online on whether the type hints are inherited)

cpmpy/solvers/choco.py
hbierlee: (just checked.. I guess cpmpy supports 3.8/3.9 still. Although I thought we dropped 3.8? And also, as we discussed, 3.9 is end-of-life anyway (https://devguide.python.org/versions/). Unless we have specific users we cater for that use end-of-life, perhaps 10.0.0 is a good point to adopt the reasonable policy that only supported python versions are supported by cpmpy?)

tias (approved):
looks good to me, please keep the ball rolling

cpmpy/solvers/choco.py
IgnaceBleukx: So the documtation does not inherit the types, if you override the function and don't explicitely re-add the type hints... (which kinda makes sense anyway).

So we do need to repeat it everywhere

cpmpy/solvers/choco.py
IgnaceBleukx: I find the Optional-syntax to be more explicit

cpmpy/solvers/choco.py
IgnaceBleukx: I added this as a typehint, to all solvers now

IgnaceBleukx:
Assuming the tests pass, I think this can be merged now

PR #777, opened 2025-10-23
Sudoku variant examples 2

OrestisLomis:
PR for the sudoku variants I made some time ago. I tried some of these with LLMs as well and they quite difficult so perhaps interesting as projects for the course as well? I also noticed a small mistake in the chess recognition example.


examples/chess-recognition.py
```diff
@@ -119,17 +119,11 @@ def softmax(vector):
     (Count(board[7], p) == 0),
     (Count(board[7], P) == 0),
     # The number of pieces of each piece type without promotion
-    ((Count(board, p) == 8) | (Count(board, P) == 8)).implies(
-        (Count(board, b) <= 2) &
-        (Count(board, r) <= 2) &
-        (Count(board, n) <= 2) &
-        (Count(board, q) <= 1)
+    (
+        (Count(board, b) + Count(board, r) + Count(board, n) + Count(board, q) <= 2+2+1+1 + 8-Count(board, p))
```
hbierlee, line 123: fine for me to include in this PR, but normally this could be split up in separate PRs (although it's a little more effort)

hbierlee: i'd update the comment to reflect the new/fixed constraint

examples/sudoku_chaos_killer.py
```diff
@@ -0,0 +1,119 @@
+import cpmpy as cp
+import numpy as np
+import time
+
+# This CPMpy example solves a sudoku by KNT, which can be found on https://logic-masters.de/Raetselportal/Raetsel/zeigen.php?id=0009KE
```
hbierlee: if there's a copy-paste description, you could add it. These comments can become nice parts of the documentation if you add them as doc-strings at the top of the module/file (see e.g. chess-recognition)

examples/sudoku_chaos_killer.py
```diff
@@ -0,0 +1,119 @@
...
+# sudoku cells
+cell_values = cp.intvar(1,SIZE, shape=(SIZE,SIZE))
+# decision variables for the regions
+cell_regions = cp.intvar(1,SIZE, shape=(SIZE,SIZE))
+# regions cardinals
+region_cardinals = cp.intvar(1,SIZE, shape=(SIZE,SIZE))
+
+
+def killer_cage(idxs, values, regions, total):
+    # the sum of the cells in the cage must equal the total
```
hbierlee: I can recommend getting in the habit to make these comments also into doc-strings

examples/sudoku_variants/sudoku_chaos_killer.py
```diff
@@ -0,0 +1,119 @@
...
+cell_regions = cp.intvar(1,SIZE, shape=(SIZE,SIZE))
+# regions cardinals
+region_cardinals = cp.intvar(1,SIZE, shape=(SIZE,SIZE))
+
+
+def killer_cage(idxs, values, regions, total):
+    # the sum of the cells in the cage must equal the total
+    # all cells in the cage must be in the same region
+    constraints = []
+    constraints.append(cp.sum([values[r, c] for r,c in idxs]) == total)
```
hbierlee, line 41: just for fun, I've been trying more numpy-type features lately. Could this become `cp.sum(values[idxs]) == total`?

examples/sudoku_chaos_killer.py
```diff
@@ -0,0 +1,119 @@
...
+m += cell_regions[SIZE-1, SIZE-1] == SIZE
+
+sol = m.solve()
+
+print("The solution is:")
+print(cell_values.value())
+print("The regions are:")
+print(cell_regions.value())
+print("The region cardinals are:")
+print(region_cardinals.value())
```
hbierlee: for all of these, a good assert is to add the (unique?) solution at the bottom, and check that value matches it. that way, constraints can be simply changed/improved and you immediately know whether the solution is still correct.

examples/sudoku_schrodingers_rat.py
```diff
@@ -0,0 +1,347 @@
...
+        for c in range(path.shape[0]):
+            neighbours = get_reachable_neighbours(r, c)
+            non_emitter_neighbours = [n for n in neighbours if tuple(n) not in EMITTERS]
+            emitter_neighbours = [n for n in neighbours if tuple(n) in EMITTERS]
+            if (r,c) == (2,8):
+                # The path starts on emitter. It doesn't have any previous cells so the first 3 vectors in the sequence must be fully -1. As for the the last 3, they are taken from the neighbour. Vice versa also applies.
+                constraints.append(cp.all([cp.all(sequence[r,c,:3].flatten() == -1)]))
+                constraints.append(cp.any([cp.all([path[nr,nc] == 2, sequence[r,c,3,0] == cells[nr,nc], sequence[r,c,4,0] == nr, sequence[r,c,5,0] == nc, cp.all(sequence[r,c,3,1:] == sequence[nr,nc,3,:19]), cp.all(sequence[r,c,4,1:] == sequence[nr,nc,4,:19]), cp.all(sequence[r,c,5,1:] == sequence[nr,nc,5,:19]), sequence[nr,nc,0,0] == cells[r,c], sequence[nr,nc,1,0] == r, sequence[nr,nc,2,0] == c, cp.all(sequence[nr,nc,0,1:] == sequence[r,c,0,:19]), cp.all(sequence[nr,nc,1,1:] == sequence[r,c,1,:19]), cp.all(sequence[nr,nc,2,1:] == sequence[r,c,2,:19])]) for nr, nc in neighbours]))
```
hbierlee: if these things represent some kind of path, could this path be kept in a separate data structure to make this somewhat more insightfull?

Other way to make this somewhat more readable is to use an auto-formatter like `ruff` on just this files. It will introduce a lot of newlines, but it will be much easier to read and edit.

examples/sudoku_schrodingers_rat.py
```diff
@@ -0,0 +1,347 @@
+import time
+import cpmpy as cp
+import numpy as np
+
+from cpmpy.tools.explain import optimal_mus, mus
```
hbierlee: unused imports?

hbierlee (commented):
Main thoughts:

- [ ] Obviously, I'm not checking the constraints in depth. But if there is a unique solution, adding an assert will convince me that it's correct.
- [ ] Plus, if solved in <5 seconds, can these examples tested by test_examples?
- [ ] CI is failing still
- [ ] Also I think you meant to request me for review, not assign me. The assigned person is the one doing the coding.

examples/sudoku_variants/sudoku_chaos_killer.py
OrestisLomis: that doesn't work, I think something along the lines `cp.sum(values[list(zip(*idxs))) == total` should work, although the last time I did that the CI complained about the * operator, because it was introduced quite recently, so I'm not gonna attempt to make this magic work

Issue #747, opened 2025-09-23
Decompose before simplify_boolean for Xor


ThomSerg:
The transformation of the following model fails for gurobi:
```python
cp.Xor([cp.BoolVal(True), cp.BoolVal(True), p])
```

```console
...
File "/cw/dtailocal/thomass/cpmpy_fuzzing/.fuzz/targets/cpmpy/cpmpy/transformations/linearize.py", line 599, in canonical_comparison
    assert not is_num(lhs), f"lhs cannot be an integer at this point! Got {lhs}."
           ^^^^^^^^^^^^^^^
AssertionError: lhs cannot be an integer at this point! Got 2.
```

The `xor` could have been simplified to `~cp.Xor[cp.BoolVal(True), p]` or even further to `p`. Instead it first goes through `decompose_in_tree` resulting in `((boolval(True)) + (boolval(True)) == 1) + (p) == 1` (we use recursive sums as decomposition for xor). This will eventually lead to the expression `(1 + 1 == 1)`, resulting in the above error.

This could benefit from a normalisation before decompose?

(somewhat linked to #703 since again `Xor` is a special global and thus currently does not have an implementation in `simplify_boolean`)

PR #778, opened 2025-10-24
Fix constants in Xor

IgnaceBleukx:
Some changes to how we handle constants in Xor, fixing both #620 and #747 

Trivially true comparisons are now simplified away during `simplify_bool`, and we cann `simplify_bool` on the decomposition of Xor.
This is not needed for any other global constraint, as it is the only one that takes Boolean arguments.
(Actually also IfThenElse does, but it's decomposition uses `implies` constraints, and the simplifications are built-in to that constructor)

### DIFF
--- a/tests/test_expressions.py
+++ b/tests/test_expressions.py
@@ -6,7 +6,8 @@ from cpmpy.exceptions import IncompleteFunctionError
 from cpmpy.expressions import *
 from cpmpy.expressions.variables import NDVarArray
 from cpmpy.expressions.core import Comparison, Operator, Expression
-from cpmpy.expressions.utils import eval_comparison, get_bounds, argval
+from cpmpy.expressions.utils import eval_comparison, get_bounds, argval, all_pairs
+
 
 class TestComparison(unittest.TestCase):
     def test_comps(self):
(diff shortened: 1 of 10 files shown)


IgnaceBleukx:
I was a bit too quick with adding the trivially true/false cases in simplify_bool.
Implemented it properly now by computing bounds on the comparison itself.

Can probably do with a re-review

tests/test_expressions.py
```diff
@@ -450,6 +451,29 @@ def test_bounds_unary(self):
...
+
+                for expr in [Comparison(comp, x,y), Comparison(comp, y,x)]:
+                    lb, ub = expr.get_bounds()
+
+                    if lb == 0 == ub:
+                        self.assertEqual(cp.Model(expr).solveAll(), 0)
+                    elif lb == 1 == ub:
+                        self.assertEqual(cp.Model(expr).solveAll(), total_vals)
+                    else:
+                        self.assertNotEqual(cp.Model(expr).solveAll(), total_vals)
```
ThomSerg, line 476: use `assertLess()`?
What with `assertMore(..., 0)`?

ThomSerg (approved):
Maybe a small change to the test suite.

Also some tests seem to be failing due to the simplification now resulting in different expressions after transformation. The hard-coded checks fail.

tias:
solved in #908 

what remains in this PR is that the bounds of Comparison now check for trivial true/falseness; we do not currently plan to put this optimisation in main

PR #780, opened 2025-10-24
solver parameterised pytest

ThomSerg:
Attempt to parametrise our pytest test-suite. The default behavior stays the same: generic tests get run on the default solver `OR-Tools` and the more solver specific tests / across solver tests check which backends are available on the current system.

But sometimes you want to just run the testsuite on a single solver without having to uninstall all other solvers / create a fresh environment. Now you can pass an optional argument `--solver`:

```
python -m pytest ./tests --solver exact
```

This will have three consequences:
- the "generic tests" will now be run on `exact` instead of the default `ortools`
- the solver specific tests not targeting exact will be filtered
- the across solver tests will be limited to only running on `exact`

In general, I opted for filtering instead of skipping tests. So the non-`exact` tests will not count towards the total number of tests. I believe we should reserve "skipping" for tests which don't get run due reasons of which we want to inform the user, e.g. missing dependencies which they need to install. When the user provides the `--solver` option, they already know that tests targeting other solvers won't be run so it would just clutter the results if we were to skip instead of filter those tests.

To parameterise a unittest class for the "generic" tests, simply decorate it with:
```python
@pytest.mark.usefixtures("solver")
```
After which `self.solver` will be available, matching the user provided solver argument.

All solver specific tests can now be decorated with:
```python
@pytest.mark.requires_solver("<SOLVER_NAME>")
```
And will automatically be skipped if a `--solver` argument has bee provided which doesn't match `SOLVER_NAME`.

For the across-solver tests which use generators (those in `test_constraints.py`), the `pytest_collection_modifyitems` hook will filter out parameterised pytest functions which have been instantiated with a solver different than the user provided one. Both the argument `solver` and `solver_name` get filtered on.

There are still some smaller places (see `test_solveAll.py`) where `cp.SolverLookup.base_solvers()` is used more directly, which can't be filtered without making changes to the test itself (not possible with one of the decorators / callback functions)

As a further improvement, it might be possible to merge the following two **(Already did it ;) )**
```python
@pytest.mark.requires_solver("minizinc")
@pytest.mark.skipif(not CPM_minizinc.supported(), reason="MinZinc not installed")
```
So do the skipping if solver is not available also through the first mark and skip the tests more centrally in `pytest_collection_modifyitems`.

Using this parameterisation with solver different from OR-Tools revealed some issues with our testsuite related to #779


tests/conftest.py
```diff
@@ -0,0 +1,329 @@
...
+        if filter_not_installed:
+            warnings.warn('Option "all" already expands to all installed solvers. Ignoring filter for "filter_not_installed".')
+    else:            
+        solvers = original_solvers.copy()
+        # Filter out non-installed solvers if requested
+        if filter_not_installed:
+            solvers = [s for s in solvers if s in cp.SolverLookup.supported()]
+    
+    # If solver was provided but all were filtered out, return empty list (not None)
+    # This distinguishes "no solver specified" from "solver specified but not available"
```
IgnaceBleukx, line 76: I don't fully understand the difference I think

tests/conftest.py
```diff
@@ -0,0 +1,329 @@
...
+    if solver_option:
+        # Parse without filtering to check original list
+        parsed_solvers_unfiltered = _parse_solver_option(solver_option, filter_not_installed=False)
+        if parsed_solvers_unfiltered:
+            installed_solvers = cp.SolverLookup.supported()        # installed base solvers
+            not_installed_solvers = list(set(parsed_solvers_unfiltered) - set(installed_solvers))
+            if not_installed_solvers:
+                warnings.warn(
+                    f"The following solvers are not installed and will not be tested: {', '.join(not_installed_solvers)}. "
+                    f"Only installed solvers will be used for testing.",
```
IgnaceBleukx, line 238: Maybe print the list of solvers that will be tested here?

tests/README.md
```diff
@@ -0,0 +1,177 @@
...
+```bash
+pytest tests/test_model.py::TestModel::test_ndarray
+```
+
+### Parallelisation
+
+Through the `pytest-xdist` pytest plugin, running tests can be parallelised. 
+E.g. running with 40 workers:
+```console
+pytest -n 40 tests/test_model.py
```
IgnaceBleukx, line 29: I use `-n auto` these days ; )

tests/README.md
```diff
@@ -0,0 +1,177 @@
...
+#### All Installed Solvers
+Run tests with all installed solvers:
+```bash
+pytest --solver=all
+```
+
+This automatically detects all installed solvers from `SolverLookup` and parametrises non-solver-specific tests to run against each one.
+
+#### Skip Solver Tests
+Skip all solver-parametrised tests (only run tests that don't depend on solver parametrisation):
```
IgnaceBleukx: "I.e., tests that do not rely on solving a model. Examples are tests that evaluate constructors of expressions"

tests/test_examples.py
```diff
@@ -96,4 +109,17 @@ def test_advanced_example(example):
...
+        # TODO: this is a hack to skip tests when the solver is not installed, 
+        #       for now no better way to do this without having to manually label all 
+        #       examples with the required solvers / dependencies
+        error_msg = str(e).lower()
+        if ("install" in error_msg and ("package" in error_msg or "solver" in error_msg)) or \
+           ("not installed" in error_msg) or \
+           ("not available" in error_msg):
+            pytest.skip(f"Skipped, solver not installed: {e}")
+        # Re-raise other exceptions
+        raise
```
IgnaceBleukx: raise e

tests/README.md
ThomSerg: Oh, good to know. Will add to the documentation.

tests/conftest.py
ThomSerg: It's the difference between `pytest ./tests` and `pytest ./tests --solver pumpkin` but without having pumpkin installed. In the former case, `solvers` will be 'None'. In the latter, it will be `[]`

tests/test_expressions.py
```diff
@@ -126,11 +132,6 @@ def test_weightedadd_weighted_iv(self):
         self.assertIsInstance(expr4, Operator)
         self.assertEqual(expr4.name, 'wsum')
 
-    def test_weightedadd_int(self):
```
IgnaceBleukx, line 129: Not sure why this one is removed?

IgnaceBleukx (approved):
This is an absolute huge diff to properly review, but I can't find any obvious issues with it.
The conftest.py feels a bit like magic still, but I suppose that's just the way pytest works...

There are still tests failing when running with `--solver=all`, but that is to be expected, right? For me, this can be merged as is, with the surfaced issues being fixed in subsequent PRs

tests/test_expressions.py
ThomSerg: the test was duplicated, so `test_weightedadd_int` is still there

ThomSerg:
As mentioned a couple of times, the parametrisation of our entire test suite (as opposed to a limited section as currently on master) reveals a bunch of small bugs / edge cases we'll have to fix (some are just as simple as not running a certain test for a certain solver because it doesn't support it). So currently the default behavior of pytest (without `--solver`) has changed as opposed to master; all tests are only run on ortools, except for the solver dependent ones. This is not great, since (as Tias mentioned internally) this no longer fully tests our entire transformation pipeline. Currently the only way to somewhat replicate master is to call `--solver=all` with all the solvers from the github runner. But due to all the newly parametrised test, this has a bunch of failing tests (that can be easy to fix, but don't fit in this PR). What I think might be the best approach is to add a very simple but temporary pytest marker to add to those tests that we want run on all solvers (like master). That way running without `--solver` should have the exact same behavior as master, with minimal code change for adding those small markers in some places. This PR would then basically put all the infrastructure in place for running all tests on all solvers, but for now not run it yet by default (just simulate current master). Then in a later PR we can start fixing those issues, up to a point where they are no longer needed and we can simply run with `--solver=all` and the full benefit of this PR can be exploited.

So for example this could look like:
```python
@pytest.mark.run_all_by_default()     <- only marker to add in selective places to replicate current behavior on master, can have different name
@pytest.mark.generate_constraints.with_args(bool_exprs)
@skip_on_missing_pblib(skip_on_exception_only=True)
def test_bool_constraints(solver, constraint):
    ...
 ```

 This seems like a minimal effort temporary fix to be able to merge this PR, with the goal to fix these remaining issues in future PRs and being able to run even more tests parametrised across all solvers.

tias:
so on my machine all test failures (ignore examples) are pysdd related (which properly throws a not implemented error on any integer expression).

The temporary fix for that can be simply to remove pysdd from 'all':
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -62,6 +62,7 @@ def _parse_solver_option(solver_option: Optional[str] , filter_not_installed: bo
     # Expand "all" to all installed solvers
     if "all" in original_solvers:
         solvers = [name for name, solver in cp.SolverLookup.base_solvers() if solver.supported()] 
+        solvers = [name for name in solvers if name != "pysdd"]  # TODO: add int2bool pipeline for it

in a later PR, we will make pysdd like pysat and pindakaas also have int2bool and we can reactivate this...


However, while deactivating pysdd, I noticed:
* tests/test_trans_simplify.py claims to have 0 tests... on master it has 6
* other files have the same, the way I ran it doesnt make clear which (alphabetically inbetween test_tocnf and test_transf_comp; perhaps more)
* examples dont seem to timeout, e.g. for choco its been running for 21 minutes. Not sure if we expect them to timeout? or we have to manually be more aggressive and disabling some of them for certain solvers?

tias:
I see.

- For pysat, you will have to install pblib as it says.
- For multiplication and reified alldifferent, that is for when make make mult a global and do #808 ... 

I am unsure what the best way forward is...

The tests pass with the default option of only testing with ortools. So lets just keep using that on github, and then we fix these things after the release?

tias:
Impossible to merge, kept for reference in tracker #1076

PR #781, opened 2025-10-28
Add encoding variables as expressions to CSE

hbierlee:
By adding encoding variables to CSE, e.g. for the direct encoding for `x` which encodes `b == ( x == 2 )`, adding `cse[x==2] = b` ensures `b` is used whenever we see the expression `x==2`. 

In the following test, we see how this improves the encoding of `(x == 0) | (x == 2)` for `x in 0..2`:

```py
old: [(BV1) or (BV2), sum([⟦x == 0⟧, ⟦x == 1⟧, ⟦x == 2⟧]) == 1, (BV1) -> (⟦x == 0⟧), (~BV1) -> (sum([0, 1, 2, -3] * (⟦x == 0⟧, ⟦x == 1⟧, ⟦x == 2⟧, BV3)) <= -1), (~BV1) -> (sum([0, 1, 2, -1] * (⟦x == 0⟧, ⟦x == 1⟧, ⟦x == 2⟧, BV3)) >= 0), sum([1, -1] * (BV1, ~BV3)) <= 0, (BV2) -> (⟦x == 2⟧), (~BV2) -> (sum([0, 1, 2, -1] * (⟦x == 0⟧, ⟦x == 1⟧, ⟦x == 2⟧, BV4)) <= 1), (~BV2) -> (sum([0, 1, 2, -3] * (⟦x == 0⟧, ⟦x == 1⟧, ⟦x == 2⟧, BV4)) >= 0), sum([1, -1] * (BV2, ~BV4)) <= 0]
new: [(⟦x == 0⟧) or (⟦x == 2⟧), sum([⟦x == 0⟧, ⟦x == 1⟧, ⟦x == 2⟧]) == 1]
```

However, this does mean integer variables need to be collected and encoded before the other transformations. This may be an encoding inefficiency, plus it does mean we cannot inspect the constraint that holds the variable (an inequality constraint normally uses order encoding, an (dis)equality the direct encoding).

Another problem is that the `int2bool` transformation requires boilerplate across solvers. This should be solved in a separate PR, I think. This optimization is now only added to pindakaas, but should also happen for the other solvers which use `int2bool`.

### DIFF
--- a/cpmpy/transformations/int2bool.py
+++ b/cpmpy/transformations/int2bool.py
@@ -19,8 +19,8 @@ def int2bool(cpm_lst: List[Expression], ivarmap, encoding="auto", csemap=None):
     """Convert integer linear constraints to pseudo-boolean constraints. Requires `linearize` transformation.
 
     :param: cpm_lst: list of constraints to transform
-    :param: ivarmap: dictonary mapping integer variables to their encoding
-    :param: encoding: choice of encoding: "direct", "order", "binar", or "auto", which makes encoding choices based on constraint comparator and domain size
+    :param: ivarmap: dictionary mapping integer variables to their encoding
+    :param: encoding: choice of encoding: "direct", "order", "binary", or "auto", which makes encoding choices based on constraint comparator and domain size
     :param: csemap: To enable CSE
     """
     assert encoding in (
@@ -202,7 +202,7 @@ def _decide_encoding(x, cmp=None, encoding="auto"):
         return "binary"
     elif cmp in ("==", "!="):
         return "direct"  # equalities suit the direct encoding
-    else:  # we use the order encoding for inequalities, en when we do not have `cmp`
+    else:  # we use the order encoding for inequalities, and as default when we do not have `cmp`
         return "order"
 
 


IgnaceBleukx:
Yes, I think indeed the core issue lies in linearize here and we should make the encoding variables there for categorical variables.

I quite like the two steps, not sure if it should become part of linearize or become it's own transformation? Probably linearize is fine

tias:
This will not be merged, should be implemented in separate PRs as discussed above.

Leaving the PR open for the discussion part.

cpmpy/transformations/int2bool.py
```diff
@@ -77,26 +85,26 @@ def _encode_expr(ivarmap, expr, encoding):
...
-            ivarmap[x.name] = IntVarEncOrder(x)
+            ivarmap[x.name] = IntVarEncOrder(x, csemap=csemap)
         elif encoding == "binary":
-            ivarmap[x.name] = IntVarEncLog(x)
+            ivarmap[x.name] = IntVarEncLog(x, csemap=csemap)
         else:
             raise NotImplementedError(encoding)
 
-        return (ivarmap[x.name], ivarmap[x.name].encode_domain_constraint())
+        return (ivarmap[x.name], ivarmap[x.name].encode_domain_constraint(csemap=csemap))
```
tias: is this really needed? the BVs are already created at this point?

cpmpy/transformations/int2bool.py
```diff
@@ -194,28 +202,29 @@ def _decide_encoding(x, cmp=None, encoding="auto"):
...
 
-    def __init__(self, x, n, name):
-        """Create encoding of integer variable `x` with `n` Boolean variables named by `name`."""
-        if _dom_size(x) == 0:
-            raise EMPTY_DOMAIN_ERROR
+    NAMED = False
+    """Enable to name the encoding variables semantically for debugging purposes (e.g. `BV[x == 42]` for a direct encoding variable of `x`)"""
 
+    def __init__(self, x, x_enc, csemap=None):
+        """Create encoding of integer variable `x` over the given Boolean expressions, `x_enc`. E.g. the direct encoding for `x` should provide `x_enc = ( x == 1, x == 2, ..)`. Any literals created (e.g. b == ( x == 1 )`) are added to the `csemap` if provided."""
```
tias: does this constructor need to know the integer expressions?

I think it should just receive boolean variables, and whoever creates the magic 'x_enc' now, should create the bools for us?

cpmpy/transformations/int2bool.py
```diff
@@ -279,12 +289,12 @@ class IntVarEncDirect(IntVarEnc):
     Uses a Boolean 'equality' variable for each value in the domain.
     """
 
-    def __init__(self, x):
+    def __init__(self, x, csemap=None):
         """Create direct encoding of integer variable `x`."""
         # Requires |dom(x)| Boolean equality variables
-        super().__init__(x, _dom_size(x), f"EncDir({x.name})")
+        super().__init__(x, (x == d for d in _dom(x)), csemap=csemap)
```
tias: right, so, wouldn't it make more sense that the IntVarEncDIrect constructor calls get_or_make_var, and then only gives the bvs to its superclass (which is just a container that does not know what it receives)?

cpmpy/transformations/int2bool.py
```diff
@@ -381,15 +391,16 @@ class IntVarEncLog(IntVarEnc):
...
-        super().__init__(x, bits, f"EncBin({x.name})")
+        super().__init__(x, (cp.boolvar(name=f"bit({x},{k})") for k in range(bits)), csemap=csemap)
+        # TODO possibly...: super().__init__(x,  ((( ((x - x.lb) ** k) % 2) == 0) for k in range(bits)), csemap=csemap)
 
-    def encode_domain_constraint(self):
+    def encode_domain_constraint(self, csemap=None):
         """Return binary encoding domain constraint (i.e. upper bound is respected with `self._x<=self._x.ub`. The lower bound is automatically enforced by offset binary which maps `000.. = self._x.lb`)."""
         # encode directly to avoid bounds check for this seemingly tautological constraint
-        return self.encode_comparison("<=", self._x.ub, check_bounds=False)
+        return self.encode_comparison("<=", self._x.ub, check_bounds=False, csemap=csemap)
```
tias: I guess thsi is the edge case that makes it look like encode_dom requires csemap, but what expression is this really? it doesn't look like something that can ever need a cse...

cpmpy/transformations/int2bool.py
hbierlee: There's two points, one is that the encoding objects definitely need the original integer variable to e.g. know the bounds to know what the encoding variables mean, to know the integer name, or to set the integers value, etc. 


the second point about passing `x_enc` we can discuss in the other comment

cpmpy/transformations/int2bool.py
hbierlee: I think this design works well: `IntVarEnc` implements exactly the shared behavior of all three encodings, and is given a list of expressions (either BVs for `IntVarEncBin` still, or `x==d`'s for IntVarEncDir) which encode the given integer variable. What would be upside of taking this shared behavior out of IntVarEnc?

One thing I can think off now is that IntVarEncBin does not require (for now) to call `get_or_make_var`, but that could still be prevent in this design

cpmpy/transformations/int2bool.py
hbierlee: yes true, I removed this

cpmpy/transformations/int2bool.py
tias: It hides the logic... its too clever... it requires to have the entire codebase in your mind when working on the code. The class stores Boolean variables in self._xs, but what are they? You check the constructor... nothing, what is in _xs, what does it mean? You have to trace that a generator of expressions is given to the superclass and that this superclass creates Boolean variables for you, which you have to trust are put in xs in that order.

Its more obvious if the constructor itself would give the x and xs to the superclass, which would just store it. In your own class, you then manipulate these xs, and you know what they mean, because you created them yourself.

PR #782, opened 2025-10-31
Expand CNF using encoding back-end

hbierlee:
This PR expands the `to_cnf` function to all CPMpy models (insofar supported by other Boolean solvers). It does so by passing all constraints through `pindakaas` transform and encoding, and reading the Pindakaas encoding object back to CPMpy.

TODO

- [x] constraints which are only clauses do not have to pass through pindakaas

### DIFF
--- a/cpmpy/tools/dimacs.py
+++ b/cpmpy/tools/dimacs.py
@@ -39,7 +39,7 @@ def write_dimacs(model, fname=None, encoding="auto"):
     """
 
     constraints = toplevel_list(model.constraints)
-    constraints = to_cnf(constraints)
+    constraints = to_cnf(constraints, encoding=encoding)
 
     vars = get_variables(constraints)
     mapping = {v : i+1 for i, v in enumerate(vars)}


tests/test_tool_dimacs.py
```diff
@@ -60,7 +64,8 @@ def test_write_cnf(self):
         m += a <= 0
 
         cnf_txt = write_dimacs(m)
-        gt_cnf = "p cnf 3 3\n1 2 3 0\n-2 -3 0\n-1 0\n"
+        # TODO note the order is slightly unexpected, because of an optimization in `to_cnf` which puts simple clauses before encoded constraints (i.e.) sums
```
IgnaceBleukx: A solution to this would be to compare sets of lines, which is basically what we want for this tool anyway
E.g.
```python
gt_clauses = set(gt_cnf.split("\n")[1:])
txt_clauses = set(cnf_txt.split("\n")[1:])
self.assertSetEqual(gt_clauses, txt_clauses)
```

hbierlee:
Thanks for picking this up, I lost it out of sight. The new changes look good to me. So ok to merge if tests pass.

PR #783, opened 2025-10-31
Pysat pb reification optimisation, let pysat handle the conditionals

OrestisLomis:
This PR is such that conditionals can natively be added when using pysat to translate PB constraints.
The PR depends on two other PRs, one in pysat itself https://github.com/pysathq/pysat/pull/206 which adds this functionality within pysat and another one in PBLIB https://github.com/rjungbeck/pypblib/pull/6, which cleans up some trivial constraints. The latter one is not a blocker, it's just a testcase that would need to be changed.

### DIFF
--- a/tests/test_pysat_wsum.py
+++ b/tests/test_pysat_wsum.py
@@ -104,14 +104,14 @@ class TestEncodePseudoBooleanConstraint:
         pb_clauses = solver._pysat_pseudoboolean(cons1, conditional=sel_var)
         
         # Check the result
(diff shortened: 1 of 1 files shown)


hbierlee (commented):
This can actually be made into a mergeable PR with the following trick: expect this assertion (using `with pytest.raises(..)`, [see doc](https://docs.pytest.org/en/7.1.x/how-to/assert.html)) to raise an `AssertionError` (as it currently would). This passes. Once pypblib is updated and on our end upgraded, this test fails, and can be update the test. Maybe over-engineered, but avoids needing to remember this PR forever.

PR #784, opened 2025-10-31
cvc5 SMT solver

ThomSerg:
Adds support for the cvc5 SMT solver as backend. cvc5 provides a "pythonic" API which closely replicates Z3's API (but no support for optimisation). Unfortunately this API does not expose all cvc5's features, like unsat core extraction. There is also a very verbose C-style python API which exposes much more, maybe something to look at in the future but would require more work (the current wrapper is +- the same as the Z3 one).

There are still two TODOs since I'm not all too certain if I haven't missed any global constraints that cvc5 would support + what constraints it supports in a reified context.


ThomSerg:
One of the tests will fail on cvc5, one that was also previously linked with a Z3 bug?
```python
x = cp.intvar(0, 1)
m = cp.Model((x >= 0.1) & (x != 1))
s = cp.SolverLookup.get("z3", m)
self.assertFalse(s.solve()) # upgrade z3 with pip install --upgrade z3-solver
```

IgnaceBleukx:
Might be more insteresting to just interface to https://github.com/pysmt/pysmt and call cvc5 through that? That saves us the hassle of keeping up with future API changes as well

tias:
no, lets do CVC5 straight, to update?

PR #786, opened 2025-11-03
Z3 negate maximisation objective

ThomSerg:
As encountered during the last two XCSP3 competitions, Z3 internally negates the objective function when maximising (turning the problem into minimisation). But when calling `sol.evaluate(obj).as_long()` to get the objective value after a solve call, this abstraction gets broken and Z3 returns the negated value. This in turn resulted in incorrect values for `solver.objective_value()`. Added code to undo the negation.


IgnaceBleukx (commented):
Can you add a test as well? Other than that it is good to go

tias:
yes, great, can be merged

PR #787, opened 2025-11-03
`solution_callback` for Hexaly

ThomSerg:
Hexaly has support for intermediate solution callbacks. Added the necessary code on CPMpy's side for the `solution_callback` argument of a `.solve()` call, similar to OR-Tools.

Also very small fix for setting solver-native arguments, `setattr(self.hex_solver, arg, val)` should have been `setattr(self.hex_solver.param, arg, val)`.


IgnaceBleukx:
Ok I see, yes let's merge this already so people can use the Hexaly-specific callback. If and when we revive the other PR, we can add structured testing for this as well

PR #789, opened 2025-11-10
Nurserostering benchmark

IgnaceBleukx:
Similar to the RCPSP and JSPLib data loaders, I also implemented a loader and model for the nurse rostering instances from schedulingbenchmarks.org.

Model is largely ported from the XCP-Explain tutorial repo, but cleaned it up quite a bit.

### DIFF
--- a/examples/nurserostering.py
+++ b/examples/nurserostering.py
@@ -16,12 +16,14 @@ import pandas as pd
 
 try:
     from faker import Faker
-except ImportError:
+except ImportError as e:
     print("Install `faker` package using `pip install faker`")
+    raise e
 try:
     from natsort import natsorted
-except ImportError:
+except ImportError as e:
     print("Install `natsort` package using `pip install natsort`")
+    raise e
 
 pd.set_option('display.max_columns', 500)
 pd.set_option('display.width', 5000)


examples/nurserostering.py
```diff
@@ -0,0 +1,283 @@
...
+
+        Args:
+            index (int): Index of the instance to retrieve
+
+        Returns:
+            Tuple[Any, Any]: A tuple containing:
+                - The filename of the instance
+                - Metadata dictionary with file name, track, year etc.
+        """
+        if isinstance(index, int) and (index < 0 or index >= len(self)):
```
hbierlee: personal preference, not(0 <= index < len(self))?

examples/nurserostering.py
```diff
@@ -0,0 +1,283 @@
...
+            raise IndexError("Index out of range")
+
+        # Get all instance files and sort for deterministic behavior # TODO: use natsort instead?
```
hbierlee: think TODO can be removed?

examples/nurserostering.py
```diff
@@ -0,0 +1,283 @@
...
+import copy
+import pathlib
+from io import StringIO
+from os.path import join
+from typing import Tuple, Any
+from urllib.request import urlretrieve
+from urllib.error import HTTPError, URLError
+import zipfile
+
+from faker import Faker
```
hbierlee: where are this and the natsort requirements added?

examples/nurserostering.py
```diff
@@ -0,0 +1,283 @@
...
+        if isinstance(index, int) and (index < 0 or index >= len(self)):
+            raise IndexError("Index out of range")
+
+        # Get all instance files and sort for deterministic behavior # TODO: use natsort instead?
+        files = natsorted(list(self.instance_dir.glob("*.txt"))) # use .txt files instead of xml files
+        file_path = files[index]
+
+        filename = str(file_path)
+        if self.transform:
+            # does not need to remain a filename...
```
hbierlee: unclear on what this means?

examples/nurserostering.py
```diff
@@ -0,0 +1,283 @@
...
+        kwargs = {"header":0, "index_col":0} | kwargs
+        df = pd.read_csv(StringIO(data), *args, **kwargs)
+        return  df.rename(columns=lambda x: x.replace("#","").strip())
+    return datatype(data, *args, **kwargs)
+
+def parse_scheduling_period(fname):
+    fake = Faker()
+    fake.seed_instance(0)
+
+    with open(fname, "r") as f:
```
hbierlee, line 139: think these "r" are the default open mode

examples/nurserostering.py
hbierlee: (I guess we don't really have a way to add example-specific requirements. Would it warrant something in setup.py (like, pip install .[examples], which just adds all requirements)?

examples/nurserostering.py
```diff
@@ -0,0 +1,283 @@
...
+    dataset = NurseRosteringDataset(root=".", download=True, transform=parse_scheduling_period)
+    print("Dataset size:", len(dataset))
+    data, metadata = dataset[0]
+
+    for key, value in data.items():
+        print(key,":")
+        print(value)
+
+    model, nurse_view = nurserostering_model(**data)
+    assert model.solve()
```
hbierlee, line 271: the known optimal solution for the first instance is 607 according to https://schedulingbenchmarks.org/nrp/. Add that as an assert? It will be a better way to find whether the code is correct, rather than me trying to read it carefully. Does add a bit of dependence on the website

examples/nurserostering.py
```diff
@@ -0,0 +1,283 @@
...
+
+    # Max number of working weekends for each nurse
+    weekends = [(i - 1, i) for i in range(1,horizon) if (i + 1) % 7 == 0]
+    for i, nurse in staff.iterrows():
+        n_weekends = cp.sum((nurse_view[i,sat] != FREE) | (nurse_view[i,sun] != FREE) for sat,sun in weekends)
+        model += n_weekends <= nurse['MaxWeekends']
+
+    # Days off
+    for _, holiday in days_off.iterrows(): # could also do this vectorized... TODO?
+        i = staff.index[staff['ID'] == holiday['EmployeeID']][0]
```
hbierlee: is argmax faster than index to find the first occurrence?

examples/nurserostering.py
IgnaceBleukx: Hmm yeah no that's tricky indeed... I could wrap this into a try-catch and give a more informative error message?

examples/nurserostering.py
IgnaceBleukx: I didn't know of the argmax trick, quite neat!

PR #790, opened 2025-11-15
Add transversal example

nandorsieben:
Example computing minimal transversals of a family of sets.


ThomSerg:
Thank you very much for the contribution of a nice example on hitting sets! I've added some comments to make clear what is going on and reference you as the author. Will be merging this one soon.

PR #793, opened 2025-11-27
Refactor decompose for global functions

IgnaceBleukx:
As discussed offline, global functions can be decomposed into a numerical value, and some defining constraints.

This PR accomplishes that by removing the `decompose_comparison` function we have now.
It also massively simplifies the `decompose_in_tree` code as there are way fewer cases to check.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -20,4 +20,4 @@ jobs:
 
       - name: Test with pytest
         run: |
-          python -m pytest -v tests/
+          python -m pytest -n auto tests/


cpmpy/expressions/globalfunctions.py
```diff
@@ -42,8 +42,8 @@ class my_global(GlobalFunction):
             def __init__(self, args):
                 super().__init__("my_global", args)
 
-            def decompose_comparison(self):
-                return [self.args[0] + self.args[1]] # your decomposition
+            def decompose(self):
```
tias, line 51: I think this whole toplevel doc deserve a slight update.

E.g. instead of "then preferably define it with a standard comparison decomposition,"
something like
"then implement a decompose function that returns a tuple of two arguments:
1) a single expression representing what the global function computes (this is often a sum over auxiliary variables, or a single auxiliary variable)
2) a list of constraints that define the auxiliary variables used in the first argument

To make maximum use of simplification and common subexpression elimination, we recommend that you use nested expression as much as possible and avoid creating auxiliary variables unless really needed"

cpmpy/expressions/globalfunctions.py
```diff
@@ -42,8 +42,8 @@ class my_global(GlobalFunction):
...
+                return [self.args[0] + self.args[1]], [] # your decomposition
```
tias: I guess this global is "MySum" or so, and then this should now be `return (self.args[0]+self.args[1]), []`

cpmpy/expressions/globalfunctions.py
tias: Below in the abstract class, you write "Returns a decomposition into smaller constraints.
            Returns a numerical expression and a list of defining constraints."
            
something consistent with that is also fine : )

cpmpy/expressions/globalfunctions.py
```diff
@@ -126,20 +142,15 @@ def value(self):
...
         """
-        Decomposition if it's part of a comparison
-
-        Returns two lists of constraints:
-
-        1) constraints representing the comparison
-        2) constraints that (totally) define new auxiliary variables needed in the decomposition,
-           they should be enforced toplevel.
+        Decomposition of Minimum constraint.
+        Returns
```
tias: this is a generic comment, ideally this is only needed in the abstract class?

Here it could be something more specific that actually explains this not entirely trivial decomp, like
"Can only be decomposed by introducing an auxiliary variable and constraining it is smaller than each variable, while at the same time not being smaller then all (e.g. it needs to be (larger or) equal to one of them)"

cpmpy/expressions/globalfunctions.py
```diff
@@ -126,20 +142,15 @@ def value(self):
...
+        1) a numerical value to replace the constraint, and
+        2) a list of defining constraints, which should be enforced toplevel
         """
-        lb, ub = self.get_bounds()
-        _min = intvar(lb, ub)
-        return [eval_comparison(cpm_op, _min, cpm_rhs)], \
-               [cp.any(x <= _min for x in self.args), cp.all(x >= _min for x in self.args), ]
+        _min = intvar(*self.get_bounds())
+        return _min, [cp.all(x >= _min for x in self.args), cp.any(x <= _min for x in self.args)]
```
tias: I would find it more intuitive to read it with the order flipped, maybe also vectorized, e.g.
`return _min, [cp.all(_min <= self.args), cp.any(_min >= self.args)]`

cpmpy/expressions/globalfunctions.py
```diff
@@ -197,30 +203,25 @@ def __init__(self, expr):
...
-            is_pos = boolvar()
-            return [eval_comparison(cpm_op, newarg, cpm_rhs)], \
-                    [is_pos == (arg >= 0), is_pos.implies(arg == newarg), (~is_pos).implies(-arg == newarg)]
+        if lb >= 0: # always positive
+            return arg, []
+        if ub <= 0: # always negative
+            return -arg, []
 
+        _abs = intvar(*self.get_bounds())
+        assert _abs.lb == 0
```
tias: I don't find this obvious... if i read get_bounds, I do see that it enforces that, but... do we really need this assert here?

cpmpy/expressions/globalfunctions.py
```diff
@@ -197,30 +203,25 @@ def __init__(self, expr):
...
 
+        is_pos = arg >= 0
```
tias: a reader has to pay attention that this is not a Boolean check, it is an expression...
maybe add `# CPMpy expression that checks of the argument is positive`

cpmpy/expressions/globalfunctions.py
```diff
@@ -197,30 +203,25 @@ def __init__(self, expr):
...
+        return _abs, [is_pos.implies(arg == _abs), (~is_pos).implies(arg == -_abs)]
```
tias: there are probably LP-friendly ways to do this, which we should keep in mind for linearize later

cpmpy/expressions/globalfunctions.py
```diff
@@ -379,31 +380,27 @@ def __init__(self, arr):
...
         lbs, ubs = get_bounds(self.args)
         lb, ub = min(lbs), max(ubs)
 
-        constraints = []
-
-        # introduce boolvar for each possible value
-        bvars = boolvar(shape=(ub+1-lb,)) # shape is tuple to ensure it is a 1D array
+        n_values = 0
+        for v in range(lb, ub+1):
+            n_values += cp.any(a == v for a in self.args)
```
tias: its a bit of an unusual coding style
how about `return cp.sum(cp.any(self.args == v) for v in range(lb, ub+1))

cpmpy/expressions/globalfunctions.py
IgnaceBleukx: I flipping the order if probably more intuitive, but we can't do vectorized if the `_min` is the first arg of the comparison

IgnaceBleukx:
I very much like this doc and style rewrite already.
For the typehinting: you can use the "builtin" types like `list[Expression]` but it requires Python>=3.9; the "old" way of doing type hints is by using `typing.List[Expression]` but that will be decprecated soon-ish according to [this stackoverflow article](https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc).
So we should use `list[...]` and bump the minimum Python version to 3.9.

To check the typehints: you need to pip-install `mypy` and run ` mypy cpmpy/expressions/globalfunctions.py`

tias:
So the decompose objective...

its needed because `decompose_in_tree` assumes toplevel and hence returns a list, I think?

But decompose_objective does not contain supported_reified, is that on purpose or by accident?

It now requires a solver to duplicate this information in two places; perhaps we should set it in the constructor, so that anybody can look up what this solver supports?

IgnaceBleukx:
Yes and no...
So `decompose_in_tree` expects to start in a Boolean context, while for objectives this is obviously not true.
I actually wasn't sure whether we need a supported_reified here; I suppose we do indeed...
Because now, if you do `model.minimize(AllDifferent(x,y,z) + count(x,y,z))`, the solver could indeed support a reified alldiff, so no need to decompose, we just reify it later in `flatten_objective`.

As to where we need to define which solver supports what: I think it would be great if we indeed make this an attribute of the object itself. It would also allow users to turn on/off supportedness for some constraints.

Do you want to add such a refactor in this PR too?

tias:
great; tests are passing again too : )

I find it surprising that you added 3 categories (cons, reif cons, funcs), but decompose_in_tree only takes two...

I can see why you did it, e.g. the funcs are automatically also supported reified, but it still signals something is off... either you would expect to give our decomposer 2 categories and so do the solvers have; or the solvers ahve 3 categories but then we give the 3 categories to the decomposer?

(also, decompose_objective only still accepts 1 it seems)

tias:
well, the decompose_in_tree docs are at least explicit:

        :param supported: a set of supported global constraints or global functions
        :param supported_reified: a set of supported reified global constraints (globals with Boolean return type only)

and a bit lower

        Supported numerical global functions remain in the expression tree as is. They can be rewritten using
        :func:`cpmpy.transformations.reification.reify_rewrite`
        The following ``bv -> NumExpr <comp> Var/Const`` can be rewritten as  ``[bv -> IV0 <comp> Var/Const, NumExpr == IV0]``.
        So even if numerical constraints are not supported in reified context, we can rewrite them to non-reified versions if they are total.


so I would not grow this PR even larger... we can stick to the two arguments.

(if you are going to change to 2 arguments... can you please declare class attributes at the very top of the class, just below the class docstring, and just above any following @staticmethod; otherwise they are too invisible given Python's flexible class definitions)

tias:
perhaps it should be a frozenset even... (it can still be overwritten as a whole)

tias (approved):
Assuming all tests passed above, this sizable but valuable refactor can be merged, thx!

PR #794, opened 2025-12-02
Abs none

ThomSerg:
Calling `.value()` on and expression containing `abs` will currently throw an error if it contains a None. This can occur when a model is UNSAT and one tries to call `.value()` on one of its expressions afterwards (one containing an `abs`). Fixed implementation such that the `abs` will correctly also return None.

### DIFF
--- a/tests/test_expressions.py
+++ b/tests/test_expressions.py
@@ -638,7 +638,7 @@ class TestBuildIns(unittest.TestCase):
         self.x[0]._value = 0
         self.assertEqual(gt.value(), 0)
         self.x[0]._value = None
-        self.assertEqual(gt.value(), None)
+        self.assertIsNone(gt.value())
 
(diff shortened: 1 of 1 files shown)


tests/test_expressions.py
```diff
@@ -628,6 +628,18 @@ def test_max(self):
...
+        gt = Abs(self.x[0])
+        self.assertEqual(str(gt), str(cp.abs(self.x[0])))
+        self.x[0]._value = 1
+        self.assertEqual(gt.value(), 1)
+        self.x[0]._value = -1
+        self.assertEqual(gt.value(), 1)
+        self.x[0]._value = 0
+        self.assertEqual(gt.value(), 0)
+        self.x[0]._value = None
+        self.assertEqual(gt.value(), None)
```
IgnaceBleukx: Better to use `assertIsNone` here

PR #796, opened 2025-12-03
Linearize unsupported constraints for Pumpkin

### LINKED ISSUE #795: pumpkin does not support exponentiation or modulo division
Currently the following models give a NotImplementedError:

`
model = cp.Model()  

vars = cp.intvar(0,10,shape=2)

model += ((vars[0] % 4) == vars[1])

model.solve(solver="pumpkin")
`

`
model = cp.Model()  

vars = cp.intvar(0,10,shape=2)

model += ((vars[0] % 4) == vars[1])

model.solve(solver="pumpkin")
`

This is because we do not linearize constraints for pumpkin, but exponentiation or modulo division are also not natively supported in pumpkin.
I'm making a PR that linearizes the constraints and in the future this can maybe use the native constraints in pumpkin

OrestisLomis:
fixes #795

### DIFF
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -85,6 +85,10 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum"}, reified=False, c
 
     newlist = []
     for cpm_expr in lst_of_expr:
+        if cpm_expr.name in supported:
+            newlist.append(cpm_expr)
+            continue
+        
         # Boolean literals are handled as trivial linears or unit clauses depending on `supported`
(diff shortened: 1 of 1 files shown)


tias:
Orestis, can you confirm you do this because of `div` and `mod`?

I agree that now their decomposition is hidden in linear, but that this can be worth exposing more generally, e.g. making them a GlobalFunction. This would also be in line with how we are formalizing it in the journal paper (that div and mod are not primitives because rewritable to mult)...

IgnaceBleukx:
I'm actually already working on a refactor for div and mod to become global functions, but the branch is rebased on #793; so haven't opened a PR yet due to the huge diff ; )

PR #807, opened 2025-12-12
Adding Division and Modulo as global functions

IgnaceBleukx:
Division and Modulo were previously operators, but as discussed offline and in #796 it makes more sense for them to be global functions, with an appropriate decomposition.

This branch is based on #793, so blocked by that one before going in.

TODO: make power operator also a global function. (What kind of decomposition do we want? Just a cascaded `*` operation?)

### DIFF
--- a/tests/test_expressions.py
+++ b/tests/test_expressions.py
@@ -6,7 +6,8 @@ from cpmpy.expressions import *
 from cpmpy.expressions.variables import NDVarArray
 from cpmpy.expressions.core import Comparison, Operator, Expression
 from cpmpy.expressions.utils import eval_comparison, get_bounds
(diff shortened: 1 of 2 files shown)


tias:
That's great...

The simplification in linearize is very welcome, with the decomposition much more readable now.

Also that now the 'value' function is not spread out over core is also very welcome. Now we can group the documentation on what exactly this integer division that we have is, same for modulo.

So I (read up and) updated said documentation...


One thing I wonder is all these disabled tests in linearize... do we test equivalent things as part of test_expression or similar? or is disabling them a net loss?
if we test similar things elsewhere, maybe just delete it instead of comment it out?

Other than that: good to merge!

PR #808, opened 2025-12-22
Document linearize and have complete and separated mul decomposition

tias:
So I wanted to better characterize linearize and add documentation.

Then I felt like 'mul' should be treated separately

Then I added multiplication of int*int as well.

But it is not finished... tests are choking on bool*bool especially, which I kept at bool level (half undone already) but then there are bools popping up at places where there is a (trivial) comparison.

Some other notes:

- I wonder if we should make mul between two vars a global constraint after all... then we can take this out of linearize? e.g. there might be a CP solver that does not implement mul...

- We switched away from doing 'decompose_comparison' on GlobalFunctions, in the hope of avoiding unnecessary auxliaries in nested expressions, but I now realize that this way we may include unneeded auxiliary variables for toplevel comparisons, e.g. now `b*v >= 5` will be decomposed as `decomp-of(b*v = aux), aux >= 5`...

- Maybe the previous point is not too bad, because aux>=5 is just a bound update, not a live constraint. But for `b*v = r` we would do `b*v = aux, aux=r`, which happens when flatten flattens some sum that contains `b*v`, which is another argument for making mul of two vars a global constraint: then we can decompose it before flatten?

### DIFF
--- a/cpmpy/solvers/pindakaas.py
+++ b/cpmpy/solvers/pindakaas.py
@@ -238,8 +238,7 @@ class CPM_pindakaas(SolverInterface):
         cpm_cons = no_partial_functions(cpm_cons, safen_toplevel={"div", "mod", "element"})
         cpm_cons = decompose_in_tree(
             cpm_cons,
-            supported=self.supported_global_constraints
-            | {"alldifferent"},  # alldiff has a specialized MIP decomp in linearize
+            supported=self.supported_global_constraints | {"alldifferent"},  # alldiff has a specialized MIP decomp in linearize
             supported_reified=self.supported_reified_global_constraints,
             csemap=self._csemap,
         )


cpmpy/transformations/linearize.py
```diff
@@ -1,56 +1,63 @@
...
-
-
-
+Transform constraints to a linear form.
+
+This transformation is necessary for Integer Linear Programming (ILP) solvers, and also
+for translating to Pseudo-Boolean or CNF formats.
+
+There are a number of components to getting a good linearisation:
+- **Decomposing global constraints/functions** in a 'linear friendly' way.
```
IgnaceBleukx, line 8: Yes, but we currently don't do this yet in this PR? (Only for AllDiff)

cpmpy/transformations/linearize.py
```diff
@@ -1,56 +1,63 @@
...
+  A common pattern is that constraints that enforce domain consistency on integer variables,
+  that they should be decomposed over a Boolean representation of the domain. This is the
```
IgnaceBleukx: broken sentence

cpmpy/transformations/linearize.py
```diff
@@ -1,56 +1,63 @@
...
+  case for :class:`~cpmpy.expressions.globalconstraints.AllDifferent`, :class:`~cpmpy.expressions.globalfunctions.Element`
```
IgnaceBleukx, line 11: certainly also for `Count`

cpmpy/transformations/linearize.py
```diff
@@ -1,56 +1,63 @@
...
+
+- **Linearising multiplication** of variables, e.g. bool*bool, bool*int and int*int.
+  Because '*' is a core Operator and not a global constraint, we need to do it with a helper function here.
+
+- **Disequalities** e.g. `sum(X) != 5` should be rewritten as `(sum(X) < 5) | (sum(X) > 5)`
+  and further flattened into implications and linearised.
+
+- **Implications** e.g. `B -> sum(X) <= 4` should be linearised with the Big-M method.
+
+- **Domain encodings** e.g. of `A=cp.intvar(0,4)` would create binary variables and constraints
```
IgnaceBleukx, line 25: This is a bit surprising to see in the docs here for me.
The other points are "constraint xxx is not linear, and should be linearized as follows...."; but here there is no "input" constraint?

cpmpy/transformations/linearize.py
```diff
@@ -1,56 +1,63 @@
...
+  like `B0 -> A=0`, `B1 -> A=1`, `B2 -> A=2`, etc and `sum(B0..4) = 1`.
```
IgnaceBleukx: `B0..B4`

cpmpy/transformations/linearize.py
```diff
@@ -1,56 +1,63 @@
...
+  However each `B0 -> A=0` would require 2 Big-M constraints. Instead we can linearise the entire
+  domain encoding with two constraints: `sum(B0..4) = 1` and `A = sum(B0..4 * [0,1,2,3,4])`.
+  We could even go as far as eliminate the original integer decision variable.
+
+After linearisation, the output can be further transformed:
+- Remove negated boolean variables (:class:`~cpmpy.expressions.variables.NegBoolView`) so only
```
IgnaceBleukx: negative literals? The variable itself is always positive

cpmpy/transformations/linearize.py
```diff
@@ -1,56 +1,63 @@
...
+Helper functions:
+- :func:`canonical_comparison`: Canonicalizes comparison expressions by moving variables to the
+  left-hand side and constants to the right-hand side.
+- :func:`linearize_mul_comparison`: Linearizes multiplication comparisons (bool*bool | bool*int | int*int) <cmp> rhs.
+
+Post-linearisation transformations:
+- :func:`only_positive_bv`: Transforms constraints so only boolean variables appear positively
+  (no :class:`~cpmpy.expressions.variables.NegBoolView`).
+- :func:`only_positive_bv_wsum`: Helper function that replaces :class:`~cpmpy.expressions.variables.NegBoolView`
+  in var/sum/wsum expressions with equivalent expressions using only :class:`~cpmpy.expressions.variables.BoolVar`.
```
IgnaceBleukx: :class:`~cpmpy.expressions.variables.BoolVarImpl_`
Otherwise the link in the doc will not work

cpmpy/transformations/linearize.py
```diff
@@ -63,18 +70,142 @@
 from ..expressions.utils import is_bool, is_num, eval_comparison, get_bounds, is_true_cst, is_false_cst, is_int
 
 from ..expressions.variables import _BoolVarImpl, boolvar, NegBoolView, _NumVarImpl
+from ..transformations.int2bool import _encode_int_var
+
+def linearize_mul_comparison(cpm_expr: Comparison, supported: Set[str] = {}, reified: bool = False, csemap: Optional[Dict[Any, Any]] = None) -> List[Expression]:
```
IgnaceBleukx: The type of csemap is `Dict[Expression, _NumVarImpl]` I think

cpmpy/transformations/linearize.py
```diff
@@ -63,18 +70,142 @@
...
+        leni2 = (i2.ub - i2.lb) + 1
+        if leni2 < leni1:
+            # swap, call i1 the smallest integer
+            i1,i2 = i2,i1
+
+        # encode i1 with Booleans (with temproary ivarmap)
+        encoding = "direct"
+        if min(leni1,leni2) >= 8:  # arbitrary heuristic
+            encoding = "binary"  # results in fewer Bools
+        i1_enc, cons = _encode_int_var({}, i1, encoding)  # {}: no ivarmap used
```
IgnaceBleukx, line 139: So here there is no csemap-check?
That means that if we linearize two constraints (x * y <= 4) and (x * z <= 6); we will Booleanize x twice?

cpmpy/transformations/linearize.py
```diff
@@ -63,18 +70,142 @@
...
+
+        # channel i1 to the Bools
+        (encpairs, offset) = i1_enc.encode_term()
+        cons += [i1 - cp.sum(v*bv for v,bv in encpairs) == offset]
+
+        # Build the sum: aux = sum(v * (b_v * i2) for v,bv in encoding)
+        terms = []
+        for v, b_v in encpairs:
+            mymul = b_v * i2
+            myaux = cp.intvar(*mymul.get_bounds())
```
IgnaceBleukx: this should go through get_or_make_var so we use the csemap

IgnaceBleukx: I guess that should also fix my previous comment? Not entirely sure

cpmpy/transformations/linearize.py
```diff
@@ -63,18 +70,142 @@
...
+
+            # the definition of myaux, to linearize
+            cons += linearize_mul_comparison(mymul == myaux, supported=supported, reified=reified, csemap=csemap)
```
IgnaceBleukx: Could also remove this linearize call, as we are linearizing again in the return anyway

cpmpy/transformations/linearize.py
```diff
@@ -63,18 +70,142 @@
...
+        * Boolean expressions:
+          - ``GenExpr`` (when :func:`~cpmpy.expressions.core.Expression.is_bool()`)
+
+        * Numeric comparisons:
+          - ``GenExpr == Var/Constant`` (when GenExpr is numeric)
+          - ``GenExpr <= Var/Constant`` (when GenExpr is numeric)
+          - ``GenExpr >= Var/Constant`` (when GenExpr is numeric)
+
+    **Indicator constraints, if '->' is in 'supported':**
+        The left-hand side is always a boolean variable or its negation, and
```
IgnaceBleukx: Boolean

cpmpy/transformations/linearize.py
```diff
@@ -63,18 +70,142 @@
...
-        reified: whether the constraint is fully reified
+        lst_of_expr: List of CPMpy expressions to linearize. Must be in 'flat normal form'
+            with only boolean variables on the left-hand side of implications.
+        supported: Set of constraint and variable type names that are supported by the target
+            solver, e.g. `{"sum", "wsum", "->", "and", "or", "alldifferent"}`.
+            :class:`~cpmpy.expressions.globalconstraints.AllDifferent` has a special linearization
+            and is decomposed as such if not in `supported`.
+            Any other unsupported global constraint should be decomposed using
+            :func:`cpmpy.transformations.decompose_global.decompose_in_tree()`.
+        reified: Whether the constraint is fully reified. When True, nested implications
```
IgnaceBleukx: Not fully reified? We only support half-reified in linearize

cpmpy/transformations/linearize.py
```diff
@@ -367,18 +489,20 @@ def only_positive_bv(lst_of_expr, csemap=None):
...
-
-        It might add a constant term to the expression, if you want the constant separately, use :func:`only_positive_bv_wsum_const`.
-        
-        Arguments:
-        - `cpm_expr`: linear expression (sum, wsum, var)
-        
-        Returns tuple of:
-        - `pos_expr`: linear expression (sum, wsum, var) without NegBoolView
+    Replaces a var/sum/wsum expression containing :class:`~cpmpy.expressions.variables.NegBoolView`
+    with an equivalent expression using only :class:`~cpmpy.expressions.variables.BoolVar`.
```
IgnaceBleukx: also BoolVarImpl_ here

cpmpy/transformations/linearize.py
```diff
@@ -391,20 +515,23 @@ def only_positive_bv_wsum(expr):
...
-        
-        Arguments:
-        - `cpm_expr`: linear expression (sum, wsum, var)
-        
-        Returns tuple of:
-        - `pos_expr`: linear expression (sum, wsum, var) without NegBoolView
-        - `const`: The difference between the original expression and the new expression, 
-                   i.e. a constant term that must be added to pos_expr to be an equivalent linear expression.
+    Replaces a var/sum/wsum expression containing :class:`~cpmpy.expressions.variables.NegBoolView`
+    with an equivalent expression using only :class:`~cpmpy.expressions.variables.BoolVar`,
```
IgnaceBleukx: BoolVarImpl_

cpmpy/transformations/linearize.py
```diff
@@ -367,18 +489,20 @@ def only_positive_bv(lst_of_expr, csemap=None):
...
-        - `cpm_expr`: linear expression (sum, wsum, var)
-        
-        Returns tuple of:
-        - `pos_expr`: linear expression (sum, wsum, var) without NegBoolView
+    Replaces a var/sum/wsum expression containing :class:`~cpmpy.expressions.variables.NegBoolView`
+    with an equivalent expression using only :class:`~cpmpy.expressions.variables.BoolVar`.
+
+    It might add a constant term to the expression. If you want the constant separately,
+    use :func:`only_positive_bv_wsum_const`.
+
```
IgnaceBleukx, line 635: I'm wondering if this doc needs an example.

"E.g., the weighed sum `3*~x + 4*y` is transformed to `3 + -3*x + 4*y`. If you want the constant 3 to be separate, use `only_positive_bv_wsum_const` instead."

tests/test_trans_linearize.py
```diff
@@ -231,6 +231,47 @@ def assert_cons_is_true(cons):
...
+        cons_cnt = cp.Model(cons).solveAll(display=assert_cons_is_true(cons))
+        self.assertEqual(lin_cnt, cons_cnt)
+
+        # Test inequality: x * y >= z
+        cons = x * y >= z
+        lin_cons = linearize_constraint([cons], supported={"sum", "wsum"})
+        lin_cnt = cp.Model(lin_cons).solveAll(display=assert_cons_is_true(cons))
+        cons_cnt = cp.Model(cons).solveAll(display=assert_cons_is_true(cons))
+        self.assertEqual(lin_cnt, cons_cnt)
+
```
IgnaceBleukx, line 264: Also test != and reified mult?

cpmpy/transformations/linearize.py
tias: no, cons is not re-linearized and this one is 'shorter' this way

tias:
First extracting mul as a global (and fixing it) would have been the easier route retrospectively...

This rewrite should be a good basis for further improvements that will simplify it (e.g. doing globals before linearize, incl mul). 

Some things that surfaced:
* should linearize output a single boolvar or only a sum? it already is allowed to output a boolval, so I think it shoudl also be alloed to output a boolvar, will take away some silly recursions
* int2bool does not use csemap, that should be fixed...
* should implications also be in csemap, or in a separate cremap... we should make a decision at some point
* flatten should only half-reify when positive, that will make the mul decompositions (the output) notably simpler

The implimentation style of having a function with a 'condition' argument is quite clean... its what Henk used for pindakaas. We could consider it when rewriting flatten too. (perhaps we should rename it from 'implication_literal' to 'condition' indeed, or actually in pindakaas its 'conditions' as multiple are accepted, which makes sense...)

I don't think we should add more to this PR though

tias:
merged #811 in, now only the 4 panics from #810 remain.

The slowness of pindakaas also remains; pysat is notably faster, but still takes quite some time on some tests too.

However, another thing I notice is memory use: when running test/test_constraints with only pysat, memory climbs up linearly, at 30% it was around 15Gb or so... so something is not cleaning up properly. Is it that the python garbage collector is not being run, so it is not cleaning up all our Booleans? or is it because maybe we are not closing the solver objects properly, so they keep their memory alive? or they keep our python objects alive?

I have 0 experience with memory and Python, if somebody knows how to look into this, by all means... (running the tests with ortools stays at less then a Gb...)

PR #809, opened 2026-01-02
Fix Hexaly Callback when none provided

IgnaceBleukx:
Merged #787 slightly too fast.
On my version of hexaly (14.0) posting "None" as callback the API crashes.

Small change to only post the callback when it is indeed provided.


ThomSerg (approved):
Oeps, did encounter this before but probably forgot to push the same fix.

PR #811, opened 2026-01-05
Fix/bug810 handle pdk unsat with conditions

### LINKED ISSUE #810: pindakaas throws pdk.Unsat on a reified sum
Not CPMpy, but still reporting here... surfaced in #808 

```python
C = [(~p).implies(3*cp.boolvar() == 2)]
solv = "pindakaas"
print("***", solv, cp.Model(C).solve(solv))
# _post_constraint(3*b == 2, conditions=[~p]) raises pdk.Unsatisfiable
print("b7", b7.value(), "p", p.value())
print("+++", argval(C[0]))
print("***", "ort", cp.Model(C).solve())
```

hbierlee:
Fixes #810, plus refactors the conditions handling, and auto-formatting changes. The important changes are in `_post_constraint`. The problem was in the CPMpy `pindakaas.py` side, where trivial unsatisfiability of a constraint implied by conditions led to unsatisfiability of the whole model, rather than posting the negated conditions.

### DIFF
--- a/cpmpy/expressions/python_builtins.py
+++ b/cpmpy/expressions/python_builtins.py
@@ -142,15 +142,14 @@ def min(*iterable, **kwargs):
     return Minimum(iterable)
 
 
-def sum(*iterable, **kwargs):
+def sum(iterable, **kwargs):
     """
         sum() overwrites the python built-in to support decision variables.
 
         if iterable does not contain CPMpy expressions, the built-in is called
         checks if all constants and uses built-in sum() in that case
     """
-    if len(iterable) == 1:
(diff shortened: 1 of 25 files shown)


hbierlee:
Wow, this is from a feature which I didn't know was on: it does bounds propagation on the integer model, but in experiments this didn't help too much, so it should be off. This, plus the fact we use Totalizer i/o Adder (as I previously thought), could explain the additional memory errors over PySAT.

In any case, should be automatically fixed upstream in https://github.com/pindakaashq/pindakaas/pull/176

PR #817, opened 2026-01-14
Pytest improvements core

ThomSerg:
The PR has a similar target as #780, but tries to keep the number of changes manageable. It introduces the core infrastructure for a solver-parametrised testsuite, but only parametrises the tests that already ran on all solvers. So from the outside, all the same tests as currently on master will be run (i.e. only test_constraints, test_solverinterface, and test_solvers_solhint run on all solvers). But now this parametrisation is done more centrally in `conftest.py`, with the added ability to controll this parametrisation through the pytest argument `--solver`.

This is the start for a sequence of future PRs to improve the testsuite. 
E.g.:
- run `pytestify` to remove unittest.TastCase dependency
- make similar changes to other test files as for test_constraints, test_solverinterface, test_solvers_solhint 
- remove use of * imports
- ... (other improvements made in #780, added piecewise)

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -22,17 +22,11 @@ ADVANCED_EXAMPLES = glob(join("examples", "advanced", "*.py"))
 EXAMPLES = sorted(EXAMPLES)
 ADVANCED_EXAMPLES = sorted(ADVANCED_EXAMPLES)
 
-# Examples that get skipped (with a message)
 SKIPPED_EXAMPLES = [
                     "ocus_explanations.py", # waiting for issues to be resolved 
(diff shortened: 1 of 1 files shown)


tias (approved):
Added a small doc update so that it is clear that it impacts test_constraints and will gradually be expanded (I think it is already a bit more expanded, but this leaves that sufficiently open)

Both standard and --solvers all runs fine on my system, except (this is an issue for some time, but might as well fix it now...):

examples/advanced/decision_focussed_learning.py is hardcoded to depend on gurobi, but it is run by test_advanced_example. Can we maybe add it to the skiplist, and then have a separate function `test_advanced_example_dfl` that requires gurobi and calls only this one example?

PR #818, opened 2026-01-14
Pytestify TestCase

ThomSerg:
Used the pytestify script (https://github.com/dannysepler/pytestify) to convert the unittest TastCase classes into pytest plain classes. Follow up of #817 (which must be merged first).


ThomSerg:
Since we are only required to remove the use of TestCase when further parametrising the testsuite (for the current parametrisation not needed) this PR can be left out of the next release.

PR #824, opened 2026-01-15
Fix clause bypass issue

### LINKED ISSUE #823: `to_cnf` introduces unwanted solutions
Consider the following program (cpmpy on a8dcc3e, i.e, main on the time of writing):

```
import cpmpy as cp
from cpmpy.transformations.to_cnf import to_cnf

x = cp.boolvar(shape=2, name="x")

cons = ( (cp.sum(x) == 0) | (cp.all(x)) ) # expected solutions: [True, True], [False, False]

# === original ===
print("=== original ===")
model = cp.Model(cons)
model.solveAll(display=x)


# === with CNF ===
print("\n=== CNF ===")
model = cp.Model(to_cnf(cons))
model.solveAll(display=x)
```

From my understanding, there could very well be more solutions after translating to CNF, but they should all coincide on `x` with one of the original solutions. That is not the case here. Output:

```
=== original ===
[ True  True]
[False False]

=== CNF ===
[ True False] <---
[ True False] <---
[ True  True]
[False  True] <---
[False  True] <---
[ True  True]
[False False]
[False False]
[False False]
[False  True] <---
[ True  True]
[ True  True]
[ True False] <---
```

hbierlee:
Fixes #823

When using `to_cnf`, we bypass the pindakaas solver for simple clauses. However, this seems to introduce some bug. This PR will fix it, and the first commit removes the bypass so `to_cnf` can be used in the meantime.

### DIFF
--- a/cpmpy/transformations/to_cnf.py
+++ b/cpmpy/transformations/to_cnf.py
@@ -20,9 +20,7 @@ def to_cnf(constraints, csemap=None, ivarmap=None, encoding="auto"):
         Equivalent CPMpy constraints in CNF, and the updated `ivarmap`
     """
     if not CPM_pindakaas.supported():
-        raise ImportError(
-            f"Install the Pindakaas python library `pindakaas` (e.g. `pip install pindakaas`) package to use the `to_cnf` transformation"
-        )
+        raise ImportError(f"Install the Pindakaas python library `pindakaas` (e.g. `pip install pindakaas`) package to use the `to_cnf` transformation")
 
     import pindakaas as pdk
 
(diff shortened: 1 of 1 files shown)


hbierlee:
Well, the issue is that the auxiliary variables introduced during transformation are not collected as existing variables in `cpmpy_vars`, only the user variables are. This risks adding a constraint with some aux var `y` to pindakaas, then reading back its clauses with `y` in it, then creating a new variable for it since it is not in `cpmpy_vars`. My attempts to fix this make this too complicated, so I'd rather remove this optimization.

If for some applications, `to_cnf` becomes too slow, we can see if reading back the clauses (which does require each lit to be looked up in `cpmpy_vars`) is a bottleneck, and if so, try to fix it.

PR #825, opened 2026-01-15
solvers: throw ModuleNotFoundError when module not installed, 

tias:
catchable (e.g. in test runner)

now examples tester no longer has the gurobi failure for me

### DIFF
--- a/cpmpy/solvers/minizinc.py
+++ b/cpmpy/solvers/minizinc.py
@@ -263,7 +263,7 @@ class CPM_minizinc(SolverInterface):
         if not self.installed():
             raise ModuleNotFoundError("CPM_minizinc: Install the python package 'minizinc' to use this solver interface.")
         elif not self.executable_installed():
-            raise Exception("CPM_minizinc: Install the MiniZinc executable and make it available in path.")
+            raise ModuleNotFoundError("CPM_minizinc: Install the MiniZinc executable and make it available in path.")
         elif self.outdated():
             version = str(self.required_version[0])
             for x in self.required_version[1:]:


cpmpy/solvers/minizinc.py
```diff
@@ -261,7 +261,7 @@ def __init__(self, cpm_model=None, subsolver=None):
                           has to be one of solvernames()
         """
         if not self.installed():
-            raise Exception("CPM_minizinc: Install the python package 'minizinc' to use this solver interface.")
+            raise ModuleNotFoundError("CPM_minizinc: Install the python package 'minizinc' to use this solver interface.")
         elif not self.executable_installed():
             raise Exception("CPM_minizinc: Install the MiniZinc executable and make it available in path.")
```
ThomSerg: Should this also become ModuleNotFoundError? If somebody installs minizinc without the executable, tests will still fail. Maybe that's exactly what we want, because why would somebody install the python wrapper without then wanting to use it for controlling the executable?

cpmpy/solvers/hexaly.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, cpm_model=None, subsolver=None):
         - subsolver: str, name of a subsolver (optional)
         """
         if not self.supported():
-            raise Exception("CPM_hexaly: Install the python package 'hexaly' to use this solver interface.")
+            raise ModuleNotFoundError("CPM_hexaly: Install the python package 'hexaly' to use this solver interface.")
```
ThomSerg, line 103: Now looking at it, doesn't Hexaly always require a license? And also the installation of an executable? So it seems to be missing both an `executable_installed()` and an `license_ok()` check. Will be for separate PR.

cpmpy/solvers/minizinc.py
tias: missed this one

cpmpy/solvers/hexaly.py
ThomSerg: See #826

PR #830, opened 2026-01-17
Save `end_is_none` flag in Cumulative and NoOverlap

### LINKED ISSUE #828: None argument in Cumulative (demand) not allowed by safening
Expression for the capacity of a cumulative constraint seems to trigger safening, which doesn't allow for the "None" argument in the constraint.

Minimal example:
```python
import cpmpy as cp

def original_model():
    p = cp.boolvar(name='p')
    start_array_0 = cp.intvar(0, 10, name='start_array_0')
    start_array_1 = cp.intvar(0, 10, name='start_array_1')
    start_array_2 = cp.intvar(0, 10, name='start_array_2')

    constraints = [
        (p).implies((p > cp.Cumulative([start_array_0, start_array_1, start_array_2], [1, 4, 3], None, [4, 5, 7], 10))),
    ]
    return cp.Model(constraints)


def model():
    p = cp.boolvar(name='p')
    start_array_0 = cp.intvar(0, 10, name='start_array_0')
    start_array_1 = cp.intvar(0, 10, name='start_array_1')
    start_array_2 = cp.intvar(0, 10, name='start_array_2')
    _IV0 = cp.intvar(0, 0, name='_IV0')

    constraints = [
        (p).implies((p > cp.Cumulative([start_array_0, start_array_1, start_array_2], [1, 4, 3], None, [4, 5, 7], cp.sum([10, (-_IV0)])))),
    ]
    return cp.Model(constraints)


if __name__ == '__main__':
   model().solve()
   original_model().solve()
```

```console
  File "/home/thomserg/Documents/programming/fuzz-tester/bug.py", line 29, in <module>
    model().solve()
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/model.py", line 194, in solve
    s = SolverLookup.get(solver, self)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/solvers/utils.py", line 148, in get
    return solver_cls(model, subsolver=subname, **init_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/solvers/ortools.py", line 140, in __init__
    super().__init__(name="ortools", cpm_model=cpm_model)
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/solvers/solver_interface.py", line 95, in __init__
    self += cpm_model.constraints
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/solvers/ortools.py", line 455, in add
    for con in self.transform(cpm_expr):
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/solvers/ortools.py", line 420, in transform
    cpm_cons = no_partial_functions(cpm_cons, safen_toplevel=frozenset({"div", "mod"})) # before decompose, assumes total decomposition for partial functions
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/transformations/safening.py", line 109, in no_partial_functions
    new_args = no_partial_functions(args, _toplevel, _nbc, safen_toplevel=safen_toplevel)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/transformations/safening.py", line 109, in no_partial_functions
    new_args = no_partial_functions(args, _toplevel, _nbc, safen_toplevel=safen_toplevel)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/transformations/safening.py", line 109, in no_partial_functions
    new_args = no_partial_functions(args, _toplevel, _nbc, safen_toplevel=safen_toplevel)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/transformations/safening.py", line 99, in no_partial_functions
    assert isinstance(cpm_expr, Expression), f"each `cpm_expr` should be an Expression at this point, not {type(cpm_expr)} {lst_of_expr}"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: each `cpm_expr` should be an Expression at this point, not <class 'NoneType'>
```

IgnaceBleukx:
Fixes #828; not sure if this is the best way to do it, but seems like the most obvious for now.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -1215,7 +1215,7 @@ class Cumulative(GlobalConstraint):
 
         # ensure demand doesn't exceed capacity
         lb, ub = min(start), max(end)
-        np_start, np_end = np.array(start), np.array(end) # eases check below
+        np_start, np_end = np.asanyarray(start), np.asanyarray(end) # eases check below
         for t in range(lb, ub+1):
             if capacity < sum(demand * ((np_start <= t) & (np_end > t))):
                 return False
@@ -1420,7 +1420,7 @@ class CumulativeOptional(GlobalConstraint):
 
         # ensure demand doesn't exceed capacity
         lb, ub = min(start), max(end)
-        np_start, np_end, np_present = np.array(start), np.array(end), np.array(is_present) # eases check below
+        np_start, np_end, np_present = np.asanyarray(start), np.asanyarray(end), np.asanyarray(is_present) # eases check below
         for t in range(lb, ub+1):
             if capacity < sum(demand * (np_present & (np_start <= t) & (np_end > t))):
                 return False
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -1197,18 +1197,18 @@ class Cumulative(GlobalConstraint):
         """
 
         if self.end_is_none:
-            start, dur, demand, capacity = argvals(self.args)
+            start, duration,demand, capacity = argvals(self.args)
             if any(a is None for a in start + dur + demand + [capacity]):
                 return None
             end = [start[i] + dur[i] for i in range(len(start))]
         else:
-            start, dur, end, demand, capacity = argvals(self.args)
+            start, duration,end, demand, capacity = argvals(self.args)
             if any(a is None for a in start + dur + end + demand + [capacity]):
                 return None
                 
         if any(d < 0 for d in dur):
             return False
-        if any(s + d != e for s,d,e in zip(start, dur, end)):
+        if any(s + d != e for s,d,e in zip(start, duration,end)):
             return False
 
         if any(d < 0 for d in demand):
@@ -1403,18 +1403,18 @@ class CumulativeOptional(GlobalConstraint):
         """        
 
         if self.end_is_none:
-            start, dur, demand, capacity, is_present = argvals(self.args)
+            start, duration,demand, capacity, is_present = argvals(self.args)
             if any(a is None for a in start + dur + demand + [capacity] + is_present):
                 return None
             end = [s + d for s,d in zip(start, dur)]
         else:
-            start, dur, end, demand, capacity, is_present = argvals(self.args)
+            start, duration,end, demand, capacity, is_present = argvals(self.args)
             if any(a is None for a in start + dur + end + demand + [capacity] + is_present):
                 return None
 
-        if any(p and d < 0 for d,p in zip(dur, is_present)):
+        if any(p and d < 0 for d,p in zip(duration,is_present)):
             return False
-        if any(p and s + d != e for s,d,e,p in zip(start, dur, end, is_present)):
+        if any(p and s + d != e for s,d,e,p in zip(start, duration,end, is_present)):
             return False
 
         if any(p and d < 0 for d,p in zip(demand, is_present)):
@@ -1496,13 +1496,13 @@ class NoOverlap(GlobalConstraint):
                 return None
             end = [s + d for s,d in zip(start, dur)]
         else:
-            start, dur, end = argvals(self.args)
-            if any(a is None for a in [start, dur, end]):
+            start, duration,end = argvals(self.args)
+            if any(a is None for a in [start, duration,end]):
                 return None
        
         if any(d < 0 for d in dur):
             return False
-        if any(s + d != e for s,d,e in zip(start, dur, end)):
+        if any(s + d != e for s,d,e in zip(start, duration,end)):
             return False
         for (s1,d1), (s2,d2) in all_pairs(zip(start,dur)):
             if s1 + d1 > s2 and s2 + d2 > s1:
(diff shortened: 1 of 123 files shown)


IgnaceBleukx:
Now that I think about it, probably the right way of doing this, is to let the expressions decide which arguments should recusively be transformed.
E.g., each expression would get a `get_expr_args` and `update_expr_args` function, which returns the arguments that __could__ be an expression. This will have a few advantages: for e.g., Table, we won't ever recurse into the table, even if the array contains subexpression (and thus `has_subexpr` will be True).
For Cumulative, we could just return [start,dur,height,capacity] if the end variable is None.

tias:
so no, after more playing with Typing, the realisation is now:

* Expression subclasses are perfectly OK to store their own non-expression data in their subclass; no need to feed data to the superclass args
* so, Cumulative (and nooverlap), should store a `end_is_none` flag in their subclass instead of passing a 'None' to superclass args, and the code should be updated to check that...

then transformations remain untouched as intended, and you don need `novar` things.

IgnaceBleukx:
Updated this PR to indeed have a `end_is_none` flag saved on the expression object. Whenever end is None, it's simply not passed to the superclass.

Keeping it as draft for now, as there will some merge issues with #676, so leaving it unmerged until that one is in master.

TODO:
- [x] Wait for #676  to be in master
- [x] Update CPO interface

cpmpy/expressions/globalconstraints.py
```diff
@@ -1183,31 +1195,30 @@ def value(self) -> Optional[bool]:
...
+        if any(s + d != e for s,d,e in zip(start, duration,end)):
             return False
 
         if any(d < 0 for d in demand):
             return False
 
         # ensure demand doesn't exceed capacity
         lb, ub = min(start), max(end)
-        start, end = np.array(start), np.array(end) # eases check below
+        np_start, np_end = np.array(start), np.array(end) # eases check below
```
tias: we probably want to do 'np.asanyarray' here, so NDVarArrays get returned without copy?

cpmpy/expressions/globalconstraints.py
```diff
@@ -1116,12 +1120,17 @@ def _consistency_constraints(self) -> list[Expression]:
...
         - start + duration == end
         """
-        start, duration, end, demand, capacity = self.args
-        cons = [d >= 0 for d in duration]  # enforce non-negative durations
-        cons += [h >= 0 for h in demand]  # enforce non-negative demand
 
-        if end is not None:
+
+        cons = []
+        if self.end_is_none:
```
tias: I suddenly realize we don't actually need a separate flag...

if len(args) == 4:  # no end
else: # with end

cpmpy/solvers/choco.py
```diff
@@ -645,30 +645,35 @@ def _get_constraint(self, cpm_expr):
                 expr, table = self.solver_vars(cpm_expr.args)
                 return self.chc_model.member(expr, table)
             elif cpm_expr.name == "cumulative":
-                start, dur, end, demand, cap = cpm_expr.args
-                # Choco allows negative durations, but this does not match CPMpy spec
-                dur, extra_cons = get_nonneg_args(dur)
-                # Choco allows negative demand, but this does not match CPMpy spec
+                assert isinstance(cpm_expr, Cumulative) # ensure hasattr end_is_none
+                if cpm_expr.end_is_none:
```
tias: implicit_end == (len(cpm_expr.args) == 4)
if implicit_end:
   ...

cpmpy/solvers/ortools.py
```diff
@@ -590,14 +590,17 @@ def _post_constraint(self, cpm_expr, reifiable=False):
                 return self.ort_model.AddAutomaton(array, cpm_expr.node_map[start], [cpm_expr.node_map[n] for n in accepting], 
                                                    [(cpm_expr.node_map[src], label, cpm_expr.node_map[dst]) for src, label, dst in transitions])
             elif cpm_expr.name == "cumulative":
-                start, dur, end, demand, cap = cpm_expr.args
+                assert isinstance(cpm_expr, Cumulative) # ensure hasattr end_is_none
+                if cpm_expr.end_is_none:
+                    start, dur, demand, cap = cpm_expr.args
+                    end = [intvar(*get_bounds(s+d)) for s,d in zip(start, dur)]
```
tias: ortools also has an `new_fixed_size_interval_var()`

I'm a bit confused, I can't find the docs I'm used to... the docs I do find are all messy, and they use underscore based names; while we use CamelBack names? e.g. `NewIntervalVar()` vs `new_interval_var()`?
https://or-tools.github.io/docs/pdoc/ortools/sat/python/cp_model.html

so, my original point is: this should just be about first creating the right kind of interval var, then posting the constraint. In case `end_is_none` and demand is constant, the better thing to do is to create a `new_fixed_size_interval_var()` rather then a `new_interval_var()`

tias (commented):
great, and we should do that indeed.

I just had one simplication realisation: we don't need the flag, we can just check `len(args)`... this would be simpler/cleaner I think...

Would you be OK changing it to that, or do you prefer it this way?

(also smth ortools specific)

cpmpy/solvers/ortools.py
IgnaceBleukx: Yes, OR-Tools refactored their API to snake_case, acually not sure how long the camelCase is still supported. We should find out which version introduced the snake case, update our API and bump the min requirements. I think I read somewhere there should be a `sed` file to automatically replace all occurences.

For the issue here: I'll use the fixed intervals indeed if duration is constant.

PR #836, opened 2026-01-19
Decompose linear

IgnaceBleukx:
Starting from #835, allow for `decompose_custom` which is populated with a new transformation "decompose_linear" containing a few specialized decompositions.

Only draft for now, and some open questions:
- For non-Boolean solvers, do we also add the constraint `encoding == intvar` or do we eliminate the intvar all together?
- Where should we put the decompose code? We should be able to access ivarmap, so for me it made the most sense to put them into the linearize file. Alternatively, we can detect `bv == (iv == val)` style constraints after flattening, and we don't do the encoding explicitely during decompose.

### DIFF
--- a/cpmpy/solvers/pindakaas.py
+++ b/cpmpy/solvers/pindakaas.py
@@ -136,7 +136,7 @@ class CPM_pindakaas(SolverInterface):
                 user_vars.update(self.ivarmap[x.name].vars())
         return user_vars
 
-    def solve(self, time_limit:Optional[float]=None, assumptions:Optional[List[_BoolVarImpl]]=None):
+    def solve(self, time_limit: Optional[float] = None, assumptions: Optional[List[_BoolVarImpl]] = None):
         """
         Solve the encoded CPMpy model given optional time limit and assumptions, returning whether a solution was found.
 
@@ -157,10 +157,7 @@ class CPM_pindakaas(SolverInterface):
         solver_assumptions = None if assumptions is None else self.solver_vars(assumptions)
 
         t = time.time()
-        with self.pdk_solver.solve(
-            time_limit=time_limit,
-            assumptions=solver_assumptions,
-        ) as result:
+        with self.pdk_solver.solve(time_limit=time_limit, assumptions=solver_assumptions) as result:
             self.cpm_status.runtime = time.time() - t
 
             # translate pindakaas result status to cpmpy status
@@ -222,9 +219,7 @@ class CPM_pindakaas(SolverInterface):
             return self._varmap[cpm_var.name]
         elif isinstance(cpm_var, _IntVarImpl):  # intvar
             if cpm_var.name not in self.ivarmap:
-                enc, cons = _encode_int_var(
-                    self.ivarmap, cpm_var, _decide_encoding(cpm_var, None, encoding=self.encoding)
-                )
+                enc, cons = _encode_int_var(self.ivarmap, cpm_var, _decide_encoding(cpm_var, None, encoding=self.encoding))
                 self += cons
             else:
                 enc = self.ivarmap[cpm_var.name]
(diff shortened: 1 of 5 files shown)


tias:
Thought some more about it; your proposal is the better way forward.

And we don't need answers to the half-reif question to move forward with this improvement to our current linearisation.

IgnaceBleukx:
Ok managed to fix the element issue, small oversight in the decomposition on my side ; )

What other globals would we want to make a linear friendly decomposition for?
Table/NegTable, InDomain, Count and GCC come to mind. (Maybe Among?)

Also the the ExceptN variants probably?

tias:
Great to see this happening.

Why do you use encode/intvarmap instead of the more straightforward (var == val) expressions, which CSE has to pick up later then?
The decompose_linear is called before flatten right, so the decompositions could look like any other decomposition?

(this would push the responsability of handling the proper encoding of the var and the 'keep_integer' to later in the stack)

tias:
chatted with an agent about my rough ideas; decided a new transformation instead of doing two-stage could work well.

Small test example:
```
import cpmpy as cp
x = cp.intvar(1, 3, shape=3, name=("a", "b", "c"))
cons = cp.AllDifferent(x)
cp.SolverLookup.get("exact").transform(cons)
```
gives
```[sum([BV0, BV1, BV2]) <= 1, sum([BV3, BV4, BV5]) <= 1, sum([BV6, BV7, BV8]) <= 1, sum([BV0, BV3, BV6]) == 1, sum([1, 2, 3, -1] * [BV0, BV3, BV6, a]) == 0, sum([BV1, BV4, BV7]) == 1, sum([1, 2, 3, -1] * [BV1, BV4, BV7, b]) == 0, sum([BV2, BV5, BV8]) == 1, sum([1, 2, 3, -1] * [BV2, BV5, BV8, c]) == 0]```

We could give the new transformation your original `keep_integer` argument so a solver can indicate the last two are not needed?

File: cpmpy/transformations/linearize.py [hunk 1]
```diff
@@ -598,3 +580,162 @@ def only_positive_coefficients(lst_of_expr):
...
+        if not is_num(rhs):
+            continue
+
+        var = _extract_var_from_lhs(lhs)
+        if var is None:
+            continue
+
+        if expr.name == "==":
+            val = int(rhs)
+            encodings.setdefault(var, []).append((val, bv, idx))
```
IgnaceBleukx: Hmm I didn't know this trick!
I guess it avoids the akward `if var not in encodings: encodings[var] = []; encodings[var].append...`.
So it actually saves us a call to `__hash__`! In a future PR we should do this everywher (esp in csemap and `solver_var`)

cpmpy/transformations/linearize.py
```diff
@@ -598,3 +580,162 @@ def only_positive_coefficients(lst_of_expr):
...
+        for _, _, i in triples:
+            to_remove.add(i)
+        pairs = sorted(distinct_pairs, key=lambda p: p[0])
+        values = [p[0] for p in pairs]
+        bvs = [p[1] for p in pairs]
+        added.append(Comparison("==", Operator("sum", bvs), 1))
+        added.append(Comparison("==", Operator("wsum", [values, bvs]), var))
+
+    to_remove = frozenset(to_remove)
+    newlist = [constraints[i] for i in range(len(constraints)) if i not in to_remove] + added
```
IgnaceBleukx: Style: I prefer `i, cons in enumerate(constraints)`

tias:
Yes, it should use the ivarmap/int2bool infrastructure, absolutely.

At this point its more a prototype that shows that we don't need ivarmaps in the decompositions, but that we can have these more intuitive decompositions and then 'catch' the separate (iv=v)s and linearize them together instead of individual big-Ms.

So we should split the PR in 2: one the actual decompose_linear, but with the clean decompositions (this one); another with the new 'linearize_reified_variables()' transformation (properly implemented and tested)

hbierlee:
(the duplicate sum is just an artifact of testing, the decomposition itself should be ok to use, except for that it does not work for negated. I'm not sure if there are workarounds for that)

IgnaceBleukx (commented):
Great! So this PR actually became quite minimal changes, with only alldiff getting a new decomposition.
Only open question is indeed the table decomposition.

(I can't approve it as I opened the initial PR...)

PR #839, opened 2026-01-29
Non-strict var name check

ThomSerg:
Added some internal tools to have greater control over variable name checks.
When working on file readers and writers, I encountered the issue that we couldn't read the model files that we ourselves have exported, due to the use of IV and BV decision variable namings. In this PR I added a couple of helper functions to allow for these variable names as long as they haven't been used before.

Now, file readers can wrap their logic in 
```python
with ignore_strict_variable_name_check():
    ... create CPMpy model based on file contents here ...
```
This will temporarily allow the use of restricted variable names (but still checks for variable name collisions).

Additionally I also extracted the "update variable counters from model" logic used when loading models from `.pickle` files to make it reusable (`_update_variable_counters`) (file readers will also need this) 

For now, readers will still have to update the counters themselves after exiting `ignore_strict_variable_name_check`, i.e. after creating the CPMpy model they should call:
```python
_update_variable_counters(model)
```

But I might have found a way to do this automatically (that could potentially also resolve the issues when copying CPMpy models across processes, but to be seen / no promises) (different approach than Ignace's). Will be for a follow-up PR.

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -799,7 +799,7 @@ def _is_invalid_name(name):
     and the variables' counter is greater than the index, i.e. the name is already in use.
 
     Toggle the strict mode with `_enable_strict_variable_name_check()` and `_disable_strict_variable_name_check()`,
-    or use the context manager `ignore_strict_variable_name_check()`.
+    or use the context manager `_ignore_strict_variable_name_check()`.
     """
     if name.startswith(_IV_PREFIX):
         if _VAR_STRICT_NAME_CHECK:
@@ -833,7 +833,7 @@ def _disable_strict_variable_name_check():
     _VAR_STRICT_NAME_CHECK = False
 
 
-def ignore_strict_variable_name_check():
+def _ignore_strict_variable_name_check():
     """
     Context manager to temporarily disable strict variable name check.
     """
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -68,6 +68,7 @@ from .utils import is_num, is_int, flatlist, is_boolexpr, is_true_cst, is_false_
 _BV_PREFIX = "BV"
 _IV_PREFIX = "IV"
 _VAR_ERR  = f"Variable names starting with {_IV_PREFIX} or {_BV_PREFIX} are reserved for internal use only, chose a different name"
+_VAR_STRICT_NAME_CHECK = True
 
 def BoolVar(shape=1, name=None):
     """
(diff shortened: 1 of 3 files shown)


cpmpy/expressions/variables.py
```diff
@@ -790,5 +791,58 @@ def _genname(basename, idxs):
...
+def _enable_strict_variable_name_check():
+    global _VAR_STRICT_NAME_CHECK
+    _VAR_STRICT_NAME_CHECK = True
+
+def _disable_strict_variable_name_check():
+    global _VAR_STRICT_NAME_CHECK
+    _VAR_STRICT_NAME_CHECK = False
+
+
+def ignore_strict_variable_name_check():
```
IgnaceBleukx: I would also make this one private and let it start with an underscore.
Users really are not supposed to use any of this right? Only if they REALLY know what they are doing ; )

Normally, end-users just use the API features we provide (reading from pickle etc)

cpmpy/model.py
```diff
@@ -40,6 +41,33 @@
...
+    from cpmpy.expressions.variables import _BoolVarImpl, _IntVarImpl, _BV_PREFIX, _IV_PREFIX # avoid circular import
+            
+    vs = get_variables_model(model)
+    bv_counter = 0
+    iv_counter = 0
+    for v in vs:
+        if v.name.startswith(_BV_PREFIX):
+            try:
+                bv_counter = max(bv_counter, int(v.name[2:])+1)
+            except:
```
IgnaceBleukx: A comment here would be appropriate -- I suppose it is the case if we have `BVandthensomethingnonint` as name?

cpmpy/expressions/variables.py
```diff
@@ -790,5 +791,58 @@ def _genname(basename, idxs):
...
+    """
+    Context manager to temporarily disable strict variable name check.
+    """
+    class IgnoreStrictVariableNameCheck:
+        def __enter__(self):
+            _disable_strict_variable_name_check()
+        def __exit__(self, exc_type, exc_value, traceback):
+            _enable_strict_variable_name_check()
+            # _update_variable_counters() # TODO: add automatic support for this later (different PR)
```
IgnaceBleukx: So for that you would need access to the model here?

IgnaceBleukx (approved):
Very elegant use of the context manager : )
I would just make the function call also private, to really indicate this is non-trivial usage of the library.

cpmpy/expressions/variables.py
ThomSerg: Yes, with the current update function you would indeed. So currently the only approach to do it somewhat automatic is to create you model before entering the context, and then call:
```python
with ignore_strict_variable_name_check(model=model):
    ...
```
The context manager then has the model and can call the update for you on exit.

cpmpy/model.py
ThomSerg: I suppose, didn't write this code. Will add the comment.

tias (approved):
fine by me...

will also be of benefit in other file readers I understand?

If Ignace has no further comments then can go in...

maybe do add your example from the comment in the Docstring

ThomSerg:
Added it to the milestone, since it is needed for the upcoming IO (otherwise we can't read models back in that we've written ourselves).

tias:
Since my last comment was "If Ignace has no further comments then can go in..." I'll invite him

IgnaceBleukx:
Great, I pushed some minor changes and updated tests.

Two general comments:
1. Currently the context manager works as a toggle between strict and non-strict mode; while I would expect it to just enable it? What was the reasoning behind this? I would be fine if it is just syntactic sugar for enable() ... disable()
2. The error you get now is not entirely correct in non-strict mode; but I'm also not sure if we should add the bloat to deal with different exceptions depending on the mode... 

So first point is the real open question here, otherwise fine to merge

ThomSerg:
1. Yeah, no real reason, it's just what I thought its behavior should be when used nested. For example in the case of the dimacs parser, which in turn calls the wcnf parser when it detects weighted instances. But even in that parser's current implementation, nested `_ignore_strict_variable_name_check()` is not needed. So there is currently no real need for it being a toggle instead of just off->on.
2. Yes, that's a limitation. Only solution I see is a) returning a 2-tuple with the reason or b) let `_is_invalid_name()` do the exception raising (instead of returning False).

IgnaceBleukx:
Ok, then I would prefer it to indeed be a switch going from off -> on; I can imagine some horribly difficult to trace bugs when it would be used nested... Can you do the changes? Afterwards good to merge for me, no need to address the special exception message

ThomSerg:
I did the following:
- convert from relative toggle to on-off switch
- raise exception when used nested, since otherwise there is unexpected behavior (which the toggle prevented):
    ```python
    #strict
    with _ignore_strict_variable_name_check():
        # non-strict
        with _ignore_strict_variable_name_check():
            # non-strict
        # strict <- unexpected
    # strict
    ``` 
    So easiest to not allow nested use, enforcing this by raising an exception.

PR #840, opened 2026-01-30
Pytorch-compatible datasets

ThomSerg:
Cherry picked the dataset code for `benchmark_datasets` branch.

Proposes / improves upon the Pytorch-compatible dataset class from our work on xcsp3. 

Features:
- Interface is compatible with pytorch datasets, allowing you to use them with pytorch dataloaders (e.g. train/test split, ...). 
- Some syntactic sugar has been added on top to make it more suitable for CP / CPMpy. Most of the logic is reusable and has been placed in an abstract base class `_Dataset`.
- New datasets need to implement a very limited set of methods
    - \_\_init\_\_ (to allow for additional dataset specific arguments, like year, track, competition, ...)
    - category (benchmark-level metadata, i.e. year, track, ...)
    - download (how the dataset should be downloaded, helper functions are provided to ease implementation)
    - open (optional, how to open the files, e.g. for when files have been lzma/xz compressed

Some notes:
- split up the download with a helper `_download_sequential`. Some datasets provide a `range` html header, allowing you to significantly speed up the download by parallelising. `_download_parallel` is not included in this PR, but I have working code and we can always add it later
- metadata is currently still a bit "wild west". Most datasets currently only have dataset-level metadata provided by `def category(self)`. Some, like jsplib, also have instance level metadata. They add this metadata on access by overwriting `_getitem_` (extracting the metadata on the spot by reading the file, and injecting the result into the other metadata). I don't really like this approach, so I've been experimenting with an alternative where metadata gets extracted upon first download (only once) and gets stored in metadata sidecar files (i.e. `<instance_name>.meta.json`). Then on iterating over the dataset this metadata gets added (and you can also inspect it by looking at the files). The only noticeable change with this new system would be that when adding a new dataset, you can optionally overwrite a method to define how metadata extraction should happen. But to keep this PR simple, this is not included.
- similar to `open`, I've also been experimenting with another helper method `read` that would call an appropriate data format reader from our collection. But for this our readers first need to be merged. So you could call
    ```python
    dataset = XCSP3Dataset()
    for instance, metadata in dataset:
        model = dataset.read(instance)  # <- calls `read_xcsp3(instance, open=dataset.open)` internally
   ```
   This way you can much easier write code that transparently works for any cpmpy dataset.

TODO:
- [ ] examples have to be updated to make use of these new dataset classes
- [ ] add documentation

### DIFF
--- a/cpmpy/tools/dataset/__init__.py
+++ b/cpmpy/tools/dataset/__init__.py
@@ -1,0 +1,7 @@
+from .miplib import MIPLibDataset
+from .jsplib import JSPLibDataset
+from .psplib import PSPLibDataset
+from .nurserostering import NurseRosteringDataset
+from .xcsp3 import XCSP3Dataset
+from .opb import OPBDataset
+from .mse import MSEDataset
--- a/cpmpy/tools/dataset/_base.py
+++ b/cpmpy/tools/dataset/_base.py
@@ -169,7 +169,10 @@ class _Dataset(ABC):
     @staticmethod
     def _download_sequential(url: str, filepath: pathlib.Path, total_size: int, desc: str,
                              chunk_size: int = 1024 * 1024):
-        """Download file sequentially (fallback method)."""
+        """
+        Download file sequentially with progress bar.
+        """
+        
         import sys
         
         req = Request(url)
--- a/cpmpy/tools/dataset/_base.py
+++ b/cpmpy/tools/dataset/_base.py
@@ -1,0 +1,222 @@
+"""
+Dataset Base Class
+
+This module defines the abstract `_Dataset` class, which serves as the foundation
+for loading and managing benchmark instance collections in CPMpy-based experiments.  
+It standardizes how datasets are stored, accessed, and optionally transformed.
+"""
+
+from abc import ABC, abstractmethod
+import os
+import pathlib
+import io
+import tempfile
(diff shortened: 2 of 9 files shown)


tias:
Minor comments:
- tools.dataset in singular? I would have expected tools.datasets? (tools is also plural)
- same for _Dataset.category(), it returns a dict of all categories if I understand the doc right?

Perhaps docs could be slightly improved, e.g.
in _Dataset header: ink to pytoch dataset and say which 3 functions make it pytorch compatible, that the rest is added by us?
I think _Dataset.__getitem__ deserves a doc too, e.g. why it returns two things...

If I run tools/jsplib.py I get an No Such file or dir '' in _base 153 because it gets str(master.zip) whose parent is '' instead of '.' and so it breaks...

i would suggest to rename mse to maxsateval, mse is very cryptic (and mean square error to an ML head like me)

nurserostering has two if name == main's, one regular, one extensive

nurserostering and jsplib contain the code for making the CPMpy model but the others not, is that on purpose?

the part in your message in this PR about what is needed to do to add a new dataset would be good to add to the docs, e.g. of _Dataset


So, I think the design is good, idd a few more doc updates and then ready?
Having the examples use it can be done in a separate PR potentially too

ThomSerg:
Superseded by #900 and #1037.

PR #841, opened 2026-01-30
Datasets metadata

ThomSerg:
Follow-up of #840, standardising instance-level metadata extraction and storage as `.meta.json` sidecar files.
(still also includes the changes from #840)

Instead of extracting metadata on the spot (during iteration over the dataset), this PR proposes to perform this ahead of time on initial download (since depending on what metadata you want to add, this can take some time). New datasets only have to implement 
```python
def collect_instance_metadata(self, file: str) -> dict:
    ... do your one-time processing of the file here ...
```
For each instance, metadata gets stored in a separate metadata sidecar file `<instance_name_without_extensions>.meta.json`

For example, for jsplib' instance `abz5`, `abz5.meta.json`:
```json
{
  "name": "abz5",
  "jobs": 10,
  "machines": 10,
  "optimum": 1234,
  "bounds": {
    "upper": 1234,
    "lower": 1234
  }
}
```
This data was already getting collected before, from scratch on each access of the dataset. Now it is stored in a reusable metadata file.

### DIFF
--- a/cpmpy/tools/dataset/__init__.py
+++ b/cpmpy/tools/dataset/__init__.py
@@ -1,0 +1,7 @@
+from .miplib import MIPLibDataset
+from .jsplib import JSPLibDataset
+from .psplib import PSPLibDataset
+from .nurserostering import NurseRosteringDataset
+from .xcsp3 import XCSP3Dataset
+from .opb import OPBDataset
+from .mse import MSEDataset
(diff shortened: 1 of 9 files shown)


tias:
yes, can be good to make it at download time indeed; and following a standard is also fine if it is not too much extra work or code

ThomSerg:
Superseded by #900 and #1037.

Issue #511, opened 2024-08-01
Export model to mps/lp file


sin3000x:
I really like the feature that we can translate a CP model to a Gurobi one, and then I can call `grb_model.write()` to export the model to a standard mps format. As far as I know, the ortools CP-SAT interface does [NOT](https://groups.google.com/g/or-tools-discuss/c/v17bqI0GzHc/m/IB1imWT5AAAJ) support this.

Is it possible to export `cpmpy.Model` to a mps file directly even when Gurobi is not available? This will make it really convenient to use cpmpy with an unsupported solver, while avoiding writing a new interface.

tias:
That is a good point.

But indeed only Gurobi is a solver we currently connect to that can do that export.

We have a pull request/branch with SCIP, but it is not finished nor actively worked on, so no specific timeline from our side.

We are happy to take pull requests with new functionality : )

OrestisLomis:
compleed by #842

PR #842, opened 2026-01-30
IO module with readers and writers

ThomSerg:
Code cherry picked from `benchmark_datasets` branch.

Adds a new module under tools to centralise our readers and writers.

Also added the helper functions `read` and `write`

```python
from cpmpy.tools.io import read, write

model = read("file.mps", format="mps")
model = read("file.mps") # <- attempts to auto-determine format based on file extension

write(model, "file.mps", format="mps) # this is a special case where format will actually always be ignored, since internally SCIP is used and that one only does auto format detection. (probably needs an assert check then?)
write(model, "file.mps")  # <- attempts to auto-determine format based on file extension
```

But maybe we want to integrate this with the already existing `model.from_file` and `model.to_file`, just as an alternative format to `.pickle`

TODO
- [x] add context manager from #839 
- [x] add documentation
- [x] add tests


tias:
I find 'io' cryptic, better 'formats' or so?

I'm not sure about adding the tools to model.from_file and to_file... these two do not provide portable formats, the pickle thing is really just a way to dump the model object.

I see the opb.py file is very involved and closely intertwined with the transformations, should it be that 'far away' in 'io'?
we also have tools.dimacs and tranformations.to_cnf...

perhaps the transforming part (to cnf, to opb, to mip) should be in transformations, and the actual input/output handling of the formats in io/formats?

mypy.ini
```diff
@@ -64,6 +64,12 @@ ignore_missing_imports = True
 [mypy-pyscipopt.*]
 ignore_missing_imports = True
 
+[mypy-pandas.*]
+ignore_missing_imports = True
+
+[mypy-faker.*]
```
tias, line 72: add comment # used in tools/io/nurserostering.py
also for pandas: nurserostering and rcpsp

cpmpy/transformations/to_cnf.py
```diff
@@ -58,3 +66,56 @@ def to_cpmpy_clause(clause):
...
+        weights, xs = obj.args
+    else:
+        raise NotImplementedError(f"DIMACS: Non supported objective {obj} (yet?)")
+
+    terms, enc_cons, k = _encode_lin_expr(ivarmap, xs, weights, encoding, csemap=csemap)
+    const += k
+
+    extra_cons = safe_cons + decomp_cons + flat_cons + enc_cons
+
+    # remove terms with coefficient 0 (`only_positive_coefficients_` may return them and RC2 does not accept them)
```
tias, line 117: judging from this comment, you copy-pasted it from RC2; can we now switch it around and have RC2 use this function to get its objective?

cpmpy/expressions/variables.py
```diff
@@ -63,6 +63,7 @@
 import warnings # for deprecation warning
 from functools import reduce
 from typing import Any, Literal, Optional, overload
+from contextlib import contextmanager
```
tias: you don't seem to be using it, nor here nor in io.utils?

cpmpy/tools/io/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+from .jsplib import load_jsplib        
+from .nurserostering import load_nurserostering
+from .rcpsp import load_rcpsp
+
+# Standard format loaders and writers
+from .opb import load_opb, write_opb
+from .scip import load_scip, write_scip
+from .wcnf import load_wcnf
+from .dimacs import load_dimacs, write_dimacs
+from ..xcsp3 import read_xcsp3
```
tias: Ignace started doing this `__all__` in other inits, check with him if we need this here too?

cpmpy/tools/io/dimacs.py
```diff
@@ -0,0 +1,343 @@
...
+        p_header: bool = False, header : Optional[str] = None, 
+        open: Callable = builtins.open, 
+        annotate: Optional[Callable] = None
+    ) -> str:
+    """
+    Writes a CPMpy model to DIMACS format.
+    Uses the "to_cnf" transformation from CPMpy.
+
+    .. todo::
+        TODO: implement pseudoboolean constraints in to_cnf
```
tias: no longer a todo?

cpmpy/tools/io/jsplib.py
```diff
@@ -0,0 +1,161 @@
...
+    assert task_to_machines.shape == task_durations.shape
+
+    n_jobs, n_tasks = task_to_machines.shape
+
+    start = cp.intvar(0, task_durations.sum(), name="start", shape=(n_jobs,n_tasks)) # extremely bad upperbound... TODO
+    end = cp.intvar(0, task_durations.sum(), name="end", shape=(n_jobs,n_tasks)) # extremely bad upperbound... TODO
+    makespan = cp.intvar(0, task_durations.sum(), name="makespan") # extremely bad upperbound... TODO
+
+    model = cp.Model()
+    model += start + task_durations == end
```
tias: please use model.add(cons) instead of model += cons, thanks

cpmpy/tools/io/jsplib.py
```diff
@@ -0,0 +1,161 @@
...
+                              end[task_to_machines == machine])
+
+    model += end <= makespan
+    model.minimize(makespan)
+
+    return model, (start, makespan)
+
+
+
+def main():
```
tias: dimacs doesnt have this, many of the others do, is there a pattern?

tias: aren't/couldn't all these be superseded by a main function in loader.py?

cpmpy/tools/io/scip_formats.py
```diff
@@ -0,0 +1,375 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+##
+## scip.py
```
tias, line 4: its weird to me that this one is called 'scip', which is also a solver... and has a solver file (now, not when you wrote it I think).

maybe we should name it scip_formats or so?

cpmpy/tools/io/utils.py
```diff
@@ -0,0 +1,151 @@
+import warnings
+import os
+from contextlib import contextmanager
+from io import StringIO, TextIOBase
+from pathlib import Path
+from typing import Union, TextIO, Callable, Iterator, cast
+import builtins
+
+# mapping file extensions to appropriate format names
+_format_map = {
```
tias, line 10: I think the same already exists in loader...

I'm a bit lost if we need both 'loader' and 'utils'? I think they share a similar purpose

ThomSerg: There is a difference between mapping a format name to a loader function and mapping a format name to a file extension. `_extension_map`, mapping extensions to format names, is the inverse of `_format_map`. Originally I had xcsp3 in this list, the only example where the file extension is not identical to the format name (.xml), but had to remove that one since xml is not specific to xcsp3. I use `_format_map` and `_extension_map` for respectively getting the extension for a format when writing and auto-determining which format to use when loading based on the extension.

ThomSerg: It seems like I use `_format_map` (mapping format to extension) for scip temp files when calling load_scip on a string (scip only accepts paths and thus a temp file with correct extension must be created)

cpmpy/tools/io/annotate_bool.py
```diff
@@ -0,0 +1,206 @@
+from typing import Any, Callable
+
+from cpmpy.transformations.int2bool import IntVarEnc, IntVarEncDirect, IntVarEncOrder, IntVarEncLog
+
+
```
tias, line 11: an annotation in minizinc is used to annotate anything... we don't want to raise the expectation that we support something like that.

Maybe the file should be called more precisely annotate_bool.py or so? Also, can it get a simple module docstring?

cpmpy/tools/io/annotate.py
```diff
@@ -0,0 +1,206 @@
+from typing import Any, Callable
+
+from cpmpy.transformations.int2bool import IntVarEnc, IntVarEncDirect, IntVarEncOrder, IntVarEncLog
+
+
+AnnotationCallable = Callable[[list[Any], dict[str, IntVarEnc]], list[str]]
```
tias: should be named AnnotatorCallable, because it is for the annotator, not for an annotation?

But in general... its weird that we define this here, but the file does not define anything like it. Maybe you want/need a class here? A `Annotator` on which you can call annotate() -> list[str]?

cpmpy/tools/io/annotate.py
```diff
@@ -0,0 +1,206 @@
...
+"""Callable shape for writer annotation hooks: ``annotate(vars, ivarmap) -> list[str]``."""
+
+
+def _safe_name(v):
```
tias: first two are -> str? rest is nicely typed

cpmpy/tools/io/annotate_bool.py
tias: with a list of functions...

cpmpy/tools/io/annotate.py
```diff
@@ -0,0 +1,206 @@
...
+            # xs[i] is the bit at position i (value contribution: 2^i, offset by lb)
+            for i, bv in enumerate(enc._xs):
+                reverse[id(bv)] = {
+                    "source_name": orig_name,
+                    "encoding": "binary",
+                    "bit": i,
+                }
+    return reverse
+
+def annotate_cpmpy(vars: list[Any], ivarmap: dict[str, IntVarEnc]) -> list[str]:
```
tias: for readability of the file, I prefer that the annotate_ functions are the top functions in the file, and the _smth helpers are below it

cpmpy/tools/io/annotate.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    Return CPMpy-style variable annotations for Boolean encoding variables.
```
tias: why do we need our own style of annotation? we don't need it, we are a programming library... can we stick to the existing ones without introducing a new one?

cpmpy/tools/io/jsplib.py
```diff
@@ -0,0 +1,117 @@
...
+    assert task_to_machines.shape == task_durations.shape
+
+    n_jobs, n_tasks = task_to_machines.shape
+
+    start = cp.intvar(0, task_durations.sum(), name="start", shape=(n_jobs,n_tasks)) # extremely bad upperbound... TODO
+    end = cp.intvar(0, task_durations.sum(), name="end", shape=(n_jobs,n_tasks)) # extremely bad upperbound... TODO
+    makespan = cp.intvar(0, task_durations.sum(), name="makespan") # extremely bad upperbound... TODO
+
+    model = cp.Model()
+    model.add(start + task_durations == end)
```
tias: afaik the NoOverlap will take care of this?

cpmpy/tools/io/nurserostering.py
```diff
@@ -0,0 +1,433 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+##
+## nurserostering.py
+##
+"""
+Helper functions for the Nurse Rostering format.
+
+The Nurse Rostering format is a textual format to represent Nurse Rostering problems.
```
tias, line 9: is there some URL you can point to for more info? like you did for DIMACS

cpmpy/tools/io/opb.py
```diff
@@ -0,0 +1,392 @@
...
+            raise ValueError(f"Header declares {nr_vars_declared} variables but found {len(vars)} unique variables while parsing")
+        
+        if len(model.constraints) != nr_constraints_declared:
+            raise ValueError(f"Header declares {nr_constraints_declared} constraints but found {len(model.constraints)} constraints while parsing")
+
+        return model
+
+def annotate_extended(vars, ivarmap: dict[str, Any]) -> list[str]:
+    """
+    Annotate CPMpy variables as OPB ``x<int>`` identifiers.
```
tias: this reads identical to annotate_x.... and annotate_x says: only accept smth, use annotate_x, but what is the difference with this one?

cpmpy/tools/io/rcpsp.py
```diff
@@ -0,0 +1,173 @@
...
+    durations = [jobs[j]["duration"] for j in job_order]
+    horizon = sum(durations)  # worst case, all jobs sequential on a machine
+    makespan = cp.intvar(0, horizon, name="makespan")
+
+    start = cp.intvar(0, horizon, name="start", shape=len(job_order))
+    end = cp.intvar(0, horizon, name="end", shape=len(job_order))
+
+    # ensure capacity is not exceeded
+    for resource in resource_names:
+        model += cp.Cumulative(
```
tias: some more .add()'s here... and here you correctly dont enfore start+dur=end

cpmpy/tools/io/utils.py
tias: okay; but now they are all mapping to self... so you could just have a list? and if its scip specific, maybe it should be in the scip file?

no strong feelings, just... a bit unexpected

cpmpy/tools/xcsp3/parser.py
```diff
@@ -0,0 +1,150 @@
...
+    # Print results
+    print("Status:", model.status())
+    if result is not None:
+        if model.has_objective():
+            print("Objective:", model.objective_value())
+    else:
+        print("No solution found.")
+
+# Backward compatibility alias
+read_xcsp3 = load_xcsp3
```
tias: do we need this? the function is marked deprecated but still exists; this one is subtle and weird... we will probabaly forget it

docs/api/tools/io.rst
```diff
@@ -0,0 +1,77 @@
...
+     - ✓
+     -
+
+
+Format-specific loaders and writers:
+
+.. toctree::
+    :maxdepth: 1
+
+    io/opb
```
tias, line 73: its easier to see that you did not forget any when you order them alphabetically

docs/modeling.md
```diff
@@ -622,6 +622,47 @@ For solvers other than "ortools", you will need to **install additional package(
     ModuleNotFoundError: CPM_gurobi: Install the python package 'cpmpy[gurobi]' to use this solver interface.
 ```
 
+## Input and output
```
tias, line 625: that's a great addition

docs/modeling.md
```diff
@@ -622,6 +622,47 @@ For solvers other than "ortools", you will need to **install additional package(
...
+
+The generic `load()` and `write()` helpers choose the right parser or writer based on the file extension when possible, or you can pass the format explicitly.
+
+```python
+import cpmpy as cp
+from cpmpy.tools.io import load, write
+
+# Load an instance from a file
+model = load("instance.opb")
+jobshop_model = load("ft06.jsp", format="jsplib")
```
tias: if you demo this, loading two models at the same time, then you are encouraging/likely to have unintned shared variables here?

tests/test_tools_io_scip_formats.py
```diff
@@ -0,0 +1,44 @@
+from collections import namedtuple
+from types import SimpleNamespace
+
+import cpmpy as cp
+
+from cpmpy.tools.io.scip_formats import _load_scip_objective
```
tias, line 6: file should be test_tools_io_scip?   scip is not a tool, its a solver, so some name standardisation could be nice...

the exception will then be test_tool_dimacs, which could stay, or could be renamed to test_tools_io_dimacs.py

tests/test_to_opb.py
```diff
@@ -0,0 +1,113 @@
+"""
```
tias, line 1: test_to_opb as filename? the toopb reads weird to me...

tias: we also have some test_transf_smth....

tias (commented):
Surprisingly substantial and complete.

Some minor comments, or naming suggestions, also open questions

cpmpy/tools/io/opb.py
ThomSerg: I converted it to become internal, no longer user accessible.

cpmpy/tools/xcsp3/parser.py
ThomSerg: indeed, forgot to remove

docs/modeling.md
ThomSerg: thanks!

tests/test_tools_io_scip_formats.py
ThomSerg: about the dimacs part, the old reader and writer (tools.dimacs) were apparently still there. Removed the duplicate code and added deprecation warnings.

cpmpy/tools/io/annotate.py
ThomSerg: Done, but bigger change so that parts needs reviewing again.

cpmpy/expressions/variables.py
```diff
@@ -846,22 +845,20 @@ def _is_invalid_name(name: Any) -> bool:
...
     elif name.startswith(_BV_PREFIX):
         if _get_strict_variable_name_check():
             return True
         else:
             id = int(name[len(_BV_PREFIX):])
             if _BoolVarImpl.counter > id:
-                return True
-            else:
-                return False
+                # TODO: reduce warning frequency? Now reported for every variable
```
tias: if you remove {name} and rewrite it a bit in a generic way, then python will just track all the identical warnings and e.g. report 'this warning, x times...'

cpmpy/tools/io/annotate_bool.py
```diff
@@ -0,0 +1,232 @@
+"""
+Annotate Boolean encoding variables.
+
+Some target formats don't support integer variables and thus require them to be encoded into a set of 
+Boolean variables. For example, a direct (or one-hot) encoding of an integer variable `x` in the range 
+`[0, 10]` requires 11 Boolean variables, one for each possible value of `x`. 
+
+When writing to file, this mapping between the integer variable and the Boolean variables is typically lost. 
+Through annotation, we can keep track of this mapping and later recover it.
```
tias: If we also save an annotation to the file, a mapping from Boolean variable name to integer variable name, then an external tool can recover it later

cpmpy/tools/io/annotate_bool.py
```diff
@@ -0,0 +1,232 @@
...
+
+Some target formats don't support integer variables and thus require them to be encoded into a set of 
+Boolean variables. For example, a direct (or one-hot) encoding of an integer variable `x` in the range 
+`[0, 10]` requires 11 Boolean variables, one for each possible value of `x`. 
+
+When writing to file, this mapping between the integer variable and the Boolean variables is typically lost. 
+Through annotation, we can keep track of this mapping and later recover it. 
+
+Additionally, some formats don't support arbitrary variable naming. In DIMACS for example, 
+variables are referenced by integer IDs. Annotation again allows us to keep track of the original
```
tias: Storing an annotation in the file allows again to map it back later if needed

cpmpy/tools/io/annotate_bool.py
```diff
@@ -0,0 +1,232 @@
...
+"""
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+from cpmpy.transformations.int2bool import IntVarEnc, IntVarEncDirect, IntVarEncOrder, IntVarEncLog
+
+
+class BooleanEncodingAnnotator(ABC):
+
```
tias: I don't think we typically have an empty line? the docstring immediatley follows the class decl?

cpmpy/tools/io/annotate_bool.py
```diff
@@ -0,0 +1,232 @@
...
+class SugarAnnotator(BooleanEncodingAnnotator):
+
+    """
+    Return Sugar-style annotations for Boolean encoding variables.
+
+    Sugar-style convention:
+      order:  p<var>,<a>      meaning <var> <= a
+      direct: p<var>=<a>      meaning <var> = a
+      binary: p<var>#<i>      meaning bit i of <var>
```
tias, line 57: this would not allow for non-power-of-2 offsets?

tias: ah, the example below clarifies that the offset is ignored and has to be recomputed from the integer model I guess

cpmpy/tools/io/annotate_bool.py
```diff
@@ -0,0 +1,232 @@
...
+            # xs[i] is the bit at position i (value contribution: 2^i, offset by lb)
+            for i, bv in enumerate(enc._xs):
+                reverse[id(bv)] = {
+                    "source_name": orig_name,
+                    "encoding": "binary",
+                    "bit": i,
+                }
+    return reverse
+
+
```
tias, line 231: I think this is now very clean

cpmpy/tools/io/scip_formats.py
```diff
@@ -0,0 +1,363 @@
...
+    """
+    def __init__(self, model: cp.Model, problem_name: Optional[str] = None):
+        if not self.supported():
+            raise Exception(
+                "SCIP: Install SCIP IO dependencies: cpmpy[io.scip]")
+
+        super().__init__(model)
+        self.scip_model.setProbName(problem_name)
+
+def _add_header(path: Union[str, os.PathLike], format: str, header: Optional[str] = None):
```
tias: this is not part of _SCIPWriter? I find it a very generic name for a free-floating function...

cpmpy/tools/io/scip_formats.py
```diff
@@ -0,0 +1,363 @@
...
+
+    else:
+        warnings.warn(f"Unsupported format for header: {format}")
+        return
+
+    with open(path, "w") as f:
+        f.writelines(lines)
+
+
+def write_scip(
```
tias: write_scip_format?

tests/test_tools_io.py
```diff
@@ -0,0 +1,771 @@
+"""
```
tias, line 1: that's crazy extensive, I hope the testcode never breaks : )

tests/test_tools_io_scip_formats.py
```diff
@@ -0,0 +1,44 @@
+from collections import namedtuple
```
tias, line 1: rename file test_tools_io_scip_formats?

tias (approved):
Great, really substantial improvement to the library. In fact there is a (2 year old?) issue asking for an MPS writer somehwere : D #511 

Some suggestions for doc improvements and some comments int he review, all very minor.

Good to go in! (I will not be able to re-review anyway, so do proceed).

Ideally if somewhere afterwards you could close all the file format PRs we have : D

cpmpy/tools/io/annotate_bool.py
ThomSerg: indeed, sugar annotation ignores the offset

Issue #837, opened 2026-01-26
On flattening, linearization and cse


IgnaceBleukx:
So we have been discussing offline a bit on how we can improve our flattening and linearization, and it seems that centrally, it all boils down to improving our handling of common subexpressions.

As we will need quite some functionality centered around CSE, it probably makes sense to make it a custom object that mimics the current implementation with dictionaries.

The purpose of this issue is to dicuss on what and how exactly should be implemented in this map, with some initial ideas.

### Making a new auxiliary variable
When we do flattening, we need to make new aux variables for subexpressions.
Currently, this is done through `def get_or_make_var()` in `flatten_model.py`; but it probably makes sense to implement this functionality directly into the csemap object.
This has the advantage that we can easility give it the Boolean context as well.

To get this somewhat streamlined, we probably want 3 seperate dictionaries: one for positive context, one for negative and one for integer expressions.

I've tried to come up with a prototype implementation in this snippet:

```python

class CSEMap:

    def get_or_make_var(self, expr, context="mixed"):

        if not expr.is_bool(): # integer expression
            if expr not in self._int_map:
                self._int_map[expr] = intvar(*get_bounds(expr))
            return self._int_map[expr]
        
        # Boolean expressions
        if (context == "positive" or context == "mixed") and expr not in self._pos_map:
            new_var = self._neg_map.get(expr, boolvar())
            self._pos_map[expr] = new_var
        
        if (context == "negative" or context == "mixed") and expr not in self._neg_map:
            new_var = self._pos_var.get(expr, boolvar())
            self._neg_map[expr] = new_var

        # return the Boolean variable
        if context == "positive" or context == "mixed":
            return self._pos_map[expr]
        else:
            return self._neg_map[expr]
```

### Separate defining constraints-generating function
Different from the currrent `def get_or_make_var` in flattening, the above does not actually return the constraint `var == expr` next to the newly generated variable.
I think it would make sense (mainly because of my next point), to have a seperate function that materializes these defining consrtaints.
```python

class CSEMap:

    def get_defining_constraints(self):

        cons = []
        # materialize positive context ones
        for expr, var in self._pos_map.items():
            if (expr, var, "posittive") not in self._materialized:
                cons.append(var.implies(expr))
                self._materialized.add((expr, var, "positive"))

        for expr, var in self._neg_map.items():
            if (expr, var, "negative") not in self._materialized:
                cons.append((~var).implies(~expr))
                self._materialized.add((expr, var, "negative"))
        
[...]

PR #843, opened 2026-02-01
Handle reified != predicates in custom CSEMap

IgnaceBleukx:
Refactor of the `get_or_make_var` function in flatten_constraint to use a custom CSEMap object.
Should be identical API-wise compared to what we currently have, but opens the door for many cse-specific improvements we discussed offline and in #837.

### DIFF
--- a/tests/test_trans_linearize.py
+++ b/tests/test_trans_linearize.py
@@ -645,7 +645,6 @@ class TestLinearizeReifiedVariablesThreshold:
         a = self.a
         out = []
         for con in [(a == 1) | (a != 2), (a == 3) | (a == 2)]:
-            cpm_cons = toplevel_list(con)
             cpm_cons = flatten_constraint(con, csemap=self.csemap)
             out += linearize_reified_variables(cpm_cons, min_values=2, csemap=self.csemap, ivarmap=ivarmap)
 


tias:
That is nice and clean, less impactful on code structure then I thought and the type hints is a good touch.

For the design of the object, I don't see a benefit in splitting up into an `_int_map` and `_bool_map`.

This PR is a pre for supporting cse on Boolean expressions (BE) that are reified in positive context; so that 'BE' needs not be replaced by 'b' with 'b == BE' but by 'b' with 'b -> BE'.

So I think that the CSE object in the future will have two maps, an `_eq_map` and an `_impl_map`.

- Any expression that exists in _eq_map can be replaced by its var, even if it is a BE in positive context.
- `_impl_map` can only contain Boolean expressions

So looking ahead, we will not need a separate bool/int map for equalities, just a single `_eq_map` that is always checked first...?

Looking further ahead, I guess we would add to get/set/get_or_make an argument like 'positive' or 'implied' which would (except for 'set') still first check the eqmap, and only then use implmap?

IgnaceBleukx:
So for the intmap and boolmap: when drafting out the future of CSE in #837, I opted for 3 maps actually: intmap, posmap and negmap.
These three maps cover intexpr == iv, bv => expr and bv <= expr respectively.

In my mind, integer equality is really different from Boolean full reification, so that's also why I split them here.

For the future API, indeed such an extra argument would be the way to go. Do you want to add this in this PR already? (But keep it unused)

Issue #844, opened 2026-02-01
Add support for OR-Tools v9.15 and Python 3.14


lsorber:
[OR-Tools v9.15](https://github.com/google/or-tools/releases/tag/v9.15) adds support for Python 3.14.

Would be nice if CPMpy could add support for both OR-Tools v9.15 and Python 3.14!

CPMpy currently has an upper bound on OR-Tools ([which should be avoided when possible](https://iscinumpy.dev/post/bound-version-constraints/)):
https://github.com/CPMpy/cpmpy/blob/6e74bb7bed666d50c045ec8a8243b917d039fece/requirements.txt#L2

PR #848, opened 2026-02-04
Remove version upper limits

ThomSerg:
Removes hard dependency version upper limits, as discussed in #844. Added a bit of documentation, hinting towards the use of `uv` when an exactly reproducible install is required.


tias:
Thanks a lot Thomas, great follow-up!

PR #850, opened 2026-02-07
Multiplication as globalfunction, with decomposition

tias:
Closes the gap of not supporting multiplication for linear/pb/sat solvers.

Linearising a multiplication is likely not going to be very efficient, but at least it is possible now...

Tests for all solvers (that I have installed) pass except pindakaas:
```
FAILED tests/test_constraints.py::test_reify_imply_constraints[pindakaas-((3 * (x)) != (boolval(True))) -> (p)] - pyo3_runtime.PanicException: assertion failed: x.size() > 0
FAILED tests/test_constraints.py::test_reify_imply_constraints[pindakaas-((boolval(True)) != (3 * (x))) -> (p)] - pyo3_runtime.PanicException: assertion failed: x.size() > 0
FAILED tests/test_constraints.py::test_reify_imply_constraints[pindakaas-(3 * (x) != 1) -> (p)] - pyo3_runtime.PanicException: assertion failed: x.size() > 0
FAILED tests/test_constraints.py::test_reify_imply_constraints[pindakaas-(1 != 3 * (x)) -> (p)] - pyo3_runtime.PanicException: assertion failed: x.size() > 0
```
@hbierlee this looks to be on your side?
It seems to be similar to what we found in the other linearize branch, where you seem to see that the reified part is not domain consistent, but that shouldn't crash this hard...

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -743,7 +743,7 @@ def _wsum_should(arg):
      all substractions are transformed into less readable wsums)
     """
     name = getattr(arg, 'name', None)
-    return name == 'wsum' or (name == 'mul' and arg.is_lhs_num is False)
+    return name == 'wsum' or (name == 'mul' and arg.is_lhs_num)
 
 def _wsum_make(arg):
     """ Internal helper: prep the arg for wsum
@@ -757,7 +757,7 @@ def _wsum_make(arg):
     elif name == 'sum':
         return [1]*len(arg.args), arg.args
     elif name == 'mul':
-        if arg.is_lhs_num is False:
+        if arg.is_lhs_num:
             return [arg.args[0]], [arg.args[1]]
     elif name == '-':
         return [-1], [arg.args[0]]


cpmpy/expressions/core.py
```diff
@@ -582,10 +573,21 @@ class Operator(Expression):
         'sum': (0, False),
         'wsum': (2, False),
         'sub': (2, False), # x - y
-        'mul': (2, False),
         '-':   (1, False), # -x
     }
-    printmap = {'sum': '+', 'sub': '-', 'mul': '*'}
+    printmap = {'sum': '+', 'sub': '-'}
+
+    def __new__(cls, name=None, arg_list=None):
```
IgnaceBleukx: What is the advantage of doing this here instead of in `__init__`?

cpmpy/expressions/globalfunctions.py
```diff
@@ -305,6 +306,121 @@ def get_bounds(self) -> tuple[int, int]:
...
+        """
+        is_lhs_num = False
+        if is_num(x):
+            is_lhs_num = True
+        elif is_num(y):
+            (x, y) = (y, x)
+            is_lhs_num = True
+
+        super().__init__("mul", (x, y))
+        self.is_lhs_num = is_lhs_num
```
IgnaceBleukx, line 334: This should be resetted when we call `update_args` on the expression

cpmpy/expressions/globalfunctions.py
```diff
@@ -305,6 +306,121 @@ def get_bounds(self) -> tuple[int, int]:
...
+
+    def __repr__(self):
+        x, y = self.args
+
+        def wrap_bracket(arg):
```
IgnaceBleukx: re-defining this function will have quite some overhead everytime we call `__repr__` no? 
I would just put it in an if-then-else in the format below

cpmpy/expressions/globalfunctions.py
```diff
@@ -305,6 +306,121 @@ def get_bounds(self) -> tuple[int, int]:
...
+
+        # encoding a == sum(i*bv_i) with exactly one bv_i
+        domain = list(range(lb_a, ub_a + 1))
+        bvs = cp.boolvar(shape=(len(domain),))
+        encoding = [cp.sum(bvs) == 1, a == Operator("wsum", (domain, bvs))]
+
+        # encoding z_i == bv_i*b via (bv_i)->(z_i == b) & (~bv_i)->(z_i == 0)
+        z_lb = min(0, lb_b) # make sure it can take 0
+        z_ub = max(0, ub_b) # make sure it can take 0
+        zs = cp.intvar(z_lb, z_ub, shape=(len(domain),))
```
IgnaceBleukx: So I have some doubts about this encoding for pure Boolean solvers as it introduced many integer variables; which will in turn have to be encoded themselves.
I'm not sure how to code it in, but I think we should have a custom decomposition for pure Boolean solvers:
```python
sum(x * y * ((a == x) & (b == y)) for x in range(a.lb,a.ub+1) for y in range(b.lb, b.ub+1))
```
We could do some more fancy stuff even, as multiple x * y combinations will map to the same value anyway; but let's keep that for another PR.

cpmpy/expressions/globalfunctions.py
```diff
@@ -305,6 +306,121 @@ def get_bounds(self) -> tuple[int, int]:
...
+
+        # let a be the one with the smallest domain, leading to the fewest auxiliariy variables
+        lb_a, ub_a = get_bounds(a)
+        lb_b, ub_b = get_bounds(b)
+        if ub_b - lb_b + 1 < ub_a - lb_a + 1:
+            a, b = b, a
+            lb_a, lb_b = lb_b, lb_a
+            ub_a, ub_b = ub_b, ub_a
+
+        # encoding a == sum(i*bv_i) with exactly one bv_i
```
IgnaceBleukx: Why do the encoding here manually instead of using the `var == val` as in #836?

tests/test_constraints.py
```diff
@@ -77,20 +71,17 @@ def numexprs(solver):
...
-                yield Operator(name, [3, BOOL_ARGS[0]])
-
-        elif name == "mul" and "mul-bool" not in EXCLUDE_OPERATORS.get(solver, {}):
-            yield Operator(name, BOOL_ARGS[:arity])
         elif arity != 0:
             yield Operator(name, NUM_ARGS[:arity])
         else:
             yield Operator(name, NUM_ARGS)
 
+    yield cp.Multiplication(3, NUM_ARGS[0])
```
IgnaceBleukx: These should go int he `globalfunction` generator

cpmpy/expressions/core.py
tias: you can return a different class from your own (which is weird from a typing perspective but OK). This is the pre-constructor of Operator that returns a Multiplication.

cpmpy/expressions/globalfunctions.py
tias: great catch

cpmpy/expressions/globalfunctions.py
tias: yes

cpmpy/expressions/globalfunctions.py
tias: I agree that pure Boolean solvers could benefit from a different decomposition... which will be possible with your decompose_custom : )

and yes, for a different PR

cpmpy/expressions/globalfunctions.py
tias: oh, idd, I was too much in the low-level linears

tias:
I don't get the UX, and I think its too convoluted.

I want to give this to agents, and it should swamp their tokens. Just an update per file that is finished is already more than enough.

Also, the defaults dont seem to be read anymore? it fires up with maximum threads and all instances per problem

tests/test_constraints.py
tias: Could you do that part?

tias:
pysat does not seem to free its memory, I think its memouts by pysat;

 I have a patch that tells it to delete the solver objct, but the memory usage keeps increasing monotonically (will open a PR).

tias:
the remaining issue in pysat/pb.py, its pblib that is keeping the memory and not being asked to delete it... (now flight really departs)

if master does not fail you can keep it there... it was supposed to be a patch to be merged anyway

PR #851, opened 2026-02-09
Add `to_gcnf` transformation and GDIMACS reader/writer

hbierlee:
With @OrestisLomis, a GDIMACS reader/writer supported by a `to_gcnf` function. The writer optionally follows the specification (https://satisfiability.org/competition/2011/rules.pdf) that groups of clauses have to be disjoint. Tested using roundtrips for most cases (i.e. `write(read(gdimacs)) == gdimacs`), except for missing groups, which could be fixed still. Perhaps interesting for @ThomSerg.

The `to_gcnf` is regarded the same as the `make_assump_model`, but also returns the hard clauses separately. This does make it a lot easier for some cases, but now it's a bit inconsistent, so I'm not sure what to do with that yet.

Also includes a fix for the solution generator in to_cnf tests.


tias:
Lets wait with this till #840 .. #842 is in

ThomSerg:
Updated the PR to be more inline with the current IO module in master. Not too sure if the IO tool should be called `gdimacs` as is currently the case or `gcnf` (like the `wcnf` io tool)

OrestisLomis:
I think GCNF is fine, tho I will have to check for a more efficient version of this which I had implemented. Not sure if this PR has the correct version.

hbierlee:
I think this was based on the more efficient version you had. I believe the only difference will be that it does the `disjoint` step separately, which I think makes for nicer code still (and shouldn't be less efficient complexity-wise)

ThomSerg:
Also it doesn't 100% integrate with the rest of the IO due to the write function having a different signature than the rest. The rest has `write(model, ...)` whilst this one requires `write(soft, hard, ...)`

OrestisLomis:
@ThomSerg You already had a requested review, so I couldn't actually request one. I did update some of the code though, so please have another look when you can!

ThomSerg (commented):
Looks good. just not too sure about the user experience with the soft/hard split. What if they have a model already, which they then would like to convert to gcnf, but some of the constraints already in the model should be hard. This is currently not possible? The hard constraints should be kept out of the model?

OrestisLomis:
Yes, a user can keep hard constraints out of the model (the model is just considered the soft constraints). I think we can make this a bit more clear in the docs.

Edit: I actually think the docs are already clear about this tbh.

PR #852, opened 2026-02-10
Upgrade pindakaas 0.4.0

hbierlee:
Changes for the upcoming release of pindakaas, blocked by one additional issue uncovered by the CPMpy tests (https://github.com/pindakaashq/pindakaas/pull/186)


tias (approved):
we should increase the requirements/setup.py to 0.4.0 then?

PR #853, opened 2026-02-10
time_transformations script, following pysat transfs

tias:
(also an unncessary transformation in pysat)


tias:
I don't get the UX, and I think its too convoluted.

I want to give this to agents, and it should swamp their tokens. Just an update per file that is finished is already more than enough.

Also, the defaults dont seem to be read anymore? it fires up with maximum threads and all instances per problem

tias:
thx! very clean this way

PR #854, opened 2026-02-10
Update Pumpkin interface to version 0.3.0

IgnaceBleukx:
Update interface to new Pumpkin version.

API changes:
- Proof argument should be given at initialization time (similar to exact)

Fixes and improvements:
- Post safening constraints when using partial function as objective
- Check if solver found UNSAT with just root-level propagation (before solve is called)
- Disable Table constraints and friends when prooflogging
- Use `bool.as_integer` instead of helper function `solver.boolean_as_integer`
- Removed tag argument where possible.

New version is not on pip yet, but can be tested and used locally already by compiling from source

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ solver_dependencies = {
     "gcs": ["gcspy>=0.1.9"], # first version to pass all tests
     "cpo": ["docplex>=2.28.240"],
     "pumpkin": ["pumpkin-solver>=0.3.0"], # CPMpy requires features only available from Pumpkin version >=0.3.0
-    "pindakaas": ["pindakaas>=0.2.1"],
+    "pindakaas": ["pindakaas==0.2.1"],
     "cplex": ["docplex>=2.28.240", "cplex>=20.1.0.4"],
 }
 solver_dependencies["all"] = list({pkg for group in solver_dependencies.values() for pkg in group}) 


cpmpy/solvers/pumpkin.py
```diff
@@ -81,8 +81,8 @@ def supported():
         try:
             import pumpkin_solver as psp
             pum_version = CPM_pumpkin.version()
-            if Version(pum_version) < Version("0.2.2"):
-                warnings.warn(f"CPMpy uses features only available from Pumpkin version >=0.2.2 "
+            if Version(pum_version) < Version("0.3.0"):
+                warnings.warn(f"CPMpy uses features only available from Pumpkin version >=0.3.0 "
```
hbierlee, line 85: why shouldn't we just rely on the lower bound in setup.py?

cpmpy/solvers/pumpkin.py
```diff
@@ -103,13 +103,15 @@ def version() -> Optional[str]:
...
-    def __init__(self, cpm_model=None, subsolver=None):
+    def __init__(self, cpm_model=None, subsolver=None, proof=None, seed=None):
         """
         Constructor of the native solver object
 
         Arguments:
             cpm_model: Model(), a CPMpy Model() (optional)
             subsolver: None, not used
+            proof (str, optional): path to the proof file
+            seed (int, optional): ranomd seed for the solver
```
hbierlee: typo

cpmpy/solvers/pumpkin.py
```diff
@@ -136,8 +152,18 @@ def native_model(self):
         """
         return self.pum_solver
 
+    def _unsat_at_rootlevel(self):
+        self.cpm_status = SolverStatus(self.name)
+        self.cpm_status.exitstatus = ExitStatus.UNSATISFIABLE
+        self.cpm_status.runtime = 0 # TODO: use post-time instead?
```
hbierlee, line 158: I think 0 makes sense, if this is essentially determined before we call solve (e.g. when constraints are posted in initialization)

cpmpy/solvers/pumpkin.py
```diff
@@ -156,31 +182,39 @@ def solve(self, time_limit:Optional[float]=None, prove=False, proof_name="proof.
         from pumpkin_solver import SatisfactionResult, SatisfactionUnderAssumptionsResult
         from pumpkin_solver.optimisation import OptimisationResult, Direction
 
+        if self.pum_solver.is_inconsistent():
+            return self._unsat_at_rootlevel()
```
hbierlee, line 190: not sure if this has to be a separate function, it's not re-used.

cpmpy/solvers/pumpkin.py
```diff
@@ -156,31 +182,39 @@ def solve(self, time_limit:Optional[float]=None, prove=False, proof_name="proof.
...
+
         # ensure all vars are known to solver
         self.solver_vars(list(self.user_vars))
 
+        if "proof" in kwargs or "prove" in kwargs or "prove_location" in kwargs or "proof_name" in kwargs:
+            raise ValueError("Proof-file should be supplied in the constructor, not as a keyword argument to solve."
+                             "`cpmpy.SolverLookup.get('pumpkin', model, proof='path/to/proof.drcp')`")
```
hbierlee: it might be nice to have this explicit error, but we'll also just get an exception from pumpkin. Since the user is basically passing pumpkin-kwargs, I'd be fine to get pumpkin-errors

cpmpy/solvers/pumpkin.py
```diff
@@ -611,20 +631,29 @@ def add(self, cpm_expr_orig):
...
+                if isinstance(cpm_expr, Operator) and cpm_expr.name == "->": # found implication
+                    bv, subexpr = cpm_expr.args
+                    for pum_cons in self._get_constraint(subexpr):
+                        self.pum_solver.add_implication(pum_cons, self.solver_var(bv))
+                else:
+                    for pum_cons in self._get_constraint(cpm_expr):
+                        self.pum_solver.add_constraint(pum_cons)
+            return self
+
+        except RuntimeError as e:
```
hbierlee, line 651: do you still get this with your guard at the start of the function? Probably yes. If so, can you just check is_inconsistent rather than rely on string matching?

hbierlee (commented):
looks good, minor comments/questions

cpmpy/solvers/pumpkin.py
IgnaceBleukx: Hmm good point, we have it in all solvers like this currently

cpmpy/solvers/pumpkin.py
IgnaceBleukx: Allright, yes I agree

cpmpy/solvers/pumpkin.py
IgnaceBleukx: No true, I just wanted to keep the solve function itself not to bloated with this edge-case stuff

cpmpy/solvers/pumpkin.py
IgnaceBleukx: Yes, I added the explicit error as this is an API change from what was implemented before... 
So if anyone was already using this, it might be a bit of surprise finding it not working anymore

IgnaceBleukx: Happy to remove it after the next release or so, but at first I would definitely keep this verbose error, explaining how to fix it

PR #855, opened 2026-02-11
Linearize reif vars

tias:
spin-off from #836 

the new decomp had to be completely rewritten... now reads from CSEMap to be sure that the bv is intended to define the var==val.

also, it looks like to_cnf did not use cse currently : /

### DIFF
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -58,7 +58,7 @@ from ..expressions.variables import _BoolVarImpl, NegBoolView, _IntVarImpl, _Num
 from ..transformations.comparison import only_numexpr_equality
 from ..transformations.flatten_model import flatten_constraint, flatten_objective
 from ..transformations.get_variables import get_variables
-from ..transformations.linearize import linearize_constraint, only_positive_bv, only_positive_bv_wsum, decompose_linear, decompose_linear_objective
+from ..transformations.linearize import linearize_constraint, only_positive_bv, only_positive_bv_wsum, decompose_linear, decompose_linear_objective, linearize_constraint, linearize_reified_variables
 from ..transformations.reification import only_implies, reify_rewrite, only_bv_reifies
 from ..transformations.normalize import toplevel_list
 from ..transformations.safening import no_partial_functions, safen_objective


tias:
Looks like I can make it show only the diff wrt #836, nice

cpmpy/transformations/linearize.py
```diff
@@ -621,3 +622,84 @@ def get_linear_decompositions():
...
+    # Collect bv -> (var == val)'s in csemap
+    var_vals = {}  # var: [val, bv]
+    for expr, bv in csemap.items():
+        if expr.name == '==':
+            var,val = expr.args
+            if isinstance(var, _NumVarImpl) and is_int(val):
+                var_vals.setdefault(var, []).append((val, bv))
+    
+    # Make the integer encodings in integer linear friendly way
+    my_ivarmap = ivarmap if ivarmap is not None else {}
```
IgnaceBleukx, line 650: Hmm so this will make duplicated encodings if we don't post all constraints at once... E.g.,
```python
solver = cp.SolverLookup.get("exact")
solver += (x == 2) + (y == 3) <= 1
solver += (x == 5) + (y == 4) <= 1
```
in the second call, we will make a new `my_ivarmap` and thus x and y will be re-encoded.
If we're just solving a model, it should be fine as we are posting all constraints at once then

cpmpy/transformations/linearize.py
```diff
@@ -621,3 +622,84 @@ def get_linear_decompositions():
...
+    If ivarmap is not None, the encoding is added to ivarmap and only sum(bvs)==1
+    (the domain constraint) is posted; the solver can then choose to eliminate the
+    vars, or post the wsums itself anyway.
+
+    Apply AFTER flatten_constraint and BEFORE only_implies and linearize_constraint.
+    """
+    # Collect bv -> (var == val)'s in csemap
+    var_vals = {}  # var: [val, bv]
+    for expr, bv in csemap.items():
+        if expr.name == '==':
```
IgnaceBleukx, line 644: Should we also do it for `!=`?

cpmpy/transformations/linearize.py
```diff
@@ -621,3 +622,84 @@ def get_linear_decompositions():
...
+        # store the bvs that no longer need to be reified
+        for val, bv in vals:
+            bv_map[bv] = (var, val)
+
+    if len(bv_map) > 0:
+        # Now clean up and remove the '(var == val) == bv' constraints:
+        newcons = []
+        for con in constraints:
+            if con.name == '==' and con.args[0].name == '==':
+                # potential '(var == val) == bv'
```
IgnaceBleukx, line 685: Is this bv guaranteed to always be the second arg?

IgnaceBleukx (commented):
Great! I just have a concern about the dummy ivarmap being used (see comment)

cpmpy/transformations/linearize.py
tias: If Henk fixes int2bool to use csemap (as he apparently has on his own branch), than this should no longer be an issue...

cpmpy/transformations/linearize.py
tias: A != will not define the direct-encoding boolean?

After a few versions, this is the safest option: just detect direct encoding literals and replace those. Fancier things can be done later...

cpmpy/transformations/linearize.py
tias: yes, by flat normal form

cpmpy/transformations/linearize.py
IgnaceBleukx: Hmm I see, but we'll be doing/undoing a lot of work then no? In int2bool we'll check if any of the bvars already exist based on the csemap, and if we have already encoded an integer using this new transformation, all bvars already exist indeed. But that feels like a lot of work instead of doing this upfront?

I'm still fan of keeping ivarmap in the solver interface, and passing the `keep_integer` parameter along to the transformation.

cpmpy/transformations/linearize.py
IgnaceBleukx: Ok; I was thinking that when using categoricals, a user would write either `==` or `!=` as subexpr. 
But we can add that in later indeed.

IgnaceBleukx: We would just need to put (val, ~bv) in the `var_vals` map I think?

IgnaceBleukx: Hmm ok yeah no I tried a couple of things but it gets tricky quite quickly...
E.g., if you have both `a == 1 : BV0` and `a != 1 : BV1` in the csemap, then we need to detect these are actually the opposite of each other, chose one of the Booleans to represent `a = 1`, and then link everything together...
Seems out of scope for now, but can probably be made a lot easier if we have our custom CSEMap working.

E.g., the csemap could have a quick check if we already have a Boolean for the negated expression (at least for comparisons) and return the negated bool if that's the case. Out of scope for this PR in any case

cpmpy/transformations/linearize.py
tias: for now, as is documented, you either supply an ivarmap and dont want integer encodings, or you don't, in which case we will indeed recreate the BoolVarEnc objects...

I agree it would be better that all solvers have an ivarmap and indicate if they want to keep_integer or not, but I think we should move ahead without that for now...

PR #859, opened 2026-02-22
Deletion of solver objects and use of model in test_cons

hbierlee:
First added by @tias for #850, it has three changes:

- use `model` in test_constraints, which seems good to me
- delete solver objects after `model.solve*`, also seems good, but doesn't this happen automatically since `s` goes out of scope? 
- and added `__del__` methods for three solvers; I also think that at least for PySAT, it does have its own `__del__` methods, so these should also be called automatically?

### DIFF
--- a/cpmpy/model.py
+++ b/cpmpy/model.py
@@ -198,6 +198,7 @@ class Model(object):
         ret = s.solve(time_limit=time_limit, **kwargs)
         # store CPMpy status (s object has no further use)
         self.cpm_status = s.status()
+        del s  # free solver (and its memory)
         return ret
 
     def solveAll(self, solver:Optional[str]=None, display:Optional[Callback]=None, time_limit:Optional[int|float]=None, solution_limit:Optional[int]=None, **kwargs):
@@ -228,6 +229,7 @@ class Model(object):
         ret = s.solveAll(display=display,time_limit=time_limit,solution_limit=solution_limit, call_from_model=True, **kwargs)
         # store CPMpy status (s object has no further use)
         self.cpm_status = s.status()
+        del s  # free solver (and its memory)
         return ret
 
     def status(self):
--- a/cpmpy/solvers/cplex.py
+++ b/cpmpy/solvers/cplex.py
@@ -162,6 +162,11 @@ class CPM_cplex(SolverInterface):
         """
         return self.cplex_model
 
+    def __del__(self):
+        """Release CPLEX resources by calling cplex_model.end()."""
+        if self.cplex_model is not None:
+            self.cplex_model.end()
+
     def solve(self, time_limit:Optional[float]=None, **kwargs):
         """
             Call the cplex solver
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -152,6 +152,10 @@ class CPM_gurobi(SolverInterface):
         """
         return self.grb_model
 
+    def __del__(self):
+        """Release native model memory by calling grb_model.dispose()."""
+        if self.grb_model is not None:
+            self.grb_model.dispose()
 
     def solve(self, time_limit:Optional[float]=None, solution_callback=None, **kwargs):
         """
--- a/cpmpy/solvers/pysat.py
+++ b/cpmpy/solvers/pysat.py
@@ -208,6 +208,13 @@ class CPM_pysat(SolverInterface):
         # initialise everything else and post the constraints/objective
         super().__init__(name="pysat:"+subsolver, cpm_model=cpm_model)
 
+    def __del__(self):
+        if self.pysat_solver is not None:
+            self.pysat_solver.delete()  # requires explicit delete
+            self.pysat_solver = None
+        if self.pysat_vpool is not None:
+            self.pysat_vpool = None
+
     @property
     def native_model(self):
         """
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -307,11 +307,12 @@ def test_bool_constraints(solver, constraint):
     """
         Tests boolean constraint by posting it to the solver and checking the value after solve.
     """
+    m = Model(constraint)
     if ALL_SOLS:
-        n_sols = SolverLookup.get(solver, Model(constraint)).solveAll(display=lambda: verify(constraint))
+        n_sols = m.solveAll(solver=solver, display=lambda: verify(constraint))
         assert n_sols >= 1
     else:
-        assert SolverLookup.get(solver, Model(constraint)).solve()
+        assert m.solve(solver=solver)
         assert argval(constraint)
         assert constraint.value()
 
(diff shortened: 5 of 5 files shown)


tias:
as also reported by Ignace in #808, there is still a big open issue with pysat.

More specifically with pypblib in pysat/pb.py, where it seems like pysat is not (doing or able to) free pypblib's memory...

tias:
Reported: https://github.com/pysathq/pysat/issues/219

tias:
OK so pysat is due to a deeper issue with pypblib.

It still makes sense to merge this, no? It could avoid some other leaks?

@hbierlee if you agree, please merge, thx

hbierlee:
My suspicion is that implementing `__del__` in our solver interface will not do anything, because surely e.g. gurobi already implements a `__del__` method calling `model.dispose` (and additional destructors we don't know about, like `env.dispose` exists too apparently). We can check this for open-source solvers, like PySAT `Solver` definitely already calls `delete` (https://github.com/pysathq/pysat/blob/5fe41fd675e506304b3d7a140136ccd46ea055b4/pysat/solvers.py#L375-L380).

Now, it shouldn't really hurt, so if you just want to be rather safe than sorry, then this looks good to merge for me.

Otherwise, and this is the part that would be a new issue/PR if it sounds interesting enough, I think it's more useful is to just have a test (or dev script) which can check for leaks, so that we *know* when there is a leak and whether some code fixes it or not. I've made this one based on the discussion in pysat. Surely it can be improved, but the idea is to fail if an arbitrary function (in this case `make_solver`, but it can be anything, doesn't even have to be a CPMpy model) leaks more than say 0.25% per iteration. It should probably be a dev script, because running it in the CI might be flaky (especially with parallel tests; there are ways to ensure the test is run by itself, or even in a subprocess).

```py
def test_pypblib_memory_leak(solver):
    """Weighted PB encoding should not leak native memory across repeated solver creation."""
    import gc, os
    print("testing", solver)

    def get_rss_kb():
        with open("/proc/self/status") as f:
            for line in f:
                if line.startswith("VmRSS:"):
                    return int(line.split()[1])
        return None

    if get_rss_kb() is None:
        pytest.skip("RSS measurement not available (non-Linux)")

    def make_solver():

        # size = 10
        # x = cp.intvar(1, size, shape=size)
        # s = cp.SolverLookup.get(solver, cp.Model([cp.AllDifferent(x)]))
        # s.solve(time_limit=1)
        # return s

        size = 20
        cons = 5
        bvs = cp.boolvar(shape=size)
        weights = list(range(1, size + 1))

        m = cp.Model()
        for _ in range(cons):
            m.add(cp.sum(w * b for w, b in zip(weights, bvs)) <= sum(weights) // 2)

        s = cp.SolverLookup.get(solver, m)
        s.solve()
        return s

    n = 500
    rss = []

    def get_growth():
        g = ((rss[-1] - rss[0]) / rss[0])
        per_iter = g / len(rss)
        return (g * 100, per_iter * 100)

    def print_status():
        ratio, per_iter = get_growth()
        print(f"{len(rss)}: {rss[0]}..{rss[-1]} = {rss[-1] - rss[0]} growth ({ratio:.4f}%, or {per_iter:.4f}% per iter)")


    for i in range(1, n+1):
        _BoolVarImpl.counter = 0
        _IntVarImpl.counter = 0
        obj = make_solver()
        del obj
        gc.collect()
        rss.append(get_rss_kb())
        if i % 100 == 0:
            print_status()
            # print(f"  iter {i}/{n}: RSS={rss[-1]} KB ({get_growth():.2f}%)")

    _, per_iter = get_growth()
    passed = per_iter < 0.25
    if solver == "pysat":
        if passed:
            assert False, "issue resolved?"
        else:
            pytest.xfail("pypblib has a known memory leak (pysathq/pysat#219)")

    assert passed, "Memory leak?"
```

Here are two runs, the first with `__del__` implemented and the 2nd without. Clearly pysat is leaking with almost `0.6%`, but the other solvers not so much, and the `__del__` does not really make a difference:

```
# WITH __del__
tests/test_solvers.py testing ortools
100: 180288..180856 = 568 growth (0.3151%, or 0.0032% per iter)
200: 180288..181340 = 1052 growth (0.5835%, or 0.0029% per iter)
300: 180288..181644 = 1356 growth (0.7521%, or 0.0025% per iter)
400: 180288..181680 = 1392 growth (0.7721%, or 0.0019% per iter)
500: 180288..181736 = 1448 growth (0.8032%, or 0.0016% per iter)
.testing pysat
100: 182064..281084 = 99020 growth (54.3875%, or 0.5439% per iter)
200: 182064..389160 = 207096 growth (113.7490%, or 0.5687% per iter)
300: 182064..497240 = 315176 growth (173.1128%, or 0.5770% per iter)
400: 182064..605300 = 423236 growth (232.4655%, or 0.5812% per iter)
500: 182064..713420 = 531356 growth (291.8512%, or 0.5837% per iter)
xtesting gurobi
100: 721520..721528 = 8 growth (0.0011%, or 0.0000% per iter)
200: 721520..721548 = 28 growth (0.0039%, or 0.0000% per iter)
300: 721520..721548 = 28 growth (0.0039%, or 0.0000% per iter)
400: 721520..721548 = 28 growth (0.0039%, or 0.0000% per iter)
500: 721520..721548 = 28 growth (0.0039%, or 0.0000% per iter)
.testing cplex
100: 722636..722716 = 80 growth (0.0111%, or 0.0001% per iter)
200: 722636..723100 = 464 growth (0.0642%, or 0.0003% per iter)
300: 722636..723496 = 860 growth (0.1190%, or 0.0004% per iter)
400: 722636..723884 = 1248 growth (0.1727%, or 0.0004% per iter)
500: 722636..724308 = 1672 growth (0.2314%, or 0.0005% per iter)
.

# WITHOUT __del__
tests/test_solvers.py testing ortools
100: 168008..168564 = 556 growth (0.3309%, or 0.0033% per iter)
200: 168008..168956 = 948 growth (0.5643%, or 0.0028% per iter)
300: 168008..169468 = 1460 growth (0.8690%, or 0.0029% per iter)
400: 168008..169880 = 1872 growth (1.1142%, or 0.0028% per iter)
500: 168008..170488 = 2480 growth (1.4761%, or 0.0030% per iter)
.testing pysat
100: 172088..277420 = 105332 growth (61.2082%, or 0.6121% per iter)
200: 172088..385528 = 213440 growth (124.0296%, or 0.6201% per iter)
300: 172088..493588 = 321500 growth (186.8230%, or 0.6227% per iter)
400: 172088..606112 = 434024 growth (252.2105%, or 0.6305% per iter)
500: 172088..714200 = 542112 growth (315.0202%, or 0.6300% per iter)
xtesting gurobi
100: 722896..722904 = 8 growth (0.0011%, or 0.0000% per iter)
200: 722896..722928 = 32 growth (0.0044%, or 0.0000% per iter)
300: 722896..722928 = 32 growth (0.0044%, or 0.0000% per iter)
400: 722896..722928 = 32 growth (0.0044%, or 0.0000% per iter)
500: 722896..722932 = 36 growth (0.0050%, or 0.0000% per iter)
.testing cplex
100: 724080..724444 = 364 growth (0.0503%, or 0.0005% per iter)
200: 724080..724444 = 364 growth (0.0503%, or 0.0003% per iter)
300: 724080..724444 = 364 growth (0.0503%, or 0.0002% per iter)
400: 724080..724688 = 608 growth (0.0840%, or 0.0002% per iter)
500: 724080..725116 = 1036 growth (0.1431%, or 0.0003% per iter)
.
```

tias:
OK, so this was an attempt to fix PySAT's (actually, pyPBlib's) leak, which we can not fix.

Thanks for the follow-up script and analysis Henk.

I think we can safely keep expecting that solver developers do their own best in cleaning up memory, so I'm closing this for now.

PR #862, opened 2026-03-11
Unified prooflogging

IgnaceBleukx:
Proposal to unify the way we do prooflogging in CPMpy.

It makes the most sense to pass the proof-file at construction time, as for incremental solvers, the same proof-file is reused accros solve-calls.

There seems to be something weird with the tests for Exact, I think it has something to do with the file not being closed as long as the solver object is alive...

Apart from fixing that test, the only todo is adding PySAT. Feedback on the initial design is already welcome!


IgnaceBleukx:
Let's keep this on hold until Exact releases with new Prooflogging support via Python

IgnaceBleukx:
Updated this PR again, now with a more uniform inerface.

Some changes are breaking tho, so if they are kept in, the next release should be v1.1
The main difference compared to the previous iteration is verifcation of the proof is no longer supported directly after solve.
Instead, you need to do seperate calls to `solver.solve()` and `solver.verify()`. The verifification itself is handled in a new helper function under tools/verify.py and allows to use your own verifier (defaults to drat-trim for DRAT proofs and veripb for veripb proofs).

I'm not entirely sure if the changes to pysat are something we want to keep, but there is no way of getting the cnf file from the actual underlying solver object. I'm also not sure how/if assumptions are supported in the proof format.

cpmpy/solvers/gcs.py
```diff
@@ -341,9 +302,10 @@ def solveAll(self, display:Optional[Callback]=None, time_limit:Optional[float]=N
             timeout=time_limit, 
             solution_limit=solution_limit, 
             callback=sol_callback, 
-            prove=prove, 
-            proof_name=proof_name, 
-            proof_location=proof_location, **kwargs)
+            prove=self._proof is not None,
+            proof_name=self._proof,
```
OrestisLomis: Why not split the path here as done in ```.solve()```?

cpmpy/solvers/pumpkin.py
```diff
@@ -701,3 +694,13 @@ def solution_hint(self, cpm_vars: List[_NumVarImpl], vals: List[int]):
         """
         if self.pum_solver.is_inconsistent() is False: # otherwise, not guaranteed all variables are known
             self._solhint = {self.solver_var(v) : val for v, val in zip(cpm_vars, vals)} # store for later use in solve
+
+        
+    def verify(self, verifier, verifier_args: list[str] = [], time_limit: Optional[float] = None, display_output: bool = False) -> bool:
+        raise NotSupportedError("Pumpkin does not support external proof verification, only through the MiniZinc interface.")
```
OrestisLomis, line 700: Could we then not patch in the verification through `CPM_minizinc`? Though it might be a bit tricky to get the .fzn to match the posted model to pumpkin.

cpmpy/solvers/pysat.py
```diff
@@ -295,6 +310,22 @@ def solve(self, time_limit:Optional[float]=None, assumptions:Optional[Iterable[_
         else:  # another?
             raise NotImplementedError(has_sol)  # a new status type was introduced, please report on github
 
+        if self._proof is not None:
+            nvars = self.pysat_solver.nof_vars()
+            with open(f"{self._proof}.cnf", "w") as f:
```
OrestisLomis, line 317: You could perhaps add the assumptions here as hard clauses? I think it could work for a single proof + assumptions call, but might get very tricky when doing incremental calls since I think the proofs are cumulative. So at that point it's probably not even worth it..

cpmpy/solvers/pysat.py
```diff
@@ -591,3 +635,38 @@ def _pysat_pseudoboolean(self, cpm_expr, conditional=None):
             return self._pb.PBEnc.equals(**pysat_args).clauses
         else:
             raise ValueError(f"PySAT: Expected Comparison to be either <=, ==, or >=, but was {cpm_expr.name}")
+
+
+    def verify(self, verifier: str = "drat-trim", verifier_args: list[str] = [], time_limit: Optional[float] = None, display_output: bool = False):
```
OrestisLomis, line 642: I'm slightly confused as to why the 'verifier' is a user parameter. The proof was generated for a specific verifier right, so why should there be a parameter to specify this (and potentially make a mistake)?

cpmpy/solvers/pysat.py
OrestisLomis, line 280: Should we not have a guard that blocks assumptions when proof logging is enabled just like in pumpkin?

OrestisLomis (commented):
I think this is already quite nice. Some comments:
- Assumptions and proof-logging don't match (there seem to be some proof formats like IDRUP where the I stands for incremental, but those are not available through PySAT) so we need a guard to make sure only one of the two is enabled.
- Speaking of 'enabling' I think it would also be nice to not only declare the proof in the constructor, but have a method to enable/disable prooflogging.

cpmpy/solvers/gcs.py
IgnaceBleukx: Yeah good point, this should just follow the same style as .solve

cpmpy/solvers/pumpkin.py
IgnaceBleukx: Pfoe, that will be very hacky... Not really something I want to get into

cpmpy/solvers/pysat.py
IgnaceBleukx: Its probably better to just raise an error if there are assumptions while proof is not None. Its the same as in Pumpkin I believe

cpmpy/solvers/pysat.py
IgnaceBleukx: There are several tools out there to verify the same proof format, this is the most flexible approach and allows the user to specify the exact path to the verifier tool they want to use.

cpmpy/solvers/pysat.py
IgnaceBleukx: Yes : )

cpmpy/solvers/pumpkin.py
OrestisLomis: yeah that can definitely be ignored for this PR, was just thinking about the possibility

cpmpy/solvers/pysat.py
OrestisLomis: alright I was not aware about multiple verifiers for the same proof format existing!

PR #868, opened 2026-03-17
New solver: highs, open source ILP solver

tias:
latest PR version on the updated `examples/advanced/test_incremental_solving.py` has 1.8-2.2 speed-up, so moderately incremental


tests/test_solvers.py
```diff
@@ -1227,3 +1228,20 @@ def test_scip_special_cardinality():
...
+
+    m = cp.Model([x + 2 * y >= 10], minimize=x + y)
+
+    s = SolverLookup.get("highs", m)
+    assert s.solve()
+
+    # check variable types and a feasible objective
+    assert 0 <= x.value() <= 10
+    assert 0 <= y.value() <= 10
+    assert x.value() + 2 * y.value() >= 10
```
OrestisLomis: why not actually test for the specific objective? i.e. `y==5` and `x==0` according to the minimisation

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+    Install the Python bindings from PyPI:
+
+    .. code-block:: console
+
+        $ pip install highspy
+    
+    Detailed installation instructions available at:
+    https://ergo-code.github.io/HiGHS/dev/interfaces/python/
+
+    The rest of this documentation is for advanced users.
```
OrestisLomis, line 24: Add list of classes CPM_highs like for other solvers

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+from ..transformations.normalize import toplevel_list
+from ..transformations.reification import only_bv_reifies, only_implies, reify_rewrite
+from ..transformations.safening import no_partial_functions, safen_objective
+
+
+class CPM_highs(SolverInterface):
+    """
+    Interface to HiGHS' Python API (`highspy`).
+
+    Creates the following attributes (see parent constructor for more):
```
OrestisLomis, line 67: not really part of this PR, but I think the doc of each method should stand on its own without having to look at what its parent method does.

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+    supported_reified_global_constraints = frozenset()
+
+    @staticmethod
+    def supported() -> bool:
+        # try to import the package
+        try:
+            import highspy  # noqa: F401
+            return True
+        except ModuleNotFoundError:
+            return False
```
OrestisLomis, line 88: add general exception as well (as done for other solvers)

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+        """
+        Constructor of the native solver object
+
+        Arguments:
+        - cpm_model: a CPMpy Model()
+        """
+        if not self.supported():
+            raise ModuleNotFoundError("CPM_highs: Install the python package 'highspy' to use this solver interface.")
+        if subsolver is not None:
+            raise NotSupportedError("HiGHS does not support subsolvers")
```
OrestisLomis: change this to an assert to be more consistent with other solvers

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+            elif isinstance(cpm_var, _IntVarImpl):
+                hvar = self.highs.addIntegral(lb=cpm_var.lb, ub=cpm_var.ub)
+            else:
+                raise NotSupportedError(f"Not a known HiGHS variable type: {cpm_var}")
+
+            self._varmap[cpm_var] = hvar.index
+
+        return self._varmap[cpm_var]
+
+    def solver_vars_1d(self, cpm_vars):
```
OrestisLomis, line 161: should this be part of this PR? Maybe we should at least wait for #939

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+            solver_var = self._varmap.get(cpm_var, None)
+            if solver_var is not None:
+                # fast path
+                res.append(solver_var)
+            else:
+                # slow path, will check the varmap again
+                res.append(self.solver_var(cpm_var))
+        return res
+
+    def _row_from_linexpr(self, lhs) -> tuple[npt.NDArray[np.int32], npt.NDArray[np.float64], int|float]:
```
OrestisLomis: I find it weird that the param is called lhs.. elsewhere it gets called with a lhs, but here it is just an argument of the function so I think its best to refactor

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+                solvars = self.solver_vars_1d(lhs.args)
+                idx = np.array(solvars, dtype=np.int32)
+                coeffs = np.ones(idx.size, dtype=np.float64)
+            elif lhs.name == "wsum":
+                ws, vs = lhs.args
+                solvars = self.solver_vars_1d(vs)
+                idx = np.array(solvars, dtype=np.int32)
+                coeffs = np.asarray(ws, dtype=np.float64)
+            else:
+                raise NotImplementedError(f"HiGHS: unsupported operator {lhs}")
```
OrestisLomis: this error message seems to suggest the user did something wrong by providing unsupported argument, but it seems like this is actually a path we don't expect to end up in. I think it's best to update the error message to reflect that something might've gone wrong (as is the case on e.g. line 241)

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+            Eagerly add a constraint to the underlying solver.
+
+            Any CPMpy expression given is immediately transformed (through `transform()`)
+            and then posted to the solver in this function.
+        """
+        # track user vars and ensure newly seen ones have solver columns
+        get_variables(cpm_expr, collect=self.user_vars)
+
+        for con in self.transform(cpm_expr):
+            if isinstance(con, Comparison):
```
OrestisLomis, line 250: I think the `_add_transformed_constraint()` pattern in gurobi and scip is actually more readable and more extendable; e.g. I will want to add IIS computation for HiGHS in a separate PR and it makes more sense for this pattern to be available.

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+
+        self.highs.changeObjectiveOffset(const)
+
+        sense = highspy.ObjSense.kMinimize if minimize else highspy.ObjSense.kMaximize
+        self.highs.changeObjectiveSense(sense)
+
+    def has_objective(self):
+        return self._obj_cols is not None
+
+    def solve(self, time_limit=None, **kwargs):
```
OrestisLomis, line 334: what about solution callback? doc: https://ergo-code.github.io/HiGHS/stable/callbacks/

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+        """
+            Call the HiGHS solver.
+
+            Arguments:
+            - time_limit: maximum solve time in seconds (float, optional)
+            - kwargs:     any keyword argument, mapped to HiGHS options via ``setOptionValue``.
+                          Unknown/invalid options are ignored with a warning.
+
+            Notable HiGHS options (see upstream option list):
```
OrestisLomis: link to said upstream list, best to add: https://ergo-code.github.io/HiGHS/dev/options/definitions/

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+              according to HiGHS defaults).
+        """
+        import highspy
+
+        # ensure all vars are known to solver
+        self.solver_vars_1d(self.user_vars)
+
+        # edge case, empty model, ensure the solver has something to solve
+        if not len(self.user_vars):
+            self.add(intvar(1, 1) == 1)
```
OrestisLomis, line 360: why make a dummy model and not just report that the model is empty?

cpmpy/solvers/highs.py
```diff
@@ -0,0 +1,429 @@
...
+
+        start = time.time()
+        status = self.highs.run()
+        end = time.time()
+
+        info = self.highs.getInfo()
+        model_status = self.highs.getModelStatus()
+
+        self.cpm_status = SolverStatus(self.name)
+        # prefer solver-reported runtime when available, else fallback to wall-clock
```
OrestisLomis: separate from this PR, but why do we prefer solver reported runtime? Is the wall-clock not more representative for the user?

OrestisLomis (changes requested):
Some small req changes wrt clarity, consistency and a few cases of incompleteness. In general it looks good to me.

cpmpy/solvers/highs.py
tias: I'm fine with having it in here

cpmpy/solvers/highs.py
tias: I think we should only do it in yet another function when we have a need for it

cpmpy/solvers/highs.py
tias: for a follow-up PR (or at least not in my scope for this PR)

cpmpy/solvers/highs.py
tias: 'report' how? it should return SAT I guess? I don't actually konw what other solvers do if they get an empty model... that could be good to check

cpmpy/solvers/highs.py
tias: we shoud have both yes... its a known incompleteness in our implementation

tias: (e.g. the wallclock should include also posting time... there is a PR on that)

cpmpy/solvers/highs.py
OrestisLomis: other solvers seem to work fine (i.e. just report sat), without a need to handle the edgecase. In hindsight it is probably fine to just make a dummy model instead of engineering a way to report SAT without creating a model.

cpmpy/solvers/highs.py
tias: no, there is no point catching an exception to just raise it again... removing this from the template (should be removed from all solvers too, not for this PR)

tias:
all addressed

OrestisLomis (approved):
Alright, I agree with the counterpoints and the rest of the comments are resolved. Good to merge

PR #876, opened 2026-03-20
demonstrates overload for precise typing

tias:
as follow-up to the discussion in #874 

do we want this? its a lot of new syntax that was unnatural for me to read at first, but it shows that you can't nested lists

(there are 7-or-so other *args usages like this in globalconstraints)

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -216,9 +216,9 @@ class AllDifferent(GlobalConstraint):
     """
 
     @overload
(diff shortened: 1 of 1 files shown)


IgnaceBleukx:
As briefly discussed offline: the core issue is here that NDVarArray is also an expression, right?
So `ListLike[ExprLike]` can also have nested lists because of this.

But when do we actually allow for such nesting? I can't think of any case where arbitrary nesting in the constructor is allowed... So my preference would be to fix the type of NDVarArray such that `ListLike[ExprLike]` always represents a flat list of expressions or constants.

I find the syntax of the overloads quite hard to parse, but maybe it's just a matter of getting used to

tias:
by now we use it in a select nr of places, this demonstration no longer needed

PR #879, opened 2026-03-23
strict typing and implemetation of Comparison

tias:
A first attempt to make the typing and implementation more precise.

Reveals many interesting things:
- the test suite also posted `Const CMP Expr`, which is never possible for a user (python reverse-posts the comparison) so we should not support nor test it
- the test suite generates `BoolExpr == 4` which is invalid and we now raise...
- it seems like the testsuite also generates Incomplete `x // y`s which we now trip over because value() no longer uses argval() which silently hid that; maybe we should explicitly catch expected Incomplete's in the test suite? or they are generated by accident (no partial transform)?

Also, we now, as a hot path, try to convert any non-Expression to integer. If it fails, ValueError, if it succeeds we have our int. But this 'silently' converts integers to floats... (and we have a Z3 test that used a float) We don't support floats so we shouldn't test floats, but do we find it acceptable to silently convert floats? I do, but we should document it then...
(the alternative is 2 additional isinstance checks on every non-Expression in the comparison RHS)

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -241,29 +241,63 @@ class Expression(object):
         return Operator('->', [self, other])
 
     # Comparisons
-    def __eq__(self, other):
-        # BoolExpr == 1|true|0|false, common case, simply BoolExpr
-        if self.is_bool() and is_num(other):
-            if other is True or other == 1:
-                return self
-            if other is False or other == 0:
-                return ~self
-        return Comparison("==", self, other)
+    def __eq__(self, other: object):
+        if isinstance(other, Expression):
+            return Comparison("==", self, other)
+        else:
+            other_int = int(other) # type: ignore  # raises ValueError if not convertible to int
+            # BoolExpr == 1|true|0|false, common case, simply BoolExpr
+            if self.is_bool():
+                if other_int == 1:
+                    return self
+                elif other_int == 0:
+                    return ~self
+                else:
+                    raise ValueError(f"Comparison {self} == {other_int} is not valid. Expected Boolean Expression or Boolean constant, but got {other_int}.")
+            return Comparison("==", self, other_int)
 
-    def __ne__(self, other):
-        return Comparison("!=", self, other)
+    def __ne__(self, other: object):
+        if isinstance(other, Expression):
+            return Comparison("!=", self, other)
+        else:
+            other_int = int(other) # type: ignore  # raises ValueError if not convertible to int
+            # BoolExpr != 1|true|0|false, common case, simply BoolExpr
+            if self.is_bool():
+                if other_int == 0:
+                    return self
+                elif other_int == 1:
+                    return ~self
+                else:
+                    raise ValueError(f"Comparison {self} == {other_int} is not valid. Expected Boolean Expression or Boolean constant, but got {other_int}.")
+            return Comparison("!=", self, other_int)
 
-    def __lt__(self, other):
-        return Comparison("<", self, other)
+    def __lt__(self, other: object):
+        if isinstance(other, Expression):
+            return Comparison("<", self, other)
+        else:
+            other_int = int(other) # type: ignore  # raises ValueError if not convertible to int
+            return Comparison("<", self, other_int)
 
-    def __le__(self, other):
-        return Comparison("<=", self, other)
+    def __le__(self, other: object):
+        if isinstance(other, Expression):
+            return Comparison("<=", self, other)
+        else:
+            other_int = int(other) # type: ignore  # raises ValueError if not convertible to int
+            return Comparison("<=", self, other_int)
 
-    def __gt__(self, other):
-        return Comparison(">", self, other)
+    def __gt__(self, other: object):
+        if isinstance(other, Expression):
+            return Comparison(">", self, other)
+        else:
+            other_int = int(other) # type: ignore  # raises ValueError if not convertible to int
+            return Comparison(">", self, other_int)
 
-    def __ge__(self, other):
-        return Comparison(">=", self, other)
+    def __ge__(self, other: object):
+        if isinstance(other, Expression):
+            return Comparison(">=", self, other)
+        else:
+            other_int = int(other) # type: ignore  # raises ValueError if not convertible to int
+            return Comparison(">=", self, other_int)
 
     # Boolean Operators
     # Implements bitwise operations & | ^ and ~ (and, or, xor, not)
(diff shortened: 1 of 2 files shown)


cpmpy/expressions/core.py
```diff
@@ -241,29 +241,63 @@ def implies(self, other):
...
+        else:
+            other_int = int(other) # type: ignore  # raises ValueError if not convertible to int
+            # BoolExpr == 1|true|0|false, common case, simply BoolExpr
+            if self.is_bool():
+                if other_int == 1:
+                    return self
+                elif other_int == 0:
+                    return ~self
+                else:
+                    raise ValueError(f"Comparison {self} == {other_int} is not valid. Expected Boolean Expression or Boolean constant, but got {other_int}.")
```
IgnaceBleukx, line 256: According to our own spec, it is actually valid? You can use a Boolean expression as an integer anywhere you like. It's just that it is replaced with`false` by the `simplify_boolean` transformation.

cpmpy/expressions/core.py
```diff
@@ -241,29 +241,63 @@ def implies(self, other):
...
     # Comparisons
-    def __eq__(self, other):
-        # BoolExpr == 1|true|0|false, common case, simply BoolExpr
-        if self.is_bool() and is_num(other):
-            if other is True or other == 1:
-                return self
-            if other is False or other == 0:
-                return ~self
-        return Comparison("==", self, other)
+    def __eq__(self, other: object):
```
IgnaceBleukx, line 244: Shouldn't other be `ExprLike`?
Also, we can add the return type `Expression`?

cpmpy/expressions/core.py
```diff
@@ -241,29 +241,63 @@ def implies(self, other):
...
+        else:
+            other_int = int(other) # type: ignore  # raises ValueError if not convertible to int
+            # BoolExpr != 1|true|0|false, common case, simply BoolExpr
+            if self.is_bool():
+                if other_int == 0:
+                    return self
+                elif other_int == 1:
+                    return ~self
+                else:
+                    raise ValueError(f"Comparison {self} == {other_int} is not valid. Expected Boolean Expression or Boolean constant, but got {other_int}.")
```
IgnaceBleukx, line 271: Copy-paste err in eror message

cpmpy/expressions/core.py
IgnaceBleukx: We can't return `False` here just yet tho, as we need to keep all variables until we arrive at the solver

cpmpy/expressions/core.py
```diff
@@ -241,29 +241,63 @@ def implies(self, other):
...
+            return Comparison("!=", self, other_int)
 
-    def __lt__(self, other):
-        return Comparison("<", self, other)
+    def __lt__(self, other: object):
```
IgnaceBleukx, line 274: For these we can add return type `Comparison`

cpmpy/expressions/core.py
```diff
@@ -532,7 +566,7 @@ class Comparison(Expression):
     """
     allowed = {'==', '!=', '<=', '<', '>=', '>'}
 
-    def __init__(self, name: str, left: ExprLike, right: ExprLike) -> None:
+    def __init__(self, name: str, left: Expression, right: Expression|int) -> None:
```
IgnaceBleukx, line 569: I agree with this change, but we should add something inthe docstring saying that these comparisons should not be created manually as we don't do any type checking in the constructor.
The fact that `left` is never an integer is purely due to the way how Python implements operator overloading

cpmpy/expressions/core.py
```diff
@@ -559,16 +594,22 @@ def __bool__(self) -> bool:
...
-
-        if any(a is None for a in arg_vals): return None
-        if   self.name == "==": return arg_vals[0] == arg_vals[1]
-        elif self.name == "!=": return arg_vals[0] != arg_vals[1]
-        elif self.name == "<":  return arg_vals[0] < arg_vals[1]
-        elif self.name == "<=": return arg_vals[0] <= arg_vals[1]
-        elif self.name == ">":  return arg_vals[0] > arg_vals[1]
-        elif self.name == ">=": return arg_vals[0] >= arg_vals[1]
-        return None # default
+        lhs_val = self.args[0].value()
```
IgnaceBleukx, line 597: Why not use argvals here? There are certainly cases where this left-hand side can be partial?
On the current master, the IncompleteFunctionErr raising was stopped at the Comparison level; but now it can travel further up when calling `.value()`
I would have to think a bit harder if it can also result in a wrong answer from the value function, but it's definetly unexpected:

```python
x,y = cp.intvar(0,5, shape=2, name=tuple("xy"))

x._value = 3
y._value = 0

comp = x//y == 2

print(comp.value()) # raises IncompleteFunctionErr
print(argval(comp)) # ok this works fine

bv = cp.boolvar()
bv._value = True

expr = comp == bv
print(expr)

print(expr.value()) # also raises IncompleteFunctionErr, used to work fine on master
print(argval(expr))
```

tests/test_constraints.py
```diff
@@ -105,9 +105,10 @@ def comp_constraints(solver):
                     continue
                 for x,y in [(numexpr,rhs), (rhs,numexpr)]:
                     # check if the constraint we are trying to construct is always UNSAT
-                    if any(eval_comparison(comp_name, xb,yb) for xb in get_bounds(x) for yb in get_bounds(y)):
+                    if any(eval_comparison(comp_name, xb,yb) for xb in get_bounds(x) for yb in get_bounds(y)) \
```
IgnaceBleukx, line 108: We should update this test to use the `utils.eval_comparison` function, as this is also the way uses will/should make the comparisons themselves

IgnaceBleukx (changes requested):
Indeed many interesting aspects!

I agree we should not support floats, but the silent cast gives me some anxiety... Numpy used to have a `np.can_cast` function for checking if the cast is safe, but it no longer works for Python ints/floats; so not sure if there still is an efficient way of handling floats in a safe manner (i.e., raising an error/warning)

For the .value()/argval change, see my comment above, I strongly believe we should either catch the IncompleteFunctionErr in the `.value()` function of `Comparison`; or stick with using `argval` there too.

cpmpy/expressions/core.py
tias: we have to be equal or more general than the superclass typing, which is 'object'

tias:
I'm going to leave this hanging to pick up later (probably in separate PRs)

So, identified issues:

- floats in tests: we have no intention to support them in our standard Expressions anymore, we should update the tests to not use this
- floats should not be silently convert in expressions (probably)
- Comparison can be typed more strictly, but the test generation does put constants as LHS (users cant)
- value() on comparison with undefined functions... what is the right behaviour? (catch the Incomplete or let it propagate up?)
- smth with a test update, last comment

tias:
Too old and too many things at the same time.

Still tracked in #913 for when we get to Comparison eventually

PR #880, opened 2026-03-23
Add IIS-based MUS algorithm

hbierlee:
Use computeIIS of Gurobi to compute a MUS. We still need to make an assump model to ensure that each soft constraint, represented as a group of linear constraints, is individually turned off or on in the IIS. I don't think there is a way around it.

It would be very interesting to benchmark it (I did on pigeonhole v pysat, but there is no contest there of course).

Unlike the other MUS functions, it also has time/mem limit. Let me know if either of these should be omitted to be consistent with the other approaches which do not have time/mem limits, or if I should add a separate PR to add such limits to the other algorithms. Obviously, at least a time limit would be a usefull feature, I think. [edit: remove mem_limit for simplicity]


cpmpy/solvers/gurobi.py
```diff
@@ -320,6 +320,7 @@ def _make_numexpr(self, cpm_expr):
             Used especially to post an expression as objective function
         """
         import gurobipy as gp
+        print(cpm_expr, cpm_expr.name)
```
OrestisLomis: remove print statement

cpmpy/tools/explain/mus.py
```diff
@@ -330,6 +331,47 @@ def optimal_mus_naive(soft, hard=[], weights=None, solver="ortools", hs_solver="
     """
     return ocus_naive(soft, hard, weights, meta_constraint=True, solver=solver, hs_solver=hs_solver)
 
-   
 
+def mus_iis(soft, hard=[], solver="gurobi"):
+    """
+        Compute a MUS using a MIP solver's native Irreducible Inconsistent Subsystem (IIS) algorithm (MIP equivalent of MUS)
```
OrestisLomis: I find this wording slightly weird, which I guess is for future proofing reasons. It reads as several solvers may be used, but actually this implementation is gurobi specific.

cpmpy/tools/explain/mus.py
```diff
@@ -330,6 +331,47 @@ def optimal_mus_naive(soft, hard=[], weights=None, solver="ortools", hs_solver="
...
+
+        :param soft: soft constraints, list of expressions
+        :param hard: hard constraints, optional, list of expressions
+        :param solver: which ILP solver to use (only `gurobi` supported)
```
OrestisLomis: The solver is referred to as a MIP solver above (and I think that is what we stick to in general).

More in general, CPMpy doesn't really have a notion of MIP so perhaps ILP would be better?

tests/test_tools_mus.py
```diff
@@ -84,6 +84,22 @@ def test_wglobal(self):
...
+        mus_cons = self.mus_func(soft=soft, hard=hard)
+        assert len(set(mus_cons)) == 1
+        mus_naive_cons = self.naive_func(soft=soft, hard=hard)
+        assert len(set(mus_naive_cons)) == 1
+
+
+@pytest.mark.requires_solver("gurobi")
+class TestIIS(TestMus):
+    def setup_method(self):
+        self.mus_func = lambda soft, hard=[], solver="ortools": mus_iis(soft, hard=hard)
```
OrestisLomis: Slightly confusing to use "ortools" here as a default value imo even if it is a ghost default value anyway

OrestisLomis (commented):
Some small doc comments, looks good otherwise when keeping in mind that gurobi (and IIS) may give wrong results

cpmpy/tools/explain/mus.py
hbierlee: I think I prefer the future-proof version, it is meant as an IIS based alg, not a Gurobi alg.

tias (changes requested):
Please put the doc changes in a separate PR which will be trivial to accept (making reviewing this one cleaner too).

An ILP modeller would never introduce BigM's on their constraints to then use a native IIS method on only auxiliary 01 vars... thats not the spirit of ILP modellling.


Let's think it more through... so in CPMpy we can have constraint groups (either because the current linearisation does not keep it as one Generalized, or because it really is a conjunction of other constraints).

So how can we support native IIS on linear/generalized constraints, while still allowing groups of constraints? Can we post-process the result, instead of pre-processing the input?

OrestisLomis:
If not a solver method then how would you make the code modular and change depending on the solver? I'm writing the code now and I will make a PR. 

To be clear the solver method is supposed to be private, of course that is impossible to enforce in python, but the underscore kinda indicates that. Hence you're not supposed to call ```_native_mus``` on a solver object that you've already instantiated. Not sure if it is possible to assert that the internal model of a solver is indeed empty, but an assert at the start of ```_native_mus``` could throw an error if the solver's internal model is not empty.
The idea is that you call ```mus_native(soft, hard= [], solver=<solver>)``` as you would for any of the other mus methods in ```mus.py```. Then this method creates a fresh solver object (as also done in the other mus methods) and extracts the MUS natively.

hbierlee:
Static function sounds good, or even classmethod then.

Your post-processing idea sounds interesting. If I understand correctly, we use non-group MUS as a first step, then group MUS as a 2nd. So say we have 3 "groups" `(a,~a),(a),(b)`, so the non-group model is seen as `a,~a,a,b`, then we find `~a,a` (constraint 2, 3). We expand to the groups, `(a,~a),(a)`, which may grow our MUS back to a hopefully small US. Then in principal we just call any group MUS algorithm on this (we have to do more than a check if we want to find the actual group MUS). This subsequent group MUS call still requires us to reify the remaining groups, restart the solve non-incrementally, and use another solver (unless we use the current `mus_iis` implementation for this purpose). Alternatively, since Gurobi supports removing constraints, I could implement naive deletion-based MUS using e.g. `model.remove((a))`. Let me know if you had something simpler in mind?

I'm definitely interested in finding improvements, but I'm not sure why the current approach is so unnatural, as it's the exact same thing we do for all other MUS approaches with `make_assump_model`. My above optimization idea would help for models which have singleton groups (e.g. linear models), I think that might be worth a 2nd try too. Also note that if we have `n` constraints over `k` groups, and `k` is much smaller than `n`, then computing the non-group MUS on `n` constraints could perform more poorly than group MUS on `k` groups (even with the overhead of reification, if we think about deletion-based MUS, we'd get O(n) i/o O(k)).

hbierlee:
From offline discussion

- still need to make this part of `native_mus` introduced in #909 
- the newly proposed non-group/group MUS approach is more complex
- however, the current approach is not appropriate for ILP without the optimization for singleton groups. This needs to be added (and benchmarked to see if it's effective)

hbierlee:
In a weird twist, our "safe" assumption method is also not clearing that particular unit test. This may be an upstream bug. I've made a [forum post](https://support.gurobi.com/hc/en-us/community/posts/45775175552273-IISConstrForce-1-not-guaranteeing-IISConstr-1-in-computeIIS-output). In the meantime, it should be ok to continue with the optimization.

cpmpy/tools/explain/mus.py
hbierlee: now makes more sense since native_mus was added

hbierlee:
Now added the singleton group optimization, requiring an added private function to get to the Gurobi constraints. The alternative there would be to use getConstr in-between adding soft constraints, which would then require `grb_model.update`, which is actually quite slow.

I think it can be reviewed, but if approved, we should probably wait with merging until we resolved the (potentially upstream) issue that I linked in the previous comment.

cpmpy/solvers/gurobi.py
```diff
@@ -513,6 +520,66 @@ def solution_hint(self, cpm_vars:List[_NumVarImpl], vals:List[int|bool]):
...
+        soft = toplevel_list(soft, merge_and=False)
+        hard = toplevel_list(hard)
+
+        # Post hard constraints to Gurobi
+        s = cls(cp.Model(hard))
+
+        grb_soft = []  # each soft constraint will be represented by a *single* gurobi constraint
+        for con in soft:
+            get_variables(con, collect=s.user_vars)  # manually add soft constraint user vars to model
+            match s.transform(con):
```
tias: I guess we now have Python 3.10 as minimal requirement, so this could be acceptable...

but we don't use it anywhere yet, and here its just two simple cases... may I ask to use if/elif instead to follow more the current style of the rest of the codebase?

cpmpy/solvers/gurobi.py
```diff
@@ -513,6 +520,66 @@ def solution_hint(self, cpm_vars:List[_NumVarImpl], vals:List[int|bool]):
...
+            (s.native_model.getConstrs(), "IISConstrForce"),
+            (s.native_model.getGenConstrs(), "IISGenConstrForce"),
+            # we don't use these latter two at the time of writing, but future-proof anyway
+            (s.native_model.getQConstrs(), "IISQConstrForce"),
+            (s.native_model.getSOSs(), "IISSOSForce"),
+        ]:
+            for con in cons:
+                con.setAttr(attr, 1)
+
+        # now add each soft constraint or the assumption representing it
```
tias: I'm confused here... you are already adding constraints on line 549?

or no wait, you only add the reified, not the soft one?

OK, this is too far away from the real logic...

you declare grb_soft on line 541 before the loop, so please build it during that loop... I mean, add s._add_transfo(c) to grb_soft instead of first adding a c and then overwriting it here 10 lines below

cpmpy/solvers/gurobi.py
```diff
@@ -513,6 +520,66 @@ def solution_hint(self, cpm_vars:List[_NumVarImpl], vals:List[int|bool]):
...
+
+        # Post hard constraints to Gurobi
+        s = cls(cp.Model(hard))
+
+        grb_soft = []  # each soft constraint will be represented by a *single* gurobi constraint
+        for con in soft:
+            get_variables(con, collect=s.user_vars)  # manually add soft constraint user vars to model
+            match s.transform(con):
+                case [con]:  # if a constraint is represented by a single Gurobi constraint, it can be added as-is
+                    grb_soft.append(con)
```
tias: see comment below, so grb_con = s._add_transformed(con); grb_soft.append(grb_con)?

cpmpy/solvers/gurobi.py
```diff
@@ -513,6 +520,66 @@ def solution_hint(self, cpm_vars:List[_NumVarImpl], vals:List[int|bool]):
...
+                case cons:  # if a constraint is represented by multiple Gurobi constraints, `C`
+                    assumption = cp.boolvar()  # we introduce an auxiliary assumption variable, `a`
+                    s.add(assumption.implies(cp.all(cons)))  # we add `a -> /\ C` as a *hard* constraint (requires re-transform due to the added implication)
+                    grb_soft.append(assumption >= 1)  # then `a>=1` will be the single soft constraint whether group `C` is in the MUS
+
+        # update required to avoid `gurobipy._exception.GurobiError: GenConstr has not yet been added to the model`
+        s.native_model.update()
+
+        # force all hard constraints (including reified soft constraints) into the IIS
+        for cons, attr in [
```
tias: I find this much less readable then 4 for loops over the specific get*s with that specific attribute

cpmpy/solvers/gurobi.py
```diff
@@ -513,6 +520,66 @@ def solution_hint(self, cpm_vars:List[_NumVarImpl], vals:List[int|bool]):
...
+        except gp.GurobiError as e:
+            if e.errno == gp.GRB.Error.IIS_NOT_INFEASIBLE:
+                raise AssertionError("MUS: model must be UNSAT")
+            raise
+
+        return [
+            # return a MUS consisting of each soft constraint..
+            soft_i for soft_i, grb_soft_i in zip(soft, grb_soft)
+            # ..if its representing Gurobi soft constraint is in the IIS
+            if any(getattr(grb_soft_i, attr, False) for attr in ("IISConstr", "IISGenConstr", "IISQConstr", "IISSOS"))
```
tias: cryptic cryptic...
please build the list with a for loop (giving you space to put comments where we typically put them, instead of inside a list comprehension)
then return the list

cpmpy/solvers/gurobi.py
```diff
@@ -399,6 +400,15 @@ def add(self, cpm_expr_orig):
...
       # transform and post the constraints
       for cpm_expr in self.transform(cpm_expr_orig):
+          self._add_transformed(cpm_expr)
+
+      return self
+
+    __add__ = add  # avoid redirect in superclass
+
+    def _add_transformed(self, cpm_expr):
+        """Post a single already-transformed constraint to the Gurobi model. Returns the Gurobi constraint."""
```
tias: maybe add a comment 'also used in mus_native which needs access to the expressions'? so the reader knows why it differs from the standard pattern

tias (changes requested):
good; I do have some requests wrt code readability/maintainability

cpmpy/solvers/gurobi.py
hbierlee: I've now done some effort to avoid this to make the code more readable, see latest version. There is a good reason we first just collect soft constraints without adding them to Gurobi, namely because I first add *all* hard constraints so I can set their `IISConstrForce` attribute by iterating over all Gurobi constraints, and then add the soft constraints.

Please see for an updated version which does it the old way below. I'm fine with either, some parts are simpler, others more complex. Note we have to should call `model.update()` as it can be expensive (but is required before changing constraint attributes). Let me know which version you prefer, the current one or the old one (in which case I can revert).

https://github.com/CPMpy/cpmpy/blob/8b72ba3024f7ae561703a840e671c110f08479ab/cpmpy/solvers/gurobi.py#L535-L593

hbierlee:
@tias I've processed the review, see one outstanding comment. I think you can just say A or B, then approve, then I will merge.

cpmpy/solvers/gurobi.py
```diff
@@ -513,6 +520,92 @@ def solution_hint(self, cpm_vars:List[_NumVarImpl], vals:List[int|bool]):
...
+
+        Args:
+            soft: List of soft constraints over which a MUS needs to be found
+            hard: List of hard constraints that always need to be satisfied
+
+        Returns a MUS (list of constraints from soft).
+        """
+
+        # TODO unsure if needed?
+        soft_cons = toplevel_list(soft, merge_and=False)
```
OrestisLomis: in this PR I would definitely leave it like this, could be a question in general wether this is needed or not tho

hbierlee:
There is now a ticket (https://support.gurobi.com/hc/en-us/requests/116323) for the confirmed IIS bug in Gurobi (wrt forcing constraints into the IIS). May be worth just waiting until it's resolved before we merge/release IIS-based MUS. Even if fixed in gurobi 13.0.2, if we this, it may be worth putting a warning in case someone has gurobi<13.0.2 (or more extremely, upping the bound for gurobi>=13.0.2, but it'd prefer the warning).

hbierlee:
I've added detection for the IIS issue, and I can confirm it is actually triggering on one bigger test case (not checked-in). Once the issue is fixed in gurobi, we could skip this check, or we can keep it in for safety.

This added a small `in_iis` inner helper function, which allowed me to make the return a very simple comprehension.

tias (approved):
great, I just added some of the doc to the docstring of the class method.

lets hope gurobi fixes the bug soon...

PR #884, opened 2026-03-25
Read integers with `round` i/o `int` for Gurobi

### LINKED ISSUE #858: Tolerance issue for gurobi leading to incorrect/out-of-domain assignments
In the following model, we get integer variables with domains `1..3` assigned to `0` by the Gurobi interface, since Gurobi assigns them e.g. `0.999999`, and we use `int` to convert values from Gurobi, which floors/truncates them.

```py
import cpmpy as cp

x = cp.intvar(1, 3, shape=2, name="x")
_BV202_r = cp.intvar(0, 1, shape=3, name="_BV202_r")
_BV203_r = cp.intvar(0, 1, shape=3, name="_BV203_r")
m = cp.Model()
m += cp.sum([_BV202_r[0], _BV202_r[1], _BV202_r[2]]) == 1
m += cp.sum([3, 3, 1, -1] * cp.cpm_array([_BV203_r[0], _BV203_r[1], _BV203_r[2], x[1]])) == 0
m += cp.sum([_BV203_r[0], _BV203_r[1], _BV203_r[2]]) == 1

m.solveAll(solver="gurobi", solution_limit=1000, display=x, Seed=42)
```

Outputs:

```
❯ python examples/tols.py 
[None 1]
[None 3]
[None 3]
[None 3]
[None 3]
[None 3]
[None 0]
[None 3]
[None 0]
```

Instead, we should simply use e.g. `round` to get more correct results.

For additional safety, we could add an integrality check (as is done for the objective, or it could be done using `np.isclose ` as proposed in #579). Currently, we silently covert the objective to a float if it is not integer. We could:

a) do the same for integer variables, or
b) use python round in any case, or
c) round using the currently set `IntFeasTol` (there is no Gurobi-official way to do this, but I've seen mention of `np.ceil(x - 0.5 + IntFeasTol)`)

Based on my (recently developed) understanding of tolerances, with e.g. the default `IntFeasTol` set to 1e-5, we should never see e.g. 0.98, and if we do, something is going wrong and relaxing/tightening the tolerances would not resolve it. So, in that case, for option b/c, we could additionally raise an error if the integrality check fails (although I'm not sure what advice to give to the user in this case).

PS: a different issue is that `x[0]` is assigned to `None` since it does not occur in any constraint. This can be problematic for e.g. solution counts.

hbierlee:
Fixes #858 

We should at least use `round` to read integers, because now whenever Gurobi decides to slightly assign an integer variables to a value below rather than above the integer (e.g 0.99999 v. 1.000001), we'll floor/truncate to the (likely) wrong assignment. However, to be even safer, we'll have to look into IntegralityFocus.

### DIFF
--- a/cpmpy/solvers/TEMPLATE.py
+++ b/cpmpy/solvers/TEMPLATE.py
@@ -245,6 +245,7 @@ class CPM_template(SolverInterface):
             # fill in variable values
             for cpm_var in self.user_vars:
                 sol_var = self.solver_var(cpm_var)
+                # [GUIDELINE] for ILP-solvers, ensure the value is integer and use `round()`
                 cpm_var._value = self.TPL_solver.value(sol_var)
                 raise NotImplementedError("TEMPLATE: back-translating the solution values")
 
--- a/cpmpy/solvers/cplex.py
+++ b/cpmpy/solvers/cplex.py
@@ -252,7 +252,7 @@ class CPM_cplex(SolverInterface):
             if self.has_objective():
                 obj_val = self.cplex_model.get_objective_expr().solution_value
                 if round(obj_val) == obj_val: # it is an integer?:
-                    self.objective_value_ = int(obj_val)
+                    self.objective_value_ = round(obj_val)
                 else: #  can happen with DirectVar or when using floats as coefficients
                     self.objective_value_ = float(obj_val)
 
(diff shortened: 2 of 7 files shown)


IgnaceBleukx (approved):
Yes, the code seems fine to me, I've added a note on the rounding in the template for solver interfaces.

I'm not sure about the test... it definately is not watertight, so might be better to just remove it?

tias:
I think the template change is missing? please merge afterwards (with the test, its OK)

PR #891, opened 2026-03-27
MDD decomposition for Table constraint 

WoutPiessens:
The MDD decomposition for the Table global constraint has been added.

At this moment, the decomposition only happens for top-level Table constraints. This decomposition should also be valid for positively reified Table constraints, however.


WoutPiessens:
We will split this PR into three separate PRs:
1) Add a MDD global constraint 
2) Add the Table decomposition using MDD
3) Work on alternative decompositions for global constraints that are only valid when the constraint is positively reified (or top-level)

hbierlee:
One thing I thought about today:

As I said offline, my view is that adding the MDD as a new (public) global is _not_ for the sake of modularity (you can achieve that by a private class MDD as well, e.g. an inner class of Table decompose to make it _really_ private). The real reason to make the MDD a public global constraint should be because it's useful to users (e.g. used in modelling, or supported by solvers, and so on). However, a public MDD global does come with more obligations, though, namely maintenance, better documentation, and (here's the new thing I thought about) the requirement to make the flow decomposition *also* fully reifiable. So just something to be aware off :)

tias:
Is this one superseded by #924 and #981?

If so, please close, if not, please comment @WoutPiessens

PR #897, opened 2026-03-31
exprs with *args constructors, avoid flatlist

tias:
a pattern to avoid flatlist in some constructors

to be merged after #895


tias:
that will be resolved by: #958

could still be worth splitting the current PR up

tias:
outdated, better to redo from master. Kept for reference in tracker #913

PR #900, opened 2026-04-01
Datasets core

ThomSerg:
A first PR in a larger sequence of upcoming ones, bringing the work on datasets, IO and benchmarks from branch `benchmark_datasets` into master. Tried to keep it as minimal as possible, also with just a single dataset implementation; XCSP3Dataset. In some places you will notice some placeholders / things that don't seem needed right now but that future PRs will use to build upon. Some of these are labelled with "TODO".


cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,787 @@
...
+
+        return {
+            "name": cls.name,
+            "description": cls.description,
+            "homepage": cls.homepage,
+            "citation": citations,
+        }
+
+
+class IndexedDataset(Dataset):
```
OrestisLomis: Why is this functionality not directly in ```Dataset```? I do not really see what kind of data would not be implementable in an indexable way. Is it for instance generator datasets?

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,787 @@
...
+    - Model-defined instances: files directly encode variables/constraints/objective
+      (for example XCSP3, OPB, DIMACS, FlatZinc). In this case, users typically
+      pass a loader as ``transform``, converting the raw file instance into a model.
+    - Data-only instances: files encode problem data for a fixed family, but no
+      model. In this case, subclasses should override ``parse()`` and users can
+      enable ``parse=True`` to obtain parsed intermediate data structures
+      (for example table/dict structures for RCPSP-style scheduling data), then
+      build a model separately or via a transform.
+    """ 
+    # TODO documentation to add / improve
```
OrestisLomis, line 257: Is this TODO still something to add for this PR?

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,787 @@
...
+            target_transform (callable, optional): Optional transform applied to the metadata dictionary.
+            download (bool): If True, downloads the dataset if it does not exist locally (default=False).
+            parse (bool): If True, run ``self.parse(instance_path)`` before
+                applying ``transform``. Intended for data-only datasets that do
+                not directly encode a model in the source file.
+            extension (str): Extension of the instance files. Used to filter instance files within the dataset directory.
+            **kwargs: Advanced options. Currently supports:
+                - ignore_sidecar (bool): If True, do not read/write metadata
+                  sidecars and collect metadata on demand at iteration time
+                  using ``collect_instance_metadata()`` (default: False).
```
OrestisLomis: consisteny -> ```(default=False)``` like for the ```download``` option

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,787 @@
...
+
+        Raises:
+            ValueError: If the dataset directory does not exist and `download=False`,
+                or if the requested year/track combination is not available.
```
OrestisLomis: Why are year/track mentioned here?

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,787 @@
...
+                raise ValueError("Dataset not found. Please set download=True to download the dataset.")
+            else:
+                self.download()
+                if not self._ignore_sidecar:
+                    self._collect_all_metadata()
+                files = self._list_instances()
+                print(f"Finished downloading {len(files)} instances")
+
+        files = self._list_instances()
+        if len(files) == 0:
```
OrestisLomis, line 314: Can we check that the data is not tampered with? Maybe for starters if ```__len__``` is actually hardcoded to be a specific number you can change this line with ```len(files) != len(self)```. The tradeoff is that you have to add a lot of hardcoded values for each year/track/... combination for some benchmarks which might not be desirable. What if the benchmark changes size for some reason?

TL;DR I think this is fine, but perhaps think a little about more data correctness guarantees

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,787 @@
...
+    def _list_instances(self) -> list:
+        """
+        List all instance files, excluding metadata sidecar files.
+
+        Returns a sorted list of `pathlib.Path` objects for all instance files
+        matching the dataset's extension pattern.
+        """
+        return sorted([
+            f for f in self.dataset_dir.rglob(f"*{self.extension}")
+            if f.is_file() and not str(f).endswith(self.METADATA_EXTENSION)
```
OrestisLomis, line 466: Maybe add a filter for .md files? Perhaps a readme is also part of a downloaded benchmark, which you obviously don't want to actually include in your instance list.

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,787 @@
...
+        """
+        meta_path = self._metadata_path(file_path)
+        metadata_error = None
+        try:
+            instance_meta = self.collect_instance_metadata(str(file_path))
+        except Exception as e:
+            instance_meta = {}
+            metadata_error = str(e)
+
+        # Derive instance name (strip format-specific extensions) TODO
```
OrestisLomis: not a todo anymore?

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,787 @@
...
+        Arguments:
+            url (str): The original URL to download from (used as fallback).
+            target (str): The target filename to download.
+            destination (str, optional): The destination path to save the file.
+            desc (str, optional): Description to show in the progress bar.
+                                  If None, uses the filename.
+            chunk_size (int): Size of each chunk for download in bytes (default=1MB).
+
+        Returns:
+            str: The destination path where the downloaded file is saved.
```
OrestisLomis: returns an os.PathLike

cpmpy/tools/datasets/xcsp3.py
```diff
@@ -0,0 +1,156 @@
...
+import os
+import lzma
+import zipfile
+import pathlib
+import io
+
+from cpmpy.tools.datasets.core import FileDataset
+
+
+class XCSP3Dataset(FileDataset):  # torch.utils.data.Dataset compatible
```
OrestisLomis, line 17: Is XCSP3 really the next step in the tree? Would it not be useful to have ```XMLFileDataset``` as an intermediate? I'm not sure if there are other datasets using this format, but looking towards the PB comp or the SAT comp there are definitely non competition benchmarks that also use OPB or CNF as a file format, so having the fileformat as an intermediate seems to be useful.

OrestisLomis (changes requested):
I think it is really nice! Some questions about the class hierarchy remain for me, please have a look. Few more small comments as well

cpmpy/tools/datasets/core.py
ThomSerg: Any dataset in which the instances must be consumed in order (where random access is not possible). This can for example happen when instances are generated on the spot and this generation process is sequence dependent, e.g. it uses a initial random seed and then goes on from there. Or maybe a very large dataset that is for some reason in one very big file, where reading the entire file into memory is not efficient and it is better to just stream instance by instance. I do agree that for now we don't have a usecase for it, but I wanted to prevent that in the future we or any other contributor couldn't add one of these alternative dataset types due to us hard-coding indexability into the base class. With a very generic and limited base class, these future additions are still a possibility.

cpmpy/tools/datasets/core.py
ThomSerg: No, not for this PR. Just hints for what's to come in future PR's. Maybe I shouldn't have used the keyword TODO.

cpmpy/tools/datasets/core.py
ThomSerg: That's the `rglob(f"*{self.extension}")` part, only collect the files matching the dataset's extension type. For example `.xml.lzma` for xcsp3

cpmpy/tools/datasets/xcsp3.py
ThomSerg: Not sure what this would add. Generic CNF / OPB support is already provided by the loaders, which are separate from the dataset. Datasets encapsulate the specifics of how to download a dataset archive, how to extract it and how the instances inside it are structured and how they must be iterated over. A `CNFFileDataset` would just have an empty implementation, nothing special to add to that class. In the mega PR there is the addition of some utilities to create a dataset object on top of just a collection of local files, without the download stuff. This is needed for iterating over a dataset that has been translated and written back to file.

cpmpy/tools/datasets/core.py
ThomSerg: After internal discussion, I removed the intermediate `IndexedDataset`. All datasets now inherit from `Dataset`, which requires implementation for indexability and iterability. Datasets that are only one of the two are more exception than the norm, so one idea is to just keep the code simple.  Though this will cause issues for those datasets (streaming datasets, generator-based datasets), since they will have to through an exception that random access is not allowed. And it's not that nice of a design that the user has to take into account that index-based access could throw an exception for some datasets. In my opinion the previous code for keeping indexability out of the base dataset class is not a lot, but I can also see the other point of view in terms of complexity and maintainability.

OrestisLomis: I do agree that forcing someone who wants to implement a `Dataset` to throw errors for abstract functions he doesn't need is maybe not the cleanest design, however I would argue it is also not necessary. Both the cases you mentioned can naturally be iterated over, the indexing is perhaps a little trickier and maybe not so clean, but still possible by naively just starting from the startpoint every time you want to index an instance. This can be improved with hashing/caching perhaps, but up to the implementor.

OrestisLomis (approved):
I think it is alright as is now. See comment for more detailed discussion.

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,786 @@
...
+try:
+    from tqdm import tqdm as _tqdm
+    tqdm = _tqdm
+except ImportError:
+    pass
+
+import cpmpy as cp
+
+
+def _format_bytes(bytes_num):
```
tias, line 97: if you import typing I kind-of expect typed functions... e.g. this is float or int?

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,786 @@
...
+
+    Used to display download progress.
+    """
+    for unit in ['bytes', 'KB', 'MB', 'GB', 'TB']:
+        if bytes_num < 1024.0:
+            return f"{bytes_num:.1f} {unit}"
+        bytes_num /= 1024.0
+
+
+class classproperty:
```
tias, line 109: that seems like sugar coating... do we care?

things like this can stop us from using mypyc in the future, while file parsing/loading is something where C code can really shine...

cpmpy/tools/datasets/core.py
tias: (response to Orestis' response:)

its OK to work with what we have...

we check/reload the metadata for each instance on disk in any case if I understand the code lower well

tias: (continued response to my response to Orestis' response)

(with what we have = if I download the data, and chose to delete some tracks to save space, I don't expect my code to complain but rather just to use what is on the disk)

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,786 @@
...
+        # Instance reference is a string path (PyTorch-style); parse/transform may replace it.
+        data: Any = str(file_path)
+
+        # Built-in parse stage: parse the instance file into intermediate data structures.
+        # Mostly meant for datasets where files represent data and modeling is separate.
+        if self._parse:
+            data = self.parse(file_path)
+
+        if self.transform:
+            # TODO maybe revisit this flow of execution once CPMpy model feature extraction has been added
```
tias, line 498: is this an old todo, or is this an actual todo for a soon-but-later PR? I think we describe the flow well in the paper, and there is no actually intended todo left? (one can always refactor anything, but that should not be written in the code)

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,786 @@
...
+                    iterator = as_completed(futures)
+                
+                for future in iterator:
+                    try:
+                        future.result()
+                    except Exception as e:
+                        fp = futures[future]
+                        print(f"Error collecting metadata for {fp.name}: {e}")
+
+    def _collect_one_metadata(self, file_path):
```
tias, line 603: euh... we have 'instance_metadata()', why do we need this?  it looks duplicate... (its also untyped)

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,786 @@
...
+        """
+        Download file sequentially with progress bar.
+        """
+        filepath = pathlib.Path(filepath)
+
+        # Ensure parent directory exists
+        filepath.parent.mkdir(parents=True, exist_ok=True)
+        
+        req = Request(url)
+        with urlopen(req) as response:
```
tias, line 710: aren't you overdoing it? If I understand this correctly you implement your own pretty-print downloader, but I'm not sure I want to maintain a custom pretty printer in a combinatorial optimisation library...

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,786 @@
...
+    .. code-block:: python
+
+        dataset = from_files("path/to/dataset_files", extension=".txt")
+        for x, y in dataset:
+            print(x, y)
+    """
+    class FromFilesDataset(FileDataset):
+        # Plain class attributes so that dataset_metadata() (a classmethod
+        # that reads cls.name / cls.description / ...) works correctly.
+        name = ""
```
tias, line 770: I think this is the better, cleaner way... just plain class attributes, should be like that in the superclass too?

tias: if its required, you can check in the constructor that it should be non-empty...

cpmpy/tools/datasets/core.py
```diff
@@ -0,0 +1,786 @@
...
+        def __init__(self, dataset_dir: os.PathLike, extension: str = ".txt"):
+            # Set name from the directory so metadata contains something useful.
+            self.name = pathlib.Path(dataset_dir).name
+            super().__init__(dataset_dir=dataset_dir, extension=extension)
+
+        def categories(self) -> dict:
+            return {}
+
+        def download(self) -> None:
+            raise NotImplementedError("from_files() datasets are already local; downloading is not supported.")
```
tias, line 784: I would say `pass  # already in local files`? There is no error here...

tias (commented):
great!

also nicely extensive test-suite.

Some questions/remarks where I think it can be a bit cleaner/simpler, but you know better what is still coming so responses very welcome

cpmpy/tools/datasets/core.py
tias: (read other comments lower first)

ThomSerg: My reasoning for this was to be able to define "abstract" on class fields, so that when a user is developing a new dataset they get immediate feedback when they forgot to overwrite one of these fields (at import time instead of construction time). With that decorator it acts just like an abstractmethod. But I do also fear that this will cause issues with mypyc. An alternative would be to just use normal fields with placeholder values that should be replaced (or leave empty) and then use the `__init_subclass__` method to check, before construction, that the user did so.

```python
class Dataset(ABC):
    name: ClassVar[str]
    description: ClassVar[str]
    homepage: ClassVar[str]
    citation: ClassVar[List[str]] = []

    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        if cls is Dataset:
            return
        for attr in ("name", "description", "homepage"):
            if attr not in cls.__dict__:
                raise TypeError(f"{cls.__name__} must define class attribute {attr!r}")
```

cpmpy/tools/datasets/core.py
OrestisLomis: alright makes sense

cpmpy/tools/datasets/core.py
ThomSerg: So +- what you also propose later on. And thinking of it, no need to use the special `__init__subclass__`, could indeed just check in the regular constructor.

ThomSerg: Actually I do like `__init_subclass__`, allows for definition-time exceptions instead of construction-time, just like the previous `classproperty`.

cpmpy/tools/datasets/core.py
ThomSerg: Cleaned it up a lot to remove all this duplicate code. Have also removed parallel metadata extraction (it was not really used and added a lot of extra code complexity). Might revisit if the future, but fine for now.

ThomSerg:
Based on the feedback, I simplified the dataset classes. Also added it to the CPMpy API documentation. 

I do feel that we should also add a "Datasets and Benchmarking" page in the advanced section of the docs, giving some examples on how to use the datasets in practice, loading them into CPMpy models and solving, translating to a different format, etc. But that's for a separate PR.

tias (approved):
Yes, great!

- docs: I agree. I do see good and extensive docs in datasets/core.py; I would not recommend writing a lot of new documentation that is already present in the files; rather I would suggest that you add a short motivating paragraph in the user docs and then point to these module docs?

- I see XCSP3 dataset, but I don't see a change to the current tools/xcsp3; I would expect that the tool switches to using this dataset reader?
  or is that not unlinkable from the IO PR (in which case we can just continue merging this now)

ThomSerg:
Resolved the two comments
- Added minimal user docs with reference to extensive api docs
- old xcsp3 tools now make use of new dataset

tias (approved):
I made a readme update and 2 'raise' fixes/changes in benchmark.py of xcsp3

Can you check them? If good, then please do merge!

PR #902, opened 2026-04-02
Ruff autoformatting expressions/ folder

IgnaceBleukx:
As discussed offline, I played around with `ruff` a little bit.
You can go very far in configuring its rules, some are quite pedantic imho...

This is just a PR to show how our `expressions/` folder could look like if we were to add autoformatting.
I've enabled the recommended doc-styles and standard pyflakes code formatting, but enabled some rules of it as they were too incompatible with the current code base...

If we decide to go with autoformatting, there is also the question on how to integrate it continuously... Formatting the entire code base at once will destroy our git history, but after chatting with a friend, they suggested me a few options:
1. Only format changed files when pushing new changes
2. Only format changed lines (with git diff) when pushing new changes -- can cause some weirdness as ruff often needs more context for proper formatting
3. Do a big formatting commit, but ignore it in the git blame history ```git config blame.ignoreRevsFile .git-blame-ignore-revs```

We can also setup the autoformatting in the CI of course so we don't all have to do it locally.

### DIFF
--- a/cpmpy/expressions/__init__.py
+++ b/cpmpy/expressions/__init__.py
@@ -1,18 +1,18 @@
 """
-    Classes and functions that represent and create expressions (constraints and objectives)
+Classes and functions that represent and create expressions (constraints and objectives)
 
-    ==================
-    List of submodules
-    ==================
-    .. autosummary::
-        :nosignatures:
+==================
+List of submodules
+==================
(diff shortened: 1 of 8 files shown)


tias:
nothing this ambitious, better code hygiene always welcome.

Issue #342, opened 2023-06-02
What to rewrite at construction time?


IgnaceBleukx:
As its popping up in several pull requests, I think it's time to decide what to simply/rewrite at the construction time of our Expressions and what to leave for transformations later in the pipeline.

## Neutral-arguments
An obvious optimization we implemented atm is ignoring neutral arguments to expressions.
For example, `x + 0` becomes just `x`. `1 * x` becomes `x` etc.
But also for boolean operators, for example, `x | False` will become `x`.
Personally, I think it's fine to do these simplifications, but we have to be consistent. 
I think many neutral arguments are filtered away atm in the `__xxx__` methods, but then we should also do this in the constructor of the `Expression`s? For example, `Operator('sum', [x,0])` will remain `x + 0`.
I think the easiest way to accomplish that is to subclass every operator type as Helene suggested in #334.

## Falsifying arguments
The second case I can think of is when an expression will evaluate to `False`. For example `x & False` can become `False`, but also `any([])`. Analogously, some expressions simply evaluate to `True`, such as `all([])` or `AllDifferent([])`
I know we do this for some operators/global constraints atm but probably not very consistently.
I think this is a little trickier to make a decision on what to evaluate and what to leave. For example, what with comparisons where the bounds will always render a constraint `True` or `False`? Those cases are much less trivial to comprehend for a user...

## Rewriting of operators
Lastly, we can also rewrite some operators at construction time to simpler/other operators.
One that we have implemented now is `x - y` will become `x + (-y)`. This I don't really like tbh... it either should become a weighted sum or just stay `x - y`.
Then there is the merging of n-ary operators. Such as `(x + y) +  z` becomes `sum([x,y,z])`. 
On the one side, this is logical to do as probably a user will write something like `x + y + z` which is equivalent to `(x + y) +  z`
What about `10x + 20y - 30z`? Should we make it a weighted sum? I guess it makes sense to do so if we rewrite `x - y` to a sum already, but it's not exactly what the user wrote.
But then the current master branch,  `1 * x - 1 * y becomes `x + (-y) while `1*x + -1*y becomes `wsum([1,-1],[x,y])` so that's counter-intuitive with the first optimization of zero-arguments.
Another rewriting of `n-ary` operators involves the reordering of arguments and merging of constants. I.e, `10 + x + 20` **could** become `30 + x`. I'm not sure we should do it, but it's quite appealing to cover cases like: `10x + (4 * y * 3)` which would become `10x + 12y`.

[...]

PR #907, opened 2026-04-03
some more typing of core

tias:
one notable change is about 'implies' and simplification...

one principle is that we typically don't simplify expressions at creation time, only at transformation time (subject to some exceptions).

I now make this more explicit, making 'simplify' an option to implies (which we don't activate anywhere at the moment). So for the user we wont delete expressions and variables unexpectedly, and for us, if we want it to activate we can.

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -240,7 +240,7 @@ class Expression(object):
         return True
 
     def value(self) -> Optional[int]:
-        return None  # default
+        raise NotImplementedError(f"`value` is not yet implemented for type {self}")
 
     def get_bounds(self) -> tuple[int, int]:
         if self.is_bool():
@@ -263,7 +263,7 @@ class Expression(object):
 
         Args:
             other (ExprLike): the right-hand-side of the implication
-            simplify (bool): if True, simplify True/False constants (might remove expressions & their variables from user-view)
+            simplify (bool): if True, simplify by eliminating True/False constants (might remove expressions & their variables from user-view)
 
         Returns:
             Expression: the implication constraint or a BoolVal if simplified
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -141,7 +141,7 @@ class Expression(object):
         """
         self.name = name
         if not isinstance(arg_list, tuple):
-            warnings.warn(f"DEPRECATED: Argument list of {name} is not a tuple, updated the constructor!", UserWarning)
+            warnings.warn(f"DEPRECATED: Argument list of {name} is not a tuple, update the constructor!", UserWarning)
             arg_list = tuple(arg_list)
         self._args = arg_list
         self._has_subexpr = has_subexpr
@@ -193,7 +193,7 @@ class Expression(object):
         return hash(self.__repr__())
 
     def has_subexpr(self) -> bool:
-        """ Does it contains nested :class:`Expressions <cpmpy.expressions.core.Expression>` (anything other than a :class:`~cpmpy.expressions.variables._NumVarImpl` or a constant)?
+        """ Does it contain nested :class:`Expressions <cpmpy.expressions.core.Expression>` (anything other than a :class:`~cpmpy.expressions.variables._NumVarImpl` or a constant)?
             Is of importance when deciding whether certain transformations are needed
             along particular paths of the expression tree.
             Results are cached for future calls and reset when the expression changes
@@ -382,7 +382,7 @@ class Expression(object):
             return self
         return Operator("sum", [other, self])
 
-    # substraction
+    # subtraction
     def __sub__(self, other):
         # if is_num(other) and other == 0:
         #     return self
@@ -406,16 +406,16 @@ class Expression(object):
             return self
         return cp.Multiplication(other, self)
 
-    # matrix multipliciation TODO?
+    # matrix multiplication TODO?
     #object.__matmul__(self, other)
 
     # other mathematical ones
     def __truediv__(self, other):
-        warnings.warn("We only support floordivision, use // in stead of /", SyntaxWarning)
+        warnings.warn("We only support floordivision, use // instead of /", SyntaxWarning)
         return self.__floordiv__(other)
 
     def __rtruediv__(self, other):
-        warnings.warn("We only support floordivision, use // in stead of /", SyntaxWarning)
+        warnings.warn("We only support floordivision, use // instead of /", SyntaxWarning)
         return self.__rfloordiv__(other)
 
     def __floordiv__(self, other):
@@ -563,7 +563,7 @@ class BoolVal(Expression):
     
 
     def has_subexpr(self) -> bool:
-        """ Does it contains nested Expressions (anything other than a _NumVarImpl or a constant)?
+        """ Does it contain nested Expressions (anything other than a _NumVarImpl or a constant)?
             Is of importance when deciding whether certain transformations are needed
             along particular paths of the expression tree.
         """
@@ -574,7 +574,7 @@ class BoolVal(Expression):
 
         Args:
             other (ExprLike): the right-hand-side of the implication
-            simplify (bool): if True, simplify True/False constants (might remove expressions & there variables from user-view)
+            simplify (bool): if True, simplify True/False constants (might remove expressions & their variables from user-view)
 
         Returns:
             Expression: the implication constraint or a BoolVal if simplified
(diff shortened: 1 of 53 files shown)


cpmpy/expressions/core.py
```diff
@@ -233,35 +233,59 @@ def has_subexpr(self):
...
         if self.is_bool():
-            return 0, 1 #default for boolean expressions
+            return 0, 1  # default for boolean expressions
         raise NotImplementedError(f"`get_bounds` is not implemented for type {self}")
 
-    # keep for backwards compatibility
     def deepcopy(self, memodict={}):
+        """ DEPRECATED: use copy.deepcopy() instead
+
+        Will be removed in stable version.
```
IgnaceBleukx, line 253: Could probably be removed by now... We'll do a major release next time anyway

cpmpy/expressions/core.py
```diff
@@ -233,35 +233,59 @@ def has_subexpr(self):
...
-        if is_true_cst(other):
-            return BoolVal(True)
-        if is_false_cst(other):
-            return ~self
-        return Operator('->', [self, other])
+    def implies(self, other: ExprLike, simplify: bool = False) -> "Expression":
+        """Implication constraint: ``self -> other``.
+
+        Python does not offer relevant syntax for implication, call this method instead.
+        For double implication, use equivalence ``self == other``.
```
IgnaceBleukx: Double reification?

cpmpy/expressions/core.py
```diff
@@ -233,35 +233,59 @@ def has_subexpr(self):
...
+
+        Args:
+            other (ExprLike): the right-hand-side of the implication
+            simplify (bool): if True, simplify True/False constants (might remove expressions & there variables from user-view)
```
IgnaceBleukx: typo there -> their

IgnaceBleukx (approved):
Great, that should do away with a lot of weird cases in the transformation stack...
Maybe also add the optional argument to `expressions.utils.implies`? We use it when we're not sure the lhs is an expression

cpmpy/expressions/core.py
tias: I agree, but I think such removals should have a separate commit

cpmpy/expressions/core.py
```diff
@@ -233,35 +233,59 @@ def has_subexpr(self):
...
+    def is_bool(self) -> bool:
         """ is it a Boolean (return type) Operator?
             Default: yes
         """
         return True
 
-    def value(self):
-        return None # default
+    def value(self) -> Optional[int]:
+        return None  # default
```
IgnaceBleukx, line 243: Shouldn't this raise a NotImplemented error instead? It seems weird to have a default value for a method that every subclass absolutely should override...

cpmpy/expressions/core.py
```diff
@@ -233,35 +233,59 @@ def has_subexpr(self):
...
-        if is_true_cst(other):
-            return BoolVal(True)
-        if is_false_cst(other):
-            return ~self
-        return Operator('->', [self, other])
+    def implies(self, other: ExprLike, simplify: bool = False) -> "Expression":
+        """Implication constraint: ``self -> other``.
+
+        Python does not offer relevant syntax for implication, call this method instead.
+        For double reification (<->), use equivalence ``self == other``.
```
IgnaceBleukx, line 262: Choco has `expr.reify_with`, which we could also add maybe? That would just redirect to `==`
(Not for this PR, but smt to think about in the future)

cpmpy/expressions/core.py
```diff
@@ -233,35 +233,59 @@ def has_subexpr(self):
...
+
+        Args:
+            other (ExprLike): the right-hand-side of the implication
+            simplify (bool): if True, simplify True/False constants (might remove expressions & their variables from user-view)
```
IgnaceBleukx, line 266: "might **eliminate**..." is more accurate?

cpmpy/expressions/core.py
```diff
@@ -427,19 +451,19 @@ def __rpow__(self, other: Any):
...
 
     def __pos__(self):
         return self
 
     def __abs__(self):
         return cp.Abs(self)
 
-    def __invert__(self):
-        if not (is_boolexpr(self)):
+    def __invert__(self) -> "Expression":
```
IgnaceBleukx, line 463: So this `__invert__` is typed, because we call it in implies, right? And that one is also typed.
Next PR should probably type the other `__xxx__` methods?

cpmpy/expressions/core.py
```diff
@@ -539,27 +563,36 @@ def __rxor__(self, other):
...
+
+        Args:
+            other (ExprLike): the right-hand-side of the implication
+            simplify (bool): if True, simplify True/False constants (might remove expressions & their variables from user-view)
+
+        Returns:
+            Expression: the implication constraint or a BoolVal if simplified
 
+        Simplification rules:
+            - BoolVal(True) -> other :: other (BoolVal-ified if needed)
```
IgnaceBleukx, line 583: We can always apply this one, it does not remove user vars

cpmpy/expressions/utils.py
```diff
@@ -208,17 +209,23 @@ def get_bounds(expr):
             return int(expr), int(expr)
         return math.floor(expr), math.ceil(expr)
 
-def implies(expr, other):
+# first to are declarations for typing purposes only
```
IgnaceBleukx, line 212: "two"

IgnaceBleukx (approved):
Minor comments on docs, and one potential simplification in implies that we can always do, regardless of the `simplify` arg

cpmpy/expressions/core.py
tias: ah right, its also a standard Python one. I didn't mean to override standard Python ones, as they might have a signature that we should mimick, removing the type

cpmpy/expressions/core.py
tias: yes... but then the doc needs to be updated,

you are suggesting that we do:
- if self is true, we post other
- otherwise, if simplify then we remove other else we post the reification as is

while now, we do 'simplify' -> simplify it; 'no simplify' -> post as is

tias (commented):
see 2 remaining comments

cpmpy/expressions/core.py
IgnaceBleukx: Yes, I would be in favor of updating the docs and always doing the simplification.
It is in line with what we do for sum/prod etc, when there is a neutral element.
E.g., `x + 0` remains `x`, instead of writing a sum.

It also ties to #342

cpmpy/expressions/core.py
IgnaceBleukx: Type is still here, missing commit? : )

cpmpy/expressions/variables.py
```diff
@@ -773,8 +773,10 @@ def __xor__(self, other):
     def __rxor__(self, other):
         return self._vectorized(other, '__rxor__')
 
-    def implies(self, other):
-        return self._vectorized(other, 'implies')
+    def implies(self, other, simplify=False):
```
IgnaceBleukx: Type this one with `NDVarArray`?

cpmpy/tools/xcsp3/globals.py
```diff
@@ -962,9 +962,12 @@ def decompose(self):
             for t in range(num_tasks):
                 cons.append(ends[t] == end[t])
 
-            for j in range(num_tasks):
-                sj = start[j]
-                cons.append(capacity >= demand[j] + cp.sum([(start[i] <= sj) & (sj < ends[i]) for i in range(num_tasks) if i != j]))
+            for j, (start_j, demand_j) in enumerate(zip(start, demand)):
```
IgnaceBleukx: Not sure why this is changed in this PR? It does not seem to be related to the typing in core?

IgnaceBleukx: Just to comment on the change: we probably don't even want to support a constant demand in the DynamicCumulative, it is only used in XCSP3. And actually, we also select which decomposition to use dynamically in master now?

IgnaceBleukx (commented):
A few more things

cpmpy/expressions/core.py
tias: Done.

Getting the typing right makes me realize/fear we should introduce a stricter
`BoolExprLike: TypeAlias = Union["Expression", bool, np.bool_]  # subtype of ExprLike (bool subtype int)`

which should be a separate PR as it will affect all logical constraints/operators/functions?

IgnaceBleukx (approved):

cpmpy/expressions/core.py
IgnaceBleukx: Why would we need that? I would expect methods on expressions to always return an expression? Even if it is a BoolVal, it should be a CPMpy `Expression` object and not a Python/Numpy bool

tias: for user-facing functions like implies: can take a BoolExpr, can not take an int...

tias: (so for the argument, not the return type idd)

IgnaceBleukx: Aha yes I see; indeed...

Issue #444, opened 2024-01-12
2d-element constraint can have out-of-bounds indices


IgnaceBleukx:
The current implementation of the Element constraint has a bug:

```
import cpmpy as cp

iv = cp.intvar(0,10, shape=(3,3))
a,b = cp.intvar(0,10, shape=2)

m = cp.Model()
m += iv[a,b] == 3
```

this model can have the solution `a = 0,  b=4, iv[1,1] = 3` which is not what the user intended.
I think the index variables should be constrained to only take value within the bounds of the repsective axis.

Dimosts:
Again, we face the partial or total function dilemma.

If we want it as partial function, we can just check the indices when we convert it to 1-D, and throw an exception if they are not within the bounds.

If we want it as total function, we have to return constraints forcing the variables to be within the bounds along with the element constraint. This would also work in reified context.

Wout4:
Since the multidimensional case gets mapped to the one-dimensional, this will also be 'fixed' by #393 where we disallow partial functions

IgnaceBleukx:
Re-iterating on this issue. I think the correct way to do things is to make a new global constaint "MultiDimElement" or somtehing, and ensure the decomposition is used when a solver does not support multi-dim element. 
That allows to safen the expression too such as we do in #515 .

Wout4:
Ah yes we lose the dimensional information be converting to a 1D-element constraint. [0, 4] becomes the same as [1,1] in a 3x3 matrix since 0*3 + 4 == 1*3 + 1. MultiDimElement global does seem like the best option then

PR #910, opened 2026-04-07
typing with args of global functions (except multiplication, element)

tias:
skipped multiplication because of a float test that needs to be rewritten first

element because I expect some hard-to-type multidim related things that might be revised in #444 anyway

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -204,8 +204,8 @@ class Expression(object):
             return self._has_subexpr
         
         # micro-optimisations, cache the lookups
-        _NumVarImpl = cp.variables._NumVarImpl
-        _NDVarArray = cp.variables.NDVarArray
+        _NumVarImpl = cp.expressions.variables._NumVarImpl
+        _NDVarArray = cp.expressions.variables.NDVarArray
 
         # Initialize stack with direct access to private _args
         stack = list(self._args)
@@ -413,8 +413,7 @@ class Expression(object):
             raise TypeError("Power operator: modulo not supported")
         if not isinstance(other, (int, np.integer)):
             raise TypeError(f"Power operator requires a constant integer exponent, not: {other}")
-        if other == 0:
-            return 1
+        #if other == 0:  # do not simlify to 1, it might remove user variables from view
         if other == 1:
             return self
         return cp.Power(self, other)
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -413,7 +413,7 @@ class Expression(object):
             raise TypeError("Power operator: modulo not supported")
         if not isinstance(other, (int, np.integer)):
             raise TypeError(f"Power operator requires a constant integer exponent, not: {other}")
-        #if other == 0:  # do not simlify to 1, it might remove user variables from view
+        #if other == 0:  # do not simplify to 1, it might remove user variables from view
         if other == 1:
             return self
         return cp.Power(self, other)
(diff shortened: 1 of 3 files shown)


cpmpy/expressions/core.py
```diff
@@ -413,6 +413,8 @@ def __pow__(self, other: Any, modulo: Optional[int] = None):
             raise TypeError("Power operator: modulo not supported")
         if not isinstance(other, (int, np.integer)):
             raise TypeError(f"Power operator requires a constant integer exponent, not: {other}")
+        if other == 0:
```
IgnaceBleukx: We will remove user vars with this enabled

cpmpy/expressions/globalfunctions.py
```diff
@@ -283,13 +308,13 @@ def decompose(self) -> tuple[Expression, list[Expression]]:
...
         arg = self.args[0]
-        lb, ub = get_bounds(arg)
+        lb, ub = arg.get_bounds()
         if lb >= 0: # always positive
             return arg, []
         if ub <= 0: # always negative
             return -arg, []
 
-        _abs = intvar(*self.get_bounds())
+        _abs = intvar(*self.get_bounds())  # not just lb,ub, see implementation
```
IgnaceBleukx: "Not just lb,ub of arg"?

cpmpy/expressions/globalfunctions.py
```diff
@@ -472,24 +503,31 @@ def decompose(self):
...
+        if isinstance(y, Expression):
+            y_lb, y_ub = y.get_bounds()
+            if y_lb <= 0 <= y_ub:
+                defining.append(y != 0)
+                warnings.warn(f"Division constraint is unsafe, and will be forced to be total by this decomposition. If you are using {self} in a nested context, this is not valid, and you need to safen first using cpmpy.transformations.safening.no_partial_functions")
 
         r = intvar(*get_bounds(x % y))  # remainder
         _div = intvar(*self.get_bounds())
-        return _div, safen + [(x == (y * _div) + r), abs(r) < abs(y), abs(y) * abs(_div) <= abs(x)]
+        defining.append(Comparison("==", x, (y * _div) + r))
```
IgnaceBleukx: Just checking: this explicit `Comparsion` consruct is becase `__eq__` is not typed, right?

IgnaceBleukx (changes requested):
Yes, the `xxx_listlike` helper functions make a lot of sense!
I suppose we'll update the `.value() `funcs of Operator and GlobalConstraints in future PRs in a similar style?

One blocking comment on the addition in `__pow__`, otherwise looks good to go

cpmpy/expressions/globalfunctions.py
tias: lb,ub is defined 5 lines above; updated the comment

cpmpy/expressions/core.py
tias: hmm, yes..

its also tricky in decompose, because it needs an expression...

There I now return 0*base + 1... I could do the same here instead? I'm not sure; the current one is more explicit, it leaves the Pow there as the user wrote it (a semi-design principle we often respect)

cpmpy/expressions/globalfunctions.py
tias: kind-of (actually because 'x' can be constant, so it does not see that rhs is always Expr and so that it will be Expr), added a comment

tias:
> I suppose we'll update the `.value() `funcs of Operator and GlobalConstraints in future PRs in a similar style?

Yes, for future PRs...

tias:
spun-out min/max and rest outdated; better to redo starting from the helpers in master. Kept for reference in tracker #913

PR #912, opened 2026-04-07
Support Gurobi expression trees

hbierlee:
Initial draft for supporting expression trees. I've had to make many changes to flatten to prevent it from flattening expressions which are supported by the Gurobi expression trees. Now that the Gurobi side is working, I'm making a 2nd pass to make things more principled and stable again, as of course now the other solvers would break. However, it will remain difficult, as we no longer have Flat Normal Form for the later Gurobi transformations.


hbierlee:
z3 is good to look at. The challenge for gurobi is that some expressions should be flattened (e.g. max) but not other (e.g. times, pow,..). I definitely hope to simplify by a partial use of flatten.

Actually the changes to `flatten` are pretty small (although it requires knowing context in `normalized_num_expr`), it's all the transformations after that rely on FNF which are more annoying.

hbierlee:
I've added a `into_tree` transformation to avoid changes to the other transformations. Of course, this did lead to a lot of duplicated transformations from flatten/linearize.

What `into_tree` tries to do is essentially a limited form of flattening, namely, leave any supported node (e.g. for gurobi: pow, mul, sum, ..) but replace unsupported nodes (e.g. abs(x)) by a Boolean/integer reification. See test_gurobi for some examples of expected input/output/gurobi LP. Ignore changes to conftest/test_constraints, they will be removed in the final PR.

The current problem is primarily that a lot of behavior from flatten/linearize is duplicated, which would be unmaintainable. The question is of course whether it is possible to make gurobi expression tree work together with the existing code. An simpler example of this problem is `linearize_reified_variables` (not currently enabled). Calling it after `into_tree` won't work because we don't have FNF, so either it also has to be made recursive, or I have to duplicate its behaviour.

Another more minor problem is that `into_tree` has gurobi-specific additional steps in it, namely ensuring linearizing, adding indicator constraints are of the form `NegBoolView -> LinExpr CMP LinExpr` (with CMP in `<=,=,>=`) and general constraints of the form `y = f(x)` with `x` only variables. This should probably go into separate functions for clarity (although it will have to call `into_tree` recursively, e.g. if we have `abs(non_linear_expr)`, then we want `y=abs(z), z = into_tree(non_linear_expr)`).

Some additional things added which are not exactly in scope: support for the and/or gurobi general constraints.

cpmpy/expressions/globalconstraints.py
```diff
@@ -291,6 +291,14 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
             cons.append(cond.implies(cp.any([x == a for a in n]))) # equivalent to (x in n) | (y in n) | (x != y)
         return cons, []
 
+    def decompose_linear(self) -> tuple[list[Expression], list[Expression]]:
```
tias, line 294: please put in separate PR

hbierlee:
from offline, draft of expression tree (likely to be updated via edits):


```
INPUT 
⟨Model ⟩ ::= [ ⟨BoolExpr ⟩ (, ⟨BoolExpr ⟩)* ], minimize= ⟨IntExpr ⟩

⟨BoolExpr ⟩ ::=
| ⟨Bool ⟩
| ⟨BoolVar ⟩
| ¬ ⟨BoolExpr ⟩
| ⟨BoolExpr ⟩ ∧ | ∨ | ⇒ | ⇔ ⟨BoolExpr ⟩
| ⟨IntExpr ⟩ ≤ | < | ≥ | > | = | != ⟨IntExpr ⟩

⟨IntExpr ⟩ ::=
| ⟨Int⟩
| ⟨IntVar ⟩
| ⟨BoolExpr ⟩
| - ⟨IntExpr ⟩
| ⟨IntExpr ⟩ + | - | × ⟨IntExpr ⟩
| ⟨GlobalFunction⟩ [ ⟨Arg⟩ (, ⟨Arg⟩)* ]

⟨IntVar ⟩ ::= ⟨Int⟩ ⟨Int⟩ ⟨String⟩
⟨BoolVar ⟩ ::= ⟨String⟩

⟨Arg⟩ ::= ⟨BoolExpr ⟩ | ⟨IntExpr ⟩ | [ ⟨Arg⟩ (, ⟨Arg⟩)* ]
⟨GlobalFunction⟩ ::= Min | Max | Abs | Pow | Div

TARGET (using gurobi terms)

⟨Model ⟩ ::= [ ⟨Con ⟩ (, ⟨Con ⟩)* ], minimize= ⟨QuadExpr ⟩

⟨Var ⟩ ::=
| ⟨BoolVar ⟩
| ⟨IntVar ⟩

# note: no disequality or strict inequality
⟨Sense ⟩ ::= ≤ | ≥ | = 

⟨Con ⟩ ::=
| ⟨bool ⟩
| ⟨GenCon ⟩
| ⟨IndCon ⟩
| ⟨NodeExpr⟩ ⟨Sense⟩ ⟨NodeExpr⟩

# indicator constraint: note that the consequent should be a *flat* linear expression (of linear terms)
⟨IndCon ⟩  ::= ⟨BoolVar = 0|1 ⟩ ⇒ ⟨LinExpr ⟩ ⟨Sense ⟩ ⟨LinExpr ⟩
⟨LinExpr ⟩ ::= ⟨LinTerm ⟩ | ⟨LinTerm ⟩ + ⟨LinTerm ⟩
⟨LinTerm ⟩ ::= int | ⟨Var ⟩ | ⟨int ⟩ × ⟨Var ⟩

# general (and non-linear) constraints have to be posted in the form `y = f(x)`, where `x` is a list of variables
⟨GenCon⟩ ::= 
| ⟨IntVar ⟩ = ⟨Min | Max | Abs | Pow | Div ⟩ [ ⟨Var⟩ (, ⟨Var⟩)* ]
| ⟨BoolVar ⟩ = ⟨And | Or ⟩ [ ⟨BoolVar⟩ (, ⟨BoolVar⟩)* ]

⟨NodeExpr ⟩ ::= 
| ⟨NodeExpr⟩ ⟨+ | - | × | div | pow⟩ ⟨NodeExpr⟩
| ⟨Var⟩ 
| ⟨int⟩ 

# TODO not sure how to deduplicate this, only difference with a regular node is that we cannot do pow. Current version likely not completely safe 
⟨QuadExpr ⟩ ::= 
| ⟨QuadExpr⟩ ⟨+ | - | × | div⟩ ⟨QuadExpr⟩
| ⟨Var⟩ 
| ⟨int⟩ 
```

hbierlee:
pseudo-code from meeting with tias:

```
def gurobi_flatten(lst_expr: list[Expression], csemap...) -> list[Expression]
    todo: list[Expression] = []
    newlist: list[Expression] = []
    for c in lst_expr:
        if c.name == "and":
            # Simple general constraint compatible
            newargs, toplevel = _gurobi_flat_args(c.args, csemap)
            todo.extend(toplevel)
            cc = copy.copy(c)
            cc.update_args(newargs)
            newlist.append(cc)
        if c.name == "->":
            # Indicator compatible
            lhs, rhs == c.args
            lvar, lcons = csemap.get_or_make(lhs)
            todo.extend(lcons)
            # rhs, make-wsum... + flat_args?
            # special handling of var -> (expr == expr) and var -> (expr -> expr)?
        elif isinstance(c, Comparison):
            # special handling of != and < and >
            # and maybe also of var == (expr == expr) and var == (expr -> expr)?
            ...
        else:
            raise

    if len(todo) > 0:
      newlist += gurobi_flatten(todo)

    return newlist

def _gurobi_flat_args(args: Sequence[Any]) -> tuple[list[Any], list[Expression]]:
    # args can only be const/var
    toplevel: list[Expression] = []
    newargs: list[Any] = []
    for a in args:
    if const: append
    elif var: append
    else:
        newvar, cons = csemap.get_or_make(lhs)
        newargs.append(newvar)
        toplevel.extend(cons)

    return newargs, toplevel

def _gurobi_nested_args(args: Sequence[Any]) -> tuple[list[Any], list[Expression]]:
    # args can be anything but min/max/abs/CMP/->, make those a variable
    toplevel: list[Expression] = []
    newargs: list[Any] = []
    for a in args:
        if isinstance(a, GlobalFunction):
            if a.name in ["min","max","abs"]:
                newvar, cons = csemap.get_or_make(a)
                newargs.append(newvar)
                toplevel.extend(cons)
                continue
        elif a.name == "->":
            # get_or_make again...?
            continue
        elif isinstance(a, Comparison):
            # get_or_make again...?
            continue
        newargs.append(a)

    return newargs, toplevel
```

IgnaceBleukx:
Following our offline discussion: I'm wondering if we should "undo" flatten intead of skipping/reimplementing parts of flatten?
I.e., when you get the set of flat constraints, you could iterate over them and:
1. If you encounter a +, -, comparison.... that supports arbitrary nested child nodes, replace any auxiliary variable in expression with its original expression (you can figure this out using the csemap)
2. When you encounter something like a max or a logical constraint, just skip that one, we need its arguments to be flat

Would that work? And would it be more/less work than the current into_tree transformation?

tias:
In the sketched pseudo-code, the tricky part is the `expr -> (expr == expr)` and `expr == (expr == expr)` like parts... I think what we can do is for both, first make each expr 'flat_args' (e.g. define an aux var for each non-const-non-var arg), and then we can simply call the regular flatten on this new `flatexpr -> (flatexpr == flatexpr)` and `flatexpr == (flatexpr == flatexpr)`.

It will not recursively keep flattening, because there is nothing to recursively flatten... and the new toplevel, possibly nested, aux == expr will get the full recursive gurobi expression tree treatment...

hbierlee:
Little report on progress. This is only about the results of the current implementation, not about the above discussion on how to make this branch mergeable/maintainable (there's a lot of work there).

First the good news. As discussed offline, the current implementation is stable (passes the XCSP checker on the CSP25 instances), and produces fewer (nested) constraints more quickly. E.g. see below the transformation results for the first instance of each problem class of CSP22to25. Here, `b` means the current `b`ranch and `m` means `m`aster. Note it's sorted by the `tot_ratio` (ratio of number of constraints), then `t_ratio` (transform time ratio), where lower (<1.0) means this branch does better, so that instances go from good to bad for `b`:

```
                 problem  time_s_b  time_s_m  t_ratio    vars_b    vars_m  var_ratio   total_b   total_m  tot_ratio
         RamseyPartition     14.62    120.00     0.12  68250.00       NaN        NaN  90003.00       NaN        NaN --- timeout master 
             WordDesign1     21.16    120.00     0.18 175460.00       NaN        NaN 175895.00       NaN        NaN
               Fillomino     26.61    120.00     0.22 147402.00       NaN        NaN 150625.00       NaN        NaN
            Heterosquare     29.65    120.00     0.25 409124.00       NaN        NaN   7251.00       NaN        NaN
       PegSolitaireTable     38.72    120.00     0.32 301355.00       NaN        NaN 299063.00       NaN        NaN
               Crossword     45.10    120.00     0.38 213600.00       NaN        NaN 213109.00       NaN        NaN
    TilingRythmicCanons1     19.62     23.02     0.85 261873.00 266245.00       0.98  17853.00 170233.00       0.10 --- both transform, but fewer constraints for `b`, e.g. 10% of `m`
             LangfordBin      0.04      0.05     0.74    440.00    840.00       0.52     70.00    470.00       0.15
                Hadamard      0.01      0.04     0.25     68.00    340.00       0.20     78.00    350.00       0.22
             KnightTour2      4.18      9.89     0.42  31160.00  44507.00       0.70  11592.00  42043.00       0.28
        Superpermutation      0.01      0.04     0.29    189.00    351.00       0.54     82.00    244.00       0.34
FastMatrixMultiplication      0.07      0.37     0.18    788.00   1643.00       0.48    861.00   2379.00       0.36
                Eternity      1.38      3.88     0.36  12540.00   6600.00       1.90  10704.00  26088.00       0.41
             Pentominoes      0.45      0.82     0.56   6132.00   4896.00       1.25   2664.00   6372.00       0.42
      PythagoreanTriples      0.82      1.15     0.71   7189.00   5576.00       1.29   7557.00  17830.00       0.42
            PegSolitaire      3.27      3.29     0.99  13768.00  12495.00       1.10  16529.00  37483.00       0.44
      SquarePackingSuite      0.25      0.28     0.89   1470.00   2267.00       0.65   2226.00   4812.00       0.46
           CoveringArray      0.06      0.10     0.55    832.00    704.00       1.18    456.00    904.00       0.50
               Accordion      0.20      0.25     0.79    991.00   1551.00       0.64   1208.00   2288.00       0.53
                  Takuzu      9.19     14.52     0.63  59010.00  53220.00       1.11  60720.00 107250.00       0.57
           ChainReaction      0.06      0.08     0.79   1256.00    648.00       1.94    231.00    383.00       0.60
              QuasiGroup      1.21      2.77     0.43   8768.00  12864.00       0.68  12616.00  20808.00       0.61
                  Molnar      0.00      0.01     0.50     26.00     30.00       0.87     36.00     56.00       0.64
          NurseRostering      3.17      5.78     0.55  26910.00  12660.00       2.13  23525.00  34575.00       0.68
        SportsScheduling      0.46      0.48     0.96   2988.00   2204.00       1.36   1951.00   2735.00       0.71
              Ortholatin      0.16      0.27     0.61   2200.00   1575.00       1.40   1580.00   2205.00       0.72
     SolitaireBattleship      4.63      4.13     1.12  31512.00  13802.00       2.28  24349.00  33800.00       0.72
                Nonogram      4.43      8.25     0.54  45081.00  17140.00       2.63  34571.00  47340.00       0.73
                   Slant      0.50      0.30     1.66  12394.00   1789.00       6.93   2619.00   3556.00       0.74
      RotatingWorkforce2      1.72      1.20     1.44  11279.00   5649.00       2.00  10283.00  13661.00       0.75
                    SEDF      8.82     10.97     0.80  48240.00  42354.00       1.14  62148.00  80238.00       0.77
             CostasArray      0.78      1.41     0.56   6494.00   6272.00       1.04   9692.00  12486.00       0.78
      AnotherMagicSquare      0.00      0.00     1.00     32.00     40.00       0.80     28.00     36.00       0.78
                    EFPA      0.71      0.73     0.97   6497.00   6007.00       1.08   7187.00   9074.00       0.79
          MisteryShopper      1.38      2.08     0.66  14545.00  12429.00       1.17  10900.00  13644.00       0.80
      NumberPartitioning      0.02      0.02     0.89    420.00    440.00       0.95     83.00    103.00       0.81
           Cryptanalysis      1.70      1.98     0.86  11740.00  11180.00       1.05  18429.00  22165.00       0.83
            CalvinPuzzle      0.06      0.16     0.39    852.00    852.00       1.00    404.00    480.00       0.84
        SolitairePattern     13.17     18.68     0.71 100230.00  53373.00       1.88 103966.00 116019.00       0.90
             DiamondFree      2.51      3.67     0.68   5592.00   5679.00       0.98  35540.00  39107.00       0.91
                     FRB      2.88      4.04     0.71   6670.00   4495.00       1.48  37396.00  39218.00       0.95
         AverageAvoiding      0.35      0.14     2.53   1561.00   1578.00       0.99   2342.00   2359.00       0.99
             AlmostMagic      0.03      0.10     0.35    884.00    884.00       1.00    189.00    189.00       1.00 --- same number of constraints, `b` faster
             HyperSudoku      0.07      0.15     0.46    810.00    810.00       1.00    441.00    441.00       1.00
           BlockedQueens      0.22      0.29     0.76   2541.00   2541.00       1.00   2786.00   2786.00       1.00
                Coloring      0.85      0.99     0.86   6164.00   6164.00       1.00  11432.00  11432.00       1.00
                  Hidato     13.71     15.65     0.88  93763.00  50243.00       1.87  87445.00  87445.00       1.00
               CrazyFrog    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN --- both TF timeout
                Dominoes    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN
                 Fischer    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN
             MagicSquare    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN
                RoomMate    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN
               Rostering    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN
       RotatingRostering    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN
          Subisomorphism    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN
              WordSquare    120.00    120.00     1.00       NaN       NaN        NaN       NaN       NaN        NaN
                 PseudoB      0.32      0.21     1.51   3575.00   3575.00       1.00   3707.00   3707.00       1.00 --- same number of constraints, `m` faster
                  Primes      0.01      0.00     1.75     91.00     91.00       1.00     40.00     40.00       1.00
       NonTransitiveDice      0.23      0.07     3.04    488.00    487.00       1.00    914.00    913.00       1.00
                 Crypto1     25.46     32.04     0.79 191298.00 189458.00       1.01 220792.00 218952.00       1.01 --- `m` yields fewer constraints
           LotteryDesign      0.17      0.19     0.92   1126.00    853.00       1.32   1860.00   1839.00       1.01
            BinaryPuzzle      2.94      1.13     2.60  32640.00  17040.00       1.92  34100.00  33700.00       1.01
         AntimagicSquare      0.02      0.03     0.61    422.00    420.00       1.00    100.00     98.00       1.02
                 Hamming      0.81      1.70     0.47   6489.00   4779.00       1.36  10507.00   9747.00       1.08
                  Soccer      0.63      1.35     0.47   7023.00   4220.00       1.66   7463.00   6917.00       1.08
           GracefulGraph      0.13      0.28     0.49   3288.00   3243.00       1.01    364.00    319.00       1.14
        PoolBallTriangle      0.01      0.01     1.20    260.00    250.00       1.04     76.00     66.00       1.15
            AztecDiamond      1.24      0.61     2.02  16500.00   6300.00       2.62  13900.00  11300.00       1.23
           SocialGolfers      2.21      1.10     2.00  15505.00   5605.00       2.77  20301.00  15077.00       1.35
           CarSequencing     22.76      7.85     2.90   6600.00  17400.00       0.38  82685.00  50285.00       1.64

Mean time ratio (branch/master): 0.89  --- PAR2 applied, so 11% faster for `b`
Mean total constr ratio: 0.78  --- no PAR2 on number of constraints, so the difference is >22%
Median time ratio: 0.79
Median total constr ratio: 0.79
Branch better (tot<1): 36, Master better (tot>1): 12, Equal: 7
```

Note that the aggregated mean/median ratios on constraints at the end do not account for the TF timeouts on `m` (so the difference is bigger than 22%). I do PAR2 for transformation time. There are also cases where `b` does badly, e.g. CarSequencing, where I can see the issue and it would be easily fixable. So, for these instances (generally a small one from each problem class), we see good improvement for transformation time and number of constraints, but we do see that the number of variables often seems higher. Perhaps because `b` applies the direct encoding more aggressively (whenever it sees `x == v` or `x != v`).

However, strangely, when running all instances of CSP25 on the compute server on `m` and `b` (single-threaded, same seed), while `b` indeed does transform/post 4 seconds faster on average, solving is much slower, so that the total time namely 5 seconds slower on average. While the number of solved is the same (44/200 for both), this is a sizable gap. There could be some engineering issue with the benchmarks (in fact, there are instances where `m` solves but `b` timeouts in this experiment, but on my laptop `b` solves `m` times out!). It would also be more interesting to all CSP22to25 instances, and way more interesting to see COP22to25 (but running those has a separate benchmarking issue). A more interesting explanation could be that Gurobi works better with many linear constraints rather than few nested constraints.

tias:
Thanks for the detailed analysis Henk.

In nested form, some instances can be posted that could not otherwise. I think that alone is a good take-away.

But for that we need a clean maintainable implementation. Something for somebody to pick up later; for now I'm closing this experimental branch.

PR #916, opened 2026-04-08
Push down negation before decompose

IgnaceBleukx:
Re-order the transformations such that negation is pushed down into the arguments of a "not" operator before we call decompose.

This allows to implement efficient implementations of negated global constraints (e.g., `~Table` becomes Negtable).
We can also do this for some more (e.g., ~AllDiff(x) becomes NValue(x) < len(x)). We should probably do some small experiments to see what is more efficient? Also depends on the decomp of NValue of course if that one is not supported...

Downside is that no transformation can explicitly introduce negation, and needs to call `recurse_negation` themself. This turned out to be simpler to implement than I first thought.

Efficiency-wise, there might be some gains because we now have one transformation less to be called in each recursive`flatten_constraint` call. I then also noticed we call simplify_boolean in each recursive flatten call, I've added a flag to disable it when we can. Not sure if this should remain in this PR


tias:
you will have to rebase/merge master in I guess, now that push down negation rewrite is in (in a different form)

cpmpy/expressions/globalconstraints.py
```diff
@@ -68,6 +68,7 @@ def decompose(self):
     ..
 
     You can also implement a `.negate()` method if the global constraint has a better way to negate it than negating the decomposition.
+    The expression returned by `.negate()` should be equivalent to the negation of the global constraint, and should introduce any `not` operators.
```
tias: should 'not', or even stronger: is not allowed to introduce any 'not' operators...

tias: also maybe more concretely, 'instead, use SOME_FUNC if you want to negate an expression'

cpmpy/transformations/flatten_model.py
```diff
@@ -122,7 +122,7 @@ def flatten_model(orig_model, csemap=None):
             return cp.Model(*basecons, maximize=newobj)
 
 
-def flatten_constraint(expr, csemap=None):
+def flatten_constraint(expr, csemap=None, do_simplify=True):
```
tias, line 125: explain arguments in doc pls

tias (commented):
almost there.

I guess you checked that no global cons/func introduces a negation, even on the latest master?

Also, we since have 'scip' as a solver, can you check that all solvers on master that need changing are changed?

IgnaceBleukx:
Processed the review, there were indeed a couple of cases where the decomposition/negation of a global might introduce negation.

Updated scip and docs

IgnaceBleukx:
Also updated the newly added highs solver, should be ready to go in after tests complete?

cpmpy/transformations/decompose_global.py
```diff
@@ -113,6 +115,13 @@ def decompose_in_tree(lst_of_expr: list[Expression],
                 changed = True
                 if len(arg_toplevel) > 0:
                     todolist.extend(arg_toplevel)
+                    
+                if expr.name == "not": # cannot leave negation here, push down in the arguments of the decomposition
```
tias: if we pushed down negation before, and decompose can not contain negation, then there can be no negation here?

or, if we do catch remaining negations here, then we do not have the strict requirment that decompose's can not contain negation?

cpmpy/transformations/flatten_model.py
```diff
@@ -426,10 +432,10 @@ def normalized_boolexpr(expr, csemap=None):
             (lhs,lcons) = get_or_make_var(expr.args[0], csemap=csemap)
             # TODO, optimisation if args1 is an 'or'?
             (rhs,rcons) = get_or_make_var(expr.args[1], csemap=csemap)
-            return ((~lhs | rhs), lcons+rcons)
+            return ((recurse_negation(lhs) | rhs), lcons+rcons)
         if expr.name == 'not':
             flatvar, flatcons = get_or_make_var(expr.args[0], csemap=csemap)
-            return (~flatvar, flatcons)
+            return (recurse_negation(flatvar), flatcons)
```
tias: we know its a var here, ~flatvar was fine?

cpmpy/transformations/flatten_model.py
```diff
@@ -461,7 +467,7 @@ def normalized_boolexpr(expr, csemap=None):
                 (lhs, lcons) = get_or_make_var(lexpr, csemap=csemap)
                 if expr.name == '!=' and rvar.is_bool():
                     # != not needed, negate RHS variable
-                    rvar = ~rvar
+                    rvar = recurse_negation(rvar)
```
tias: also here we know its a var?

tmp/ite.py
```diff
@@ -0,0 +1,18 @@
+import cpmpy as cp
```
tias: this should be removed from the commit

cpmpy/transformations/decompose_global.py
IgnaceBleukx: expr here is the original expression; so if we had `~AllDifferent(x,y,z)`, expr would be 'not', and the new_args would be `and(x != y, x != z, y != z)`, which we would have to negate.

I've renamed the variables now to make this more clear

IgnaceBleukx:
Ah yes there were idd some parts we can just keep the `~var` part, fixed now.

cpmpy/transformations/decompose_global.py
tias: I don't get it yet...

so negate leaves `~AllDiff`? What else does it leave with a toplevel ~?

and here we are in decompose, and if we encounter ~Smth, we will call `recurse_negate(Smth)`, but in the case of ~AllDiff, our decompose is counting on  `recurse_negate()` to do the treatment of the global...

Indeed, I can see that `recurse_negate()` will check for a global and then walk its args and call `Glob.negate()` on it. But if it already has that logic, why did we not use it when doing push_down_negation??


E.g. I would expect either:

1. `push_down_negate` does not push negations over globals; and `decompose_global` is left to do that, so it will check the ~Glob case and itself handle the pushing down

2. `push_down_negate` also pushes negation over globals; then `decompose_global` can assume the input is negation-free

cpmpy/transformations/decompose_global.py
```diff
@@ -282,10 +291,16 @@ def _decompose_in_tree_args(args: list[Any]|tuple[Any, ...],
                     rec_changed, rec_newargs, rec_toplevel = _decompose_in_tree_args(arg.args, supported=supported, supported_reified=supported_reified, csemap=csemap, decompose_custom=decompose_custom)
                     if rec_changed:
                         changed = True
-                        arg = copy.copy(arg)
-                        arg.update_args(rec_newargs)
                         if len(rec_toplevel) > 0:
                             toplevel.extend(rec_toplevel)
+
+                        if arg.name == "not": # cannot leave negation here, push down in the arguments of the decomposition
```
tias: we should first resolve the other discussion about 'not'.

After resolving that, we should note that there is an ignored case here where `arg.name == not` but `arg.has_subexpr=False` and then this not would remain untreated.

cpmpy/transformations/decompose_global.py
IgnaceBleukx: So `push_down_negation` eliminates all negation operators, apart from ~Global and ~Boolvar.
If the solver supports ~Global, nothing happens and `arg_changed` will be false here.

If the solver does not support the global in negated (i.e,, nested) context, we will decompose it. But, as all future tranformrations assume there is no negation anymore, we need to push the negation into the decomposition of the global. We need to do it here, instead of during `push_down_negation` as otherwise we don't know yet whether the solver supports the global nested or not.

PR #917, opened 2026-04-09
Custom and typed CSEMap object

IgnaceBleukx:
Implementation of a custom CSEMap object, with proper typing.
For now, there is no normalization happening yet (e.g., we probably want to save `bv <-> var != val`  as `~bv <-> var == val` as it allows us to extract the "predicate" and linearize it using the domain encoding).
Not sure if that should be part of this PR as well, or if we should focus on getting the "bare" object properly typed and up and running.

For typing: I was not able to properly type the csemap passed around in the transformations; there seemed to be a lot of issues popping up (e.g., where it was unclear if the returned expression is Boolean or not). 
For now, decompose_in_tree and friends have the argument `Optional[CSEMap] = None`, for the others, it is untyped.
Do we want complete typing in this PR for all transformations? I would be open to that at the risk of becoming a mega PR...

After discussions in #837, we decided that decompose_in_tree does something special with the csemap as it maps a global to its decomposition.
I ran some small tests on xcsp3 whether this is indeed useful, and apparently it is : ) There are quite a few repeated global functions (e.g., div/mul) used in intention constraints, for which no explicit auxiliary variable is made. Hence, we decompose it multiple times if the solver does not support it.
So there is now a special interface for saving global constraints and functions to their decompositions.

### DIFF
--- a/cpmpy/solvers/solver_interface.py
+++ b/cpmpy/solvers/solver_interface.py
@@ -27,6 +27,7 @@ from enum import Enum
 from ..exceptions import NotSupportedError
 from ..expressions.core import Expression, ListLike
 from ..expressions.variables import _NumVarImpl
+from ..transformations.cse import CSEMap
 from ..transformations.get_variables import get_variables
 from ..expressions.utils import is_any_list, argvals
 from ..expressions.python_builtins import any
@@ -87,7 +88,7 @@ class SolverInterface(object):
         # initialise variable handling
         self.user_vars = set()  # variables in the original (non-transformed) model
         self._varmap = dict()  # maps cpmpy variables to native solver variables
-        self._csemap = dict()  # maps cpmpy expressions to solver expressions
+        self._csemap = CSEMap()  # maps cpmpy expressions to solver expressions
 
         # rest uses own API
         if cpm_model is not None:
--- a/cpmpy/solvers/solver_interface.py
+++ b/cpmpy/solvers/solver_interface.py
@@ -88,7 +88,7 @@ class SolverInterface(object):
         # initialise variable handling
         self.user_vars = set()  # variables in the original (non-transformed) model
         self._varmap = dict()  # maps cpmpy variables to native solver variables
-        self._csemap = CSEMap()  # maps cpmpy expressions to solver expressions
+        self._csemap = CSEMap()  # maps cpmpy expressions to auxiliary variables
 
         # rest uses own API
         if cpm_model is not None:
--- a/cpmpy/solvers/solver_interface.py
+++ b/cpmpy/solvers/solver_interface.py
@@ -88,7 +88,7 @@ class SolverInterface(object):
         # initialise variable handling
         self.user_vars = set()  # variables in the original (non-transformed) model
         self._varmap = dict()  # maps cpmpy variables to native solver variables
-        self._csemap = CSEMap()  # maps cpmpy expressions to auxiliary variables
+        self._csemap = CSEMap()  # maps cpmpy expressions to previously created expressions (typically auxiliary variables)
 
         # rest uses own API
         if cpm_model is not None:
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -5,8 +5,10 @@ from ..expressions.variables import boolvar, intvar, _IntVarImpl
 
 
 class CSEMap:
-    csemap = dict[Expression, _IntVarImpl]()
-    decomp_map = dict[Expression, Expression]()
+
+    def __init__(self):
+        self.csemap = dict[Expression, _IntVarImpl]()
+        self.decomp_map = dict[Expression, Expression]()
 
     # pass special methods to internal csemap
     def __len__(self):
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -4,13 +4,13 @@ from ..expressions.utils import is_int
 from ..expressions.variables import boolvar, intvar, _IntVarImpl
 
 
-class CSEMap:
+class flat_map:
 
     def __init__(self):
         self.flat_map = dict[Expression, _IntVarImpl]()   # map expression to variable filled during flattening
         self.decomp_map = dict[Expression, Expression]()  # map global constraint/function to its decomposition
 
-    # pass special methods to internal csemap
+    # pass special methods to internal flat_map
     def __len__(self):
         return len(self.flat_map)
 
@@ -18,7 +18,7 @@ class CSEMap:
         return self.flat_map[expr]
 
     def __setitem__(self, attr, val):
-        raise ValueError("__setitem__ is not supported for csemap, use get_or_make_var instead")
+        raise ValueError("__setitem__ is not supported for flat_map, use get_or_make_var instead")
 
     def get(self, expr: Expression) -> Optional[_IntVarImpl]:
         try:
@@ -35,7 +35,7 @@ class CSEMap:
         return self.decomp_map.get(expr)
 
     def get_reified_predicates(self) -> dict[_IntVarImpl, list[tuple[int, _IntVarImpl]]]:
-        """collect all bv <-> var == val expressions in csemap"""
+        """collect all bv <-> var == val expressions in flat_map"""
         
         var_vals = dict[_IntVarImpl, list[tuple[int, _IntVarImpl]]]()  # var: [val, bv]
         for expr, bv in self.flat_map.items():
@@ -49,14 +49,14 @@ class CSEMap:
     def get_or_make_var(self, expr: Expression) -> tuple[_IntVarImpl, Optional[Expression]]:
         """
         Make an auxiliary variable for the given expression
-        
+
         Arguments:
             expr: Expression to make an auxiliary variable for
 
         Returns:
             tuple[_IntVarImpl, Optional[Expression]]: (variable, equality constraint)
             - variable: the auxiliary variable for the expression
-            - equality constraint: the equality constraint between the expression and the variable, or None if the expression is already in the csemap
+            - equality constraint: the equality constraint between the expression and the variable, or None if the expression is already in the flat_map
         """
 
         if isinstance(expr, _IntVarImpl):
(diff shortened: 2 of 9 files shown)


cpmpy/transformations/cse.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def save_decomposition(self, expr: Expression, newexpr: Expression):
+        """Save the decomposition of the given global constraint or global function."""
+        self.decomp_map[expr] = newexpr
+
+    def get_decomposition(self, expr: Expression) -> Optional[Expression]:
+        """Get the decomposition of the given global constraint or global function."""
+        return self.decomp_map.get(expr)
+
+    def get_reified_equalities(self) -> dict[_IntVarImpl, list[tuple[int, _IntVarImpl]]]:
```
tias: its more specific then reified equalities... get_direct_encodings?

cpmpy/transformations/cse.py
```diff
@@ -0,0 +1,63 @@
...
+        var_vals = dict[_IntVarImpl, list[tuple[int, _IntVarImpl]]]()  # var: [val, bv]
+        for expr, bv in self.csemap.items():
+            if expr.name == "==":
+                var, val = expr.args
+                if isinstance(var, _IntVarImpl) and is_int(val):
+                    var_vals.setdefault(var, []).append((val, bv))
+
+        return var_vals
+
+    def get_or_make_var(self, expr: Expression) -> tuple[_IntVarImpl, list[Expression]]:
```
tias: you always create a single Comparison, so return tuple[IV, Expression] instead of wrapping it in a list?

cpmpy/transformations/flatten_model.py
```diff
@@ -336,41 +337,32 @@ def get_or_make_var(expr, csemap=None):
...
-
-        # save expr in dict
-        if csemap is not None:
-            csemap[expr] = ivar
-        return ivar, [flatexpr == ivar] + flatcons
-
-def get_or_make_var_or_list(expr, csemap=None):
+        new_var, _new_cons = csemap.get_or_make_var(expr)
+
+    # TODO: this seems weird, we are dismissing the `new_var == expr` made by the csemap
```
tias: that is weird yes. the [flatexpr == new_var] IS the returned _new_cons and should be used instead.

But going a bit deeper:

what the original code does was first check the map, then normalize the expression and then put it in the map (we should have re-checked the map after normalization I now realize).

I think what we want is:
1. first check if its in the map, if so, return
2. if not, then flatten it (only if has_subexpr I think)
3. check if the flat version is in the map, if so put original expr in the map too and return
4. if none: call csemap get or make; also put the original expr in the map, return newvar, and comparison + flatcons

cpmpy/solvers/solver_interface.py
```diff
@@ -87,7 +88,7 @@ def __init__(self, name="dummy", cpm_model=None, subsolver=None):
         # initialise variable handling
         self.user_vars = set()  # variables in the original (non-transformed) model
         self._varmap = dict()  # maps cpmpy variables to native solver variables
-        self._csemap = dict()  # maps cpmpy expressions to solver expressions
+        self._csemap = CSEMap()  # maps cpmpy expressions to solver expressions
```
tias: the comment is wrong

tias (changes requested):
Great.

Indeed, no normalization yet, that is a separate functionality.

Good to type only the parts we touch here and not make it a Mega-PR. We can fix the other type-aware issues in follow-up.

Smart choice to put the 'map' of decompose separately in this object too, it is indeed still about CSE just not wrt creating variables... perhaps we should give the naming some more thought? e.g. both do Common Subexpr Elim...

- decomp_map does Common Decomposition Elimination (map Expr -> ExprDecomp)
- csemap does Common ?Equality? Elimination (map Expr -> Var with Var=Expr)  damn its hard to give it a name... equality? flatten? variable? 

so maybe rename the latter to equality_map or flat_map or var_map?

cpmpy/transformations/cse.py
IgnaceBleukx: I was thinking about that too, but when the expression is already in the map, we can return `[]` this way, otherwise it would have to be None, which would then need to be processed by the caller again

cpmpy/transformations/flatten_model.py
IgnaceBleukx: I have re-implemented this part of get_or_make_var to follow the four steps you outlined.
So we now make a new variable for the flat expression and post `flatexpr == IV`, and we also save `orig_expr : IV` in the map, so that if we encounter it again later on, we don't flatten again and do an early exit

IgnaceBleukx: Only recursing when `has_subexpr` is true does not work as `normalized_numexpr` also does some normalization (as the name suggests). E.g., `- var` is translated to a weighted sum, but `- var` does not have any subexpressions.

cpmpy/transformations/cse.py
tias: I think Optional[Expression] makes sense though...

its a good pattern, when getting a list that is possibly empty, to do:
if len(lst) > 0:
  smth.extend(lst)

so then there is a check in the caller anyway (and it only calls if it has to). So from that pattern's perspective, it would be
if res is not None:
  smth.append(res)

with just an append...?

cpmpy/transformations/cse.py
```diff
@@ -0,0 +1,64 @@
...
+from ..expressions.variables import boolvar, intvar, _IntVarImpl
+
+
+class CSEMap:
+
+    def __init__(self):
+        self.flat_map = dict[Expression, _IntVarImpl]()   # map expression to variable filled during flattening
+        self.decomp_map = dict[Expression, Expression]()  # map global constraint/function to its decomposition
+
+    # pass special methods to internal csemap
```
tias: csemap -- flat_map

cpmpy/transformations/cse.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def save_decomposition(self, expr: Expression, newexpr: Expression):
+        """Save the decomposition of the given global constraint or global function."""
+        self.decomp_map[expr] = newexpr
+
+    def get_decomposition(self, expr: Expression) -> Optional[Expression]:
+        """Get the decomposition of the given global constraint or global function."""
+        return self.decomp_map.get(expr)
+
+    def get_reified_predicates(self) -> dict[_IntVarImpl, list[tuple[int, _IntVarImpl]]]:
```
tias: still not convinced of the name... predicates would imply all bv <-> IntExpr == val
but here its really about var == val?

cpmpy/transformations/decompose_global.py
```diff
@@ -171,40 +174,48 @@ def _decompose_in_tree(lst_of_expr: ListLike[Any],
...
                             if rec_changed:
                                 newexpr = rec_lst_newexpr[0]
                                 toplevel.extend(rec_toplevel)
+                            else:
+                                newexpr = decomp
 
                         if (csemap is not None):
-                            csemap[expr] = newexpr
+                            csemap.save_decomposition(expr, newexpr)
+                    elif isinstance(decomp, list): # retrieved list of expr from cache above
```
tias: the signature says that it returns a single Expr? you are indeed storing the newexpr, which is cpm_all'd for globalcons?

cpmpy/transformations/cse.py
tias: you call it as "var_vals = csemap.get_reified_predicates()"

so... `get_reified_varvals()`?

tias (changes requested):
just a few more things to get it right

cpmpy/transformations/cse.py
IgnaceBleukx: Ah yes I see, question remains then if we also change the return type of `flatten_constraint.get_or_make_var`, or do we keep that one unchanged for now? 
I would be in favor of keeping that one as is, and changing it in the future rewrite of flatten?

cpmpy/transformations/decompose_global.py
IgnaceBleukx: Yes, this case is indeed redundant, removed

cpmpy/transformations/cse.py
IgnaceBleukx: Yes, reified_varvals makes sense, and nicely matches the docstring just below : )

IgnaceBleukx:
I think the remaining comments are resolved now?

cpmpy/transformations/cse.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def save_decomposition(self, expr: Expression, newexpr: Expression):
+        """Save the decomposition of the given global constraint or global function."""
+        self.decomp_map[expr] = newexpr
+
+    def get_decomposition(self, expr: Expression) -> Optional[Expression]:
+        """Get the decomposition of the given global constraint or global function."""
+        return self.decomp_map.get(expr)
+
+    def get_reified_varvals(self) -> dict[_IntVarImpl, list[tuple[int, _IntVarImpl]]]:
```
tias, line 37: I think this is more precisely a _BoolVarImpl (the last argument)?

cpmpy/transformations/cse.py
```diff
@@ -0,0 +1,75 @@
...
+    def __getitem__(self, expr: Expression) -> _IntVarImpl:
+        return self.flat_map[expr]
+
+    def __setitem__(self, attr, val):
+        raise ValueError("__setitem__ is not supported for flat_map, use get_or_make_var instead")
+
+    def get(self, expr: Expression) -> Optional[_IntVarImpl]:
+        try:
+            return self.flat_map[expr]
+        except KeyError:
```
tias, line 26: if I'm not mistaken then dicts also implement 'get' but with an addition 'default':Any argument which is what it returns if the key does not exist?

so `self.flat_map.get(expr, None)` or even better, also give yourself (and get_decomp) the default arg and use that?

tias (approved):
because its so central I had another thorough look and found 2 things that could be more precise

IgnaceBleukx:
Made some final typing-related tweaks to the gettters in the csemap.
It now uses the overload to indicate what the return type is, depending on given parameters (default is None or not).

tias (approved):
made minor tweaks, looking forward to see it merged!

PR #919, opened 2026-04-15
Fix warnings in testsuite

tias:
as noticed in the CI output

### DIFF
--- a/cpmpy/solvers/cplex.py
+++ b/cpmpy/solvers/cplex.py
@@ -94,20 +94,16 @@ class CPM_cplex(SolverInterface):
         try:
             import docplex.mp as domp
         except ModuleNotFoundError as e:
-            warnings.warn(f"CPM_cplex: Could not import docplex: {e}")
             return False
         try:
             import cplex
             return True
         except ModuleNotFoundError as e:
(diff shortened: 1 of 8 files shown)


cpmpy/solvers/cplex.py
```diff
@@ -94,20 +94,16 @@ def installed():
...
             return True
         except ModuleNotFoundError as e:
-            warnings.warn(f"CPM_cplex: Could not import cplex: {e}")
             return False
 
     @staticmethod
     def license_ok():
         if not CPM_cplex.installed():
-            warnings.warn(
-                f"License check failed, python package 'docplex' or 'cplex' is not installed! Please check 'CPM_cplex.installed()' before attempting to check license.")
```
OrestisLomis, line 110: This warning is there for other licensed solvers (gurobi, cpo, hexaly) as well. Why are we removing it just for cplex?

OrestisLomis (approved):
Looks good to me, just a small comment regarding the warnings in cplex

cpmpy/solvers/cplex.py
tias: ah, I shouldn't have then, thx for checking, will put it back.

PR #922, opened 2026-04-16
Upgrade pindakaas -> 4.2

hbierlee:
In the upcoming version of pindakaas, the crash wrt to cadical + BVA should be fixed, meaning we can re-enable BVA again running Cadical with default parameters. Lower bound is increased because older versions of pindakaas are not safe when solving incrementally.

- **Re-enable BVA**
- **Upgrade to anticipated version**


hbierlee:
Ah I already had #869 for this.

PR #924, opened 2026-04-17
MDD global constraint

WoutPiessens:
Continuing on what was discussed in #891, I added an MDD global constraint, using a flow decomposition which seems to be fully reifiable according to the tests. 

A sidenote: making the flow decomposition fully reifiable did require some adjustments to the original decomposition (which was only positively reifiable), and as a consequence it introduces many additional variables to the decomposition. Another consequence is that some post-processing simplifications are no longer possible. This motivates the approach we discussed earlier, to offer a positive decomposition for certain global constraints where possible.

Next steps will be:
1)	Implement _reduce(), allowing equivalent nodes in the MDD to be merged before decomposition. 
2)	Add the Table linear decomposition using MDD.
3)	Work on positive decompositions.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -112,6 +112,7 @@
         ShortTable
         NegativeTable
         Regular
+        MDD
         IfThenElse
         InDomain
         Xor
@@ -133,6 +134,7 @@
 
 """
 import warnings
+from collections import defaultdict
 from typing import cast, Literal, Optional, Iterable, Any, TYPE_CHECKING
 import numpy as np
 
@@ -141,7 +143,7 @@ import cpmpy as cp
 from ..exceptions import TypeError
 from .core import Expression, BoolVal, ExprLike, ListLike
 from .variables import cpm_array, intvar, boolvar, _BoolVarImpl, _IntVarImpl, NDVarArray
-from .utils import all_pairs, is_int, is_bool, STAR, get_bounds, argvals, is_any_list, flatlist, is_num, is_boolexpr, implies
+from .utils import all_pairs, is_int, is_bool, STAR, get_bounds, argvals, is_any_list, flatlist, is_num, is_boolexpr, implies, argval
 
 if TYPE_CHECKING:
     from cpmpy.solvers.solver_interface import SolverInterface
@@ -824,6 +826,151 @@ class Regular(GlobalConstraint):
             else:
                 return False
         return curr_node in accepting
+
+
+class MDD(GlobalConstraint):
+    """
+    MDD-constraint: an MDD (Multi-valued Decision Diagram) is an acyclic layered graph starting from a single node and
+    ending in one. Each path of the MDD corresponds to a solution.
+    The values of the variables in 'array' correspond to a path in the MDD formed by the transitions in 'transitions'.
+    The root node is the first node used as a start in the first transition (i.e. transitions[0][0])
+    spec:
+        - array: an array of CPMpy expressions (integer variable, global functions,...)
+        - transitions: an array of tuples (nodeID, int, nodeID) where nodeID is some unique identifier for the nodes
+        (int or str are fine)
+    Example:
+        The following transitions depict a 3 layer MDD, starting at 'r' and ending in 't'
+        ("r", 0, "n1"), ("r", 1, "n2"), ("r", 2, "n3"), ("n1", 2, "n4"), ("n2", 2, "n4"), ("n3", 0, "n5"),
+        ("n4", 0, "t"), ("n5", 1, "t")
+        Its graphical representation is:
+                  r
+              0/ |1  \2     X
+            n1   n2   n3
+            2| /2    /O     Y
+             n4     n5
+              0\   /1       Z
+                 t
+        It has 3 paths, corresponding to 3 solution for (X,Y,Z): (0,2,0), (1,2,0) and (2,0,1)
+    """
+
+    def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|str, int, int|str]]):
+        """
+        Arguments:
+            array (ListLike[Expression]): List of expressions representing the input sequence
+            transitions (ListLike[tuple[int | str, int, int | str]]): List of transition triples (source, value, destination)
+        """
+        array = flatlist(array)
+        if not all(isinstance(x, Expression) for x in array):
+            raise TypeError("The first argument of an MDD constraint should only contain variables/expressions")
+
+        def is_transition(arg):
+            """ test if the argument is a transition, i.e. a 3-elements-tuple specifying a starting state,
+            a transition value and an ending node"""
+            return len(arg) == 3 and \
+                isinstance(arg[0], (int, str)) and is_int(arg[1]) and isinstance(arg[2], (int, str))
+
+
+        if not all(is_transition(transition) for transition in transitions):
+            raise TypeError("The second argument of an MDD constraint should be collection of transitions")
+        super().__init__("mdd", (array, transitions))
+        self.root_node = transitions[0][0]
+        self.mapping = defaultdict(dict)
+        for s, v, e in transitions:
+            self.mapping[s][v] = e
+
+        self.levels = {self.root_node: 0}
+        current_nodes = [self.root_node]
+        for i in range(len(array)):
+            new_nodes = []
+            for n in current_nodes:
+                for v in self.mapping[n]:
+                    new_nodes.append(self.mapping[n][v])
+                    self.levels[self.mapping[n][v]] = i + 1
+            current_nodes = new_nodes
+
+    def _reduce(self):
+        """
+            Auxiliary function that reduces the original MDD by merging nodes with equivalent suffixes (to be implemented)
+        """
+        pass
+
+
+    def decompose(self) -> tuple[list[Expression], list[Expression]]:
+        """
+            Flow decomposition of the MDD global constraint.
+            Enforces that the condition is satisfied.
+
+            Returns:
+                tuple[list[Expression], list[Expression]]: A tuple containing the constraints representing the constraint value and the defining constraints
+        """
+        arr, _ = self.args
+
+        flow = {k: {"in" : [], "out" : []} for k in self.levels.keys()}
+        transition_counter = {(level, d): 0 for level in range(len(arr)) for d in range(arr[level].lb, arr[level].ub+1)}
+
+        for (src, edges) in self.mapping.items():
+            for (value, dst) in edges.items():
+                transition = (self.levels[src], value)
+                transition_counter[transition] += 1
+                flow[src]["out"].append((transition, transition_counter[transition]))
+                flow[dst]["in"].append((transition, transition_counter[transition]))
+
+        cons = []
+        substitution = {}
+        for key in transition_counter.keys():
+            (level, value) = key
+            if transition_counter[key] == 0:
+                continue
+            elif transition_counter[key] == 1:
+                substitution[(key, 1)] = (arr[level] == value)
+            else:
+                bvs = cp.boolvar(shape=transition_counter[key]) # edge variables must be defined
+                for n in range(1, transition_counter[key] + 1):
+                    substitution[(key, n)] = bvs[n - 1]
+
+        for i, key in enumerate(sorted(flow.keys(), key=lambda k: self.levels[k])):
+
+            if ((len(flow[key]["in"]) == 1) and (len(flow[key]["out"]) == 1) and
+                    transition_counter[flow[key]["in"][0][0]] > 1 and transition_counter[flow[key]["out"][0][0]] > 1):
+                substitution[flow[key]["out"][0]] = substitution[flow[key]["in"][0]] # substitute equivalent edge variables
+
+            elif (len(flow[key]["in"]) > 0) and (len(flow[key]["out"]) > 0):
+                cons += [cp.sum([substitution[edge] for edge in flow[key]["in"]]) ==
+                         cp.sum([substitution[edge] for edge in flow[key]["out"]])]
+
+            elif len(flow[key]["in"]) == 0:
+                cons += [cp.sum([substitution[edge] for edge in flow[key]["out"]]) == 1]
+
+            elif len(flow[key]["out"]) == 0:
+                cons += [cp.sum([substitution[edge] for edge in flow[key]["in"]]) == 1]
+
+        for key in transition_counter.keys():
+            if transition_counter[key] > 1:
+                (level, value) = key
+                cons += [cp.sum([substitution[(key, n)] for n in range(1, transition_counter[key] + 1)]) == (arr[level] == value)]
+
+        return cons, []
+
+    def value(self) -> Optional[bool]:
+        """
+        Returns:
+            Optional[bool]: True if the global constraint is satisfied, False otherwise, or None if any argument is not assigned
+        """
+        arr, transitions = self.args
+        arrvals = [argval(a) for a in arr]
+        curr_node = self.root_node
+        for v in arrvals:
+            if v is None:
+                return None
+            if curr_node in self.mapping:
+                if v in self.mapping[curr_node]:
+                    curr_node = self.mapping[curr_node][v]
+                else:
+                    return False
+            else:
+                return False
+        return True # can only have reached end node
+
 
 # syntax of the form 'if b then x == 9 else x == 0' is not supported (no override possible)
 # same semantic as CPLEX IfThenElse constraint
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -894,65 +894,80 @@ class MDD(GlobalConstraint):
         """
         pass
 
-
     def decompose(self) -> tuple[list[Expression], list[Expression]]:
         """
-            Flow decomposition of the MDD global constraint.
-            Enforces that the condition is satisfied.
+        Flow decomposition of the MDD global constraint.
+        Enforces that the condition is satisfied.
 
-            Returns:
-                tuple[list[Expression], list[Expression]]: A tuple containing the constraints representing the constraint value and the defining constraints
+        Returns:
+            tuple[list[Expression], list[Expression]]:
+                A tuple containing the constraints representing the constraint value and the defining constraints.
         """
         arr, _ = self.args
 
-        flow_in: dict[int | str, list[tuple[tuple[int, int], int]]] = defaultdict(list)
-        flow_out: dict[int | str, list[tuple[tuple[int, int], int]]] = defaultdict(list)
-        transition_counter = {
-            (level, d): 0 for level in range(len(arr)) for d in range(arr[level].lb, arr[level].ub + 1)
-        }
+        flow_in: dict[int | str, list[tuple[tuple[int, int], Expression]]] = defaultdict(list)
+        flow_out: dict[int | str, list[tuple[tuple[int, int], Expression]]] = defaultdict(list)
+        edge_vars = defaultdict(list)
 
-        for (src, edges) in self.mapping.items():
-            for (value, dst) in edges.items():
-                transition = (self.levels[src], value)
-                transition_counter[transition] += 1
-                flow_out[src].append((transition, transition_counter[transition]))
-                flow_in[dst].append((transition, transition_counter[transition]))
+        for src, edges in self.mapping.items():
+            for value, dst in edges.items():
+                tr = (self.levels[src], value)
+                edge_var = cp.boolvar()
+
+                flow_out[src].append((tr, edge_var))
+                flow_in[dst].append((tr, edge_var))
+                edge_vars[tr].append(edge_var)
 
         cons = []
-        substitution = {}
-        for key in transition_counter.keys():
-            (level, value) = key
-            if transition_counter[key] == 0:
-                continue
-            elif transition_counter[key] == 1:
-                substitution[(key, 1)] = (arr[level] == value)
+        b = cp.boolvar()
+        invalid_edge_vars = []
+
+        # Go over the nodes in the MDD, and enforce flow constraints
+        for node in self.levels.keys():
+            incoming = flow_in[node]
+            outgoing = flow_out[node]
+            missing = []
+
+            if outgoing:
+                valid_transitions = [tr for tr, _ in outgoing]
+                lvl = valid_transitions[0][0]
+
+                # Find all transition values for which there is no valid path, in order to direct them to a dummy node
+                for val in range(arr[lvl].lb, arr[lvl].ub + 1):
+                    tr = (lvl, val)
+
+                    if tr not in valid_transitions:
+                        var = cp.boolvar()
+                        edge_vars[tr].append(var)
+                        invalid_edge_vars.append(var)
+                        missing.append(var)
+
+            if incoming and outgoing:
+                cons += [
+                    cp.sum([e for _, e in incoming]) ==
+                    cp.sum([e for _, e in outgoing]) + cp.sum(missing)
+                ]
+
+            elif outgoing:
+                cons += [
+                    cp.sum([e for _, e in outgoing]) + cp.sum(missing) == 1
+                ]
+
             else:
-                bvs = cp.boolvar(shape=transition_counter[key]) # edge variables must be defined
-                for n in range(1, transition_counter[key] + 1):
-                    substitution[(key, n)] = bvs[n - 1]
+                cons += [
+                    cp.sum([e for _, e in incoming]) == b
+                ]
 
-        for node in sorted(flow_in.keys() | flow_out.keys(), key=lambda k: self.levels[k]):
+        # Link direct encoding variables with edge variables
+        for (level, value), vars_ in edge_vars.items():
+            cons += [
+                cp.sum(set(vars_)) == (arr[level] == value)
+            ]
 
-            if ((len(flow_in[node]) == 1) and (len(flow_out[node]) == 1) and
-                    transition_counter[flow_in[node][0][0]] > 1 and transition_counter[flow_out[node][0][0]] > 1):
-                substitution[flow_out[node][0]] = substitution[flow_in[node][0]] # substitute equivalent edge variables
+        # Ensure that if an invalid edge is taken, the constraint is not satisfied
+        cons += [cp.sum(invalid_edge_vars) == ~b]
 
-            elif (len(flow_in[node]) > 0) and (len(flow_out[node]) > 0):
-                cons += [cp.sum([substitution[edge] for edge in flow_in[node]]) ==
-                         cp.sum([substitution[edge] for edge in flow_out[node]])]
-
-            elif len(flow_in[node]) == 0:
-                cons += [cp.sum([substitution[edge] for edge in flow_out[node]]) == 1]
-
-            elif len(flow_out[node]) == 0:
-                cons += [cp.sum([substitution[edge] for edge in flow_in[node]]) == 1]
-
-        for key in transition_counter.keys():
-            if transition_counter[key] > 1:
-                (level, value) = key
-                cons += [cp.sum([substitution[(key, n)] for n in range(1, transition_counter[key] + 1)]) == (arr[level] == value)]
-
-        return cons, []
+        return [b], cons
 
     def value(self) -> Optional[bool]:
         """
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -921,7 +921,7 @@ class MDD(GlobalConstraint):
             tuple[list[Expression], list[Expression]]:
                 A tuple containing the constraints representing the constraint value and the defining constraints.
         """
-        arr, _ = self.args
+        arr, _, _ = self.args
 
         # MDD is extended with invalid edges, which are directed to the sink node
         extended_mapping, invalid_edges_set = self._get_complete_mdd()
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -874,7 +874,7 @@ class MDD(GlobalConstraint):
             raise TypeError("The second argument of an MDD constraint should be collection of transitions")
         super().__init__("mdd", (array, transitions))
         self.root_node = transitions[0][0]
-        self.mapping = defaultdict(dict)
+        self.mapping: dict[int | str, dict[int, int | str]] = defaultdict(dict)  # mapping from source node and transition value to destination node
         for s, v, e in transitions:
             self.mapping[s][v] = e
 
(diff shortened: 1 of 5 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+            raise TypeError("The first argument of an MDD constraint should only contain variables/expressions")
+
+        def is_transition(arg):
+            """ test if the argument is a transition, i.e. a 3-elements-tuple specifying a starting state,
+            a transition value and an ending node"""
+            return len(arg) == 3 and \
+                isinstance(arg[0], (int, str)) and is_int(arg[1]) and isinstance(arg[2], (int, str))
+
+
+        if not all(is_transition(transition) for transition in transitions):
```
IgnaceBleukx, line 873: Have a look at the Regular global constraint, no need to add the is_transition helper function here. 
Alternatively, if you think it does improve readibility like this, just add the helper function in utils, then `Regular` can also use it

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+            raise TypeError("The second argument of an MDD constraint should be collection of transitions")
+        super().__init__("mdd", (array, transitions))
+        self.root_node = transitions[0][0]
+        self.mapping: dict[int | str, dict[int, int | str]] = defaultdict(dict)  # mapping from source node and transition value to destination node
```
IgnaceBleukx, line 877: Do you really need a dict in a dict?
`dict[Tuple[int|str,int], int|str]` would also work, no? that is also what we used in Regular.

I guess the constructor function here should be more or less identical to Regular? (Apart from the levels of course)

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+        Flow decomposition of the MDD global constraint.
+        Enforces that the condition is satisfied.
+
+        Returns:
+            tuple[list[Expression], list[Expression]]:
+                A tuple containing the constraints representing the constraint value and the defining constraints.
+        """
+        arr, _ = self.args
+
+        flow_in: dict[int | str, list[tuple[tuple[int, int], Expression]]] = defaultdict(list)
```
IgnaceBleukx, line 908: I find this use of defaultdict confusing. 
Why not just have a regular dictionary? You are not providing a default argument anyway here

IgnaceBleukx: I suppose it is to "pre-allocate" the memory needed for the dict?

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+    def _reduce(self):
+        """
+            Auxiliary function that reduces the original MDD by merging nodes with equivalent suffixes (to be implemented)
+        """
+        pass
+
+    def decompose(self) -> tuple[list[Expression], list[Expression]]:
+        """
+        Flow decomposition of the MDD global constraint.
+        Enforces that the condition is satisfied.
```
IgnaceBleukx, line 900: Some intuition of the decomposition would be helpful here, maybe even a tiny example with 2 or 3 nodes, depending if that is feasible or not

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+
+        cons = []
+        b = cp.boolvar()
+        invalid_edge_vars = []
+
+        # Go over the nodes in the MDD, and enforce flow constraints
+        for node in self.levels.keys():
+            incoming = flow_in[node]
+            outgoing = flow_out[node]
+            missing = []
```
IgnaceBleukx, line 929: There is some room for simplification here: 
1. First preprocess the MDD to a full-MDD by adding the missing transitions to a sink node
2. Encode the MDD as is, not caring about whether it is "complete" or not.

Or would there be a significant difference in the final encoding?

Doing this in two steps will also make the future refactor for a "positive_only" decomposition much smoother.

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: Actually, what is in this dict? A comment would

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+
+            if outgoing:
```
IgnaceBleukx, line 931: `If outgoing is not None` if you really want to use the defaultdicts

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+                    tr = (lvl, val)
+
+                    if tr not in valid_transitions:
+                        var = cp.boolvar()
+                        edge_vars[tr].append(var)
+                        invalid_edge_vars.append(var)
+                        missing.append(var)
+
+            if incoming and outgoing:
+                cons += [
```
IgnaceBleukx, line 946: Style: just write on one line please, and just `.append`

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+    def value(self) -> Optional[bool]:
+        """
+        Returns:
+            Optional[bool]: True if the global constraint is satisfied, False otherwise, or None if any argument is not assigned
+        """
+        arr, transitions = self.args
+        arrvals = [argval(a) for a in arr]
+        curr_node = self.root_node
+        for v in arrvals:
+            if v is None:
```
IgnaceBleukx, line 981: You should always return None if any of the arguments is None, even if you can do the early exit, so you'll need an extra check before the loop for this.

tests/test_trans_linearize.py
```diff
@@ -251,7 +251,18 @@ def test_implies(self):
...
 
+        x = cp.intvar(1, 4, shape=3, name="x")
+
+        cons = cp.MDD(x, [("src", 2, "2"), ("src", 1, "1"), ("src", 4, "4"), ("src", 3, "3"),
+                          ("2", 1, "2,1"), ("1", 2, "1,2"), ("4", 3, "1,2"), ("3", 2, "3,2"),
+                          ("2,1", 1, "snk"), ("2,1", 2, "snk"), ("1,2", 3, "snk"), ("3,2", 2, "snk")])
+        lin_cons = linearize_constraint(decompose_linear([cons]))
+        print("Lin cons: ", lin_cons)
+        assert len(lin_cons) == 1
+        assert str(lin_cons) == "[sum(sum(x[0] == 2, x[0] == 1, x[0] == 4, x[0] == 3) == 1, (sum(x[0] == 2)) == (sum(x[1] == 1)), (sum(x[0] == 1)) == (sum(BV3)), (sum(x[0] == 4)) == (sum(x[1] == 3)), (sum(x[0] == 3)) == (sum(BV4)), (sum(x[1] == 1)) == ((x[2] == 1) + (BV5)), ((BV3) + (x[1] == 3)) == (sum(x[2] == 3)), sum(x[2] == 1, BV5, x[2] == 3, BV4) == 1, ((BV3) + (BV4)) == (x[1] == 2), ((BV5) + (BV4)) == (x[2] == 2)) >= 10]"
```
IgnaceBleukx, line 265: These can better be put onto new lines : )
Smt like:

```python
assert set(map(str, lin_cons)) == {<cons 1>,  # what the constraints means, if unclear
                                   <cons 2>, ...}

```

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +827,169 @@ def value(self) -> Optional[bool]:
...
+            for value, dst in edges.items():
+                tr = (self.levels[src], value)
+                edge_var = cp.boolvar()
+
+                flow_out[src].append((tr, edge_var))
+                flow_in[dst].append((tr, edge_var))
+                edge_vars[tr].append(edge_var)
+
+        cons = []
+        b = cp.boolvar()
```
IgnaceBleukx, line 922: What is this boolvar for? Is it just to indicate the constraints is True/False?
If so, I think you can just replace the boolvar in the value constraints with the expression `sum(invalid_edge_vars) == 0`

IgnaceBleukx (changes requested):
Clean implementation! 
Logic-wise, it looks sound, but it can use some slight restructuring and some more documentation here and there.

cpmpy/expressions/globalconstraints.py
WoutPiessens: The reason to work with a dict in a dict, is to make the _reduce() function (currently unimplemented) simpler. It has to identify equivalent transition functions for every node in the MDD, so it is simpler if the value of the mapping dictionary is this transition function (with as key the node), instead of splitting it up.

cpmpy/expressions/globalconstraints.py
WoutPiessens: The defaultdict is there because for every node, there is a list of all incoming / outgoing flow, which is initially empty and is gradually constructed. Alternatively I would need to check if there is already a list, if so add an element, otherwise make a new list. Now the default value is an empty list.
But I see a way to make this flow_in and flow_out a bit simpler, so I'll do that and add some comments

cpmpy/expressions/globalconstraints.py
WoutPiessens: This is true, I do it like that now

cpmpy/expressions/globalconstraints.py
WoutPiessens: Personally I think it's too complicated for a docstring, but maybe feasible for the positive decomposition

WoutPiessens: But I added some extra intuition

tests/test_trans_linearize.py
WoutPiessens: For now I left it out, since the output is solver dependent anyway

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
                 return False
         return curr_node in accepting
 
+
+class MDD(GlobalConstraint):
+    """
+    MDD-constraint: an MDD (Multi-valued Decision Diagram) is an acyclic layered graph starting from a single node and
```
IgnaceBleukx: This docstring can use some love, is this copy pasted from the xcsp3 version of the global?
Proposal:

```
An MDD of depth n is an acyclic layered graph with a single root node, and one accepting end-node.
The constraint takes as input an array of n integer variables and a graph representation using a transition table.
The MDD constraint is satisfied when the values in the array correspond to a path in the MDD ending in the accepting state.

The transitions are defined by a list of tuples (id1, value, id2).
An id is an integer or string representing a state in the automaton, and value is an integer representing the value of the variable in the sequence.
If not given explicitely, the starting node is the first node in the transition table (i.e., transitions[0][0]). 
The end node is the only node on level n.

Example: the following MDD accepts the solutions [1,1,1], [2,2,1] and [2,3,2] for the three variables x,y,z.
    cp.MDD(array=[x,y,z],
           transitions = [("A", 1, "B"), ("B", 1, "D"), ("D", 1 "F"), 
                          ("A",2,"C"), ("C", 2, "D"), ("C", 3, "E"), ("E", 2, "F")]) 
```

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+              0/ |1  \2     X
+            n1   n2   n3
+            2| /2    /O     Y
+             n4     n5
+              0\   /1       Z
+                 t
+        It has 3 paths, corresponding to 3 solution for (X,Y,Z): (0,2,0), (1,2,0) and (2,0,1)
+    """
+
+    def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|str, int, int|str]]):
```
IgnaceBleukx: We can also allow an optional argument "start"? Default would be transtitions[0][0] as it is now, but it makes it a bit more flexible for the user.

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+
+        self.levels = {self.root_node: 0}
+        current_nodes = [self.root_node]
+        for i in range(len(array)):
+            new_nodes = []
+            for n in current_nodes:
+                for v in self.mapping[n]:
+                    new_nodes.append(self.mapping[n][v])
+                    self.levels[self.mapping[n][v]] = i + 1
+            current_nodes = new_nodes
```
IgnaceBleukx, line 886: If we decide there is a single node on level n, then we should assert it to be the case here.
smth like `assert (level == n for node, level in self.levels.items())`

Alternatively, we don't assert it, but then our decomposition should work with it too.
For now it is probably the safest to be strict about this and only allow a single node on level n, we can change it later if we feel this is a feature we would like to add.

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+        and link them to all variable assignments in the array.
+
+        Returns:
+            tuple[list[Expression], list[Expression]]:
+                A tuple containing the constraints representing the constraint value and the defining constraints.
+        """
+        arr, _ = self.args
+
+        # MDD is extended with invalid edges, which are directed to the sink node
+        invalid_edges = []
```
IgnaceBleukx: make this a frozenset after the loop filling it, speeds up future lookups

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+        sink_node = max(self.levels.keys(), key=lambda x: self.levels[x])
+        for s in self.mapping.keys():
+            level = self.levels[s]
+            domain = range(self.args[0][level].lb, self.args[0][level].ub + 1)
+            for v in domain:
+                if v not in self.mapping[s]:
+                    extended_mapping[s][v] = sink_node
+                    invalid_edges.append((s, v))
+
+        # Represents the ingoing and outgoing flow for a certain node: for every node, the edge variables are listed
```
IgnaceBleukx: Doc: "ingoing and outgoing flow for each node: (id, edge_var)"

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+        flow_in: dict[int | str, list[Expression]] = defaultdict(list)
```
IgnaceBleukx, line 934: add a comment on what this does, smt like "default is empty list (no flow)"

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+        flow_out: dict[int | str, list[Expression]] = defaultdict(list)
+
+        # Used to link edge variables to direct encoding variables in a later step
+        edge_vars = defaultdict(list)
+        invalid_edge_vars = []
+
+        # Flow in and flow out calculated for every node
```
IgnaceBleukx: remove the "calculated"

IgnaceBleukx: maybe "determine flow in and flow out for each node"?

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+        for src, edges in extended_mapping.items():
+            for value, dst in edges.items():
+                level = self.levels[src]
+                edge_var = cp.boolvar()
+                flow_out[src].append(edge_var)
+                flow_in[dst].append(edge_var)
+                edge_vars[(level, value)].append(edge_var)
+
+                if (src, value) in invalid_edges:
+                    invalid_edge_vars.append(edge_var)
```
IgnaceBleukx, line 951: You don't strictly this invalid_edge_var list; could also just make a "value" list, that you will return as the first arg of the decomposition, and add `edge_var == 0` to it. No strong preference

IgnaceBleukx: Actually maybe its better the way it is currently implemented, otherwise you might have lots of constraints instead of the single sum constraint

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+
+        cons = []
+
+        # Nodes in the MDD are iterated over, and flow constraints are enforced
```
IgnaceBleukx: What about:
````
Enforce flow constraints: flow in = flow out, at most one activated in/out edge
```

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+        for node in self.levels.keys():
+            incoming = flow_in[node]
+            outgoing = flow_out[node]
+
+            if incoming and outgoing:
```
IgnaceBleukx: I find these conditions inintuitive, it has to do with the level the node is at, right?
So why not use the levels explicitely? `for node, level in self.levels.items()` and then use the level in the if-conditions?
E.g., 

```
if level == 0: cp.sum(outgoing) == 1 # root
elif level == n: cp.sum(incoming) == 1 # sink
else: cp.sum(incoming) == cp.sum(outgoing) # enforce flow for internal nodes
```

IgnaceBleukx: You can also add the redundant constraints `cp.sum(incoming) <= 1`

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+                cons.append(cp.sum(incoming) == cp.sum(outgoing))
+            elif outgoing:
+                cons.append(cp.sum(outgoing) == 1)
+            else:
+                cons.append(cp.sum(incoming) == 1)
+
+        # Link direct encoding variables with edge variables
```
IgnaceBleukx: Something more informative?
E.g., `enforce variable i to value v if a edge with label v is activate at level i`

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+        return [cp.sum(invalid_edge_vars) == 0], cons
+
+
+    def value(self) -> Optional[bool]:
+        """
+        Returns:
+            Optional[bool]: True if the global constraint is satisfied, False otherwise, or None if any argument is not assigned
+        """
+        arr, transitions = self.args
+        arrvals = [argval(a) for a in arr]
```
IgnaceBleukx: use `argvals` instead

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,159 @@ def value(self) -> Optional[bool]:
...
+        curr_node = self.root_node
+        for v in arrvals:
```
IgnaceBleukx: `if any(v is None for v in arrvals)`

IgnaceBleukx (commented):
Yes this is already a big improvement over the previous iteration!
Some comments apart from the code itself
- Should the "extension to full mdd" part become a helper function? I think it would make sense. Smt like "self.get_complete_mdd" that returns all edges?
- There are also solvers that support MDD natively. Choco for sure, potentially some other as well (have a look at the XCSP3 competition branch).

Other comments are all doc related : )

cpmpy/expressions/globalconstraints.py
WoutPiessens: It was copy pasted from xcsp3 version with some small changes, but this is better, I used it with some small changes and clarifications. Also, in the input example the reduce step (identifying equivalent nodes) has already been performed which is not required (it can also be done during decomposition), but that is fine I think.

cpmpy/expressions/globalconstraints.py
WoutPiessens: I agree that we should only allow a single node on level n, added the assert

cpmpy/expressions/globalconstraints.py
WoutPiessens: For now I'll keep it as it is

cpmpy/expressions/globalconstraints.py
WoutPiessens: That's clearer to read since we iterate over the levels dict anyway, I agree

cpmpy/expressions/globalconstraints.py
WoutPiessens: Slightly reformulated it

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,171 @@ def value(self) -> Optional[bool]:
...
+        if not all(isinstance(x, Expression) for x in array):
+            raise TypeError("The first argument of an MDD constraint should only contain variables/expressions")
+
+        _node_type = type(transitions[0][0])
+        for s, v, e in transitions:
+            if not isinstance(s, _node_type) or not isinstance(e, _node_type) or not isinstance(v, int):
+                raise TypeError(
+                    f"The second argument of an MDD constraint should be a collection of transitions ({_node_type}, int, {_node_type})")
+
+        super().__init__("mdd", (array, transitions, start))
```
IgnaceBleukx: You shouldn't save the start node as is in the arguments.
Either you save the root node (which you define on the next line); or don't put start as one of the arguments to the super class (the prefered option).

None values can cause trouble in the transformations, but as it is a constant or None, you can just save it as an attribute on the object (self.root_node as you do below).

WoutPiessens: I now do it the second way (don't add start to the arguments)

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+        for s, v, e in transitions:
+            self.mapping[s][v] = e
+
+        self.levels = {self.root_node: 0}
+        current_nodes = [self.root_node]
+        for i in range(len(array)):
+            new_nodes = []
+            for n in current_nodes:
+                for v in self.mapping[n]:
+                    new_nodes.append(self.mapping[n][v])
```
tias: something very very strange is happening here...

`v in self.mapping[n]` gives... well, 'v' suggests it should be a value, but self.mapping[n] returns a dict...

then you do self.mapping[n][v] but v is a dict, so you lookup the dict in the dict?

I would have expected code like `for value, id2 in self.mapping[id1]` or so?

tias (commented):
I will continue after this is resolved, ping me

cpmpy/expressions/globalconstraints.py
WoutPiessens: @tias given tuples (start_node, transition_value, end_node), I currently group them as follows {start_node : {transition_value : end_node}} (instead of {start_node : (transition_value, end_node)}). So there is a nested dictionary, and the keys of that nested dictionary are the transition values (that's why I use v there, I iterate over all keys of the nested dictionary).

The reason I did this is because in several methods of the MDD global, I need to reason about (start_node, transition_value) pairs where the end_node is either unknown or irrelevant, so it makes the code a bit cleaner. But it is not necessary to work this way.

tias: its still rather twisted...

it sounds like you are relying on it doing `for v in self.mapping[n].keys()` and then `self.mapping[n][v]`
but its more direct to do `for v, id2 in self.mapping[n].items(): ...`

so its not about whether you store it as a dict or not, its about how you access the dict, and that 'v in dict' returns a key while `dict[v]` returns an element is too subtle

WoutPiessens: Okay I can rewrite it like that, that is more intuitive indeed. I'll also look at other parts of the code to apply that change where possible

tias: great, for readability I also recommend using names 'id1', 'v' and 'id2' here

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+
+        super().__init__("mdd", (array, transitions))
+        self.root_node = transitions[0][0] if start is None else start
+        self.mapping: dict[int | str, dict[int, int | str]] = defaultdict(dict)  # mapping from source node and transition value to destination node
+        for id1, v, id2 in transitions:
+            self.mapping[id1][v] = id2
+
+        self.levels = {self.root_node: 0}
+        current_nodes = [self.root_node]
+        for i in range(len(array)):
```
tias: for readability, consider `level` instead of `i`

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+            new_nodes = []
+            for n in current_nodes:
+                for v in self.mapping[n]:
+                    new_nodes.append(self.mapping[n][v])
+                    self.levels[self.mapping[n][v]] = i + 1
+            current_nodes = new_nodes
+        # Check that there is exactly one sink node on level n (with n the number of integer variables)
+        assert sum(level == len(array) for node, level in self.levels.items()) == 1
+
+        self.sink_node = next(node for node, level in self.levels.items() if level == len(array))
```
tias: the above 4 lines more intuitive as

```
# Check that there is exactly one sink node on level n (with n the number of integer variables)
sink_nodes = [node for node, level in self.levels.items() if level == len(array)]
assert len(sink_nodes) == 1
self.sink_node = sink_nodes[0]
```
?

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+        self.nodes = self.levels.keys()
+        self.node_map = {n: (-1 if n == self.sink_node else i) for i, n in enumerate(self.nodes)}
```
tias: or... first build this, and then just read it off with node_map[n]?

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+        """
+        Auxiliary function that extends the MDD with invalid edges, which are directed to the sink node.
+
+        Returns:
+            tuple[dict[int | str, dict[int, int | str]], set[tuple[int | str, int]]]:
+            A tuple containing the extended mapping of the MDD and a set of invalid edges (source node, transition value) that are added to the MDD.
+        """
+        invalid_edges = set()
+        extended_mapping = copy.deepcopy(self.mapping)
+        sink_node = max(self.levels.keys(), key=lambda x: self.levels[x])
```
tias: you have a self.sink_node? this reads too much like magic btw

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+        for s in self.mapping.keys():
+            level = self.levels[s]
+            domain = range(self.args[0][level].lb, self.args[0][level].ub + 1)
```
tias: do `array, transitions = self.args` at the top of the function, less cryptic and fewer lookups needed

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+        Flow decomposition of the MDD global constraint.
+        Enforces that the condition is satisfied, by ensuring that the flow in equals the flow out for every node.
+        To do this, we introduce auxiliary boolean variables for every edge in the MDD, use them in flow constraints,
+        and link them to all variable assignments in the array.
+
+        Returns:
+            tuple[list[Expression], list[Expression]]:
+                A tuple containing the constraints representing the constraint value and the defining constraints.
+        """
+        arr, _ = self.args
```
tias: arr = self.args[0]

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+            else:
+                cons.append(cp.sum(incoming) == cp.sum(outgoing)) #enforce flow for internal nodes
+                cons.append(cp.sum(incoming) <= 1) # redundant constraint: at most one incoming edge
+                cons.append(cp.sum(outgoing) <= 1) # redundant constraint: at most one outgoing edge
+
+        # Enforce direct encoding variable arr[i] == v if an edge with label v is activated at level i
+        for (level, value), vars_ in edge_vars.items():
+            cons.append(cp.sum(vars_) == (arr[level] == value))
+
+        return [cp.sum(invalid_edge_vars) == 0], cons
```
tias: hmm hmm, I'm a bit surprised here, but also not sure...

the second arg should be defining the auxiliaries, e.g. the new edge_vars, but the value part seems so... small. Shouldn't the 'arr[level]==value' also be in the first list, rather than in `cons`?

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+
+
+    def value(self) -> Optional[bool]:
+        """
+        Returns:
+            Optional[bool]: True if the global constraint is satisfied, False otherwise, or None if any argument is not assigned
+        """
+        arr, transitions = self.args
```
tias: arr = self.args[0]

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,180 @@ def value(self) -> Optional[bool]:
...
+        argvals = [argval(a) for a in arr]
+        curr_node = self.root_node
+        if any(v is None for v in argvals):
+            return None
+
+        for v in argvals:
```
tias: more readable if called `curr_v`

tests/test_trans_linearize.py
tias, line 1: we try not to touch irrelevant files

tias (changes requested):
great, neat and clean.

Some readability comments so we can use the docs/code more directly in teaching too.

One open question on what is value and what is defining; can you discuss with Ignace and experiment a bit empirically too?

cpmpy/expressions/globalconstraints.py
WoutPiessens: Putting cp.sum(vars) == (arr[level] == value) in the first list leads to failing tests. I think this is because these constraints define how the auxiliary variables (edge vars) depend on the decision variables (direct encoding vars), therefore they must be top level (as also discussed in the half-reification paper). However, adding the flow constraints to the first list does not introduce any failing tests (these constraints only contain auxiliary variables). I'll discuss this with Ignace too.

tias: Right.

I would suggest rewording "Enforce direct encoding variable arr[i] == v if an edge with label v is activated at level i" because there is no 'direct encoding variable' for arri == v, its a subexpression in this implementation (it should be).

So more "when arr[i] == v then exactly one of the edges at level i with label v must be true, otherwise none can be true" which leads to a nicely compact `(arr[level] == value) == cp.sum(vars_)`. (which is stronger than using cp.any() because it also includes that at most one of the labelled edges can be true at once)

Coming back to my original question, with the presence of the invalid edges, there is always a solution to the flow problem, so all of those can be put toplevel and all you need is the current 'no invalid edges'. It goes a bit further than 'defines auxiliaries', and more 'auxiliaries can always take a satisfying value', but that's OK.

tias: maybe add a line documenting this though ; )

WoutPiessens:
@tias The comments have all been addressed. For the conditioning and defining constraints in the decomposition: after also discussing this with Ignace, it has been kept the same for now (with additional documentation).

tias (approved):
Great, very much approved.

I removed nodes/node_map and moved it in choco, only place where it is needed; some minor tweaks too.

@IgnaceBleukx will you still have a look? it needs your approval before it can be merged in any case

cpmpy/expressions/globalconstraints.py
```diff
@@ -825,6 +828,177 @@ def value(self) -> Optional[bool]:
...
+        cons = []
+
+        # Enforce flow constraints: flow in = flow out, at most one activated in/out edge
+        for node, level in self.levels.items():
+            incoming = flow_in[node]
+            outgoing = flow_out[node]
+
+            if level == 0:
+                cons.append(cp.sum(outgoing) == 1) # root
+                cons.append(cp.sum(outgoing) > 0)
```
IgnaceBleukx: This constraint can go, its implied by the previous one

IgnaceBleukx (approved):
Great, much improved implementation compared to last time I checked this : )

Non-blocking:

As discussed offline: for me the constraints enforcing that there is flow at all are constraining constraints, not toplevel. The intuition being

There are constraints that "define" what "flow" means: input = output at each node.
The MDD is satisfied if: there is flow from root to sink, and no invalid edges are taken.

It also simplifies the postive decomp for later: there you don't have invalid edges, so what would you enforce as constraining part? Nothing? Everything? I'm not sure... With my proposal you would put the "flow at root" and "flow at sink" constraints in the constraining part there as well.

PR #926, opened 2026-04-17
Multi dimensional element constraint

### LINKED ISSUE #444: 2d-element constraint can have out-of-bounds indices

Dimosts:
Adding Multi-Dimensional Element constraint separately than 1D Element. The main reason is to allow us to safen the indices, as the way we have it now we lose information for the different indices.


cpmpy/expressions/globalfunctions.py
```diff
@@ -706,6 +709,14 @@ def __init__(self, arr: ListLike[ExprLike], idx: ExprLike):
             raise TypeError(f"Element(arr, idx) takes an integer expression as second argument, not a boolean expression: {idx}")
         if is_any_list(idx):
             raise TypeError(f"Element(arr, idx) takes an integer expression as second argument, not a list: {idx}")
+        if isinstance(arr, NDVarArray):
```
IgnaceBleukx: NDVarArray subclasses np.ndarray so no need to do this check here too, just the one below suffices

cpmpy/expressions/globalfunctions.py
```diff
@@ -794,6 +805,106 @@ def __repr__(self) -> str:
...
+            arr = arr
+        else:
+            arr = cpm_array(arr)
+
+        if arr.ndim <= 1:
+            raise TypeError("MultiDimElement only supports multi-dimensional arrays. Use Element for 1D arrays.")
+        if len(indices) != arr.ndim:
+            raise ValueError(f"MultiDimElement expects {arr.ndim} indices, got {len(indices)}")
+
+        super().__init__("multidim_element", (arr, *tuple(indices)))
```
IgnaceBleukx: Why add each of the indices as a seperate arg? Ah because of safening probably... Maybe add a comment about it here? It seems somewhat surprising

cpmpy/expressions/globalfunctions.py
```diff
@@ -794,6 +805,106 @@ def __repr__(self) -> str:
...
+        raise CPMpyException("For using multi-dimensional Element, use comma-separated indices on the original array.")
+
+    def value(self) -> Optional[int]:
+        """
+        Returns:
+            Optional[int]: The value of the array element at the given indices, or None if any index is not assigned or the array element is not assigned
+        """
+        arr, *indices = self.args
+        vidxs = [argval(idx) for idx in indices]
+        if any(v is None for v in vidxs):
```
IgnaceBleukx, line 849: Also need to check if any of the arr is None, even if it is not the one that is indexed.
We have the policy that if any arg is None, the value of the entire global is None.

(There is probably an efficient numpy-way of checking the entire array at once by checking the dtype or smt)

cpmpy/expressions/globalfunctions.py
```diff
@@ -794,6 +805,106 @@ def __repr__(self) -> str:
...
+        flat_index = indices[-1]
+        for dim, idx in enumerate(indices[:-1]):
+            flat_index += idx * math.prod(shape[dim + 1 :])  # stride on dim: flat offset per +1 (product of later axis sizes)
+        return flat_index
+
+    def decompose(self) -> tuple[Expression, list[Expression]]:
+        """
+        Decomposition of MultiDimElement global function.
+
+        Rewritten as 1-D Element with a linear index into the flattened array.
```
IgnaceBleukx, line 878: An example would be useful here I think

cpmpy/expressions/globalfunctions.py
```diff
@@ -794,6 +805,106 @@ def __repr__(self) -> str:
...
+            return None
+        for v, dim in zip(vidxs, arr.shape):
+            if v < 0 or v >= dim:
+                raise IncompleteFunctionError(
+                    f"Index {v} out of range for dimension size {dim} while calculating value for expression {self}"
+                    + "\n Use argval(expr) to get the value of expr with relational semantics."
+                )
+        return argval(arr[tuple(vidxs)])
+
+    def _flat_index(self, shape, indices):
```
IgnaceBleukx: Add types for this helper?

cpmpy/expressions/variables.py
```diff
@@ -510,36 +510,37 @@ def clear(self) -> None:
...
-            # using index expression as single var for flat array
-            return cp.Element(arr.flatten(), flat_index)
+            index = list(index[expr_dim[0]:])
+
+            # eliminate constant indices to reduce dimensionality
+            selector = []
+            new_indices = []
+            for idx in index:
+                if isinstance(idx, Expression):
+                    selector.append(slice(None))
```
IgnaceBleukx: what does `slice(None)` do?

tests/test_globalconstraints.py
```diff
@@ -771,6 +771,28 @@ def test_multid_1expr(self):
         expr = x[1,2,a]
         assert str(expr) == "[x[1,2,0] x[1,2,1] x[1,2,2] x[1,2,3] x[1,2,4]][a]"
 
+    def test_multid_element(self):
```
IgnaceBleukx, line 774: Maybe add the tests case from this issue too?

https://github.com/CPMpy/cpmpy/issues/444

IgnaceBleukx (commented):
Great! I have some (mostly minor) comments on the code, but the CI seems to be failing for PySAT, not sure why...

Other thing is the name, personally I like `MultiDElement`, what do you think?

cpmpy/expressions/globalfunctions.py
Dimosts: Yea, I fought a bit with it, my initial thought was to add it as a list of indices. But safening as you said.

cpmpy/expressions/variables.py
Dimosts: It is basically equivalent to ":" when used as index. So, we keep it, and later when we use the elector as indices, it is like we use : (to keep that dimension for element). Putting a comment there

Dimosts:
Fixed the issue with pysat, it was a memory thing, just reduced the size of the array in the test and done. Also resolved the comments, improving based on the suggestions, and adding additional tests.

About the name, I do not have any strong preference, will change it to MultiDElement.

cpmpy/expressions/globalfunctions.py
```diff
@@ -794,6 +804,112 @@ def __repr__(self) -> str:
...
+            return None
+        for v, dim in zip(vidxs, arr.shape):
+            if v < 0 or v >= dim:
+                raise IncompleteFunctionError(
+                    f"Index {v} out of range for dimension size {dim} while calculating value for expression {self}"
+                    + "\n Use argval(expr) to get the value of expr with relational semantics."
+                )
+        return argval(arr[tuple(vidxs)])
+
+    def _flat_index(self, shape: tuple[int, ...], indices: ListLike[ExprLike]) -> ExprLike:
```
IgnaceBleukx, line 859: Actually, on second thought this helper may be a bit overkill... It's 3 lines of code which can go perfectly in the actual decomposition too?

cpmpy/expressions/globalfunctions.py
```diff
@@ -794,6 +804,112 @@ def __repr__(self) -> str:
...
+            arr = arr
+        else:
+            arr = cpm_array(arr)
+
+        if arr.ndim <= 1:
+            raise TypeError("MultiDElement only supports multi-dimensional arrays. Use Element for 1D arrays.")
+        if len(indices) != arr.ndim:
+            raise ValueError(f"MultiDElement expects {arr.ndim} indices, got {len(indices)}")
+
+        super().__init__("multidim_element", (arr, *tuple(indices))) # Indices as separate arguments to allow safening with _safen_range
```
IgnaceBleukx, line 835: This naming is now inconsistent with the class name

IgnaceBleukx (commented):
Great! Not sure if the helper function is really required?
Also, MultiD element is also supported in MiniZinc, right? So we should also add it to our interface.

I quickly checked for other solvers, they don't seem to support it.

tias:
NDElement, like we have NDVarArray?

(no real objection to MultiDElement though)

Dimosts:
No real objection or preference to the name on my side either.

On another note though, I did some experiments on using minizinc's multi dimensional element, or decomposing it ourselves.

And it seems that our decomposition is quite better. So, although minizinc does support it, I think we should not put it in its supporting list. 

My experiments used the Quadratic Assignment Problem, using a 2D element constraint in defining the objective. The main parameter is n, the number of locations, which also defines that the array (to be accessed) is nxn. I experimented with n = 6...10. 

Note that in the figure, the time is in log scale.

image

IgnaceBleukx:
Interesting, is this with gecode or OR-Tools as backend? I would still prefer to not decompose it, as users may want to use a solver with minizinc that does support native 2d element constraints.

IgnaceBleukx:
Renamed to NDElement, and merged current master into this one.
Should be ready for final review by @tias

tias (commented):
I added some stricter typecheck to the constructor; well, an assert actually, and had to update a test.

our safen has a 'safen_toplevel' that multiple solvers have to fill in with 'element', 'div', mod'... should we also add nd_element there?
  More generally, safening docs don't mention nd_element yet

we now have a decompose_linear for element, do we also need one for nd_element?

I also followed an agent recommendation to extend a test...

also small bugfix in choco's added objective safening


other than that its looking good so far...

IgnaceBleukx:
Small updates to the doc of safening, and test.

For safen_toplevel: I'm not sure... its not strictly necessary, but that is the case for all of the global functions: our decomposition forces it to be total, so that's fine.
So if we follow how we do it for `element`, then we need to add it to the safen_toplevel sets, but it should work as is too; although there will be a warning raised during decompose.

But I'm now realizing that if we do the safening using `safen_toplevel`, we are introducing additional variables into the model... So maybe the right way of doing things is to not add a global function to `safen_toplevel` if its not supported by the solver, then we just count on the decomposition to make it total.

Probably for another PR tho... shall we now just follow the same policy as for `element`?

IgnaceBleukx:
Added the nd_element to safen_toplevel for all solvers that do not support it (all but mzn).
Will merge when the tests complete; in a follow up PR we might remove this again, but food for discussion there : )

PR #927, opened 2026-04-17
mzn: fix time_limit deprecation warning

tias:
its trivial, you can just merge it if the tests pass without that warning

### DIFF
--- a/cpmpy/solvers/minizinc.py
+++ b/cpmpy/solvers/minizinc.py
@@ -308,7 +308,7 @@ class CPM_minizinc(SolverInterface):
         if time_limit is not None:
             from packaging.version import Version
             # timeout is deprecated from version 0.10.0 onwards, but cpmpy also supports older versions
-            if self.version().split("/")[0] >= Version("0.10.0"):
+            if Version(self.version().split("/")[0]) >= Version("0.10.0"):
                 kwargs['time_limit'] = timedelta(seconds=time_limit)
             else:
                 kwargs['timeout'] = timedelta(seconds=time_limit)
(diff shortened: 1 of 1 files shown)


OrestisLomis (approved):
I personally don't get a warning using mzn with 'timeout' either.. not sure if there is a special case that triggers the warning. But the PR is very simple and if it eliminates the warning then it is fine by me

tias:
https://github.com/CPMpy/cpmpy/actions/runs/24563052447/job/71816233974

Does the code still work with your minizinc version installed locally, e.g. could you runt he tests with --solver minizinc?

(I don't have it installed)

tias:
bleh... they only introduced it in their latest 0.10.0, and all versions before (we accept 0.7.0) dont have time_limit but the timeout we use...

so we should not do this. Perhaps we should check the version or so in the code...

OrestisLomis:
You could do something like 
```
            mzn_vers = list(map(int, self.version().split("/")[0].split(".")))
            if mzn_vers >= [0,10,0]:
                kwargs['time_limit'] = timedelta(seconds=time_limit)
            else:
                kwargs['timeout'] = timedelta(seconds=time_limit)
```

Although that is perhaps not the cleanest way to handle it

OrestisLomis:
This PR begs the question whether `<solver object>.version()` should not return a(n optional) Version type in the first place..

tias:
> This PR begs the question whether `<solver object>.version()` should not return a(n optional) Version type in the first place..

I think that is a good question, at least I didn't know about Version type...

Not to be changed for all solvers in this PR : )

PR #928, opened 2026-04-17
Loop optimisations

tias:
the first commit cuts 16 seconds (64 -> 48) from decompose in the dev/time_transf script

the second applies the append pattern in other places

the third is more a gimmick, precomputing a frozenset in InDomain


IgnaceBleukx (approved):
Great! Reading up on this, we should really avoid the `+` with lists... even when we extend more than one item.
Apparently, it always makes a new object, instead of extending the existing one.

PR #929, opened 2026-04-19
refactoring decompose in typed pattern

tias:
a better-typed decompose (inspired by the new pattern in negate).

It does show we need to continue typing our expressions #913 :
[ ] Element decompose can return int (even in our test suite there is a case for element), to be changed (see inline comment)
[ ] global constraint decompose can return bool, should only return Expressions

and this one should probably go after the CSEMAP #917  

This typed version should be a good basis for the positive-decompose, e.g. we only want to support that in the for loop in `decompose_in_tree` (which can also check `if expr.name == "->" and expr.args[1] isinstance GlobalConstraint`... for the halfreified positives)

### DIFF
--- a/cpmpy/transformations/decompose_global.py
+++ b/cpmpy/transformations/decompose_global.py
@@ -184,22 +184,25 @@ def _decompose_in_tree_args(args: ListLike[Any],
                     continue
                 else:
                     # nested global function, decompose
-                    if decompose_custom is not None and arg.name in decompose_custom:
-                        newarg, toplevel_exprs = cast(Tuple[Expression, List[Expression]], decompose_custom[arg.name](arg))
-                    else:
-                        newarg, toplevel_exprs = arg.decompose()
-
                     orig_for_csemap = arg
-                    arg = newarg
-                    if len(toplevel_exprs) > 0:
-                        toplevel.extend(toplevel_exprs)
+                    # this is a bit awkward, but the decompose can return a new GlobFunc to decompose...
+                    while isinstance(arg, GlobalFunction) and arg.name not in supported:
+                        if decompose_custom is not None and arg.name in decompose_custom:
+                            newarg, toplevel_exprs = cast(Tuple[Expression, List[Expression]], decompose_custom[arg.name](arg))
+                        else:
+                            newarg, toplevel_exprs = arg.decompose()
+                        arg = newarg
+                        if len(toplevel_exprs) > 0:
+                            toplevel.extend(toplevel_exprs)
 
-                    # TODO: violates type!!!
-                    # apparently in #630 we decided that decompose may return an int (e.g. for element)...
-                    # we should change that (Element constructor requires variable index; the []/__get__ override can still take anything)
+                        # TODO: violates type!!!
+                        # apparently in #630 we decided that decompose may return an int (e.g. for element)...
+                        # we should change that (Element constructor requires variable index; the []/__get__ override can still take anything)
+                        if isinstance(arg, int):
+                            # no need to recurse further, stop here
+                            newargs.append(arg)
+                            break
                     if isinstance(arg, int):
-                        # no need to recurse further, stop here
-                        newargs.append(arg)
                         continue
             
             # if it has subexprs, decompose its arguments too
--- a/cpmpy/transformations/decompose_global.py
+++ b/cpmpy/transformations/decompose_global.py
@@ -22,10 +22,10 @@ import copy
 from typing import List, AbstractSet, Optional, Dict, Tuple, Any, Callable, cast
 import numpy as np
 
-from ..expressions.core import Expression, ListLike
-from ..expressions.variables import NDVarArray
-from ..expressions.utils import is_any_list
-from ..expressions.python_builtins import all as cpm_all
+from ..expressions.core import Expression, ListLike, BoolVal, Operator
+from ..expressions.globalconstraints import GlobalConstraint
+from ..expressions.globalfunctions import GlobalFunction
+from ..expressions.variables import NDVarArray, cpm_array
 
 
 def decompose_in_tree(lst_of_expr: list[Expression],
@@ -58,22 +58,38 @@ def decompose_in_tree(lst_of_expr: list[Expression],
     if supported_reified is None:
         supported_reified = frozenset[str]()
 
-    changed, newlst_of_expr, todo_toplevel = _decompose_in_tree(lst_of_expr, supported=supported, supported_reified=supported_reified, is_toplevel=True, csemap=csemap, decompose_custom=decompose_custom)
-    if not changed:
-        return lst_of_expr
+    newlist: List[Expression] = []
+    todolist: List[Expression] = []  # these still need to be decomposed
+    for expr in lst_of_expr:
+        if isinstance(expr, GlobalConstraint) and expr.name not in supported:
+            # toplevel/positive global constraint, decompose
+            if decompose_custom is not None and expr.name in decompose_custom:
+                exprs, toplevel_exprs = cast(Tuple[List[Expression], List[Expression]], decompose_custom[expr.name](expr))
+            else:
+                exprs, toplevel_exprs = expr.decompose()
+            # both might contain globals too
+            todolist.extend(exprs)
+            if len(toplevel_exprs) > 0:
+                todolist.extend(toplevel_exprs)
+        elif isinstance(expr, bool):
+            # TODO: violates type!!!
+            newlist.append(BoolVal(expr))
+        elif expr.has_subexpr():
+            # decompose its arguments
+            changed, newargs, rec_toplevel = _decompose_in_tree_args(expr.args, supported=supported, supported_reified=supported_reified, csemap=csemap, decompose_custom=decompose_custom)
+            if changed:
+                expr = copy.copy(expr)
+                expr.update_args(newargs)
+                if len(rec_toplevel) > 0:
+                    todolist.extend(rec_toplevel)
+            newlist.append(expr)
+        else:
+            newlist.append(expr)
 
-    # new toplevel constraints may need to be decomposed too
-    while len(todo_toplevel):
-        changed, decomp, next_toplevel = _decompose_in_tree(todo_toplevel, supported=supported, supported_reified=supported_reified, is_toplevel=True, csemap=csemap, decompose_custom=decompose_custom)
-        if not changed:
-            newlst_of_expr.extend(todo_toplevel)
-            break
-
-        # changed, loop again
-        newlst_of_expr.extend(decomp)
-        todo_toplevel = next_toplevel # decompositions may have introduced nested lists or ands
-
-    return newlst_of_expr
+    # recurse on any newly generated toplevel expressions
+    if len(todolist) == 0:
+        return newlist
+    return newlist + decompose_in_tree(todolist, supported=supported, supported_reified=supported_reified, csemap=csemap, decompose_custom=decompose_custom)
 
 
 def decompose_objective(expr: Expression,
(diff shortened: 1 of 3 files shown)


cpmpy/transformations/decompose_global.py
```diff
@@ -58,22 +58,38 @@ def decompose_in_tree(lst_of_expr: list[Expression],
...
-    while len(todo_toplevel):
-        changed, decomp, next_toplevel = _decompose_in_tree(todo_toplevel, supported=supported, supported_reified=supported_reified, is_toplevel=True, csemap=csemap, decompose_custom=decompose_custom)
-        if not changed:
-            newlst_of_expr.extend(todo_toplevel)
-            break
-
-        # changed, loop again
-        newlst_of_expr.extend(decomp)
-        todo_toplevel = next_toplevel # decompositions may have introduced nested lists or ands
+    newlist: List[Expression] = []
```
IgnaceBleukx: no global `changed` flag? In the other transforms, we just return the original list of there are no changes made (saves memory as newlist will go out of scope on return)

cpmpy/transformations/decompose_global.py
```diff
@@ -58,22 +58,38 @@ def decompose_in_tree(lst_of_expr: list[Expression],
...
+    todolist: List[Expression] = []  # these still need to be decomposed
+    for expr in lst_of_expr:
+        if isinstance(expr, GlobalConstraint) and expr.name not in supported:
+            # toplevel/positive global constraint, decompose
+            if decompose_custom is not None and expr.name in decompose_custom:
+                exprs, toplevel_exprs = cast(Tuple[List[Expression], List[Expression]], decompose_custom[expr.name](expr))
```
IgnaceBleukx: don't we use `list` and `tuple` instead of typing.List and typing.Tuple?

cpmpy/transformations/decompose_global.py
```diff
@@ -58,22 +58,38 @@ def decompose_in_tree(lst_of_expr: list[Expression],
...
+            else:
+                exprs, toplevel_exprs = expr.decompose()
+            # both might contain globals too
+            todolist.extend(exprs)
+            if len(toplevel_exprs) > 0:
+                todolist.extend(toplevel_exprs)
+        elif isinstance(expr, (bool, np.bool_)):
+            # TODO: violates type!!!
+            newlist.append(BoolVal(expr))
+        elif expr.has_subexpr():
```
IgnaceBleukx, line 97: Hmm so the code path is different for toplevel globals vs toplevel non-globals?
For globals, we first decompose it, and then in a recursive call, decompose the arguments
For non-gloabals, we decompose its arguments.

In the previous version, we always first did the arguments, and then the globals themselves.
I think its ok as is but slightly unexpected at first.

cpmpy/transformations/decompose_global.py
```diff
@@ -110,21 +131,20 @@ def decompose_objective(expr: Expression,
...
-                       is_toplevel: bool,
-                       csemap: Optional[CSEMap]=None,
-                       decompose_custom:Optional[Dict[str, Callable]]=None) -> Tuple[bool, List[Expression], List[Expression]]:
+def _decompose_in_tree_args(args: ListLike[Any],
+                            supported: AbstractSet[str],
+                            supported_reified: AbstractSet[str],
+                            csemap: Optional[CSEMap]=None,
+                            decompose_custom:Optional[Dict[str, Callable]]=None) -> Tuple[bool, List[Any], List[Expression]]:
     """
+    TODO: OUTDATED DOC!!
```
IgnaceBleukx: proposal:
```
Well-typed recursive helper function to decompose unsupported global constraints and global
functions in the arguments of an Expression.

Returns:
    tuple[bool, list[Any], list[Expression]]: (changed, newargs, toplevel)
```

cpmpy/transformations/decompose_global.py
IgnaceBleukx: not actually sure what the difference is

cpmpy/transformations/decompose_global.py
```diff
@@ -143,84 +161,115 @@ def _decompose_in_tree(lst_of_expr: ListLike[Any],
...
+
+            # very special case: "and" with single argument (from simple glob.decomp)
+            if arg.name == "and" and len(arg.args) == 1:
+                arg = cast(Expression, arg.args[0])
+
+            if orig_for_csemap is not None and csemap is not None:
+                csemap.save_decomposition(orig_for_csemap, arg)
+            newargs.append(arg)
+        
+        elif isinstance(arg, list):
```
IgnaceBleukx: I had to add this one, there is an edge case for wsum, where the arg is a list, but it can contain globals in the vars...
Should we special case on wsum to avoid recursion? Currently it will also recurse into the list with weights

IgnaceBleukx (commented):
Great, had some trouble merging master into this one, but should be done now.

A couple of open questions still

cpmpy/transformations/decompose_global.py
IgnaceBleukx: We can also just type the Callable properly and avoid the cast all together here?

cpmpy/transformations/decompose_global.py
tias: we do? : )

I checked push_down_negation, there we also don't have a changed. We could though, I agree

cpmpy/transformations/decompose_global.py
tias: yes, we use list/tuple, not List.Tuple

yes, we can type the callable, sure

cpmpy/transformations/decompose_global.py
tias: yes, I see. I think I did it on purpose but I don't remember, I'm not sure which one is best or least expected...

A decomposition can create copies, so then we have to decompose in all of the copies, so perhaps it is best to do it before decompose then, hmm hmm..

cpmpy/transformations/decompose_global.py
IgnaceBleukx: Hmm yeah indeed, I though we had it in general but apparently not.
In `no_partial_functions` we do have it. So lets add it here and then also in #959 ?

https://github.com/CPMpy/cpmpy/blob/ee02564442a28db8bcd63af70be62ddace6256d1/cpmpy/transformations/safening.py#L80

cpmpy/transformations/decompose_global.py
tias: maybe I just did it because after decompose, we still have to walk over the new expression (and hence its args) because a global can use globals in its decomp; so I avoided 2 walks by only doing it afterwards

cpmpy/transformations/decompose_global.py
tias: yes, expressions can contain a list of expressions as an arg, a bit weird I did not have it...

in this function, there is no way to know you are the args of a wsum; so it would have to be higher up.

I would avoid such shortcuts in this refactor, they would require an empirical test

tias:
I've changed, it **but**

- the code was consistent: in both paths, it was first doing decompose and only then recurse over the args (so it was consistently different from what it used to be)
- the reason was indeed that after decompose, you need to decompose what you get again (for nested decompositions)
- it also made the csemap easier: you could check it, decompose the global, decompose it args, then insert it; now, you have to check the map, recurse over the args, decompose, decompose the args, then only insert it

I think the current code is harder to read, it also requires an additional 'cast', and with the csemap, the cost of 're-decomposing' globals in the args was actually non-existing... the whole reason of the map was to avoid that.

So, I would be in favor of going back...

tias:
actually two casts...

now that it is still fresh in my mind, I am going to revert it and document why we do it the way we do it (and there are some minor cleanups possible I now realize)

tias:
hmm, it was a rather extensive refactor for the args part; that looks quite a bit like what I reverted; its still a bit cleaner.

Also compatible with the other PR about 'decompose_custom_positive', I would expect, as it only touches `decompose_in_tree()`.

This is compatible with the 'supported_implication' for choco, needs a `name == '->'` check and a few lines in `decompose_in_tree()` (non-urgent?)

cpmpy/transformations/decompose_global.py
```diff
@@ -61,29 +59,53 @@ def decompose_in_tree(lst_of_expr: list[Expression],
...
+    changed = False
+    for expr in lst_of_expr:
+        if isinstance(expr, GlobalConstraint) and expr.name not in supported:
+            # toplevel/positive global constraint, decompose
+            changed = True
+            if decompose_custom is not None and expr.name in decompose_custom:
+                exprs, toplevel_exprs = cast(tuple[list[Expression], list[Expression]], decompose_custom[expr.name](expr))
+            else:
+                exprs, toplevel_exprs = expr.decompose()
+            # we merge the list toplevel rather than create an 'and', also means we can not store it in csemap
```
IgnaceBleukx: I was a bit worried at first about this, but it is actually fine for this case... It would only cover cases where a user writes the exact same global constraint toplevel, which is probably something we could/should catch before already.

Although... you could also have the case where global X is decomposed into globals Y & Z, and Y is already posted toplevel... But that's an edge case. 
Alternative is to wrap it in an `and` anyway here to put it in the csemap, and extend todist with `csemap.get(expr).args`? instead?

cpmpy/transformations/decompose_global.py
```diff
@@ -110,117 +132,171 @@ def decompose_objective(expr: Expression,
...
+                changed = True
+                if csemap is not None:
+                    decomp = csemap.get_decomposition(arg)
+                    if decomp is not None:
+                        newargs.append(decomp)
+                        continue
+                arg_orig2 = arg
+
+                # a decomposition may consist of a new GlobFunc to decompose...
+                while isinstance(arg, _GlobalFunction) and arg.name not in supported:
```
IgnaceBleukx: Why does globalfunc need another case compared to global cons?
You could also just do:
```python
val, tl = arg.decompose()
(newarg,) = decompose_in_tree_args((val,), ...)
```
This while loop feels like its trying to avoid that recursion, but is it just for performance?

cpmpy/transformations/decompose_global.py
```diff
@@ -110,117 +132,171 @@ def decompose_objective(expr: Expression,
...
+                        if len(rec_toplevel) > 0:
+                            toplevel.extend(rec_toplevel)
+
+                if csemap is not None:
+                    csemap.save_decomposition(arg_orig2, arg)
+                newargs.append(arg)
+                continue
+
+            else:  # any other expression
+                # if it has subexprs, decompose its arguments
```
IgnaceBleukx, line 265: This part is repeated in both the globalcons case and the global func case.
Should we just merge those cases into this one?
We could do that by just dropping this else case, and going directly to the `if has_subexpr`.
If it fails, the expr drops all the way down to the default case of appending

cpmpy/transformations/decompose_global.py
```diff
@@ -110,117 +132,171 @@ def decompose_objective(expr: Expression,
...
+        tuple[bool, list[Any], list[Expression]]: (changed, newargs, toplevel)  
         - ``changed`` is True if a decomposition was done (or a recursive call changed something).
-        - ``newexpr`` is the decomposed sequence (same length as ``lst_of_expr``).
+        - ``newargs`` is the decomposed sequence (same length as ``args``).
         - ``toplevel`` is the list of auxiliary constraints to post at top level.
     """
-    changed = False
-    newlist: List[Any] = []  # TODO: because of is_any_list, can be many things...
-    toplevel: List[Expression] = []
+    _Expression = Expression
```
IgnaceBleukx: Why make the aliases here?

IgnaceBleukx (commented):
Ok so if I get it, the pattern now is:

if unsupported global cons/func: decompose.
then treat the decomposition as any expression: either recursively decompose, or check the args of the decomp.

I think that's clean, easy to follow and better than what we had before : )
 but some comments on the code itself, where cleanup is possible.

cpmpy/transformations/decompose_global.py
tias: fine, added

cpmpy/transformations/decompose_global.py
tias: yes, instead of recursing, we just write a loop and so it stays in the function (the loop almost never goes into a seond iteration)

cpmpy/transformations/decompose_global.py
tias: its more subtle, the 3 cases are a bit different as I tried to document: for globcons, we use _args() on the list directly; for globcons we use _args() on the args and can overwrite the original function; in the general case we use _args() on the args and create a copy

cpmpy/transformations/decompose_global.py
tias: it was a micro optimisation (becuase the same classes are used in the loop, so Python's lookup happens each time). its less readable, I removed it again

cpmpy/transformations/decompose_global.py
IgnaceBleukx: Ah yes and because global func always returns a single expression as the value, this is quite clean to do. For global constraints we would need an extra for-loop over all parts of the decomp...

IgnaceBleukx (approved):
Great, I think we're there, and with a nice runtime improvement too!
dev/time_transformations has a decompose time doing down from 43s -> 34s on my machine.
Also memory usage is decreased by ±10%

I've added a type alias for the custom decomposition, its somewhat cryptic so fine to remove again if we don't like it.

Good to merge for me.

PR #931, opened 2026-04-20
Modernize counterfact example

OrestisLomis:
small question of whether we really want to update all uses of ```sum``` and ```max``` to ```cp.sum``` and ```cp.max``` respectively as there are some sums of non cpmpy variables in this examples where it doesnt really seem necessary (but also doesnt seem to be a problem to still use ```cp.sum```).

### DIFF
--- a/examples/advanced/counterfactual_explain.py
+++ b/examples/advanced/counterfactual_explain.py
@@ -94,7 +94,7 @@ def solve_knapsack_problem(values, weights, capacity):
 
     Based on the Numberjack model of Hakan Kjellerstrand
     """
-    x = cp.boolvar(len(values), name="x")
+    x = cp.boolvar(shape=len(values), name="x")
     model = cp.Model([cp.sum(x * weights) <= capacity], maximize=cp.sum(x * values))
     if model.solve() is not False:
         return x.value()
@@ -160,7 +160,7 @@ def make_master_problem(values, foil_idx):
     This way the variables can be used to add new constraints outside this building function.
     """
 
-    d = cp.intvar(0, INFINITY, values.shape, name="d")
+    d = cp.intvar(0, INFINITY, shape=values.shape, name="d")
     # Minimize the change to the values vector
     m = cp.Model(minimize=np.linalg.norm(values - d, ord=1))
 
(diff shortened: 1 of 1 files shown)


examples/advanced/counterfactual_explain.py
```diff
@@ -110,7 +110,7 @@ def generate_knapsack_model(n=10):
     R = 1000
     values = np.random.randint(1, R, n)
     weights = np.random.randint(1, R, n)
-    capacity = int(max([0.5 * sum(weights), R]))
+    capacity = int(cp.max([0.5 * cp.sum(weights), R]))
```
IgnaceBleukx: These are just python ints in there, no need to use the CPMpy version of sum/max

examples/advanced/counterfactual_explain.py
```diff
@@ -126,7 +126,7 @@ def generate_foil_knapsack(values, weights, capacity, x, m, tries=1):
     foil_idx = np.random.choice(n, m, replace=False)
     foil_vals = np.abs(1 - x[foil_idx])
 
-    if sum(foil_vals * weights[foil_idx]) > capacity:
+    if cp.sum(foil_vals * weights[foil_idx]) > capacity:
```
IgnaceBleukx: Same here, don't use the CPMpy sum

IgnaceBleukx (commented):
Definately don't use all CPMpy-variants, as this is an example : )
So we don't want to set the example that you should use CPMpy-specific sum/max/abs etc if you have some CPMpy code in that same file.

Just use the CPMpy-sum when you need CPMpy expressions, otherwise just Numpy/Python sums

examples/advanced/counterfactual_explain.py
```diff
@@ -94,8 +94,8 @@ def solve_knapsack_problem(values, weights, capacity):
 
     Based on the Numberjack model of Hakan Kjellerstrand
     """
-    x = boolvar(len(values), name="x")
-    model = Model([sum(x * weights) <= capacity], maximize=sum(x * values))
+    x = cp.boolvar(len(values), name="x")
```
IgnaceBleukx: also add keyword "shape="

IgnaceBleukx (approved):

examples/advanced/counterfactual_explain.py
tias: idd, its the kind of clean update we want now.

and indeed, for decision variables, we have a bit the convention that for intvar we dont have to write the required lb=, ub= for the first two parameters, but the shape= and name= keywords (the optional ones) that we always write those keywords explicitly

PR #933, opened 2026-04-21
Add additional tests to `test_solverinterface`

hbierlee:
Some additional tests which specify how solver_interface's should work. Currently, not all solvers are consistent. If we agree that the solvers should behave as the new tests indicate, we can fix the solvers (in that case, I would advocate just doing it in this PR, because everything needs to be merged altogether anyway unless I make the tests solver specific `xfail`s). If these tests are wrong, feel free to close or otherwise update them.


tias:
yes, these should be fixed: after a re-solve, variable values should be reset.

there are 2 lines that we started adding that do that, an 'else' case to filling in the values. I guess we just forgot it here

OrestisLomis:
I had already added some tests and fixes in #1001 for the value, but indeed the objective should also be handled. For Choco and Pumpkin it was relatively easy to fix, though it seems like there is just a bug with resolving an optimisation problem for choco (see https://github.com/CPMpy/cpmpy/issues/1079#issue-5110979250). Exact was a bit more tricky as there seemed to be some legacy handling of the solver status.

PR #935, opened 2026-04-22
Avoid making auxiliary variables in safening

IgnaceBleukx:
All of our transformations before flattening should try to minimize the number of auxiliary variables introduced.

In safening, we made "is_defined" Boolean variable flags, but they can just be expressions too.


tias (commented):
ah yes, good to avoid creating auxiliaries and just use it as expression indeed.

Should we add 'simplify' to the implies? (just merged)
Can you check if that slows down if we do it here?

IgnaceBleukx:
dev/time_transformation runs:

- Master: 1056.4s
- Without simplify = True: 973.75
- With simplify: 976.59

So slightly slower, probably because of the extra isinstance checks

PR #936, opened 2026-04-22
Run coverage and upload report in CI

hbierlee:
This "proof-of-concept" PR adds coverage reports to the CI, meaning we get (in each CI job) a downloadable HTML artifact which shows which lines are run during tests which are missing. A missing line which is not run could either be:

- A missing test for it
- More interesting: something we expect to run but isn't e.g. a transformation decomposes a constraint which is actually supported by the solver

Since the solver jobs are split, for this to be more useful we'd probably want to produce just one report rather than per-solver report. Furthermore, there are other test coverage tools we could consider rather than pytest-cov, I didn't really do a comparison (e.g. there's https://docs.codecov.com/docs/code-coverage-with-python, or https://github.com/marketplace/actions/code-coverage-summary). The advantage also is that then you could see the report from a website, rather than having to download the zip/html file from finding the link in the CI logs e.g. https://github.com/CPMpy/cpmpy/actions/runs/24774510169/artifacts/6576393617 for OR-tools. Finally, one could consider failing the CI if the PR degrades coverage (or something), to ensure new lines are actually tested. This can be useful but of course also annoying.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -23,3 +23,43 @@ jobs:
     with:
       solver: ${{ matrix.solver }}
       install_commands: ${{ matrix.install_commands || '' }}
+
+  coverage:
+    name: Merge coverage
+    needs: test
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install coverage
+        run: pip install coverage
+
+      - name: Download all coverage data
+        uses: actions/download-artifact@v4
+        with:
+          pattern: coverage-data-*
+
+      - name: Merge coverage reports
+        run: |
+          set -euo pipefail
+          # Each artifact is in its own directory; rename to unique filenames
+          i=0
+          for dir in coverage-data-*/; do
+            mv "${dir}.coverage" ".coverage.${i}"
+            i=$((i + 1))
+          done
+          coverage combine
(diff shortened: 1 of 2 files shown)


hbierlee:
Now it makes one report for all solvers run on CI and uploads it as artifact (https://github.com/CPMpy/cpmpy/actions/runs/24775712362/artifacts/6577555400), which you can download to get an idea. Some things I noted:


- ortools no_overlap_optional decomp is never used although there seems to be a test for it
- MiniZinc's regular case is never used because regular is not added to its supported
- The report seems to say the native_mus tests are never run
- in mus we never seem to have an unsat case, probably just lacking the test but a bit suspicious
- couple of unused cases in GCS
- a pysat merge probably gone wrong : now we're very sure Comparisons are handled
- our solver_vars are often checking for constants e.g. is_num(x): return x but this never seems to be the case. Maybe a bit dangerous to optimize out but could be handled last
- some cases for pumpkin also are never used, most notably mul

tias:
I think its too soon to add it to the CI... who is going to look at it?

It feels wasteful to start uploading artefacts on every run that nobody is commited to use.

Instead it could make sense for somebody to run it locally, inspect it and start fixing some of these things, each probabably being its own PR. But that effort should be prioritized against all other things we still want to do.

So perhaps a tracker issue where you include the instructions to run it, and your list of potential issues is more suitable?

hbierlee:
Yep, that's ok. I can just close this PR, because it's not something we have to do (and if we do it, it can be a lot nicer than now). The main thing that is useful about it is that anybody can update this PR by merging new changes, then rerun it to get a report for most solvers, without having to install all of them.

PR #939, opened 2026-04-22
more efficient Solver vars

tias:
an experiment on ortools based on a side-comment in #412 

2026-04-22T12:54:41Z,v0.10.0-69-g42d87f80,25.9,2.34,4.23,0.0,0.37,2.76,0.24,9.28,0.16,0.16,0.1,1.22,     5.03
2026-04-22T12:53:00Z,v0.10.0-70-g2fa240ed,25.03,2.28,4.18,0.0,0.37,2.72,0.24,9.42,0.16,0.16,0.1,1.2,     4.2
2026-04-22T12:58:39Z,v0.10.0-71-g15d530d6,25.18,2.47,4.35,0.0,0.39,2.77,0.24,9.56,0.16,0.17,0.1,1.24,  3.72

so in this run-once experiments the 2 commit made the 'post constraints' time for time_ortools go from 5s to 4.2s to 3.7s

could be worth applying to all solvers?

### DIFF
--- a/cpmpy/solvers/solver_interface.py
+++ b/cpmpy/solvers/solver_interface.py
@@ -19,14 +19,14 @@
         ExitStatus
 
 """
-from typing import Optional, List, Callable, TypeAlias
+from typing import Optional, List, Callable, TypeAlias, Iterable, Any
 import warnings
 import time
 from enum import Enum
 
 from ..exceptions import NotSupportedError
-from ..expressions.core import Expression, ListLike
-from ..expressions.variables import _NumVarImpl
+from ..expressions.core import Expression, ListLike, ExprLike
+from ..expressions.variables import _NumVarImpl, NDVarArray
 from ..transformations.cse import CSEMap
 from ..transformations.get_variables import get_variables
 from ..expressions.utils import is_any_list, argvals
@@ -185,6 +185,24 @@ class SolverInterface(object):
         if is_any_list(cpm_vars):
             return [self.solver_vars(v) for v in cpm_vars]
         return self.solver_var(cpm_vars)
+
+    def solver_vars_1d(self, cpm_vars: Iterable[ExprLike]) -> list[Any]:
+        """
+           Faster `solver_vars()` for 1 dimensional iterables of ExprLikes
+        """
+        res: list[Any] = []
+        for cpm_var in cpm_vars:
+            solver_var = self._varmap.get(cpm_var, None)
+            if solver_var is not None:
+                # fast path
+                res.append(solver_var)
+            elif isinstance(cpm_var, int):
+                # reasonably fast path
+                res.append(cpm_var)
+            else:
+                # slow path, will check the varmap again
+                res.append(self.solver_var(cpm_var))
+        return res
 
     def transform(self, cpm_expr):
         """
--- a/cpmpy/solvers/solver_interface.py
+++ b/cpmpy/solvers/solver_interface.py
@@ -87,7 +87,7 @@ class SolverInterface(object):
 
         # initialise variable handling
         self.user_vars = set()  # variables in the original (non-transformed) model
-        self._varmap = dict[ExprLike, Any]()  # maps cpmpy variables to native solver variables
+        self._varmap = dict()  # maps cpmpy variables to native solver variables
         self._csemap = CSEMap()  # maps cpmpy expressions to previously created expressions (typically auxiliary variables)
 
         # rest uses own API
--- a/cpmpy/solvers/solver_interface.py
+++ b/cpmpy/solvers/solver_interface.py
@@ -87,7 +87,7 @@ class SolverInterface(object):
 
         # initialise variable handling
         self.user_vars = set()  # variables in the original (non-transformed) model
-        self._varmap = dict()  # maps cpmpy variables to native solver variables
+        self._varmap : dict[ExprLike, Any] = dict()  # maps cpmpy variables to native solver variables
         self._csemap = CSEMap()  # maps cpmpy expressions to previously created expressions (typically auxiliary variables)
 
         # rest uses own API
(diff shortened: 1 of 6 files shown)


IgnaceBleukx:
Applying this to PySAT should have a huge impact I believe. There everything is a 1d-list-of-vars, so everything should go through the fast-path.

There are also solvers which allow you to make a batch of new variables at once, they can probably implement a more efficient version of the `1d-` special case?

hbierlee:
Yes, Gurobi can access multiple variable values at once (and we should override solver_vars with that).

I'm not sure why the `solver_vars_1d` should be faster than just `solver_vars`? Is it only the solver call per variable that's saved? I know function calls are not free in python, but I don't think it's worth adding a new public function over.

[edit: ok I suppose it's because solver_vars also calls solver_vars on each element which calls solver_var, so it's more like two functions per element.. hmm]

cpmpy/solvers/solver_interface.py
```diff
@@ -186,6 +186,24 @@ def solver_vars(self, cpm_vars):
...
+        """
+           Faster `solver_vars()` for 1 dimensional iterables of ExprLikes
+        """
+        res: list[Any] = []
+        for cpm_var in cpm_vars:
+            solver_var = self._varmap.get(cpm_var, None)
+            if solver_var is not None:
+                # fast path
+                res.append(solver_var)
+            elif isinstance(cpm_var, int):
```
hbierlee, line 199: so why not `is_num`? Of course it is slower but lots of bools going via the slow path could also be slower

cpmpy/solvers/solver_interface.py
```diff
@@ -186,6 +186,24 @@ def solver_vars(self, cpm_vars):
             return [self.solver_vars(v) for v in cpm_vars]
         return self.solver_var(cpm_vars)
 
+    def solver_vars_1d(self, cpm_vars: Iterable[ExprLike]) -> list[Any]:
```
hbierlee, line 189: I don't like the added `solver_vars_1d` as a new public function. I think the strategy for the different cases (e.g. checking for constants, _varmap, etc..) should be the same in both cases. Maybe a better middle ground is this:

```py
    def solver_vars(self, cpm_vars, is_1d=False):
        """
           Like `solver_var()` but for arbitrary shaped lists/tensors

           - `is_1d` : promise `cpm_vars` is a flat list, 
        """
        if is_1d:
            return [self.solver_var(v) for v in cpm_vars]
        if is_any_list(cpm_vars):
            return [self.solver_vars(v) for v in cpm_vars]
        return self.solver_var(cpm_vars)
```

hbierlee (commented):
Ok by me to add the additional dev script in this PR too. I hope in the future there is a way to make it generic for solvers (I've tried but couldn't find a nice way to do it yet).

cpmpy/solvers/solver_interface.py
tias: is_num is a real drag in hot paths... it calls a function and then calls 4 or 5 isinstance checks, including np.bool_ and np.integer....

all while in most cases (and nearly all cases once we've finished typing) it will be a regular python int

so, i'm confident this is the right fast past, only int here within the loop... if it doesnt match, just follow the slow path

cpmpy/solvers/solver_interface.py
tias: its an optimized fast path of a highly used function that often shows up when profiling solvers...

an additional supportive argument is the one of ignace: some solvers allow directly getting all vars of a 1d list, so then its an even faster fast path, which does not need this extra if-check.

from a typing/mypyc perspective, its the cleanest too; the original one allows arbitrary nested lists which is messy

I've not decided yet, we could even remove the last commit, but it seems worth it

hbierlee: I agree this does add an `is_1d` check for every `solver_vars` call (one level of nesting up than `solver_var` but ok), and it does still save a function call

What I still don't agree with, but you can just resolve this comment, is that 1) I don't see how the inner loop of solver_vars_1d should be any different from `solver_var`, e.g. if having isinstance(x, int) is advantageous in solver_vars_1d it should also be in `solver_var` and 2) getting multiple variables at once is great, but it should override the solver_interface methods in either case (now you should override two functions if you want to be sure that's used properly)

tias: `solver_var()` is the slow path and should cover all possible forms of integer, the fast path only quickly checks the most frequent one.

the inner loop is what makes the fast path a fast path: it only covers the 2 most frequent cases and does not leave the function for those; calling a function in Python is relatively slow

Maybe you could try both approaches with pindakaas and benchmark the difference? Bonus points for adding a native solver_vars_1d to pindakaas and seeying what that further saves ; )

cpmpy/solvers/solver_interface.py
```diff
@@ -186,6 +186,24 @@ def solver_vars(self, cpm_vars):
...
+            solver_var = self._varmap.get(cpm_var, None)
+            if solver_var is not None:
+                # fast path
+                res.append(solver_var)
+            elif isinstance(cpm_var, int):
+                # reasonably fast path
+                res.append(cpm_var)
+            else:
+                # slow path, will check the varmap again
+                res.append(self.solver_var(cpm_var))
```
IgnaceBleukx, line 204: Not sure if its worth it, but we could add an argument `is_new` solver_var to indicate whether the map should be checked or not. But checking the map is just a matter of hasing the name so probably not worth it.

IgnaceBleukx: Added it in a test, 0.2s faster for dev/time_ortools.py script

IgnaceBleukx (commented):
Fixed some issues and used some more `solver_vars_1d` opportunities in the ortools interface.
One open comment about a potential additional improvement.

I'll have a look on adding this to Exact and/or Pindakaas as well, given the limited number of supported constraints in those solvers that should be fairly straightforward.

cpmpy/solvers/solver_interface.py
tias: I wouldn't do it, what if somebody sets it wrongly, then a new var gets created, very subtle

tias:
not introducing a _1d after all, instead improving the fast path a bit in #992

PR #941, opened 2026-04-22
BoolExprLike for expr|bool

tias:


cpmpy/expressions/core.py
```diff
@@ -105,7 +105,7 @@
 T = TypeVar("T")
 ListLike: TypeAlias = Union[list[T], tuple[T, ...], np.ndarray]  # matches is_any_list() check
 ExprLike: TypeAlias = Union["Expression", int, np.integer, np.bool_]  # expression or int (incl np variants, e.g. user facing)
-
+BoolExprLike: TypeAlias = Union["Expression", bool, np.bool_]  # subtype of ExprLike (bool subtype int)
```
IgnaceBleukx, line 108: More accurate type is:
```
Union: ["GlobalConstraint", "BoolVal",  "BoolVarImpl_", "Comparison", "Operator", bool, np.bool_]
```
although Operator can be mixed still... Hmm so actually this only excludes GlobalFunction and intvarimpl

IgnaceBleukx (changes requested):
Great, that clears things up for implies already, but there are many more places where we can add the BoolExprLike type. Things that come to mind:
- XOR
- IfThenElse
- The "is_present" arg in Cumulative/NoOverlap

Do you think we should make this distinction part of the class hierarchy too? I actually think it would make a lot of sense... 
Actually, `Operator` is the only class that can have a mixed type still? And even for that one, it is still very restricted: everything but `sub`, `sum`, and `wsum` is Boolean.

cpmpy/expressions/core.py
tias: Yes, you are correct about excluding globalfunction/intvarimpl, but I also find it a bit tedious... I would be OK with just keeping Expression

(see comment lower on discussion about class hierarchy)

tias:
The PR already does it for IfThenElse and Xor?

Added a commit to also do it for is_present, good catch.


As to changing the class hierarchy... Retrospectively I agree that it should have been designed that way.

But changing the class hierarchy now is very impactful. I think if we start using the annotation already more explicitly, we already go a step in the right direction?

cpmpy/expressions/utils.py
```diff
@@ -211,18 +211,38 @@ def get_bounds(expr):
...
+    """
     if isinstance(expr, (cp.expressions.core.Expression, cp.expressions.variables.NDVarArray)):
         # both implement .implies()
         return expr.implies(other, simplify=simplify)
-    elif is_true_cst(expr):
-        return other
-    elif is_false_cst(expr):
+    elif is_true_cst(expr):  # True -> other :: other
+        if isinstance(other, cp.expressions.core.Expression):
+            return other
```
IgnaceBleukx, line 242: good catch!

IgnaceBleukx (approved):
My bad, I had missed the XOR, and ITE additions in the previous commits.

It is indeed already a step in the right direction. Changing the classes to "NumOperator" and "BoolOperator" can be something for CPMpy v2.0 : )

PR #943, opened 2026-04-23
modernize cp explanations example

OrestisLomis:


examples/advanced/cp_explanations.py
```diff
@@ -77,8 +77,8 @@ def omus(soft_constraints, soft_weights, hard_constraints=[], solver='ortools',
     for con in soft_constraints:
         # see if solver supports reification of 'con'
         try:
-            m = Model([boolvar().implies(con)])
-            SolverLookup.lookup(solver)(m).solve()
+            m = cp.Model([cp.boolvar().implies(con)])
+            cp.SolverLookup.lookup(solver)(m).solve()
```
tias: not your fault, but `cp.SolverLookup.lookup(solver)(m)` is bad/old style, it should be `cp.SolverLookup.get(solver, m)`; could you change this in this file (and future files that you encounter)?

tias (commented):
great! I just notice one old pattern that would be good to fix as part of this

PR #945, opened 2026-04-23
Linear decomposition for Table using MDD

WoutPiessens:
Building on #924, the MDD global constraint is now used to obtain a linear flow decomposition for the Table constraint.
To this end, an MDD is constructed that is equivalent with the given table.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -112,6 +112,7 @@
         ShortTable
         NegativeTable
         Regular
+        MDD
         IfThenElse
         InDomain
         Xor
@@ -133,6 +134,7 @@
 
 """
 import warnings
+from collections import defaultdict
 from typing import cast, Literal, Optional, Iterable, Any, TYPE_CHECKING
 import numpy as np
 
@@ -141,7 +143,7 @@ import cpmpy as cp
 from ..exceptions import TypeError
 from .core import Expression, BoolVal, ExprLike, ListLike
 from .variables import cpm_array, intvar, boolvar, _BoolVarImpl, _IntVarImpl, NDVarArray
-from .utils import all_pairs, is_int, is_bool, STAR, get_bounds, argvals, is_any_list, flatlist, is_num, is_boolexpr, implies
+from .utils import all_pairs, is_int, is_bool, STAR, get_bounds, argvals, is_any_list, flatlist, is_num, is_boolexpr, implies, argval
 
 if TYPE_CHECKING:
     from cpmpy.solvers.solver_interface import SolverInterface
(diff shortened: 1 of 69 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -578,6 +581,42 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+                    nxt = "snk"
+                else:
+                    prefix = prefix + (val,)
+                    if prefix not in prefix_to_node:
+                        prefix_to_node[prefix] = f"n{next_id}"
+                        next_id += 1
+                    nxt = prefix_to_node[prefix]
+
+                transition = (current, int(val), nxt)
+                if transition not in transitions:
```
IgnaceBleukx: Then make `transitions` a set instead. That way this check will be O(1) (now O(n))

cpmpy/expressions/globalconstraints.py
```diff
@@ -578,6 +581,42 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+
+         Returns:
+            tuple[list[Expression], list[Expression]]: A tuple containing the constraints representing the constraint value and the defining constraints
+         """
+
+        arr, tab = self.args
+
+        transitions : list[tuple[int | str, int, int | str]] = []
+        prefix_to_node = {}
+        next_id = 0
```
IgnaceBleukx: so now you have src, and then node_0, wouldn't it make more sense to have "src" and then "node_1"?

IgnaceBleukx: You don't even need the counter actually, you can just use `len(prefix_to_node) + 1` as the next id

cpmpy/expressions/globalconstraints.py
```diff
@@ -578,6 +581,40 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+         Returns:
+            tuple[list[Expression], list[Expression]]: A tuple containing the constraints representing the constraint value and the defining constraints
+         """
+
+        arr, tab = self.args
+
+        transitions : set[tuple[int | str, int, int | str]] = set()
+        prefix_to_node : dict[tuple[int, ...], str] = {}
+
+        for row in tab:
```
tias: it would be more consistent if you follow the style of MDD code you added in the parent PR, e.g.

source = 0
sink = -1
cnt = 1

and then your dict[int, dict[int, int]] that maps id1 to v: id2 pairs. Then you can 'walk' your MDD for each row (going from source, if id2 does not exist, create it, otherwise fetch it).

then you don't need the dict of tuples, everything is nicely in MDD space

tias (changes requested):
please merge master when parent MDD is in there, then:

tias:
please merge master in and see the comment I wrote above.

Also, this does not use any variable ordering? but that made a notable difference?

WoutPiessens:
Initially, my idea was to consider column orderings in a separate PR, but now I do think it is best to add at the very least the domain-based column ordering to this PR. So I will do that and also address the other comment

WoutPiessens:
@tias Now there is the newest version of the Table decomposition, with the domain-based variable ordering used (this was the best one in the paper), and also with the other feedback addressed. A slight disadvantage of the current way of working is that the MDD mapping object (dict[int, dict[int, int]]) is built in the Table decomposition, but it is not passed to the MDD object and therefore it has to be rebuilt in the MDD's constructor. But I don't think we would want to have this dict passed as an argument to MDD.

The tests failing now do not seem to be related to Table decomposition, but I will look into them.

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,59 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+        if len(arr) == 0:
+            return arr, tab
+
+        dom_size = [v.ub - v.lb + 1 for v in arr]
+        ordering = sorted(range(len(arr)), key=lambda i: dom_size[i])
+        arr = [arr[i] for i in ordering]
+        tab = tab[:, ordering]
+        return arr, tab
+
+    def decompose_linear(self) -> tuple[list[Expression], list[Expression]]:
```
IgnaceBleukx: I would make the ordering here an extra argument (same as we have for the cumulative constraint decomp).
Currently, we'll only support domain-based ordering, to be extended if we want to later (the argument can just be passed to the helper function above)

tests/test_transf_decompose.py
```diff
@@ -183,12 +183,7 @@ def test_decompose_linear(self):
         # test element
         cons = cp.cpm_array([10,20,30,40])[x[0]] == 8
         assert set(map(str, decompose_linear([cons]))) == \
-                            {"sum([20, 30, 40] * [a == 1, a == 2, a == 3]) == 8"}  # a == 0 is False (a in 1..3) 
-
-        # test table
-        cons = cp.Table(x, [[1,1], [2,3]])
-        assert set(map(str, decompose_linear([cons]))) == \
-                            {'((a == 1) and (b == 1)) or ((a == 2) and (b == 3))'}
+                            {"sum([20, 30, 40] * [a == 1, a == 2, a == 3]) == 8"}  # a == 0 is False (a in 1..3)
```
IgnaceBleukx, line 186: well, I would expect a test for the linear table decompose here?
The output should be the MDD constraint?

tests/test_constraints.py
```diff
@@ -162,6 +162,7 @@ def global_constraints(solver):
         elif name == "Table":
             yield cp.Table(NUM_ARGS, [[0,1,2],[1,2,0],[1,0,2]])
             yield cp.Table(BOOL_ARGS, [[1,0,0],[0,1,0],[0,0,1]])
+            yield cp.Table([cp.intvar(lb=1, ub=5), cp.intvar(lb=1, ub=2), cp.intvar(lb=1, ub=3)], [[1,1,3], [2,1,3], [3,2,3]])
```
IgnaceBleukx, line 166: Why add another case here? Is it to have different domains for the variables?
If so, best to add a comment above:

```
# different domain sizes for variables (test ordering in linear decomp)
yield ...
```

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,59 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+        """
+        Linear-friendly decomposition of the Table global constraint using an MDD, which is subsequently decomposed into linear flow constraints.
```
IgnaceBleukx, line 611: Feel free to add the reference to your paper here

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,59 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
         arr, tab = self.args
         return [cp.any([cp.all([ai == ri for ai, ri in zip(arr, row)]) for row in tab])], []
 
+    def _variable_ordering(self):
+        """
+        Orders the variables by their domain size, in order to obtain a potentially smaller MDD during decomposition.
```
IgnaceBleukx: If we make this a more general ordering heuristic, this doc should become:

```
Heuristically order the variables to obtain a better MDD
```

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,59 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+
+        Returns:
+            tuple[ListLike[Expression], (ListLike[ListLike[int]] | np.ndarray)]: The ordered array and table arguments
+        """
+        arr, tab = self.args
+        if len(arr) == 0:
+            return arr, tab
+
+        dom_size = [v.ub - v.lb + 1 for v in arr]
+        ordering = sorted(range(len(arr)), key=lambda i: dom_size[i])
```
IgnaceBleukx: gimmic: you can write this as `sorted(range(len(arr)), key=dom_size.__getitem__)`

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,59 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+
+        for row in tab:
+            current = 0
+            for i, val in enumerate(row):
+                if current not in mdd.keys():
+                    mdd[current] = {}
+                if i == len(row) - 1:
+                    nxt = -1
+                else:
+                    if val not in mdd[current].keys():
```
IgnaceBleukx: you don't need the `.keys()` here

IgnaceBleukx (changes requested):
Some comments on the code style.
Blocking issue for me is a missing test (you removed the original `decompose_linear` test as it no longer matches the current output, but the output should just be the MDD)

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,64 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
         arr, tab = self.args
         return [cp.any([cp.all([ai == ri for ai, ri in zip(arr, row)]) for row in tab])], []
 
+    def _variable_ordering(self, ordering):
```
tias: better give it a default 'ordering = "domain"' and also document the attribute and its valid values

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,64 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+        Linear-friendly decomposition of the Table global constraint using an MDD, which is subsequently decomposed into linear flow constraints.
+        Based on: Bierlee, H., Piessens, W., Stuckey, P., & Guns, T. (2026). Table Constraints for Integer Programming. In Leibniz International Proceedings in Informatics. Schloss Dagstuhl -- Leibniz-Zentrum fuer Informatik.
+
+         Returns:
+            tuple[list[Expression], list[Expression]]: A tuple containing the constraints representing the constraint value and the defining constraints
+         """
+
+        arr, tab = self._variable_ordering(ordering)
+
+        transitions: set[tuple[int | str, int, int | str]] = set()
```
tias: I don't think you need 'transitions'?

You can just have your 'mdd' map, and then just before return you materialize it:

`trans = [(id1, v, id2) for id1, pairs in mdd.items() for v, id2 in pairs]` (or something along that line)

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,64 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+        mdd: dict[int, dict[int, int]] = {}
+
+        count = 1
```
tias: lets be much more explicit:

```
root = 0
sink = -1
count = 1 # number of non-sink nodes

File: cpmpy/expressions/globalconstraints.py [hunk 14]
```diff
@@ -581,6 +581,64 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+
+        for row in tab:
+            current = 0
```
tias: current = root  # always start from root

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,64 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+            for i, val in enumerate(row):
+                if current not in mdd.keys():
+                    mdd[current] = {}
+                if i == len(row) - 1:
+                    nxt = -1
```
tias: nxt = sink

cpmpy/expressions/globalconstraints.py
```diff
@@ -581,6 +581,64 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+                else:
+                    if val not in mdd[current]:
```
tias: or smth nxt = mdd[current].get(val)
if nxt is None:
  nxt = count
  count += 1
  mdd[current][val] = nxt

tias (changes requested):
great, but some more cleanups possible

WoutPiessens:
There was at one point a failing test that I could not explain:

```
solver = 'z3'
constraint = regular([IV674, IV675, IV676],[('a', 1, 'b'), ('b', 1, 'c'), ('b', 0, 'b'), ('c', 1, 'c'), ('c', 0, 'b')],a,['c'])

    @pytest.mark.generate_constraints.with_args(numexprs)
    def test_objective_numexprs(solver, constraint):
    
        model = cp.Model(cp.intvar(0, 10, shape=3) >= 1) # just to have some constraints
        try:
            model.minimize(constraint)
            assert model.solve(solver=solver, time_limit=3)
>           assert constraint.value() < constraint.get_bounds()[1] # bounds are not always tight, but should be smaller than ub for sure
E           TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

However, I could not locally reproduce it, and when rerunning the tests in the CI, the test no longer failed. The reason why it may be related to this PR is because the Regular global decomposes into Table constraints.

IgnaceBleukx (approved):
Made some final tweaks:
- better typing of the mdd you construct in the table decomp
- made the test more robust
- add edge case for empty table
- convert numpy array to python list before iterating over it.

Should be good to merge after the tests complete

tias (approved):
great

PR #946, opened 2026-04-23
Global constraint name as class attribute

IgnaceBleukx:
Previously, the name of a global constraint was hard-coded in the super() call during init.
This PR sets the name of global constraints and functions as a class attribute, as it is not supposed to change between instances.

Originally inspired by the issues @hbierlee discovered with coverage testing for OR-Tools (missing comma in supported list).
This change also allows to catch these mistakes automatically by checking whether all strings in the supported set of a solver actually exist.

I'm not entirely sold on the idea myself, but I thought this may be a nice change, so happy to discuss/merge/close : )


cpmpy/solvers/minizinc.py
```diff
@@ -94,7 +94,7 @@ class CPM_minizinc(SolverInterface):
     https://minizinc-python.readthedocs.io/
     """
 
-    supported_global_constraints = frozenset({"alldifferent", "alldifferent_except0", "allequal",
+    supported_global_constraints = frozenset({"alldifferent", "alldifferent_except_0", "allequal",
```
hbierlee: here's a source of duplication then, still, which leads to typos: I'd have `cp.AllDifferent.name`, which avoids this issue by giving an error if you write `cp.AlDifferent`

cpmpy/expressions/globalfunctions.py
```diff
@@ -938,7 +949,7 @@ def __init__(self, arr: ListLike[ExprLike]):
         """
         if not is_any_list(arr):
             raise ValueError(f"NValue(arr) takes an array as input, not: {arr}")
-        super().__init__("nvalue", tuple(arr))
+        super().__init__(self.name, tuple(arr))
```
hbierlee, line 1187: can't we change the constraint constructor to make the name optional?

hbierlee (commented):
I think this will be an improvement, I'm really thinking that the proper way is to just have a `get_name()` function, since in most cases it can just return the class name, but will require too many changes plus adds some overhead I suppose

PR #947, opened 2026-04-24
flex jobshop example: use NoOverlapOptional 

tias:
(not in README yet because not in release yet)


examples/flexible_jobshop.py
ThomSerg, line 77: I know that this is not part of this PR, but this will also go over all subsolvers, e.g. for minizinc and pysat.

ThomSerg (approved):
Looks good, just a small comment about `for solvername in cp.SolverLookup.solvernames():`.

IgnaceBleukx:
Small suggestion: you don't actually need the end variables, and the lexico-objective can be computed on the fly with the bounds on the energy
```python
# --- Decision variables ---
start = cp.intvar(0, horizon, name="start", shape=num_alternatives)
active = cp.boolvar(name="active", shape=num_alternatives)

# --- Constraints ---
model = cp.Model()

# Each job must have one active alternative
for job_id, group in df_data.groupby('job_id'):
    model += (cp.sum(active[group.index]) == 1)

# No two active alternatives on the same machine may overlap
for mach_id, group in df_data.groupby('machine_id'):
    sel = group.index
    model += cp.NoOverlapOptional(start[sel], group['duration'].values, is_present=active[sel])

# --- Objectives ---
# Makespan: max over all active alternatives
end = start + df_data['duration']
makespan = cp.max(end)

# Total energy consumption
total_energy = cp.sum(df_data['energy'] * active)

# Minimize makespan first, then total energy
max_energy = total_energy.get_bounds()[1]
model.minimize(max_energy * makespan + total_energy)
```

examples/flexible_jobshop.py
tias: Yes, that seems to fit 'with all installed solvers'?

Or would you suggest something else? This is also on our main README, so its good to make a conscious choice on what we put.

tias:
Yes for skipping the 'end'; 

for the objective, I do prefer the common practice of just 'a weight', but its true that the comment suggests we wanted real lexico; will update.

ElFosco:
I am currently playing a bit with it, but I noticed something weird. I get an average solving time of 0.112.
If I replace:
`model += cp.NoOverlapOptional(start=start[sel], duration=group['duration'].values, is_present=active[sel])`
with
`model += cp.NoOverlapOptional(start=start[sel], duration=group['duration'].values, end=start[sel] + group['duration'].values, is_present=active[sel])`
The average solving time becomes 0.02.

The objective values that I get are the same, in theory they are the same, right?

IgnaceBleukx:
Aha very interesting! In the OR-Tools interface, we make the dummy end variables right before posting the constraint.
But indeed, when a task is present in multiple constraints, the end variables are made over and over again.

For this particular example, it is already better on master (we do something special when the duration is a constant); but nevertheless we should avoid making a new end variable if we already have it. 
Opening a PR for that

examples/flexible_jobshop.py
```diff
@@ -11,43 +12,41 @@
...
-num_alternatives = len(df_data.index)
-assert list(df_data.index) == list(range(num_alternatives)), "Index must be default integer (0,1,..)"
+# Compute maximal horizon (sum of slowest duration per job)
+horizon = df_data.groupby("job_id")["duration"].max().sum()
 
+# Decision `start[j]`: integer start time for each job `j`
+start = cp.intvar(0, horizon, shape=num_jobs, name="start")
+# Decision `end[j]`: integer end time for each job `j`
+end = cp.intvar(0, horizon, shape=num_jobs, name="end")
+# Decision `active[(j,m)]`: Per compatible combination, Boolean indicating if it is used
```
IgnaceBleukx: Hmm I find this comment confusing, if you read it too fast it seems like a 2d array. Not sure how to improve tho

examples/flexible_jobshop.py
```diff
@@ -11,43 +12,41 @@
...
-horizon = df_data.groupby('job_id')['duration'].max().sum()
-num_alternatives = len(df_data.index)
-assert list(df_data.index) == list(range(num_alternatives)), "Index must be default integer (0,1,..)"
+# Compute maximal horizon (sum of slowest duration per job)
+horizon = df_data.groupby("job_id")["duration"].max().sum()
 
+# Decision `start[j]`: integer start time for each job `j`
+start = cp.intvar(0, horizon, shape=num_jobs, name="start")
+# Decision `end[j]`: integer end time for each job `j`
+end = cp.intvar(0, horizon, shape=num_jobs, name="end")
```
IgnaceBleukx, line 25: So here we do have the end vars after all?
Ah yes because we have  a single start for each task, not duplicated like before.

Ok and then we indeed cannot just do `end = start + df_data['duration']` because its different for each mach...
Lets keep it like this then, having a single start/end var for each task is more important than omitting the ends

IgnaceBleukx (approved):
Fixed visualization and added the more modern `#!/usr/bin/env python` command on the header.
Good to go in

examples/flexible_jobshop.py
tias: fixed

tias:
Great, all set except for hearing back from @ThomSerg about the 'all solvers' comment.

examples/flexible_jobshop.py
ThomSerg: It's not really an issue for this example in particular. I just though it would be nice to also have a method to retrieve just the base names of the installed solvers. There is only `.base_solvers()`, but there you still need to do the supported check yourself:
```python
solvers = [solver_name for solver_name, CPM_solver in cp.SolverLookup.base_solvers() if CPM_solver.installed()]
```

 Right now, on a more complete install, the example will run on all 57 of these solvers:
```console
CPMpy version: 0.10.0
/home/thomserg/Documents/kul/cpmpy/cpmpy/solvers/pumpkin.py:86: UserWarning: CPMpy uses features only available from Pumpkin version >=0.3.0 but you have version 0.2.2
  warnings.warn(f"CPMpy uses features only available from Pumpkin version >=0.3.0 "
Solver                    Installed  Version        
--------------------------------------------------
ortools                   Yes        9.14.6206      
z3                        Yes        4.15.4.0       
minizinc                  Yes        0.10.0/2.9.5   
 ↪ chuffed                Yes        0.13.2         
 ↪ coin-bc                Yes        2.10.12/1.17.10
 ↪ cplex                  Yes                       
 ↪ gecode                 Yes        6.3.0          
 ↪ gurobi                 Yes                       
 ↪ highs                  Yes        1.12.0         
 ↪ cp-sat                 Yes        9.15.6755      
 ↪ scip                   Yes                       
 ↪ xpress                 Yes                       
gcs                       Yes        0.1.9          
gurobi                    No         -              
pysat                     Yes        1.8.dev26      
 ↪ cadical103             Yes                       
 ↪ cadical153             Yes                       
 ↪ cadical195             Yes                       
 ↪ cryptominisat5         Yes                       
 ↪ gluecard3              Yes                       
 ↪ gluecard4              Yes                       
 ↪ glucose3               Yes                       
 ↪ glucose4               Yes                       
 ↪ glucose42              Yes                       
 ↪ lingeling              Yes                       
 ↪ maplechrono            Yes                       
 ↪ maplecm                Yes                       
 ↪ maplesat               Yes                       
 ↪ mergesat3              Yes                       
 ↪ minicard               Yes                       
 ↪ minisat22              Yes                       
 ↪ minisat-gh             Yes                       
pysdd                     Yes        1.0.6          
exact                     Yes        2.2.1          
choco                     Yes        0.2.4          
pumpkin                   No         -              
cpo                       Yes        2.31.254/22.1.2.0
cplex                     Yes        2.31.254/22.1.2.0
pindakaas                 No         -              
highs                     No         -              
hexaly                    No         -              
rc2                       Yes        1.8.dev26      
 ↪ cadical103             Yes                       
 ↪ cadical153             Yes                       
 ↪ cadical195             Yes                       
 ↪ cryptominisat5         Yes                       
 ↪ gluecard3              Yes                       
 ↪ gluecard4              Yes                       
 ↪ glucose3               Yes                       
 ↪ glucose4               Yes                       
 ↪ glucose42              Yes                       
 ↪ lingeling              Yes                       
 ↪ maplechrono            Yes                       
 ↪ maplecm                Yes                       
 ↪ maplesat               Yes                       
 ↪ mergesat3              Yes                       
 ↪ minicard               Yes                       
 ↪ minisat22              Yes                       
 ↪ minisat-gh             Yes                       
scip                      Yes        6.0.0 
```
I though that for an example / the front page README this might be a bit too much and we might just want to run on the base solvers? Or maybe as you write this is actually what we want.

But I do think we should either add an optional filter argument / separate method for just the installed base solvers or we should document it in the "Modeling and Solving with CPMpy - Selecting a solver" section of the docs the above use of `.base_solvers()`

tias: Its what I intended, but your comment is valid; made a proposal in #1030

PR #949, opened 2026-04-24
Reduce MDD

WoutPiessens:
Continuing on #924 and #945, I added an auxiliary function _reduce() to the MDD global constraint that identifies equivalent nodes: in other words, nodes with the same set of completing suffixes.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -112,6 +112,7 @@
         ShortTable
         NegativeTable
         Regular
+        MDD
         IfThenElse
         InDomain
         Xor
@@ -133,6 +134,7 @@
 
 """
 import warnings
+from collections import defaultdict
 from typing import cast, Literal, Optional, Iterable, Any, TYPE_CHECKING
 import numpy as np
 
@@ -141,7 +143,7 @@ import cpmpy as cp
 from ..exceptions import TypeError
 from .core import Expression, BoolVal, ExprLike, ListLike
 from .variables import cpm_array, intvar, boolvar, _BoolVarImpl, _IntVarImpl, NDVarArray
-from .utils import all_pairs, is_int, is_bool, STAR, get_bounds, argvals, is_any_list, flatlist, is_num, is_boolexpr, implies
+from .utils import all_pairs, is_int, is_bool, STAR, get_bounds, argvals, is_any_list, flatlist, is_num, is_boolexpr, implies, argval
 
 if TYPE_CHECKING:
     from cpmpy.solvers.solver_interface import SolverInterface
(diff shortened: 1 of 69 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -960,6 +961,42 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
+
+    def _reduce(self):
+        """
+        Auxiliary function that reduces the original MDD by merging nodes with equivalent suffixes
+        Alters the mapping in-place.
+        """
+        arr, _ = self.args
+        substitutions = {}
+
+        for i in range(len(arr)-1, -1, -1):
```
IgnaceBleukx: can be more explicit and readable: `for i in reversed(range(len(arr))`

cpmpy/expressions/globalconstraints.py
```diff
@@ -960,6 +961,42 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
+            for equiv_nodes in groups.values():
+                # First node chosen as representative, others are merged with it
+                rep = equiv_nodes[0]
+                for node in equiv_nodes[1:]:
+                    substitutions[node] = rep
+                    self.mapping.pop(node, None)
+
+            # Mapping is redirected to representative nodes for the previous layer in the MDD
+            # This is needed to ensure that the groups for level i-1 are correctly identified in the next iteration
+            for node in (n for n in self.levels if self.levels[n] == i-1):
```
IgnaceBleukx: You don't need to do this here, you can instead loop over the transition function in the loop at line 980-981 and do the substitution there

IgnaceBleukx: That saves you another loop over all of the `level_nodes`

cpmpy/expressions/globalconstraints.py
```diff
@@ -960,6 +961,42 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
+        substitutions = {}
+
+        for i in range(len(arr)-1, -1, -1):
+            level_nodes = [n for n in self.levels if self.levels[n] == i]
+            groups = defaultdict(list) # All nodes with the same transition function are grouped together, and can be merged
+
+            for node in level_nodes:
+                transition_function = self.mapping[node]
+                signature = tuple(sorted(transition_function.items()))
+                groups[signature].append(node)
```
IgnaceBleukx, line 992: add a comment why you sort this?

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: maybe also a comment or smt: "loop backwards over MDD levels from sink to root" or smt?

IgnaceBleukx (changes requested):
Good, some cleanups and improvements for readibility possible.

Please also add a small tests `tests/test_globalconstraints.py` to ensure the reduce step is working correctly.

cpmpy/expressions/globalconstraints.py
```diff
@@ -960,6 +961,45 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
+                # Ordered tuple of (value, destination node) pairs, serves as a unique signature for equivalent transition functions
+                signature = tuple(sorted(transition_function.items()))
+                groups[signature].append(node)
+
+            for equiv_nodes in groups.values():
+                # First node chosen as representative node, others are merged with it
+                rep = equiv_nodes[0]
+                for node in equiv_nodes[1:]:
+                    substitutions[node] = rep
+                    self.mapping.pop(node, None)
```
tias, line 999: OK, now I'm starting to question a lot of things...

I see you delete it here from `self.mapping`, but `self.args` is `(arr, transitions)` and there is no deleting from 'transitions' here..

so then, are we storing duplicate information, e.g. the constructor is storing `self.mapping` and `self.levels`, but if this can be recreated from 'transitions' then there is actually no need to store it in our constraint? We can just create it when we need it (e.g. with a helper funciton, or just 'there' if that is the only place that needs it)...

because now, you are reducing the mapping, but your self.levels and your 'transitions' tables are untouched, which is inconsistent (and could even lead to unnecessary Booleans, or unnecessary edges in a propagator)

tias (commented):
from what I see and realize now, the constructor does too much work...

it should just store (arr, transitions), then self.root_node,
and then it can compute self.sink_node by taking set(id2) - set(id1)?

and then the `_get_complete_mdd()` would become `get_mapping(complete=True)` which would create those (mapping, levels) for decompose?

so then the question is reduce, the goal is to rewrite/remove rows from the table, right? can that be done more directly?

If not, then at the least I think you should make sure that reduce keeps self.levels and self.mapping correct too?

tias:
I don't want to complicate things... but I don't like double information either.

I checked the type: 'transitions' does not actually contain variables, it never will, its constants. `self.args` is for things that can contain variables.

So we actually don't need to store 'transitions' I think... we can just build self.mapping (and self.levels) and then keep most of the code the same (we just walk over mapping most of the time I assume), and then the reduce code just needs to make sure that removed nodes are removed from self.levels too?

WoutPiessens:
The way I see it is as follows: transitions is only an input argument to the MDD global, and is used to build self.mapping in the constructor. After that, transitions is indeed no longer needed, and it's never used outside of the constructor. self.mapping is the internal representation of the MDD, which changes when calling reduce(). But indeed, removed nodes do need to be removed from self.levels too since that is used outside of the constructor.

cpmpy/expressions/globalconstraints.py
```diff
@@ -939,13 +940,15 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
                 raise TypeError(
                     f"The second argument of an MDD constraint should be a list of transitions ({_node_type}, int, {_node_type})")
 
-        super().__init__("mdd", (array, transitions))
+        super().__init__("mdd", (array,))
+        self.transitions = transitions
```
tias: I don't think you need transitions... the choco global can just reconstruct it from self.mapping...

keeping it means you sitll have duplicated (wrong) information after reduce...

cpmpy/expressions/globalconstraints.py
```diff
@@ -939,13 +940,15 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
         self.root_node = transitions[0][0] if start is None else start
         self.mapping: dict[int | str, dict[int, int | str]] = defaultdict(dict)  # mapping from source node and transition value to destination node
         for id1, v, id2 in transitions:
             self.mapping[id1][v] = id2
 
         self.levels = {self.root_node: 0}
+        self.nodes = [self.root_node]
```
tias: I don't think you need this either... its self.levels.keys()?

cpmpy/expressions/globalconstraints.py
```diff
@@ -954,12 +957,53 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
+        """
+        Auxiliary function that reduces the original MDD by merging nodes with equivalent suffixes
+        Alters the mapping in-place.
+        """
+        arr = self.args[0]
+        substitutions = {}
+
+        # Loop backwards over MDD levels, from sink to root node
+        for i in reversed(range(len(arr))):
+            level_nodes = [n for n in self.levels if self.levels[n] == i]
```
tias: there is a better pattern for this kind of access to both the key and the value in python:
`[n for (n, lvl) in self.levels.items() if lvl == i]`

tests/test_transf_decompose.py
```diff
@@ -200,9 +200,10 @@ def test_decompose_linear(self):
         assert len(decomp) == 1
         assert isinstance(decomp[0], cp.MDD)
         # need more thorough test, order of transistions is not fixed
-        arr, transitions = decomp[0].args
+        arr = decomp[0].args[0]
+        transitions = decomp[0].transitions
```
tias: this will become `sum(len(valdict) for valdict in decomp[0].values())` without transitions

tias (changes requested):

tests/test_transf_decompose.py
WoutPiessens: This didn't work, but I will construct the transitions from the MDD mapping using a list comprehension

tias (approved):
great

cpmpy/expressions/globalconstraints.py
```diff
@@ -939,7 +940,7 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
                 raise TypeError(
                     f"The second argument of an MDD constraint should be a list of transitions ({_node_type}, int, {_node_type})")
 
-        super().__init__("mdd", (array, transitions))
+        super().__init__("mdd", (array,))
```
IgnaceBleukx, line 943: Why not just save `tuple(array)` as the args? No need to wrap it in another tuple

IgnaceBleukx: Also, you'll need to update the `__repr__` method. Now it will only print MDD(array)

tests/test_globalconstraints.py
```diff
@@ -639,6 +639,28 @@ def test_regular(self):
...
+        transitions = [("src", 0, "a1"), ("src", 1, "b1"), ("src", 2, "c1"), ("a1", 1, "a2"), ("b1", 1, "b2"), ("c1", 2, "c2"),
+                       ("a2", 0, "snk"), ("b2", 0, "snk"), ("c2", 0, "snk")]
+
+        start = "src"
+
+        solutions = [(0,1,0), (1,1,0), (2,2,0)]
+        reduced_sols = set()
+        non_reduced_sols = set()
+
+        reduced_model = cp.Model(cp.MDD(x, transitions, start=start, reduce=True))
```
IgnaceBleukx: Add a check that it is indeed reduced? E.g., expected output of the MDD?

IgnaceBleukx (approved):
Great, some final small comments that need to be addressed before merging (I'll already approve otherwise I have to re-review...)

cpmpy/expressions/globalconstraints.py
tias: (array,) is imho the correct way; the array is the first (and only) argument; otherwise there are n-ary arguments (and it gets modified if its an array).

I consider it a mistake that many of our other expressions store n-ary arrays in args

PR #950, opened 2026-04-24
Test Native MUS function

OrestisLomis:
Not sure if this is the correct way to do things, but apparently the `native_mus` tests are never run according to Henk's code cov and I confirm this by just putting an` assert False` in the function. Since the test is for exact only I wrote `@pytest.mark.requires_solver("exact")` above the testclass, which probably is related to the issue.

@ThomSerg do you know more about this?

### DIFF
--- a/tests/test_tools_mus.py
+++ b/tests/test_tools_mus.py
@@ -87,7 +87,7 @@ class TestMus:
 # add solvers that implement the native_mus method
 @pytest.mark.requires_solver("exact")       
 class TestNativeMus(TestMus):
-    # use solver from conftest.py
+    # use modern hook to add solver argument to setup method
     @pytest.fixture(autouse=True)
     def setup_method(self, solver):
         self.mus_func = lambda soft, hard=[], solver=solver: mus_native(soft, hard=hard, solver=solver)
--- a/tests/test_tools_mus.py
+++ b/tests/test_tools_mus.py
@@ -4,7 +4,7 @@ import cpmpy as cp
 from cpmpy.tools import mss_opt, marco, OCUSException
 from cpmpy.tools.explain import mus, mus_naive, quickxplain, quickxplain_naive, optimal_mus, optimal_mus_naive, mss, mcs, ocus, ocus_naive, mus_native
 
-
+@pytest.mark.requires_solver("ortools")
 class TestMus:
     def setup_method(self):
         self.mus_func = mus
@@ -25,7 +25,7 @@ class TestMus:
         assert set(self.mus_func(cons)) == set(cons[:3])
         assert set(self.naive_func(cons)) == set(cons[:3])
 
-    def test_bug_191(self):
+    def test_bug_191(self, solver):
         """
         Original Bug request: https://github.com/CPMpy/cpmpy/issues/191
         When assum is a single boolvar and candidates is a list (of length 1), it fails.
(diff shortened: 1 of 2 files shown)


hbierlee:
As discussed offline, it's indeed the missing `solver` arg from the test functions. For me it makes sense to just add that argument to all the MUS tests, as they have now become solver-specific.

OrestisLomis:
The mus functions aren't solver specific so I do not think that is the best solution; instead I only added the solver argument to the TestNative, which after some trial and error seems to be working now

tests/test_tools_mus.py
```diff
@@ -83,13 +83,16 @@ def test_wglobal(self):
...
         assert not cp.Model(ms).solve()
         # self.assertEqual(set(self.naive_func(cons)), set(cons[:2]))
+
+# add solvers that implement the native_mus method
 @pytest.mark.requires_solver("exact")       
-class TestNativeMusExact(TestMus):
-    def setup_method(self):
-        self.mus_func = lambda soft, hard=[], solver="exact": mus_native(soft, hard=hard, solver="exact")
+class TestNativeMus(TestMus):
+    # use solver from conftest.py
```
hbierlee: doesn't really explain to me how this autouse works? Since it's exceptionally used, I think the comment should be more explicit

tests/test_tools_mus.py
```diff
@@ -34,7 +34,7 @@ def test_bug_191(self):
         hard = [~bv]
         soft = [bv]
 
-        mus_cons = self.mus_func(soft=soft, hard=hard, solver="ortools") # crashes
+        mus_cons = self.mus_func(soft=soft, hard=hard) # crashes
```
hbierlee: why this change?

hbierlee (changes requested):
Well, maybe we need a third opinion, but I think that the other MUS tests might as well be good to parametrize for all solvers. They do have `solver` arguments, so we're hoping that any CPMpy solver would be able to compute the MUS, so I think this is a good opportunity to test that assumption. It would also simplify the TestNativeMus to not have the exceptional autouse decorator, but just `requires_solvers` meaning: unlike the other MUS algorithms, this one requires exact.

tests/test_tools_mus.py
OrestisLomis: since I made `TestNativeMus` take the actual solver argument instead of hardcoding it to exact this needed to change otherwise ortools will be given as argument even for the `native_mus` tests

tests/test_tools_mus.py
OrestisLomis: I don't fully understand it either to be completely honest, however from my understanding `def steup_method` is a legacy hook without the decorator and essentially only has `self` and `method` as its arguments. I.e. it does not have the full scope of the test environment. The decorator allows you to add the solver argument correctly. I will update the comment.

OrestisLomis:
I'm not sure I agree with that suggestion, but either way it is something for a different PR then. Because while the solver argument exists in the MUS function it still only accepts solvers which allow solving under assumptions. Besides the goal of the tests is to tests the MUS algorithms which are agnostic of the solver (or solving technology in general) and hence I don't see why we would duplicate these tests for all solvers.

hbierlee (commented):
Ok, but maybe the best solution is to then still parametrize the tests, but add `requires_solver("ortools")` to  `TestMUS`, and `requires_solver("exact")` to the `TestNativeMUS`? That's just the regular way to indicate which solver should run which tests, and my preference over the `autouse` addition.

Then we can easily add a few more (assumption-based) solvers to the `TestMUS` if we wish (in principal I agree its duplication because we're testing the MUS code, so but in practice the assumptions are thinly tested `test_solverinterface`; fine to have a separate mini PR for that)

 If you still do not agree, then we need a third opinion

tias:
I don't like the autouse much either, would prefer something more explicit.

For Thomas to comment on, as he knows the parameterization best (can also be at the next dev meeting)

OrestisLomis:
I do like Henk's final suggestion, also sounds like something that can be added in this PR in the end. I'm just not sure exactly how to do it without the autouse to actually have the solver available in the setup (where the mus_func is defined). Summoning @ThomSerg again.

hbierlee:
Doesn't seem easy to add the solver fixture to `setup_class`, so I think the easiest way around that is if you make `self.mus_func = mus` and pass the solver in the test methods themselves (some duplication but I think avoiding it is not worth it)

OrestisLomis:
Closed because of #1002

PR #952, opened 2026-04-24
Support and fix the `regular` constraint for MiniZinc

hbierlee:
The `regular` constraint was not in the support globals of MiniZinc, even though it was implemented. The implementation needed to be fixed as well. The convert_expression now can also handle strings, which are part of the MiniZinc language. Reification is not supported (https://github.com/MiniZinc/libminizinc/blob/ce06f39e8a7667721ed7410745d82e896bd9b16f/share/minizinc/std/fzn_regular_reif.mzn#L4).


cpmpy/solvers/minizinc.py
```diff
@@ -891,24 +892,39 @@ def zero_based(array):
...
-            # where transitions_table is a 2D array
-            array, transitions, start, accepting = expr.args
+            # MiniZinc: `regular(array[int] of var int: x, array[int,int] of opt int: d, int: q0, set of int: F)`
+            # We map CPMpy's named states to 1-indexed integers.
+            # Example:
+            #   CPMpy:   `Regular([IV0,IV1,IV2], [('a',1,'b'),('b',1,'c'),('b',0,'b'),('c',1,'c'),('c',0,'b')], 'a', ['c'])`
+            #   MiniZinc: `constraint regular([IV0,IV1,IV2], array2d(1..3, 0..1, [<>,2,2,3,2,3]), 1, {3})`
+            #            note: `d` is a 2D array `[|<>,2|2,3|2,3|]` with rows=states, cols=values
+
+            reg = cast(Regular, expr)
```
tias: it would be cleaner that you do an 'elif isinstance(expr, Regular)', then all is good and you dont need the cast... that additional isinstance here so low in the cases should have minimal effect.

WoutPiessens (approved):
The implementation looks great, one small thing: a transition dictionary object is already built in the constructor of Regular, so it can be used here.

tias (approved):
merged in master, all good if the tests pass

tias:
required a merge, esp the imports...

please have a short look Wout, also wait for the tests to finish...

WoutPiessens (commented):
"regular" was missing in the set of supported global constraints (likely due to a merge conflict). I will add it back: if the tests pass I will merge this.

PR #957, opened 2026-04-24
floatsum to still enable floats in the objective

tias:
also deprecates the use of floats in mul

if this looks good, we can apply it to the other solvers that support floats before merging (scip, highs, gurobi, cplex; maybe rc2, z3, choco, exact, hexaly?)


cpmpy/expressions/globalfunctions.py
```diff
@@ -1076,3 +1085,61 @@ def get_bounds(self) -> tuple[int, int]:
...
+    Does not inherit from Expression because it is objective only and has float .value()
+    Basically it is breaking all design rules of CPMpy...
+
+    Accepts only (numpy) floats as coefficients, and Expressions as terms.
+    """
+    name: Final = "floatsum"
+    coeffs: np.ndarray
+    terms: NDVarArray
+
+    def __init__(self, coeffs: ListLike[float|np.floating], terms: ListLike[Expression]):
```
IgnaceBleukx: Shouldn't the terms be `_NumVarImpl`? I suppose we don't do flattening as it is not an expression? So there cannot be subexpressions as part of the terms here

cpmpy/expressions/globalfunctions.py
```diff
@@ -1076,3 +1085,61 @@ def get_bounds(self) -> tuple[int, int]:
...
+            raise TypeError("FloatSum(coeffs, terms) expects at least one term")
+
+    def __repr__(self) -> str:
+        return f"FloatSum({list(self.coeffs)}, {list(self.terms)})"
+
+    def value(self) -> Optional[float]:
+        vals = argvals(self.terms)
+        if any(v is None for v in vals):
+            return None
+        return sum(c * v for c, v in zip(self.coeffs, vals))
```
IgnaceBleukx: use numpy instead? `int(np.dot(self.coeffs, vals))`

cpmpy/expressions/globalfunctions.py
IgnaceBleukx: Ah I see you do get_or_make_var in the solver interface, then its ok I suppose. Or is it? What about unsafe expressions? Does that sill work?

E.g., `FloatSum([1,2,3], [x,y, x/y])` where `y` covers 0.

Ah yes because it is toplevel, so that should be fine.

cpmpy/solvers/cplex.py
```diff
@@ -301,22 +302,35 @@ def objective(self, expr, minimize=True):
...
-                                                      supported_reified=self.supported_reified_global_constraints,
-                                                      csemap=self._csemap)
-        obj, flat_cons = flatten_objective(obj, csemap=self._csemap)
-        obj, self._obj_offset = only_positive_bv_wsum_const(obj) # remove negboolviews
-
-        self.add(safe_cons + decomp_cons + flat_cons)
-
-        # make objective function or variable and post
-        cplex_obj = self._make_numexpr(obj)
+        if isinstance(expr, FloatSum):
```
IgnaceBleukx, line 320: Why not put this part in the `_make_numexpr` helper function? It will not be triggered by the `add` code as we already have check there

cpmpy/solvers/z3.py
```diff
@@ -256,7 +257,13 @@ def solve(self, time_limit:Optional[float]=None, assumptions:Optional[Iterable[_
             # translate objective, for optimisation problems only
             if self.has_objective():
                 obj = self.z3_solver.objectives()[0]
-                self.objective_value_ = sol.evaluate(obj).as_long() 
+                obj_val = sol.evaluate(obj)
+                if z3.is_int_value(obj_val):
+                    self.objective_value_ = obj_val.as_long()
+                elif z3.is_rational_value(obj_val):
+                    self.objective_value_ = obj_val.numerator_as_long() / obj_val.denominator_as_long()
```
IgnaceBleukx: Document this?

IgnaceBleukx (changes requested):
Overall seems fine; it's very different from our expressions, but that is the point, of course.

There is quite some duplication of handling the FloatSums in the solver interfaces; maybe it's better to move it to the `_make_numexpr` helper functions?

Main thing for me is some missing docs: e.g., in the template, and in the main modeling docs, making it clear `FloatSum` and `DirectConstraint` are the ONLY things where you are allowed to use floats.

A follow-up PR should probably go through all of the `is_num` checks and replace them with `is_int` instead?

tias:
removing the is_num's, yes, but its expansive... some are these are in the constructors that we are still cleaning up anyway, so I think after properly typing the constructors is best

cpmpy/expressions/globalfunctions.py
tias: no yes, non-nested is the intention! will fix

cpmpy/expressions/globalfunctions.py
tias: idd

cpmpy/expressions/globalfunctions.py
tias: its not toplevel its nested. it does look like I intended it to be expressions, but we better do variables, this is supposed to be an exceptional class anyway

cpmpy/solvers/cplex.py
tias: no, its objective only, I think it should be in the objective() functions... I don't really mind the duplicate codes at this point

tias:
I refactored it in parallel...

It was failing on negboolviews for some solvers, and there was also the open thing of constants.

Its still a bit a non-well-typed mess: in the current form we can let the solverobject overwrite `objective()` to also accept FloatSum (I would not do this in Model... so the tests directly need to call the solverobject, we would have to document this). This is allowed by the type mechanism.

but then, the current code lets `objective_value()` return int|float, but that is not allowed by the type mechanism... so we would have to convert it to int, and we would have to document that you need to read the objective straight from the FloatSum? (for finite-precision/comparison reasons, this might actually be OK)

(both are not actually visible yet because our typing has not advanced that much yet)

tias:
- refactored to have fewer duplicate code (and also store a constant, and be able to not return negboolviews)
- typed model.minimize/maximize/objective_value and friends as int
- floatsum only in specific solver interfaces
- added docs in multiple places, including FLOBJ as new capability in capability table

please re-review

cpmpy/expressions/globalfunctions.py
```diff
@@ -51,6 +51,16 @@ def __init__(self, args):
             def decompose(self):
                 return (self.args[0] + self.args[1]), []  # the decomposition
 
+    Objective-only :class:`FloatSum`
+    -------------------------------
+
+    :class:`FloatSum` is **not** an Expression nor a GlobalFunction.
+    It is only supported by some solvers (ortools, gurobi, cplex, scip, z3, minizinc, highs, hexaly),
+    and must be passed directly to that solver's :meth:`~cpmpy.solvers.solver_interface.SolverInterface.minimize`
+    / :meth:`~cpmpy.solvers.solver_interface.SolverInterface.maximize`
```
IgnaceBleukx: For consistency: I think elsewhere in the code base we use :func: instead of :meth:?

cpmpy/expressions/globalfunctions.py
```diff
@@ -1091,3 +1110,94 @@ def get_bounds(self) -> tuple[int, int]:
...
+
+    Accepts only (numpy) floats as coefficients, decision variables (including NegBoolView) as terms,
+    and an optional float constant term (default ``0.0``).
+    """
+    name: Final = "floatsum"
+    coeffs: np.ndarray
+    vars: NDVarArray
+    const: float
+
+    def __init__(self, coeffs: ListLike[float|np.floating], vars: ListLike[_NumVarImpl], const: float|np.floating = 0.0):
```
IgnaceBleukx, line 1141: docstring for the intit function?
(I suppose just the past paragraph in the docstring of the class , but witht the "Arguments:" and "Return" thingies?)

cpmpy/expressions/globalfunctions.py
```diff
@@ -1091,3 +1110,94 @@ def get_bounds(self) -> tuple[int, int]:
...
+            # typing is wrong: numpy preserves our ndarray subclass
+            self.vars = cast(NDVarArray, flat)
+
+        if self.coeffs.size != self.vars.size:
+            raise TypeError(f"FloatSum(coeffs, terms) expects equal lengths, got {self.coeffs.size} coefficients and {self.vars.size} terms")
+        if self.coeffs.size == 0:
+            raise TypeError("FloatSum(coeffs, terms) expects at least one term")
+
+    def __repr__(self) -> str:
+        if self.const:
```
IgnaceBleukx: At first I thought this was checking for None, but its checking if it's 0, maybe add a comment?

cpmpy/expressions/globalfunctions.py
```diff
@@ -1091,3 +1110,94 @@ def get_bounds(self) -> tuple[int, int]:
...
+            return f"FloatSum({list(self.coeffs)}, {list(self.vars)}, constant={self.const})"
+        return f"FloatSum({list(self.coeffs)}, {list(self.vars)})"
+
+    def value(self) -> Optional[float]:
+        vals = argvals_intexpr(self.vars)
+        if vals is None:
+            return None
+        return float(np.dot(self.coeffs, vals) + self.const)
+
+    def components(self, negbool=False) -> tuple[np.ndarray, NDVarArray, float]:
```
IgnaceBleukx: Can/should we type the NDVarArray that is returned here with `_NumVarImpl`?

cpmpy/expressions/globalfunctions.py
```diff
@@ -1091,3 +1110,94 @@ def get_bounds(self) -> tuple[int, int]:
...
+    def __gt__(self, other: object) -> NoReturn: self._raise_objective_only()
+    def __ge__(self, other: object) -> NoReturn: self._raise_objective_only()
+    def __add__(self, other: object) -> NoReturn: self._raise_objective_only()
+    def __radd__(self, other: object) -> NoReturn: self._raise_objective_only()
+    def __sub__(self, other: object) -> NoReturn: self._raise_objective_only()
+    def __rsub__(self, other: object) -> NoReturn: self._raise_objective_only()
+    def __mul__(self, other: object) -> NoReturn: self._raise_objective_only()
+    def __rmul__(self, other: object) -> NoReturn: self._raise_objective_only()
+    def __neg__(self) -> NoReturn: self._raise_objective_only()
+    def __abs__(self) -> NoReturn: self._raise_objective_only()
```
IgnaceBleukx, line 1222: Missing a couple here: pow, truediv, floordiv, mod
https://www.pythonmorsels.com/every-dunder-method/

cpmpy/expressions/globalfunctions.py
```diff
@@ -1091,3 +1110,94 @@ def get_bounds(self) -> tuple[int, int]:
...
+                    ws.append(-w)
+                    vs.append(v._bv)
+                    const += w
+                else:
+                    ws.append(w)
+                    vs.append(v)
+            return np.asarray(ws, dtype=float), cpm_array(vs), const
+
+    def _raise_objective_only(self) -> NoReturn:
+        raise TypeError("FloatSum is objective-only. Use it directly in solver minimize()/maximize().")
```
IgnaceBleukx: Its more specific than that, maybe: `FloatSum cannot be used as a subexpression` or something?

(Btw: we probably want something similar implemented for DirectConstraint)

cpmpy/solvers/cplex.py
```diff
@@ -251,10 +255,10 @@ def solve(self, time_limit:Optional[float]=None, **kwargs):
             # set _objective_value
             if self.has_objective():
                 obj_val = self.cplex_model.get_objective_expr().solution_value
-                if round(obj_val) == obj_val: # it is an integer?:
+                if round(obj_val) == obj_val:  # its integer
                     self.objective_value_ = round(obj_val)
-                else: #  can happen with DirectVar or when using floats as coefficients
-                    self.objective_value_ = float(obj_val)
+                else:  # FloatSum objective, must be read through FloatSum.value()
```
IgnaceBleukx: Are we sure of this? Is there no other way the solver can return a float as objective value? (e.g., due to precision issues? 
I think it's safer to explicitely test if we posted a FloatSum

cpmpy/solvers/cplex.py
```diff
@@ -304,22 +314,28 @@ def objective(self, expr, minimize=True):
...
-        obj, self._obj_offset = only_positive_bv_wsum_const(obj) # remove negboolviews
-
-        self.add(safe_cons + decomp_cons + flat_cons)
-
-        # make objective function or variable and post
-        cplex_obj = self._make_numexpr(obj)
+        if isinstance(expr, FloatSum):
+            ws, vs, const = expr.components()
+            self.user_vars.update(vs)  # save user variables
+            self._obj_offset = const
```
IgnaceBleukx, line 323: We don't do anything with this offset in the case of floatsum, right? Its computed through the FloatSum.value function.

BUt actually, for the non-floatsum case, we should add this to the objective value after solving...

cpmpy/expressions/globalfunctions.py
IgnaceBleukx: Also `allow_negbool` would be more intuitive

cpmpy/solvers/ortools.py
```diff
@@ -278,10 +281,10 @@ def solve(self, time_limit:Optional[float]=None, assumptions:Optional[Iterable[_
...
             if self.has_objective():
                 ort_obj_val = self.ort_solver.objective_value
-                if round(ort_obj_val) == ort_obj_val: # it is an integer?
-                    self.objective_value_ = round(ort_obj_val)  # ensure it is an integer
-                else: # can happen when using floats as coeff in objective
-                    self.objective_value_ = float(ort_obj_val)
+                if round(ort_obj_val) == ort_obj_val:  # its integer
+                    self.objective_value_ = int(ort_obj_val)
+                else:  # FloatSum objective, must be read through FloatSum.value()
+                    self.objective_value_ = None
```
IgnaceBleukx, line 289: Yes so for OR-Tools its probably fine to do it like this, but for the MIP solvers we should do an extra whether we have posted a FloatSum I think

IgnaceBleukx (changes requested):
Good, almost there I think.

The main question mark for me is the objective value return.
I.e., are we sure that we will always have a float as objective value if and only if we have a FloatSum objective?

I tweaked some tests, but overall ok; docs also look fine.

cpmpy/expressions/globalfunctions.py
tias: Had to look up/ask Sphinx conventions: solver minimize/maximize/objective_value are class methods of the SolverInterface class, so :meth: with full qualified paths. :func: kept only for module-level functions (decompose_in_tree, boolvar, etc.). So kept meth here (with a qualified path update)

tias: also update the docs/developers.md (not on readthedocs public yet)

cpmpy/expressions/globalfunctions.py
tias: `allow_negbool`: agreed

I don't know if there is a way to type the NDVarArray as containing _NumVarImpls? it contains objects from dtype perspective... not opening that can of worms : )

cpmpy/expressions/globalfunctions.py
tias: sure

cpmpy/expressions/globalfunctions.py
tias: I'm leaving the fun of doing it for DirectConstraint to others : )

cpmpy/solvers/cplex.py
tias: oh, good catch. We did it in SolveAll(), but not yet in Solve(), fixed

cpmpy/solvers/cplex.py
tias: right... that might happen; so we actually have to store the cpmpy object like we do for exact... refactoring

IgnaceBleukx (approved):
Had to do a final cleanup for Hexaly as there was some duplicated logic (`is_satisfaction` and `objective_ is None`).

It should be good to go now, I think.
Final thing I'm wondering is: should we raise some warning/error if the user provides a FloatSum to `Model.objective`? I know it's "enforced" by the type system, but it might be good to also explicitly point the user to the docs describing how to do floatsum?

tias:
I've added an assert that the argument to Model.objective is an Expression

I don't want to mention FloatSum there... its a special thing for special occasions, not something we want to put centerfold.

tias:
A few final non-essential fixes added (some from the just merged negate before decomp).

I think its finally ready!

IgnaceBleukx (approved):
Great I think we're there now!
Now all of our expressions are float-free based on the type system, and some hard checks such as in Multiplication.

In theory, a user can of course still use floats (e.g., `Operator("wsum", [[0.1,0.2,0.3], [x,y,z]])`) but such things are not allowed by the type system, and will probably fail later in the system : )

PR #958, opened 2026-04-24
listlike: switch from invariant list[T] to covariant Sequence[t]

tias:
so list[Expression] was not a ListLike[ExprLike]...

because list is mutable, and mutable collections are type 'invariant'; what we want is 'covariant' behaviour which would treat Expression as a subclass of ExprLike
https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance

The Sequence type covers tuple and list (and others, like iterators I think) and it is covariant, so that resolves that (and accepts other things we also accept, because we just use generic sequence (iterate, select element) functions on it

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -103,7 +103,7 @@ from ..exceptions import TypeError
 
 # Common typing helpers
 T = TypeVar("T")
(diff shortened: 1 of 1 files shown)


IgnaceBleukx (approved):
Good to go. In a next PR we should check where we use ListLikes and do unnecessary `new_lst[T] = lst` assertions to make mypy happy.

PR #959, opened 2026-04-26
simplify_boolean in the new pattern (9s -> 6s flatten)

tias:


cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
-                expr.update_args(newargs)
-            newlist.append(expr)
-        elif is_bool(expr):  # unlikely base-case (Boolean constant)
-            newlist.append(int(expr) if num_context else BoolVal(expr))
-        elif isinstance(expr, list):  # nested list in args (like for wsum)
-            newlist.append(simplify_boolean(expr))
-        else:  # variables/constants/direct constraints
+        changed, newexpr = _simplify_boolean_expr(expr, num_context=num_context)
+        if changed:
+            assert not isinstance(newexpr, int)
```
IgnaceBleukx, line 79: You can also assert this by using an overload on the _simplify_boolean_expr header?
If num_context is False, then return type is bool/Expression.

I would expect the `num_context` on the main transformation to be deprecated? Users should only call this one at the toplevel, where everything is bool

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+                elif is_false_cst(a):
+                    changed = True
+                    if newargs is None:
+                        newargs = list(args[:i])
+                elif newargs is not None:
+                    newargs.append(a)
+
+            if changed:
+                if newargs is None:
+                    newargs = list(args)
```
IgnaceBleukx, line 127: Add comment on when this can happen? E.g., `# can happen when args were simplified recursively`

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+                _, newcond = _simplify_boolean_expr(recurse_negation(cond), num_context=False)
+                # always changed, because negated cond
+                return True, newcond
+
+            if changed:
+                return True, cond.implies(bool_expr)
+            return False, expr
+
+        elif expr.name == "not":
+            if isinstance(args[0], _BoolVarImpl):
```
IgnaceBleukx: This case should not happen anymore, right? We should have already pushed down negation. Only case left here should be `not(GlobalConstraint)`

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+            weights, vars = args
+            newvars = None
+            for i, v in enumerate(vars):
+                if is_bool(v):
+                    if newvars is None:
+                        newvars = list(vars[:i])
+                    changed = True
+                    newvars.append(int(v))
+                elif newvars is not None:
+                    newvars.append(v)
```
IgnaceBleukx, line 189: dummy `else` case here would be useful I think: 
```
else: # still operating on original vars
    pass
```

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
             newlist.append(expr)
     return newlist
+
+
+def _simplify_boolean_expr(expr: Expression, num_context=False) -> tuple[bool, Expression|int]:
+    changed = False
+    if isinstance(expr, Operator):
+        args = expr.args
+        if expr.has_subexpr():
+            changed, rec_args = _simplify_boolean_args(args, num_context=not expr.is_bool())
```
IgnaceBleukx, line 109: Add comment on "hack" that all operators with Boolean return type take Boolean arguments too?

cpmpy/transformations/normalize.py
IgnaceBleukx: Hmm according to the coverage tool, it still fires -- not sure from where

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+            elif rhs == 0:
+                if name == "!=" or name == ">":
+                    return True, lhs
+                if name == "==" or name == "<=":
+                    return True, recurse_negation(lhs)
+                if name == "<":
+                    return True, (0 if num_context else BoolVal(False))
+                if name == ">=":
+                    return True, (1 if num_context else BoolVal(True))
+            elif 0 < rhs < 1:
```
IgnaceBleukx, line 258: This case is deprecated now, should also be removed from the table above, and the tests

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+                newargs = list(args)
+            newexpr = copy.copy(expr)
+            newexpr.update_args(newargs)
+            return True, newexpr
+
+    elif isinstance(expr, Comparison):
+        changed, (lhs, rhs) = _simplify_boolean_args(expr.args, num_context=True)
+        name = expr.name
+        if is_num(lhs) and is_boolexpr(rhs):  # flip arguments of comparison to reduct nb of cases
+            if name == "<":    name = ">"
```
IgnaceBleukx: Style: put on newline? (Makes ruff linter happy)

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+            elif name == ">":
+                name = "<"
+            elif name == "<=":
+                name = ">="
+            elif name == ">=":
+                name = "<="
+            lhs, rhs = rhs, lhs
+            changed = True
+
+        if changed or (is_num(lhs) and is_num(rhs)):
```
IgnaceBleukx, line 285: Do you need the is_num checks here? I can't see how lhs or rhs can become a constant, if there is no `changed`? Except if the user explicitely wrote it

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+            newexpr = eval_comparison(name, lhs, rhs)
+            if is_bool(newexpr): # Result is a Boolean constant
```
IgnaceBleukx: Can we get away with just `isinstance(newexpr, bool)` instead? I think eval_comparison always returns a Python bool or an expression

IgnaceBleukx: (quick test on typing `eval_comparison` reveals it is not that trivial to implement...)

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+                return True, (int(newexpr) if num_context else BoolVal(newexpr))
+            else: # Result is an expression
+                return True, newexpr
+
+    elif isinstance(expr, (GlobalConstraint, GlobalFunction)):
+        rec_changed, rec_args = _simplify_boolean_args(expr.args, num_context=False)
+        if rec_changed:
+            newerexpr = copy.copy(expr)
```
IgnaceBleukx: `new_global` or smt? `newerer` is not very informative : )

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+
+        if changed or (is_num(lhs) and is_num(rhs)):
+            newexpr = eval_comparison(name, lhs, rhs)
+            if is_bool(newexpr): # Result is a Boolean constant
+                return True, (int(newexpr) if num_context else BoolVal(newexpr))
+            else: # Result is an expression
+                return True, newexpr
+
+    elif isinstance(expr, (GlobalConstraint, GlobalFunction)):
+        rec_changed, rec_args = _simplify_boolean_args(expr.args, num_context=False)
```
IgnaceBleukx, line 294: The original code had a todo here, best to put it back?
```python
# TODO: how to determine which args are bool/int?
```

cpmpy/transformations/normalize.py
IgnaceBleukx: We can raise a Value Error if we want here

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+        if rec_changed:
+            newerexpr = copy.copy(expr)
+            newerexpr.update_args(rec_args)
+            return True, newerexpr
+
+    return False, expr
+
+
+def _simplify_boolean_args(args: list[Any]|tuple[Any, ...], num_context=False) -> tuple[bool, list[Any]|tuple[Any, ...]]:
```
IgnaceBleukx, line 303: Add docstring

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+        changed, newexpr = _simplify_boolean_expr(expr, num_context=num_context)
+        if changed:
+            assert not isinstance(newexpr, int)
+            newlist.append(newexpr)
+        else:
             newlist.append(expr)
     return newlist
+
+
+def _simplify_boolean_expr(expr: Expression, num_context=False) -> tuple[bool, Expression|int]:
```
IgnaceBleukx, line 89: Add docstring:
```
"""
Well-typed helper function to eliminate boolean constants in an Expression.
Removes Boolean constants from logical operators, and replaces Boolean constants with ints in a numerical context (sum, comparisons, etc.)
Uses :func:`_simplify_boolean_args()` to recurse into the arguments of all other expressions.

Arguments:
    expr: Expression to eliminate boolean constants from
    num_context: Whether the expression is used as a numerical argument in another expression.

Returns:
    tuple[bool, Expression]: (changed, newexpr)

"""
```

cpmpy/transformations/normalize.py
```diff
@@ -62,183 +63,250 @@ def simplify_boolean(lst_of_expr, num_context=False):
...
+        elif is_bool(arg):
+            changed = True
+            newargs.append(int(arg) if num_context else BoolVal(arg))
+            continue
+
+        newargs.append(arg)
+
+    if changed:
+        return True, newargs
+    return False, args
```
IgnaceBleukx, line 343: I'm wondering if we should add `del newargs` here, so the garbage collector directly removes the ref to newargs? It should improve memory consumption

IgnaceBleukx (dismissed):
Primarily doc-related comments

IgnaceBleukx:
This PR merges to `solver_vars`, is that on purpose? Or should it be rebased on master?

cpmpy/transformations/normalize.py
IgnaceBleukx: Ah yes of course... #916 is not yet merged.

IgnaceBleukx (commented):
Fixed my comments and cleaned up the PR by removing solver-var-1d-related stuff.

Can you also have a look at the final result @tias ?

tias:
Pumpkin triggers a floating point error (we currently have a test that feeds a 0.5, even though we don't intend to support it).

So maybe we should merge #957 first, which removes that case and does foresee the only way we do intend to accept floats in the objective

cpmpy/transformations/normalize.py
tias: I wouldn't overload the simpl-bool-expr header.

I think I agree that we dont need num_context in this function anymore, but I also don't want to try it and possibly break things that I have forgotten by now.

here we just make sure we don't get a constant integer... seems fine

cpmpy/transformations/normalize.py
tias: it is now : )

tias: it can still fire with a 'not Global', right? but it wont need simplifying anymore...

tias (commented):
you still had two open comments that are worth resolving.

Other than that, looks ready

IgnaceBleukx:
Ok resolved the final comments and added an assert that any `not` in simplify_bool is indeed a `not(Global)`; which uncovered some cases I had missed before:
1) we also need to call `push_down_negation` on objective functions
2) in a bunch of solvers we forgot to add `push_down_negation` to the transformation stack in #916 

Should be good to go now once the tests finish

tias (commented):
the push_down_negation([obj])[0] is a bit of a hack...

should we add a helper?
```
def push_down_negation_expr(exp: Expression) -> Expression:
   ret = push_down_negation([obj])
   assert len(ret) == 1, "something meaningful"
   return ret[0]
```

other than that, please approve your own changes ; ) and feel free to merge it!

PR #960, opened 2026-04-27
modernize hyperparameter search example

OrestisLomis:
Two points: 
- Not sure if the `CPM_<solver>(model)` pattern should indeed be updated to `cp.SolverLookup.get(<solver>, model)`, but it made sense to me to update it
- Also in the 4 years since this example was added CPMpy/OR-Tools has gotten way faster which led to the difference between best and worse being neglible, so I decided to up the number of queens a bit.


OrestisLomis:
I guess a third, and more important point is: should the whole example not be updated to use the `ParameterTuner` class?

examples/advanced/hyperparameter_search.py
```diff
@@ -3,13 +3,12 @@
...
-from cpmpy.solvers import CPM_ortools, param_combinations
+import cpmpy as cp
+from cpmpy.solvers import param_combinations
 from cpmpy.transformations.flatten_model import flatten_model
 
 def main():
-    model = nqueens(n=25)
+    model = nqueens(n=50)
     # flatten once upfront, reduces overhead of multiple solves
     model = flatten_model(model)
```
IgnaceBleukx: This is from when flatten was the only transformation for ortools...
I would actually just remove this. The overhead should be negligable, and we don't count it in the runtime anyway (only solve time there)

examples/advanced/hyperparameter_search.py
```diff
@@ -3,13 +3,12 @@
...
 Generally applicable, and demonstrated on n-queens with ortools
 """
-import sys
-from cpmpy import *
-from cpmpy.solvers import CPM_ortools, param_combinations
+import cpmpy as cp
+from cpmpy.solvers import param_combinations
 from cpmpy.transformations.flatten_model import flatten_model
 
 def main():
```
IgnaceBleukx, line 11: give model as arg?
Also function name could be more informative. Like "tune_parameters" or something.

IgnaceBleukx (commented):
I would increase the number of queens a little more to 75.
Best to exclude this example from the test_examples (there is an exclude list in that file somewhere).

We should also point to our actual hyperparameter tuning tool in the docstring of this file.
E.g., 
```
CPMpy also includes a fully-fledged hyperparameter tuner, located in `cpmpy.tools.tune_solver`.
```

Or actually, shouldn't we just rework this example to use the built-in tuner? I feel that makes more sense as an example on how to use the library.

OrestisLomis:
Indeed I agree it makes more sense to update the whole example to reflect how the `ParameterTuner` class should be used. Please have a look at the issue I opened regarding this class, because I feel like it's actually a bit outdated/undersupported #963.

OrestisLomis:
@IgnaceBleukx I updated the example to use actual CPMpy tooling, also I increased the nb of queens even more to 125 because it has 'more interesting' behaviour in terms of continuously improving the runtime.

Related to this PR, I was looking at adding parameters for Exact but then we would actually have to account for this in the ParameterTuner class, since Exact doesn't take params at solve time. Is that something we want to do?

IgnaceBleukx:
Hmm in the current form in the tuner we indeed use `SolverLookup` but thats not really required as we always re-initialize it to avoid accidentally using an incremental solver an biasing the results like that.
The only reason why we do it currently is to not count the transformation time when tuning.

`cp.Model.solve` already takes care of dispatching solve parameters to the constructor of exact, so ideally we would just use that, but we should first address #803 and then we can add the tuneable parameters to exact.
Not a blocker for this PR in any case : )

examples/advanced/hyperparameter_search.py
```diff
@@ -3,59 +3,63 @@
...
+    tuner = ParameterTuner(
+        solvername="ortools", 
+        model=model, 
+        all_params=all_params, 
+        defaults=defaults
+    )
 
-    s = CPM_ortools(model); s.solve()
-    print("With default parameters:", round(s.status().runtime,2))
+    # Uncomment to use GridSearchTuner instead of ParameterTuner
```
IgnaceBleukx: Not sure this is needed, there is no real reason to not use the actual Parameter Tuner

examples/advanced/hyperparameter_search.py
```diff
@@ -3,59 +3,63 @@
...
+    # )
 
-    # Outputs:
-    # Fastest in 0.02 seconds, config: {'cp_model_probing_level': 0, 'linearization_level': 0, 'symmetry_level': 1}
-    # Comparing best -- worst: 0.02 -- 0.2
-    # With default parameters: 0.13
+    
+    # 4. Run Tuning
+    # time_limit: total budget for the whole process
+    # max_tries: limit number of configs to test
```
IgnaceBleukx: you don't run max_tries, so don't document it either

examples/advanced/hyperparameter_search.py
```diff
@@ -3,59 +3,63 @@
...
+        'cp_model_probing_level': 2,
+        'linearization_level': 1,
+        'symmetry_level': 2
+    }
 
-    print()
-    best = configs[0]
-    print("Fastest in", round(best[0],2), "seconds, config:", best[1])
-    print("Comparing best -- worst:", round(configs[0][0],2), "--", round(configs[-1][0],2))
+    # initialize tuner; can use ParameterTuner (SMBO-based) or GridSearchTuner (Exhaustive)
```
IgnaceBleukx: just `initialize tuner instance`

examples/advanced/hyperparameter_search.py
```diff
@@ -3,59 +3,63 @@
...
+    #     all_params=all_params, 
+    #     defaults=defaults
+    # )
 
-    # Outputs:
-    # Fastest in 0.02 seconds, config: {'cp_model_probing_level': 0, 'linearization_level': 0, 'symmetry_level': 1}
-    # Comparing best -- worst: 0.02 -- 0.2
-    # With default parameters: 0.13
+    
+    # 4. Run Tuning
```
IgnaceBleukx: just `run tuner: total budget of 60 seconds`

examples/advanced/hyperparameter_search.py
```diff
@@ -3,59 +3,63 @@
...
-    return Model(
-             AllDifferent(queens),
-             AllDifferent([queens[i] + i for i in range(n)]),
-             AllDifferent([queens[i] - i for i in range(n)]),
-           )
+    """ N-queens problem """
+    queens = cp.intvar(1, n, shape=n)
+    return cp.Model(
+        cp.AllDifferent(queens),
+        cp.AllDifferent([queens[i] + i for i in range(n)]),
```
IgnaceBleukx: for style points: `cp.AllDifferent(queens + np.arange(n))`

examples/advanced/hyperparameter_search.py
IgnaceBleukx, line 2: add `#!/usr/bin/env python`

IgnaceBleukx (commented):
Almost there, I feel the example is now being treated as a very verbose docstring on how to use the parameter tuner...
Rather I would like it to just be an example that USES the tuner, instead of explaining how it CAN be used (we have the docstring of the tuner for that)

So code-wise it looks good, but the comments should really be updated (see reviewer comments)

IgnaceBleukx (approved):

PR #962, opened 2026-04-27
Pass `verbose` to sub-process in order to save stats

hbierlee:
Important because if verbose is not passed to the sub-process, it will not output its various statistics (e.g. posting time, solve time, ...)


IgnaceBleukx:
@ThomSerg is this PR still relevant given the new updates and runner etc? If so, seems good to merge to me?

PR #965, opened 2026-04-29
typing of min/max

tias:
This is a small, more specialised subset of #910 mypy_globfunc (consider that one on hold)
Here we for the first time filter out the np.integers in a global, so we don't have to account for them later...

With that, we don't need the
circular imported Expression in the typed utils
(because with int|Expression, if not int, then implicitly Expression)

### DIFF
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -78,9 +78,9 @@ import numpy as np
 import cpmpy as cp
 
 from ..exceptions import CPMpyException, IncompleteFunctionError, TypeError
-from .core import Expression, Operator, ExprLike, ListLike
-from .variables import boolvar, intvar, cpm_array
-from .utils import flatlist, argval, is_num, is_int, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals, implies
+from .core import Expression, Operator, Comparison, ExprLike, ListLike
+from .variables import intvar, NDVarArray
+from .utils import argval, is_num, eval_comparison, is_any_list, is_boolexpr, get_bounds, argvals, implies, npint2int, argvals_intexpr, get_bounds_intexpr
 
 
 class GlobalFunction(Expression):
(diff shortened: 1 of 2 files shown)


cpmpy/expressions/globalfunctions.py
```diff
@@ -210,18 +223,32 @@ def __init__(self, arg_list: ListLike[ExprLike]):
...
+            if isinstance(arg_list, NDVarArray):
+                has_subexpr = arg_list.has_subexpr()
+            arg_iter = arg_list.flat
+
+        args: tuple[int|Expression, ...] = npint2int(arg_iter)
+        super().__init__("max", args, has_subexpr=has_subexpr)
+
+    @property
+    def args(self) -> tuple[int|Expression, ...]:
+        """ READ-ONLY, well-typed argument of this global function """
```
IgnaceBleukx: Add explicit comment on no np ints?

cpmpy/expressions/utils.py
```diff
@@ -153,6 +153,18 @@ def argvals(arr):
         return [argvals(arg) for arg in arr]
     return argval(arr)
 
+def argvals_intexpr(lst: Iterable[int|Expression]) -> Optional[list[int]]:
+    """ The well-typed way to get the values of a list of int|Expression, or None if any expression is not assigned """
```
IgnaceBleukx: "well-typed helper function to ...."

cpmpy/expressions/utils.py
```diff
@@ -282,3 +308,10 @@ def is_star(arg):
         Check if arg is star as used in the ShortTable global constraint
     """
     return isinstance(arg, type(STAR)) and arg == STAR
+
+
+def npint2int(iter: Iterable[ExprLike]) -> tuple[int|Expression, ...]:
+    """Convert numpy values in iterable to Python integers, return as tuple."""
+    return tuple(int(el) if isinstance(el, (np.integer, np.bool_)) else el for el in iter)
```
IgnaceBleukx: A faster way would be (in practice, we'll almost always have just expressions)

```python
NP_TYPES = frozenset({
    np.int8, np.int16, np.int32, np.int64,
    np.uint8, np.uint16, np.uint32, np.uint64,
    np.bool_
})

def npint2int(iter):
    return tuple(int(x) if type(x) in NP_TYPES else x for x in lst)
```

IgnaceBleukx (approved):
A couple of small suggestions. Looks good overall.

IgnaceBleukx:
Did the doc updates already, what do you think about the npint2int version without isinstance checks? 
It can go in either way for me

tias:
Coffee break agent compared current:

```tuple(int(el) if isinstance(el, (np.integer, np.bool_)) else el for el in iter)```

vs proposed (with lst/iter typo fix):

```
NP_TYPES = frozenset({...})
tuple(int(x) if type(x) in NP_TYPES else x for x in iter)
```

Microbenchmark on representative iterables (n=2000, best of repeats)
mixed python int + np.int64: current 205.56ms, proposed 68.51ms (3.0x faster)
np.int64 flat iterator: 187.63ms vs 88.57ms (2.1x faster)
np.bool_ flat iterator: 387.60ms vs 261.92ms (1.48x faster)
mixed Expression + np.int64: 241.35ms vs 73.82ms (3.3x faster)
NDVarArray flat (no numpy scalars): 291.50ms vs 59.62ms (4.9x faster)

Correctness matched on those datasets.

End-to-end via Minimum + Maximum constructors (n=5000)
np.int64 array: 221.74ms vs 122.23ms (1.8x faster)
np.bool_ array: 404.34ms vs 270.75ms (1.49x faster)
object array mixed Expression/np.int64: 256.99ms vs 82.89ms (3.1x faster)

Takeaway
The type(x) in NP_TYPES approach is consistently faster in realistic CPMpy usage, including Min/Max paths.

so yes, worth it

tias:
please check and merge if it passes, thx : )

PR #969, opened 2026-05-06
Canonicalize neq comparisons in csemap

IgnaceBleukx:
Some small canonicalization to improve the csemap.

If we encounter `var != val` as a subexpression, we now make a Boolean variable `bv` with the consrtaint `bv <-> var == val` and return `~bv`.
This increases the number of hits in the csemap, and greatly improves the linearize_reified_varvals transformation.

### DIFF
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -2,7 +2,7 @@ import copy
 from typing import Optional, Any, cast, overload, Literal
 from ..expressions.core import Expression, Comparison
 from ..expressions.utils import is_int
-from ..expressions.variables import boolvar, intvar, _IntVarImpl, _BoolVarImpl
+from ..expressions.variables import NegBoolView, boolvar, intvar, _IntVarImpl, _BoolVarImpl
 
 
 class CSEMap:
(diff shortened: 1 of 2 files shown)


cpmpy/transformations/cse.py
```diff
@@ -71,14 +79,34 @@ def get_or_make_var(self, expr: Expression) -> tuple[_IntVarImpl, Optional[Expre
...
+            return res, None # already made a variable for expr
 
-        elif expr.is_bool():
+        if expr.is_bool():
             bv = boolvar()
+            expr, negate = self._canonicalize_boolexpr(expr)
+
             self.flat_map[expr] = bv
+            if negate: # return the negated variable to get the original expression back
+                neg_bv = cast(NegBoolView, ~bv) # __invert__ of _BoolVarImpl returns Expression, but we know it is a NegBoolView
```
tias: you can just create the NegBoolView directly? thats what __invert__ is doing too?

cpmpy/transformations/cse.py
```diff
@@ -71,14 +79,34 @@ def get_or_make_var(self, expr: Expression) -> tuple[_IntVarImpl, Optional[Expre
...
+
+
+    def _canonicalize_boolexpr(self, expr: Expression) -> tuple[Expression, bool]:
+        """canonicalize a Boolean expression, results in more hits in the flat_map"""
+
+        if isinstance(expr, Comparison):
+            lhs, rhs = expr.args
+            if expr.name == "!=" and is_int(rhs):
+                # b <-> (expr != val) :: (~b) <-> (expr == val)
+                expr = copy.copy(expr)
```
tias: I don't know...
we can also more directly just Comparison("==", (lhs,rhs)) ?

tias (commented):
great, 2 small cleanups to consider or discard

cpmpy/transformations/cse.py
IgnaceBleukx: Yes excellent point...

cpmpy/transformations/cse.py
IgnaceBleukx: Yeah that's also fine, should be more mypyc-able too
Changed it and copied over `_has_subexpr` to the new expression

PR #971, opened 2026-05-08
Remove Gurobi TODO

hbierlee:
There was a gurobi IIS bug, which is now resolved in 13.0.2, however, I'd say we should probably leave in this assert for safety (rather than adding a version guard). It's a non-trivial check (going over each transformed hard constraint), however, it should still be much faster than the IIS call itself. So this PR just removes the TODO (otherwise we can use this PR to do the version guard to only do the check if `gurobipy<=13.0.2`)


tias (approved):
clarified the bug is on their side, and added the reference to your report for tracability

PR #972, opened 2026-05-08
Save expr == var in csemap to catch silly user mistakes

IgnaceBleukx:
As discussed offline, it would be useful to catch silly mistakes by the user, where an aux var is introduced for some expression, but then the expression is used later on as a nested expression somewhere.

To fix this, this PR saves the `csemap[expr] = var` whenever we encounter a top-level `==` constraint.

### DIFF
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -18,8 +18,8 @@ class CSEMap:
     def __getitem__(self, expr: Expression) -> _IntVarImpl:
         return self.flat_map[expr]
 
-    def __setitem__(self, attr, val):
-        raise ValueError("__setitem__ is not supported for flat_map, use get_or_make_var instead")
+    def __setitem__(self, attr: Expression, val: _IntVarImpl) -> None:
+        self.flat_map[attr] = val
 
     @overload
     def get(self, expr: Expression) -> Optional[_IntVarImpl]: ...
--- a/cpmpy/transformations/flatten_model.py
+++ b/cpmpy/transformations/flatten_model.py
@@ -229,7 +229,6 @@ def flatten_constraint(expr, csemap=None):
             exprname = expr.name  # so it can be modified
             lexpr, rexpr = expr.args
             rewritten = False
-
             # rewrite 'Var # Expr' to normalized 'Expr # Var' (where # is any comparator)
             if __is_flat_var(lexpr) and not __is_flat_var(rexpr):
                 assert (expr.name in ('==', '!=', '>', '>=', '<', '<='))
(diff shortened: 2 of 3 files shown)


cpmpy/transformations/flatten_model.py
```diff
@@ -246,6 +245,12 @@ def flatten_constraint(expr, csemap=None):
                 elif exprname == '<=':
                     exprname = '>='
 
+            # save expr == var to the csemap, so we don't make a new variable for expr later
+            if csemap is not None and exprname == '==': # cheap checks first
+                if isinstance(lexpr, Expression) and isinstance(rexpr, Expression) and \
+                    lexpr.is_bool() == rexpr.is_bool() and __is_flat_var(rexpr): # types must match
```
tias, line 251: `isinstance(rexpr, Expression) and __is_flat_var(rexpr)` sounds like `isinstance(rexpr, IntVarImpl)`?

and what if its var == expr? or is that already normalized out before this line?

cpmpy/transformations/flatten_model.py
```diff
@@ -246,6 +245,12 @@ def flatten_constraint(expr, csemap=None):
...
+                    csemap[lexpr] = rexpr # save lexpr == rexpr to the csemap
```
tias, line 252: what if lexpr is already in? we would overwrite it?
if its in, we could post `rexpr == csemap[lexpr]` and store `csemap[rexpr] == csemap[lexpr]`?

cpmpy/transformations/cse.py
```diff
@@ -18,8 +18,8 @@ def __len__(self):
     def __getitem__(self, expr: Expression) -> _IntVarImpl:
         return self.flat_map[expr]
 
-    def __setitem__(self, attr, val):
-        raise ValueError("__setitem__ is not supported for flat_map, use get_or_make_var instead")
+    def __setitem__(self, attr: Expression, val: _IntVarImpl) -> None:
```
tias, line 21: I agree we need it for the code below, but maybe add a doc "Use get_or_make_var() instead, which will create a fresh variable for you as needed. Only use this function if you must register an existing `attr == val`.

tias (changes requested):

IgnaceBleukx:
Already implemented for booleans via #1010;  too be revisited in another PR if we want to do this for int expressions too.
The main use-case was reifications anyway as they should be catched to improve the linearize_reified_varvals transform.

tias:
I did not realize these were 2 separate commits; it makes sense to do it for integers too yes; lets keep it open for a little longer to remind ourselves?

PR #980, opened 2026-05-13
Positive decomposition

WoutPiessens:
Building on #924 and #929, this is a first implementation of the separate positive decomposition for global constraints.
For now:
- decompose_positive() is called if the global constraint c is top-level or positively reified (b => c). In case the global constraint is nested in any other way, it is decomposed as it was before.
- Only the MDD global constraint supports a separate positive decomposition

### DIFF
--- a/cpmpy/transformations/decompose_global.py
+++ b/cpmpy/transformations/decompose_global.py
@@ -184,22 +184,25 @@ def _decompose_in_tree_args(args: ListLike[Any],
                     continue
                 else:
                     # nested global function, decompose
-                    if decompose_custom is not None and arg.name in decompose_custom:
-                        newarg, toplevel_exprs = cast(Tuple[Expression, List[Expression]], decompose_custom[arg.name](arg))
-                    else:
-                        newarg, toplevel_exprs = arg.decompose()
-
                     orig_for_csemap = arg
-                    arg = newarg
-                    if len(toplevel_exprs) > 0:
-                        toplevel.extend(toplevel_exprs)
+                    # this is a bit awkward, but the decompose can return a new GlobFunc to decompose...
+                    while isinstance(arg, GlobalFunction) and arg.name not in supported:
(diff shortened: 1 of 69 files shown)


tias:
reviewed in #981

cpmpy/transformations/decompose_global.py
```diff
@@ -95,9 +97,19 @@ def decompose_in_tree(lst_of_expr: list[Expression],
             changed = True
             newlist.append(BoolVal(expr))
         elif expr.has_subexpr():
+            # special case for positive reified
+            decomposed_positive = False
+            if expr.name == "->" and isinstance(expr.args[1], GlobalConstraint) and expr.args[1].name not in supported_reified:
```
tias, line 102: I think you need an explicit 'changed = True'

tias (commented):
neat.

One change needed I think.

Are we sure the 'constraining','defining' is valid for the -> of MDD? That was an open question previously; is there a way to add a test for that?

cpmpy/transformations/decompose_global.py
IgnaceBleukx: That is handled on line 121, no?

WoutPiessens:
> neat.
> 
> One change needed I think.
> 
> Are we sure the 'constraining','defining' is valid for the -> of MDD? That was an open question previously; is there a way to add a test for that?

The constraining & defining constraints for positive decomp do seem right, based on my intuition and also on the tests

cpmpy/transformations/decompose_global.py
IgnaceBleukx: My bad, I had unpushed commits... this should be handled below now (also saves a copy)

tias:
OK great, it also makes sense to me but good to see it tested.

I rewrote the 'if' part slighly to have clearer non-overlapping cases

PR #981, opened 2026-05-15
Linear decomposition for Regular using MDD

WoutPiessens:
Building on #924, #949 and #980, I implemented a linear decomposition of the Regular global constraint using the MDD global. To convert the DFA to an MDD, I used Algorithm 1 in:
"A Regular Language Membership Constraint for Finite Sequences of Variables", Gilles Pesant, 2004
to convert a DFA to a layered graph (equivalent to an MDD). The advantage here is that transitions that cannot lead to an accepting state will not be part of the resulting layered graph.

By doing this, we can also take advantage of the positive decomposition for MDDs when the Regular global is top-level or positively reified. (Same goes for Table in #945)

### DIFF
--- a/cpmpy/transformations/decompose_global.py
+++ b/cpmpy/transformations/decompose_global.py
@@ -184,22 +184,25 @@ def _decompose_in_tree_args(args: ListLike[Any],
                     continue
                 else:
                     # nested global function, decompose
-                    if decompose_custom is not None and arg.name in decompose_custom:
-                        newarg, toplevel_exprs = cast(Tuple[Expression, List[Expression]], decompose_custom[arg.name](arg))
-                    else:
-                        newarg, toplevel_exprs = arg.decompose()
-
                     orig_for_csemap = arg
-                    arg = newarg
-                    if len(toplevel_exprs) > 0:
-                        toplevel.extend(toplevel_exprs)
+                    # this is a bit awkward, but the decompose can return a new GlobFunc to decompose...
+                    while isinstance(arg, GlobalFunction) and arg.name not in supported:
+                        if decompose_custom is not None and arg.name in decompose_custom:
+                            newarg, toplevel_exprs = cast(Tuple[Expression, List[Expression]], decompose_custom[arg.name](arg))
+                        else:
+                            newarg, toplevel_exprs = arg.decompose()
+                        arg = newarg
+                        if len(toplevel_exprs) > 0:
+                            toplevel.extend(toplevel_exprs)
 
-                    # TODO: violates type!!!
-                    # apparently in #630 we decided that decompose may return an int (e.g. for element)...
-                    # we should change that (Element constructor requires variable index; the []/__get__ override can still take anything)
+                        # TODO: violates type!!!
+                        # apparently in #630 we decided that decompose may return an int (e.g. for element)...
+                        # we should change that (Element constructor requires variable index; the []/__get__ override can still take anything)
+                        if isinstance(arg, int):
+                            # no need to recurse further, stop here
+                            newargs.append(arg)
+                            break
                     if isinstance(arg, int):
-                        # no need to recurse further, stop here
-                        newargs.append(arg)
                         continue
             
             # if it has subexprs, decompose its arguments too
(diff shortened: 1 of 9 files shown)


cpmpy/transformations/decompose_global.py
```diff
@@ -61,28 +58,72 @@ def decompose_in_tree(lst_of_expr: list[Expression],
...
-
-    # new toplevel constraints may need to be decomposed too
-    while len(todo_toplevel):
-        changed, decomp, next_toplevel = _decompose_in_tree(todo_toplevel, supported=supported, supported_reified=supported_reified, is_toplevel=True, csemap=csemap, decompose_custom=decompose_custom)
-        if not changed:
-            newlst_of_expr.extend(todo_toplevel)
-            break
+    newlist: List[Expression] = []
+    todolist: List[Expression] = []  # these still need to be decomposed
+    positive_decomps = get_positive_decompositions()
```
tias, line 63: this is unexpected (to call an outside function) and different from 'decompose_custom'. 

Should we instead not have an extra argument 'decompose_positive' which is a dict like decompose custom?

cpmpy/transformations/decompose_global.py
```diff
@@ -61,28 +58,72 @@ def decompose_in_tree(lst_of_expr: list[Expression],
...
+    for expr in lst_of_expr:
+        if isinstance(expr, GlobalConstraint) and expr.name not in supported:
+            # toplevel/positive global constraint, decompose
+            positive_decomp = positive_decomps.get(expr.name)
```
tias, line 67: there are 2 different patterns in this code: for positive_decomp it is fetched upfront, for decompose_custom it checks in the if whether it is 'in' and only then calls/applies it...

these two are so similar that they should be treated similar imho

cpmpy/transformations/decompose_global.py
```diff
@@ -61,28 +58,72 @@ def decompose_in_tree(lst_of_expr: list[Expression],
...
+                global_constraint = expr.args[1]
+                positive_decomp = positive_decomps.get(global_constraint.name)
+
+                if positive_decomp is not None:
+                    # Special case: use positive decomposition
+                    exprs, toplevel_exprs = positive_decomp(global_constraint)
+
+                    newlist.extend(expr.args[0].implies(e) for e in exprs)
+                    newlist.extend(toplevel_exprs)
+                    already_decomposed = True
```
tias, line 92: the already_decomposed is a bit weird...

I'm also not sure the flow is correct; as in, 2 ifs above after decompose you add them to todolist because their args can have globals too; the code below does it recursive.

So, don't you have to follow the pattern of above? and add them to todolist?

note that you can just 'continue' if you have fully handled the case; this avoids the need for an already decomposed flag, and makes sure nothing extra will be posted accidentally either

tias (changes requested):
some changes on the decompose part required.

Also, best weight till the decompose itself is in master (Ignace is working on it) and then that we iterated on it here, before putting it in the xcsp3 branch

tias:
seems I reviewed it against the wrong branch... is it still OK like this?

cpmpy/transformations/decompose_global.py
WoutPiessens: It's possible, but the reason I did it like this is because the existence of a positive decomposition is solver independent, so it will be the same for any call to decompose_in_tree(), unlike decompose custom

cpmpy/transformations/decompose_global.py
WoutPiessens: agree

cpmpy/transformations/decompose_global.py
tias: I understand. but its positive decompositions are the same 'for now'. if we are extensible on custom for plain decompositions, might as well be extensible for positive decomps too.

With ignace we finalized the signature for decomposition for cpmpy at large, including nested and half-reif; Ideally we can do that straight

(not we have to do this not in the PR, but the parent PRs)

IgnaceBleukx:
I think we decided offline we'll not implement this as it doesn't seem to be worth it for dfa's with many loops? 
Please close if that is the case : )

PR #983, opened 2026-05-18
Small fixes for Hexaly

IgnaceBleukx:
Some small fixes for Hexaly: adding proper support for alldiff and nvalue.

We should look into how we can improve the modeling for other globals (e.g,, scheduling constraints and circuit).
Can be done in this PR or a future one.


tests/test_solvers.py
```diff
@@ -1047,22 +1047,41 @@ def test_partial_div_mod(self, solver):
             pytest.skip("solver does not support this test context")
         if solver == 'cplex':
             pytest.skip("skip for cplex, cplex supports solveall only for MILPs, and this is not linear.")
+
+        kwargs = dict()
+        if solver in ("gurobi", "cplex"):
```
tias: there is a skip for cplex just above?

cpmpy/solvers/hexaly.py
```diff
@@ -66,8 +66,9 @@ class CPM_hexaly(SolverInterface):
     https://www.hexaly.com/docs/last/pythonapi/index.html
     """
 
-    supported_global_constraints = frozenset({"min", "max", "abs", "mul", "div", "mod", "pow", "element"})
-    supported_reified_global_constraints = frozenset()
+    supported_global_constraints = frozenset({"min", "max", "abs", "mul", "div", "pow", "element", "nvalue", "nvalue_except",
```
tias: document that we don't translate to their 'mod' because (of round to 0?)

tests/test_solvers.py
tias: just put an if on hexaly; with the solution limit in the else?

PR #985, opened 2026-05-19
New solver/optal cp

OrestisLomis:
I saw Optal was sponsoring JFPC and had a lot of similiraties to CPO, so I tried to add it to CPMpy (with the help of codex). Locally all tests pass, but I'm not sure how to actually add OptalCP to the CI for testing (both in general and with the license) so initially the PR will be a draft. Another tricky thing is that I'm not sure how to figure out whether someone has the preview version installed or the full/academic version. So I can not really throw a warning/error for that.. Instead the user will just get an error whenever the masked values are accessed.


tias:
I think Ignace had a look at Optal before, so for him to check.

Do remove the is_np_int and is_np_bool from the shared function and just use an inline check in your new file

PR #986, opened 2026-05-20
add mus native cplex

OrestisLomis:
Still as a draft until the testing is fixed in #961


OrestisLomis:
So CPLEX obviously is not in CI, but the test do pass locally as well (and I confirmed that they do not pass if the code actually does something wrong). One thing we don't do with this implementation is that the transformation does not really have a clean split between value and defining constraints when done without assump vars. So here we keep all the defining constraints in the group of the constraint itself instead of putting it in the hard constraints group. We could improve this in the future, but I think it's out of scope for this PR.

IgnaceBleukx:
Hmm but it might actually give the wrong result like this, because of CSE: 
Suppose there is a constraint with a subexpression expr, for which we make a new intvar and a constraint iv == expr is added to the model. In the current implementation this new defining consrtaint will be part of the group of the constraint that generated it. But if there is another constraint that also uses expr as a subexpr, we will also replace it with iv, but without the first group iv has no meaning. So any MUS will always contain the first constraint if the second consrtaint is involved.

example:

```python
import cpmpy as cp
from cpmpy.tools.explain import mus_native, mus_naive

x = cp.intvar(-10, 10, name="x")
y = cp.intvar(-10, 10, name="y")

soft = [
    cp.abs(x) + y <= 15,  # satisfiable, not needed for the conflict
    cp.abs(x) + y >= 11,  # the real conflict with hard
]
hard = [x == 0]

print(mus_naive(soft, hard=hard))
>> [(abs(x)) + (y) >= 11]

print(mus_native(soft, hard=hard, solver="cplex"))
>> [(abs(x)) + (y) <= 15, (abs(x)) + (y) >= 11]
```

So we do need to set all of the defining constraints to hard constraints I believe (or patch `solver._csemap = None`) but that might have other performance issues naturally...

OrestisLomis:
I see.. I'll have to check Gurobi as well actually because I'm not sure it is handled differently. As a solution perhaps transformations themselves should split the output into value and defining constraints. Otherwise it seems a bit tricky for it to be extracted here efficiently.

cpmpy/solvers/cplex.py
```diff
@@ -684,3 +690,64 @@ def solveAll(self, display:Optional[Callback]=None, time_limit: Optional[float]=
...
+        # The conflict refiner result expands groups back to their native docplex
+        # constraints, so map each returned conflict element to its CPMpy source.
+        conflict_statuses = {
+            ConflictStatus.Member,
+        }
+        mus_idxs = set()
+        for conflict in refine_res.iter_conflicts():
+            if conflict.status in conflict_statuses:
+                soft_idx = native_to_soft_idx.get(conflict.element)
+                if soft_idx is not None:
```
IgnaceBleukx: Just checking, that means its a hard constraint then, right? So indeed no need to add it to the core here, but would be good to add a comment about it to clarify

cpmpy/solvers/cplex.py
```diff
@@ -684,3 +690,64 @@ def solveAll(self, display:Optional[Callback]=None, time_limit: Optional[float]=
...
+        More about the native Conflict Refiner:
+        https://www.ibm.com/docs/en/icos/22.1.0?topic=conflicts-more-about-conflict-refiner
+        """
+        from docplex.mp.conflict_refiner import ConflictRefiner, ConstraintsGroup
+        from docplex.mp.constants import ConflictStatus
+        
+        soft_cons = toplevel_list(soft, merge_and=False)
+        s = cls()
+        # Disable CSE: a shared csemap would attach defining constraints
+        # (e.g. IV == abs(x)) to the first soft that created them. Later
```
IgnaceBleukx: thats a bit verbose, would be fine to just say "disable CSE so no defining constraints are shared among transformed constraitns" or something?
Indeed good to keep the link to the PR

IgnaceBleukx (approved):
Great! Two small doc comments, otherwise good to go

cpmpy/solvers/cplex.py
OrestisLomis: I don't think 'shared' is the right term here but indeed I made it a bit less verbose.

cpmpy/solvers/cplex.py
OrestisLomis: indeed

PR #987, opened 2026-05-20
Feature/mus cpo

OrestisLomis:
Draft until #961 is resolved and this can be tested


cpmpy/solvers/cpo.py
```diff
@@ -719,6 +719,53 @@ def _make_task(self, start, dur, end, is_present):
...
+        CP Optimizer refines conflicts over native constraints. A CPMpy soft
+        constraint may expand to several native ones. In that case we post them under an 
+        indicator and only expose the indicator as soft to the refiner.
+        """
+        soft_cons = toplevel_list(soft, merge_and=False)
+        s = cls()
+        dom = s.get_docp().modeler
+
+        # Hard constraints may appear in the conflict, but are never returned.
+        for cpm_con in s.transform(hard):
```
IgnaceBleukx: why not just `s += hard`?

cpmpy/solvers/cpo.py
```diff
@@ -719,6 +719,53 @@ def _make_task(self, start, dur, end, is_present):
...
+            s.cpo_model.add(s._cpo_expr(cpm_con, boolexpr=True))
+
+        native_to_soft_idx = {}
+        for i, soft_con in enumerate(soft_cons):
+            native_soft = []
+            for cpm_con in s.transform(soft_con):
+                cpo_expr = s._cpo_expr(cpm_con, boolexpr=True)
+                # Globals such as Cumulative may return a list of native exprs.
```
IgnaceBleukx, line 754: Hmm that's some weird typing then... it should just coerce all of the expressions into an `and` I would say?
So type of `_cpo_expr` should be `list -> list` and `expr -> cpo_expr`?
Probably not the best idea to fix that here, but may be useful to keep a note somewhere about t

IgnaceBleukx: Actually, why not just wrap them in a CPO-and here? Or does cpo split them up again in the conflict refnining?

IgnaceBleukx (commented):
I'm not sure about the approach for mapping to multiple solver-level constraints.
CPO is expression-based, so we can wrap a list of expressions into a CPO-AND expression. Would that work for the conflict refiner? Or does CPO split them up again and they end up as seperate constraints after all?

cpmpy/solvers/cpo.py
OrestisLomis: That makes a lot more sense.. I tried it when I started the PR and something about it was difficult I thought, but actually it is a lot simpler in the end

cpmpy/solvers/cpo.py
OrestisLomis: Hmmm, it might've been related to wanting to not only add these constraints but also make them actual hard constraints in the MUS. Unfortunately that is impossible through the API (even though the doc says it should be possible). So actually this is an issue that remains unsolved here...

One thing we can do is to actually just throw a value error when hard is not an empty list, which is quite undesirable but I think in many cases the method would still be useful without hard constraints.

IgnaceBleukx: Yes agreed! You can also open an issue on the docplex github, with some context on this problem. It might just be a docplex thing, and not be related to cp optimizer itself.

OrestisLomis: Alright, [I did open an issue](https://github.com/IBMDecisionOptimization/docplex/issues/41), though last time I did unfortunately there was not much activity.

cpmpy/solvers/cpo.py
```diff
@@ -719,6 +719,52 @@ def _make_task(self, start, dur, end, is_present):
...
+
+            # Keep multi-native softs atomic: the refiner does not split &&.
+            soft_native = native_soft[0] if len(native_soft) == 1 else dom.logical_and(native_soft)
+            s.cpo_model.add(soft_native)
+            native_to_soft_idx[soft_native] = i
+
+        refine_res = s.cpo_model.refine_conflict(LogVerbosity='Quiet')
+        assert refine_res.is_conflict(), "MUS: model must be UNSAT"
+
+        mus_idxs = {
```
IgnaceBleukx: I find this rather hard to read...
Why not just a regular for-loop? Then you don't need the comprehension afterwards either.
Smt like
```python

core = []
cpo_constraints = refine_res.get_member_constraints() + refine_res.get_possible_constraints()
for cpo_cons in cpo_constraints:
    if cpo_cons in native_to_soft_idx:
        cpm_cons =  soft_cons[native_to_soft_idx[cpo_cons]]
        core.append(cpo_cons)

return core
```

IgnaceBleukx (commented):
Code-wise: one formatting suggestion.

If I run `pytest tests/test_tools_mus.py --solver=cpo` I get 3 errors related to hard constraints not being supported.

Best to exclude those tests for cpo I suppose; potentially add some more tests if required? Not sure

IgnaceBleukx:
Good, just make sure to merge #986 first so the cse test is also ran on cpo before its pushed to master

PR #989, opened 2026-05-20
feature/mus scip

OrestisLomis:


OrestisLomis:
Note that I already added some functionality for hard constraints, which makes this PR depend on https://github.com/scipopt/scip/issues/201, or more specifically https://github.com/scipopt/PySCIPOpt/pull/1221. I can temporarily remove this dependency (and then the ```mus_native``` would only work for MIP problems without hard constraints) and only add it back later, but since this PR is in draft anyway I thought it's fine to add it already and wait.

OrestisLomis:
Keeping it as draft as without hard constraint the use is quite limited

PR #990, opened 2026-05-20
Varname in varmap instead of variable object

IgnaceBleukx:
Store the name of variables in the varmap instead of the _NumVarImpl object.

This speeds up looking up the variable in the dict in solver_vars.

The only downside is that there is now no way of iterating over ALL CPMpy vars ever posted to the solver, only the user vars. (using the user_vars set)


tias (approved):
OK. 

I see a 4.9s -> 4.5s drop for the uncommited dev/time_ortools.py from solver_vars, so all good

Issue #938, opened 2026-04-22
Update fast path in solver_var


hbierlee:
Seems some of the solvers implement `solver_var` in a sub-optimal way, namely checking unlikely cases first.

> do we do this function the same elsewhere?
> 
> its not written with 'fast path' in mind... an alterantive would be
> ```
> solver_var = self._varmap.get(cpm_var, None)
> 
> if solver_var = None:
>   if is_num...
>   elif is _BoolVarImpl:
>     if NegBool: raise
>     else:
>        solver_var = ...
>   elif _IntVarImple:
>   self._varmap[cpm_var] = solver_var
> 
> return solver_var
> ```
> 
> if other solvers do similar like we do here, we should fix this in a separate PR 

 _Originally posted by @tias in [#412](https://github.com/CPMpy/cpmpy/pull/412/changes/BASE..1f794b68f00dc01d0fdc953d069b8f9101a81b12#r3123049472)_

E.g., see this example. Note that coverage revealed the `is_num` is never run, similarly `NegBoolView` should never run as long as the transforms are correct.

https://github.com/CPMpy/cpmpy/blob/626bfa80c6507bf2ada8aa5d665766ceb031af75/cpmpy/solvers/gurobi.py#L252-L276

PR #992, opened 2026-05-21
simpler solver_vars improvement

tias:
Not as effective as 'solver_vars_1d' from #939  , but simpler and compatible with #990 

(also includes the dev/time_ortools script)

### DIFF
--- a/cpmpy/solvers/solver_interface.py
+++ b/cpmpy/solvers/solver_interface.py
@@ -19,13 +19,13 @@
         ExitStatus
 
 """
-from typing import Optional, List, Callable, TypeAlias
+from typing import Optional, List, Callable, TypeAlias, Iterable, Any
 import warnings
 import time
 from enum import Enum
 
 from ..exceptions import NotSupportedError
-from ..expressions.core import Expression, ListLike
+from ..expressions.core import Expression, ListLike, ExprLike
 from ..expressions.variables import _NumVarImpl
 from ..transformations.cse import CSEMap
 from ..transformations.get_variables import get_variables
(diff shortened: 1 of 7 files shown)


cpmpy/solvers/solver_interface.py
```diff
@@ -178,13 +178,26 @@ def solver_var(self, cpm_var):
...
+        res: list[Any] = []
+        for cpm_var in cpm_vars:
+            if isinstance(cpm_var, _NumVarImpl):
+                if cpm_var.name in self._varmap:  # fast path
+                    res.append(self._varmap[cpm_var.name])
+                else:  # slow path
+                    res.append(self.solver_var(cpm_var))
+            elif isinstance(cpm_var, int):
+                res.append(cpm_var)
+            elif is_any_list(cpm_var):
```
IgnaceBleukx, line 194: We could also add a fast path for numpy arrays here? Not sure if it is worth it. It would flatten the np array and operate on that

IgnaceBleukx (approved):
Do we also want to change the actual solver_var implementation of each solver?
There we could also first do the quick check if var.name is in the varmap or not.
I think the first step should at least be to check if it is a _NumVarImpl or not, currently the frist step is to check if it is num, but that is unlikely and slow to check.

We can do it in a different PR tho, this one is good to go in.

cpmpy/solvers/solver_interface.py
tias: we rarely store non-1D numpy(ndvar) arrays in constraints, so, don't think its worth special casing and the return type has to be plain list

tias:
> Do we also want to change the actual solver_var implementation of each solver? There we could also first do the quick check if var.name is in the varmap or not. I think the first step should at least be to check if it is a _NumVarImpl or not, currently the frist step is to check if it is num, but that is unlikely and slow to check.
> 
> We can do it in a different PR tho, this one is good to go in.

that was the original #938 remark indeed : )

Better in a separate PR so this can already go in...?

Issue #979, opened 2026-05-12
Improve native MUS gurobi by setting lower bound of indicator vars to 1


OrestisLomis:
[According to Gurobi staff member Riley Clement](https://support.gurobi.com/hc/en-us/community/posts/45775175552273-IISConstrForce-1-not-guaranteeing-IISConstr-1-in-computeIIS-output?page=1#community_comment_46557868661649), there may be an optimisation for our native MUS approach with Gurobi. It boils down to setting the lower bound of the indicator vars to 1 instead of actually adding this as a constraint.

PR #993, opened 2026-05-21
gurobi improved MUS

OrestisLomis:
This implement the suggested improvement of #979

### DIFF
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -559,6 +559,7 @@ class CPM_gurobi(SolverInterface):
             if len(soft_con_tf) == 0:
                 # this uncommon case ensures `grb_soft_cons` maps to `soft_cons`
                 soft_con_rep = cp.BoolVal(True)
+                grb_soft_cons.append(s._add_transformed(soft_con_rep))
             elif len(soft_con_tf) == 1:
                 # if `con` represented by a single transformed constraint, it can be added as-is
                 soft_con_rep = soft_con_tf[0]
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -546,8 +546,7 @@ class CPM_gurobi(SolverInterface):
         # instantiate Gurobi solver
         s = cls()
 
-        # Gurobi minimizes the IIS over native constraints, so collect every
-        # constraint that is hard for the CPMpy-level MUS mapping and force it.
+        # we collect the Gurobi constraint objects, so we can enable their `IISConstrForce` attribute later
         grb_hard_cons = []
 
 
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -550,7 +550,6 @@ class CPM_gurobi(SolverInterface):
         grb_hard_cons = []
 
 
-        # The Gurobi IIS algorithm minimizes constraints directly, unlike assumption-based solvers. However, a user-level constraint may be transformed to a group of multiple Gurobi constraints. In this case, we have to represent this group by a *single* soft constraint, otherwise the Gurobi IIS may not map to the user-level constraint MUS. We collect `grb_soft_cons` so that `grb_soft_cons[i]` is a single soft constraint representing `soft[i]`. After calling `computeIIS`, we can read the `IISConstr` (or variations) attribute to see which are in the IIS/MUS.
         grb_soft_cons = []
 
         for soft_con in soft_cons:
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -546,21 +546,20 @@ class CPM_gurobi(SolverInterface):
         # instantiate Gurobi solver
         s = cls()
 
-        # we collect the Gurobi constraint objects, so we can enable their `IISConstrForce` (or variations) attribute later
+        # collect the Gurobi constraint objects
         grb_hard_cons = []
-
-
         grb_soft_cons = []
 
         for soft_con in soft_cons:
-            # manually transform the constraint so we can see whether `soft_con` is represented by more than one constraint
+            # transform each constraint seperately, can map to multiple Gurobi-level constraints
             soft_con_tf = s.transform(soft_con)
 
             if len(soft_con_tf) == 0:
-                # this uncommon case ensures `grb_soft_cons` maps to `soft_cons`
-                soft_con_rep = cp.BoolVal(True)
-                grb_soft_cons.append(s._add_transformed(soft_con_rep))
-            elif len(soft_con_tf) == 1:
+                raise ValueError
+                # uncommon case, just ensure `grb_soft_con` and `soft` are same length
+                soft_con_tf = [cp.BoolVal(True)]
+            
+            if len(soft_con_tf) == 1:
                 # if `con` represented by a single transformed constraint, it can be added as-is
                 soft_con_rep = soft_con_tf[0]
                 grb_soft_cons.append(s._add_transformed(soft_con_rep))
@@ -583,7 +582,7 @@ class CPM_gurobi(SolverInterface):
             # use ._add_transformed instead of .add because we need the Gurobi constraint object later
             grb_hard_cons.append(s._add_transformed(cpm_con))
 
-        # update required to avoid `gurobipy._exception.GurobiError: GenConstr has not yet been added to the model` when accessing constraint attribute.
+        # update model so we can access consrtaint attribtutes
         # model updates can be expensive, so we do this only once!
         s.native_model.update()
         for grb_con in grb_hard_cons:
@@ -605,7 +604,7 @@ class CPM_gurobi(SolverInterface):
         except gp.GurobiError as e:
             if e.errno == gp.GRB.Error.IIS_NOT_INFEASIBLE:
                 raise AssertionError("MUS: model must be UNSAT")
-            raise
+            raise e # something else happened
 
         def in_iis(grb_con):
             """Check if `grb_con` is in the IIS. The exact attribute name depends on the type of Gurobi constraint."""
--- a/cpmpy/solvers/gurobi.py
+++ b/cpmpy/solvers/gurobi.py
@@ -546,11 +546,11 @@ class CPM_gurobi(SolverInterface):
         # instantiate Gurobi solver
         s = cls()
 
-        # we collect the Gurobi constraint objects, so we can enable their `IISConstrForce` attribute later
+        # we collect the Gurobi constraint objects, so we can enable their `IISConstrForce` (or variations) attribute later
         grb_hard_cons = []
 
 
-        # The Gurobi IIS algorithm minimizes constraints directly, unlike assumption-based solvers. However, a user-level constraint may be transformed to a group of multiple Gurobi constraints. In this case, we have to represent this group by a *single* soft constraint, otherwise the Gurobi IIS may not map to the user-level constraint MUS. We collect `grb_soft_cons` so that `grb_soft_cons[i]` is a single soft constraint representing `soft[i]`. After calling `computeIIS`, we can read the relevant IIS attribute to see which are in the IIS/MUS.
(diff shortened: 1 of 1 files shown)


cpmpy/solvers/gurobi.py
IgnaceBleukx: It is needed : )
Soft can be a nested list of constraints (unless you type it otherwise; but I think this the right way of doing things)

cpmpy/solvers/gurobi.py
```diff
@@ -547,14 +547,16 @@ def mus_native(cls, soft, hard=[]):
 
         # we collect the Gurobi constraint objects, so we can enable their `IISConstrForce` attribute later
         grb_hard_cons = []
+        grb_force_cons = []
 
         # transform and add all hard constraints
         for cpm_con in s.transform(hard):
             # note: we use `s.transform`, then `_add_transformed`, to add some of the constraints to the solver, because need to introspect the transformation and require access to the Gurobi constraints. This bypasses normal creation of `user_vars`. This is safe to do since `user_vars` are not used in this algorithm, and the solver object does not leave this function.
```
IgnaceBleukx: I feel this comment is more confusing than helping... I still don't understand why we would skip adding it to the user vars? Is it just a performance improvement? `get_variables` is insanely fast, so it should be quite minor.

Aaah no I get it now: you need the Gurobi constraint object later on.
Ok, best to just document it like that then...
E.g.,
```
# use ._add_transformed instead of .add because we need the Gurobi constraint object later
```

cpmpy/solvers/gurobi.py
```diff
@@ -568,24 +570,28 @@ def mus_native(cls, soft, hard=[]):
...
-                # `soft_con_tf` is a group of multiple constraints. We introduce an assumption variable `a` and add *hard* constraint `a -> /\ tf_cons`. Then, `a` be a single soft constraint implying `soft_con`
+                # We introduce an assumption variable `a` and add *hard* constraints `a -> /\ tf_cons`.
+                # The lower bound fixing `a == 1` is the soft part Gurobi can select in the IIS.
                 assumption = cp.boolvar()
 
                 # adding `a -> /\ C` may require re-transform due to the added implication
                 additional_hard_constraint = assumption.implies(cp.all(soft_con_tf))
                 for tf_con in s.transform(additional_hard_constraint):
-                    grb_hard_cons.append(s._add_transformed(tf_con))
+                    grb_force_cons.append(s._add_transformed(tf_con))
```
IgnaceBleukx: This very much feels like you're doing things backwards.
Wouldn't it make more sense to first do all of this soft-constraints stuff, and just add the mapping you get here to the original hard constraint list?

IgnaceBleukx (changes requested):
I'm still getting to grips with the code, so also some comments on the original implementation.

If I understand correctly, the tool works as follows:
You want to find an MUS for a set of soft CPMpy constraints; there can be hard CPMpy constraints as well.
A single CPMpy constraint may require multiple Gurobi-constraints. So, we introduce a Boolean variable `a` to represent the group, and add the hard constraints
```
a -> a_1 /\ a_2 /\ ... a_n
a_1 -> grb_transformed_constraint_1
a_2 -> grb_transformed_constraint_2
...
```
the soft constraint is then `a = 1` ?

There seems to be too much bookkeeping going on. What is the difference between grb_force_cons and grb_hard_cons?

hbierlee:
Mostly correct, in my mind it is simpler: for MIP solvers (or at least for gurobi) the main interface difference with the assumption-based solvers is that instead of only Boolean literals we can directly make a MUS out of supported "soft" constraints. It's a little more general in that sense. The transformation process is identical though, we just make an assumption model (reify`a -> C` for each model constraint `C`, which is then transformes in whatever way. We don't have to intro a_1, ... This is just like `make_assump_model` right?).

There are two optimizations we then added, 1) since we can post a supported soft constraint directly, we don't introduce `a` if C is already supported (e.g. a linear), and 2) as per Riley's comment `a` can be a constant because gurobi allows the bound of `a` to be part of the MUS.

Maybe some things can be reshuffled, however, note that we should only do model.update once as it's expensive. I believe this is what causes some of the extra bookkeeping.

I think the grb_hard_cons was just renamed, but the original instantiation was left in.

cpmpy/solvers/gurobi.py
hbierlee: Fine to remove comment then

cpmpy/solvers/gurobi.py
hbierlee: Also fine with that

cpmpy/solvers/gurobi.py
```diff
@@ -547,14 +547,16 @@ def mus_native(cls, soft, hard=[]):
 
         # we collect the Gurobi constraint objects, so we can enable their `IISConstrForce` attribute later
         grb_hard_cons = []
```
hbierlee, line 550: Probably meant to be removed?

cpmpy/solvers/gurobi.py
hbierlee: I think that would also work

cpmpy/solvers/gurobi.py
```diff
@@ -547,14 +547,16 @@ def mus_native(cls, soft, hard=[]):
...
+        grb_force_cons = []
```
hbierlee: This name also makes sense of course, however, now we're mixing the terms "force" and "soft". I'd prefer "hard" from that perspective

hbierlee (commented):
The new optimization looks good to me. Ignace's comments can further simplify the implementation which would be good

cpmpy/solvers/gurobi.py
OrestisLomis: I'm not sure I get what you mean. The hard constraint machinery comes after this. Only the transformation of the hard constraint to gurobi comes before which is fine to be moved but not sure if that's what you really meant.

OrestisLomis:
Just to add to what @hbierlee already said. We indeed do not "need" the assumption variables for a MIP model. A MIP solver can just compute an IIS for the constraints as they are. The need for the assumption variables (and here it makes more sense to call them indicator variables) is to group gurobi constraints as CPMpy level constraints. So in practice a soft constraint such as `p >= 1 & p <= 2` gets transformed into a list of gurobi constraints and each of them gets reified. So we get `a_1 -> p >= 1` and `a_2 -> p <= 2`. Crucially these constraints should be hard constraints because they always hold (will clarify with an example below, this breaks for other MIP solvers which don't expose hard constraints). Then you can add a soft constraint `a_1 >= 1`.  As @hbierlee said there are two improvements. The first one is that we don't actually need to reify constraints when the group is just a singleton, this was already implemented in the original PR #880. The second one is that actually we do not need `a_1 >= 1` as a soft constraint either. The model should be simpler if we just set the lowerbound of `a_1` to 1 and then that lowerbound can be found as part of the MUS/IIS, which is something MIP solvers can do.

There is indeed quite a bit of bookkeeping going on with different type of gurobi constraints which all have different API calls. The `grb_hard_cons` and `grb_force_cons` were essentially the same thing. Indeed no reason to have had them split in two so I simplified the code.

Now why do the reified constraints need to be hard constraints? Essentially just using the same indicator variable does not force the grouping by itself. Consider the problem where we have `p >= 1 & p < 1` and `p < 1` as the constraints in our model, which for gurobi becomes `a_1 -> p >= 1`, `a_1 -> p < 1` and `p < 0`. Clearly the actual MUS should just be the constraint `p >= 1 & p < 1` . But if all the constraints `a_1 -> p >= 1`, `a_1 -> p < 1` and `p < 1` are soft what can happen is that Gurobi's IIS can return `a_1 -> p >= 1`, `p < 1` and the lowerbound of `a_1` as part of the IIS, bypassing the grouping.

OrestisLomis:
Also I've gone over the requested changes and also fixed a small bug that was left behind. Feel free to have another look.

hbierlee:
> So in practice a soft constraint such as p >= 1 & p <= 2  gets transformed into a list of gurobi constraints and each of them gets reified. So we get a_1 -> p >= 1 and a_2 -> p <= 2

Assuming this example is really a single soft constraint (a little ambiguous due to the &), the soft constraint p >= 1 & p <= 2 is added as hard constraint a-> (p >= 1 & p <= 2) which is then hopefully transformed to (a-> (p >= 1)) & (a-> (p <= 2)). If you mean these are two soft constraints you'd get two a's. However my main point is that this is all no different from assumption-based solvers when you do make_assump_model (and the reified constraints therein also become 'hard constraints ' ; it's just the assumptions which need not all hold)

OrestisLomis:
Yes, you are correct. Though it doesn't really break the example as the transformed constraints will typically still be considered as separate soft constraints internally. So unless a solver supports a conjunctive constraints they should still be hard constraints. But this discussion is more relevant for other solvers/PRs anyway so no need to go to deep into the details.

cpmpy/solvers/gurobi.py
```diff
@@ -570,19 +564,27 @@ def mus_native(cls, soft, hard=[]):
                 # if `con` represented by a single transformed constraint, it can be added as-is
                 soft_con_rep = soft_con_tf[0]
             else:
-                # `soft_con_tf` is a group of multiple constraints. We introduce an assumption variable `a` and add *hard* constraint `a -> /\ tf_cons`. Then, `a` be a single soft constraint implying `soft_con`
+                # We introduce an assumption variable `a` and add *hard* constraints `a -> /\ tf_cons`.
+                # The lower bound fixing `a == 1` is the soft part Gurobi can select in the IIS.
                 assumption = cp.boolvar()
 
                 # adding `a -> /\ C` may require re-transform due to the added implication
                 additional_hard_constraint = assumption.implies(cp.all(soft_con_tf))
```
IgnaceBleukx: Aren't you still doing the same thing twice here? I would think you can just add `cp.all(soft_cons_tf)` to the CPMpy hard constraints, and then they're handled below just like all the ohters?

cpmpy/solvers/gurobi.py
```diff
@@ -570,19 +564,27 @@ def mus_native(cls, soft, hard=[]):
...
-                # `a >= 1` will be the single soft constraint to indicate whether `soft_cons[i]` is in the MUS
-                soft_con_rep = assumption >= 1
+                soft_con_rep = s.solver_var(assumption)
+                soft_con_rep.setAttr("LB", 1)
 
-            grb_soft_cons.append(s._add_transformed(soft_con_rep))
+            if isinstance(soft_con_rep, gp.Var):
+                grb_soft_cons.append(soft_con_rep)
+            else:
+                grb_soft_cons.append(s._add_transformed(soft_con_rep))
```
IgnaceBleukx: I would move this into the if-then-else structs above?
For BoolVal(True), no need to do anything,
for single cons, do the add_transformed thingy, for assumptions vars, do the append? The map exactly with the cases here

cpmpy/solvers/gurobi.py
```diff
@@ -541,22 +541,16 @@ def mus_native(cls, soft, hard=[]):
...
-            grb_con = s._add_transformed(cpm_con)
-            grb_hard_cons.append(grb_con)
 
-        # The Gurobi IIS algorithm minimizes constraints directly, unlike assumption-based solvers. However, a user-level constraint may be transformed to a group of multiple Gurobi constraints. In this case, we have to represent this group by a *single* soft constraint, otherwise the Gurobi IIS may not map to the user-level constraint MUS. We collect `tf_soft` so that `tf_soft[i]` is a single soft constraint representing `soft[i]`. After calling `computeIIS`, we can read the `IISConstr` attribute to see which are in the IIS/MUS.
+        # The Gurobi IIS algorithm minimizes constraints directly, unlike assumption-based solvers. However, a user-level constraint may be transformed to a group of multiple Gurobi constraints. In this case, we have to represent this group by a *single* soft constraint, otherwise the Gurobi IIS may not map to the user-level constraint MUS. We collect `grb_soft_cons` so that `grb_soft_cons[i]` is a single soft constraint representing `soft[i]`. After calling `computeIIS`, we can read the `IISConstr` (or variations) attribute to see which are in the IIS/MUS.
```
IgnaceBleukx: I'm actually not sure what this comment contributes over the docstring?

IgnaceBleukx (dismissed):

cpmpy/solvers/gurobi.py
OrestisLomis: assuming you actually meant `assumption.implies(cp.all(soft_con_tf))` then yes, you are right.

Issue #815, opened 2026-01-12
Getting `.value()` for variables introduced in a transformation


MaximilianMoeller:
### Context
When computing step-wise explanations, obtaining a solution to a model is the ‘easy’ part.
The main computational challenge only starts after the fact. It consists of iteratively finding (special kinds of) MUSs of the input constraints extended by the negations of literals appearing in the previously obtained solution.

### The Setup
I’m using existing CPMpy models for my experiments, many of which contain integer variables and global constraints (e.g., `AllDifferent`). However, the formalism I use to compute the explanation steps works on the level of CNFs. Translation through `to_cnf` works seamless and through the `ivarmap` argument and the resulting `IntVarEnc`s, I can get the relations between original variables and translated variables that I need.

### The Challenge
In order to compute the explanations I need a solution to the _translation_ of the original constraints.
It takes considerably more time to solve the translated model than the original one, which I would expect, given that a solver is provided with more variables and less context about them. To speed up the development process it would be very nice to solve the original model containing integer variables and global constraints and then obtain the assignments of the translated variables from that ‘high level’ solution.
I think I can already achieve this to some extend, by using `ivarmap[original_var.name].encode_comparison('==', original_var.value())` and then (depending on the chosen encoding) deriving the values for the variables appearing in `ivarmap[original_var.name].vars()`.

But what about all the auxiliary variables that were introduced during `to_cnf`?
More generally, this doesn’t seem limited to the `to_cnf` transformation, but every transformation could not only transform constraints to ‘lower level’ constraints but also (partial?) solutions to (partial?) solutions of said ‘lower level’ constraints.

### Related
This comment almost answers my question, but it was written 2.5 years ago, so maybe something (in the architecture of CPMpy) changed that would make this easy now?👀
 
> Calling the value on the decomposition will not work when that decomposition introduces auxiliary variables.
> 
> The new variables will not have a value (as they are not passed through any solve-call) so the `.value()` of the global constraint will always be `None`.
> 
> […]

 _Originally posted by @IgnaceBleukx in [#336](https://github.com/CPMpy/cpmpy/issues/336#issuecomment-1580173729)_

PR #994, opened 2026-05-21
use order encoding for inequalities

OrestisLomis:
This PR makes sure an order encoding is used for inequalities. Do note that I had to create a fresh CSEMap in the tests because there is already a linearization with equalities in the setup and currently CPMpy does not support multiple encodings for the same variable. I do remember Henk saying this might be something that is useful, but it is clearly outside of the scope of this PR.

### DIFF
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -55,10 +55,12 @@ class CSEMap:
                                                       dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]]:
         """Collect bv <-> var == val and bv <-> var >= val expressions in flat_map."""
 
+        comps = frozenset({"==", ">="})
+
         var_vals = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var == val: var -> [(val, bv)]
         var_bounds = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var >= val: var -> [(val, bv)]
         for expr, bv in self.flat_map.items():
-            if expr.name in {"==", ">="}:
+            if expr.name in comps:
                 var, val = expr.args
                 if isinstance(var, _IntVarImpl) and is_int(val):
                     target = var_vals if expr.name == "==" else var_bounds
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -63,6 +63,18 @@ class CSEMap:
 
         return var_vals
 
+    def get_reified_varbounds(self) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
+        """collect all bv <-> var >= val expressions in flat_map"""
+
+        var_bounds = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var: [val, bv]
+        for expr, bv in self.flat_map.items():
+            if expr.name == ">=":
+                var, val = expr.args
+                if isinstance(var, _IntVarImpl) and is_int(val):
+                    var_bounds.setdefault(var, []).append((val, bv))
+
+        return var_bounds
+
     def get_or_make_var(self, expr: Expression) -> tuple[_IntVarImpl, Optional[Expression]]:
         """
         Make an auxiliary variable for the given expression
@@ -108,5 +120,5 @@ class CSEMap:
                 new_expr = Comparison("==", lhs, rhs)
                 new_expr._has_subexpr = expr._has_subexpr
                 return new_expr, True
-            
+
         return expr, False
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -51,29 +51,17 @@ class CSEMap:
         """Get the decomposition of the given global constraint or global function."""
         return self.decomp_map.get(expr, default)
 
-    def get_reified_varvals(self) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
-        """collect all bv <-> var == val expressions in flat_map"""
+    def get_reified_comparisons(self, cmp) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
+        """collect all bv <-> var cmp val expressions in flat_map, where cmp can be `==` or `>=`"""
         
         var_vals = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var: [val, bv]
         for expr, bv in self.flat_map.items():
-            if expr.name == "==":
+            if expr.name == cmp:
                 var, val = expr.args
                 if isinstance(var, _IntVarImpl) and is_int(val):
                     var_vals.setdefault(var, []).append((val, bv))
 
         return var_vals
-
-    def get_reified_varbounds(self) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
-        """collect all bv <-> var >= val expressions in flat_map"""
-
-        var_bounds = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var: [val, bv]
-        for expr, bv in self.flat_map.items():
-            if expr.name == ">=":
-                var, val = expr.args
-                if isinstance(var, _IntVarImpl) and is_int(val):
-                    var_bounds.setdefault(var, []).append((val, bv))
-
-        return var_bounds
 
     def get_or_make_var(self, expr: Expression) -> tuple[_IntVarImpl, Optional[Expression]]:
         """
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -51,17 +51,20 @@ class CSEMap:
         """Get the decomposition of the given global constraint or global function."""
         return self.decomp_map.get(expr, default)
 
-    def get_reified_varval_comparisons(self, cmp) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
-        """collect all bv <-> var cmp val expressions in flat_map, where cmp can be `==` or `>=`"""
-        
-        var_vals = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var: [val, bv]
+    def get_reified_varval_comparisons(self) -> tuple[dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]],
+                                                      dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]]:
+        """Collect bv <-> var == val and bv <-> var >= val expressions in flat_map."""
+
+        var_vals = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var == val: var -> [(val, bv)]
+        var_bounds = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var >= val: var -> [(val, bv)]
         for expr, bv in self.flat_map.items():
-            if expr.name == cmp:
+            if expr.name in {"==", ">="}:
                 var, val = expr.args
                 if isinstance(var, _IntVarImpl) and is_int(val):
-                    var_vals.setdefault(var, []).append((val, bv))
+                    target = var_vals if expr.name == "==" else var_bounds
+                    target.setdefault(var, []).append((val, bv))
 
-        return var_vals
+        return var_vals, var_bounds
 
     def get_or_make_var(self, expr: Expression) -> tuple[_IntVarImpl, Optional[Expression]]:
         """
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -52,7 +52,7 @@ class CSEMap:
         return self.decomp_map.get(expr, default)
 
     def get_reified_varvalbounds(self) -> tuple[dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]],
-                                                      dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]]:
+                                                dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]]:
         """Collect bv <-> var == val and bv <-> var >= val expressions in flat_map."""
 
         comps = frozenset({"==", ">="})
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -51,7 +51,7 @@ class CSEMap:
         """Get the decomposition of the given global constraint or global function."""
         return self.decomp_map.get(expr, default)
 
-    def get_reified_comparisons(self, cmp) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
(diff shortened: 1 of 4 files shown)


OrestisLomis:
Fairly sure the failing checks are because of the xcsp3_26 branch itself..

OrestisLomis:
@hbierlee is this what you had in mind? I'm checking on some generated instance and actually the runtime seems to be worse..

hbierlee:
Well the expected test output looks good. I can't really say anything wrt runtime without seeing the actual generated instance and its transformation with and without this change, for which solver etc. Also for the unit test it would be good to see a before and after.

I did notice that these can be expensive transformations, so you might want to look separately at transformation time and solve time. If your generated instances are trivial to solve, then this change may only make the transfo worse.

There's no channelling for this test between the int variable and the encoding (depending on the solver, that makes sense). May be good to make these tests so that the channelling is added, so we can see that it's reasonable too (e.g. right constraint, and not added multiple times)

tias:
if it doesn't seem to pay off and/or is invasive to implement then you can just skip it and look at something else...

OrestisLomis:
actually I tested a bit more extensively and it does seem to have a positive effect. The tests mentioned earlier were not entirely correct.

IgnaceBleukx:
For which kind of models/consrtaints do we expect an improvement with this?
The main thing I can think of is when we're posting a time-decomposed cumulative constraint. There we have nested (start >= val) expressions. It makes sense to use an order encoding for those variables, indeed.

Maybe run a small experiment on j60 of psplib with gurobi/exact?

I'm wondering what we should prefer: if a model has both `x == val` and `x >= val` as subexpressions, which encoding would be best to use?

OrestisLomis:
Alright, I'll check on some real instances. Now I was just testing on an explicit instance generator which basically has a lot of ```((var1 >= val1) or (var2 >= val2) or ..)``` clauses

tests/test_trans_linearize.py
```diff
@@ -648,15 +648,15 @@ def test_linearize_reified_disequalities(self):
         out = linearize_reified_variables(self.linearize((a != 1) | (a != 2)), min_values=2, csemap=self.csemap, ivarmap=self.ivarmap)
         assert str(out) == "[(~BV[a == 1]) or (~BV[a == 2]), sum(BV[a == 1], BV[a == 2], BV[a == 3]) == 1]"
 
+    def test_linearize_reified_inequalities(self):
+        """Use order encoding on inequalities and replace `a>=k` expressions"""
+        a = self.a
+        self.csemap = CSEMap()
```
hbierlee, line 655: It would be a good idea to also add a "mixed encdoing" test case where a is both direct/order encoded (I presume at this point the order matters? So then I'd also check both orders in that test, or separate tests)

cpmpy/transformations/cse.py
```diff
@@ -63,6 +63,18 @@ def get_reified_varvals(self) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]
 
         return var_vals
 
+    def get_reified_varbounds(self) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
+        """collect all bv <-> var >= val expressions in flat_map"""
+
+        var_bounds = dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]()  # var: [val, bv]
+        for expr, bv in self.flat_map.items():
+            if expr.name == ">=":
```
hbierlee: only one character difference with the other function. I'd vote for making a single function more general e.g. `get_reified_comparisons(self, cmp)`

tests/test_trans_linearize.py
```diff
@@ -648,15 +648,15 @@ def test_linearize_reified_disequalities(self):
...
     # The following tests are marked with `xfail` because they are expected to fail, because they are not yet implemented; to see the current output compared with the desired output in the test, run with `pytest --runxfail`
     
-    @pytest.mark.xfail(reason="aspirational")
-    def test_linearize_reified_inequalities(self):
-        """Use order encoding on inequalities and replace `a>=1` expressions"""
-        a = self.a
-        out = linearize_reified_variables(self.linearize((a >= 1) | (a >= 2)), min_values=2, csemap=self.csemap, ivarmap=self.ivarmap)
-        assert str(out) == "[(BV[a >= 1]) or (BV[a >= 2]), sum([1, -1] * (BV[a >= 2], BV[a >= 1])) <= 0, sum([1, -1] * (BV[a >= 3], BV[a >= 2])) <= 0]"
 
     @pytest.mark.xfail(reason="aspirational")
```
hbierlee: maybe this PR can add another aspirational for `1 >= a`  (which I suppose a future canonicalization can take off?)

cpmpy/transformations/linearize.py
```diff
@@ -675,18 +677,43 @@ def linearize_reified_variables(constraints, min_values=3, csemap=None, ivarmap=
...
+            continue  # do not encode
+
+        lb, ub = var.lb, var.ub
+        # order encoding has one Boolean per non-trivial lower bound threshold
+        vals = [(val, bv) for val, bv in vals if lb < val <= ub]
+        if len(vals) == 0:
+            continue  # do not encode
+
+        _, domain_constraint = _encode_int_var(my_ivarmap, var, "order", csemap=csemap)
+        toplevel.extend(domain_constraint)
```
hbierlee: seems to be missing the `if ivarmap is None` case for solvers which keep the integer var?

cpmpy/transformations/linearize.py
```diff
@@ -675,18 +677,43 @@ def linearize_reified_variables(constraints, min_values=3, csemap=None, ivarmap=
...
+    for var, vals in var_bounds.items():
+        if var.name in my_ivarmap and not isinstance(my_ivarmap[var.name], IntVarEncOrder):
+            continue  # another encoding already represents this variable
+
+        if len(vals) < min_values:
+            continue  # do not encode
+
+        lb, ub = var.lb, var.ub
+        # order encoding has one Boolean per non-trivial lower bound threshold
+        vals = [(val, bv) for val, bv in vals if lb < val <= ub]
```
hbierlee: we may want to test a <= 10 and a <= -2 for these out of domain values.

cpmpy/transformations/linearize.py
```diff
@@ -675,18 +677,43 @@ def linearize_reified_variables(constraints, min_values=3, csemap=None, ivarmap=
...
             bv_map[bv] = (var, val)
 
-    if len(bv_map) > 0:
-        # Now clean up and remove the '(var == val) == bv' constraints:
+    bound_bv_map = {}  # bv -> (var, val)
+    for var, vals in var_bounds.items():
+        if var.name in my_ivarmap and not isinstance(my_ivarmap[var.name], IntVarEncOrder):
+            continue  # another encoding already represents this variable
+
+        if len(vals) < min_values:
```
hbierlee: I suppose this should happen after removing trivial bounds?

cpmpy/transformations/linearize.py
```diff
@@ -675,18 +677,43 @@ def linearize_reified_variables(constraints, min_values=3, csemap=None, ivarmap=
...
+            continue  # do not encode
+
+        lb, ub = var.lb, var.ub
+        # order encoding has one Boolean per non-trivial lower bound threshold
+        vals = [(val, bv) for val, bv in vals if lb < val <= ub]
+        if len(vals) == 0:
```
hbierlee: Also the <0 can be removed if we assert min_values>=0.

cpmpy/transformations/linearize.py
hbierlee: also, the way this is done for `direct` can be replaced by an int2bool function call see https://github.com/CPMpy/cpmpy/issues/815

hbierlee (dismissed):
I think we should consider deduplicating this, on the one hand we are not expecting to see a third case of this pattern (given we're planning to do canonicalization to handle `<,!=,...`, and for the binary encoding it's not worth it), however, it's quite a lot of complex code which do the exact same thing, just for >= i/o ==.

tests/test_trans_linearize.py
IgnaceBleukx: Python already converts this by itself : )

```python
import cpmpy as cp
a = cp.intvar(0,1,name"a")
print(1 >= a)
>> a <= 1
```

cpmpy/transformations/linearize.py
OrestisLomis: Not sure if I understood correctly what you meant with the int2bool function call. I added a helper function to refactor it a bit more cleanly.

tests/test_trans_linearize.py
OrestisLomis: I'm a bit confused. This doesn't seem related to this PR, but rather #996. @IgnaceBleukx is correct that it does already get rewritten, but since it is not in canonical form the order encoding is actually not used here (as that explicitly expects comparisons of the form 'a >= k').

cpmpy/transformations/linearize.py
hbierlee: I meant that (as you can see somewhere in this comment https://github.com/CPMpy/cpmpy/issues/815#issuecomment-3760075645) we can use `_encode_lin_expr_` to get what the new `encode_value_constraint` now does

tests/test_trans_linearize.py
hbierlee: Yes it's about #996 but I thought it wouldn't hurt to just already add new aspirational tests here. If this test already exists in #996 then it's ok to skip, but otherwise it would also be fine to add it to this PR (it's related in the sense that it's just showing the limits of what's currently implemented)

Now I still don't understand if this is or is not handled by the current code without #996 ?

IgnaceBleukx:
I did not get through the code yet, but from a high level my main question is: what should we do if we have both `x == val` and `x >= val` expressions in the csemap?

Would it make sense to just have both encodings? And if so, the int2bool infrastructure capable of dealing with that?
Otherwise, we'll have to make heuristic choice on what encoding we use after all. I suppose @hbierlee will have some insights regarding this.

OrestisLomis:
There is no support for multiple encodings for one variable (there is for different variables). And this PR also does not introduce such support, that would be a bigger change to the infrastructure. Right now the first occurence takes precedence. After the rewrite you could do a weighted comparison between the number of occurences of ```var == val``` vs ```var >= val```. I'm indeed curious what @hbierlee has to say about what good weights would be. On the other hand constraints that are added incrementally could throw a spanner in the works.

tests/test_trans_linearize.py
OrestisLomis: Okay so since #996 is merged into master it indeed is fine to just have it here and I could easily check with both changes working together if it is handled correctly.. Unfortunately it is not in fact handled correctly, see test ```test_linearize_reified_inequality_misordered```. The order is flipped so actually the expression is canonical, but then it does not get the order encoding. I will investigate for a bit and otherwise mark it as aspirational.

OrestisLomis: The bounds I had chosen were trivial so min_values was not triggering.. fixed it

IgnaceBleukx:
I've changed the implementation to something that is much closer to what we had before.
Its now much more explicit what is going on in the function, also updated the docstring.

@tias should probably have a look, as he implemented the original linearize_varvals transformation

cpmpy/transformations/cse.py
```diff
@@ -51,12 +51,12 @@ def get_decomposition(self, expr: Expression, default: Any = None) -> Any:
         """Get the decomposition of the given global constraint or global function."""
         return self.decomp_map.get(expr, default)
 
-    def get_reified_varvals(self) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
-        """collect all bv <-> var == val expressions in flat_map"""
+    def get_reified_varval_comparisons(self, cmp) -> dict[_IntVarImpl, list[tuple[int, _BoolVarImpl]]]:
```
tias: normally we don't change function names but add default arguments instead, e.g.

`get_reified_varvals(self, cmp='==')`

but the function is so new that its not a blocker for me

cpmpy/transformations/linearize.py
```diff
@@ -629,38 +629,58 @@ def get_linear_decompositions():
...
 
-    var_vals = csemap.get_reified_varvals()
-    
     # Make the integer encodings in integer linear friendly way
     my_ivarmap = ivarmap if ivarmap is not None else {}
-    toplevel = []
-    bv_map = {}  # bv -> (var, val)
+
+    var_vals = csemap.get_reified_varval_comparisons("==")
+    var_bounds = csemap.get_reified_varval_comparisons(">=")
```
tias: I like the generalisation, but you are now walking the full csemap twice...

an alternative would be to collect both in one loop?
e.g. `get_reified_varvalbounds() -> (var_vals, var_bounds)`

cpmpy/transformations/linearize.py
```diff
@@ -629,38 +629,58 @@ def get_linear_decompositions():
...
+
+    # decide the encoding to use for each variable
+    var_encodings = dict() # var -> (encoding, vals)
     for var, vals in var_vals.items():
-        # check if we should linearize the reified variables
         lb, ub = var.lb, var.ub
         vals = [(val, bv) for val, bv in vals if lb <= val <= ub]  # only the valid values, in bounds!
-        if len(vals) < min_values:
-            continue  # do not encode
+        if len(vals) >= min_values and var.name not in my_ivarmap:
```
tias: if it can't be in my_ivarmap, maybe check this as first thing in the for loop; no need to collect the 'vals' then?

tias (commented):
looks solid, small if/skip possible; I do am wondering about the 2 passes...

cpmpy/transformations/linearize.py
OrestisLomis: I changed it to be in one pass, of course this does change the function signature as you already indicated but I suppose that's fine

OrestisLomis:
Alright I like your rewrite @IgnaceBleukx and applied the requested changes from @tias. Definitely still worth considering a heuristic to choose between the two encoding options @hbierlee.

hbierlee:
I don't think there's much interesting to do at this point in terms of heuristic, I would just pick whatever constraint you see first. You could add a group+count on the var-vals/bounds you see (e.g. to choose direct if we have more `x=v`'s and than `x>=v`'s). But I wouldn't go to through much trouble since a future dual/channeled encoding will likely be better.

On channeled encodings, note if you already have one encoding (e.g. direct), my idea would be to not add a full order encoding as well: you just add the other (e.g. order) encoding's literals `x>=v` lazily as they come up. Then, just channel that literal to the direct encoding i.e. `x=v <-> (x>=v /\ x<=v)`. Note you don't even need the order encoding's consistency constraint i.e. `x>=v -> x>=v-1`; note this also works the other way around where you have a full order encoding, with consistency constraint, and a partial direct encoding without consistency constraint. The only difficulty is that `x<=v` may not have been introduced. Apart from that, I think that will perform best (better than choosing between two single encodigs, or having two complete encodings), and would be easier to implement too (i.e. every integer variable has one "root" encoding, and there are just some extra literals floating around to encode remaining `x>=v`'s).

IgnaceBleukx (approved):
Great, I think we're there.

Made some final performance-oriented tweaks and added proper typing of the transformation.
Good to merge after the tests complete

PR #995, opened 2026-05-22
Fast path solver var

IgnaceBleukx:
As discussed in previous PRs, proposal for a fast path for solver var
Three main things:

1. First check if it is a variable (most common case)
2. Avoid checking for _IntVarImpl and _BoolVarImpl (just do var.is_bool())
3. Use is_int instead of is_num in final check

dev/time_ortools.py reduces from 3.7s post time to 2.7 post time

### DIFF
--- a/cpmpy/solvers/TEMPLATE.py
+++ b/cpmpy/solvers/TEMPLATE.py
@@ -55,7 +55,7 @@ from packaging.version import Version
 from .solver_interface import SolverInterface, SolverStatus, ExitStatus, Callback
 from ..expressions.core import Expression, Comparison, Operator
 from ..expressions.variables import _BoolVarImpl, NegBoolView, _IntVarImpl, _NumVarImpl
-from ..expressions.utils import is_num, is_any_list, is_boolexpr
+from ..expressions.utils import is_num, is_int, is_any_list, is_boolexpr
 from ..transformations.get_variables import get_variables
 from ..transformations.normalize import toplevel_list
 from ..transformations.safening import no_partial_functions
@@ -263,30 +263,33 @@ class CPM_template(SolverInterface):
         """
             Creates solver variable for cpmpy variable
             or returns from cache if previously created
+            or returns a constant if the variable is a constant
         """
-        if is_num(cpm_var): # shortcut, eases posting constraints
+        if isinstance(cpm_var, _NumVarImpl):
+            # [GUIDELINE] some solver interfaces explicitely create variables on a solver object
+            #       then use self.TPL_solver.NewBoolVar(...) instead of TEMPLATEpy.NewBoolVar(...)
+
+            name = cpm_var.name
+            revar = self._varmap.get(name)
+            if revar is not None:
+                return revar
+
+            # not yet created, make a new solver var
+            if cpm_var.is_bool():
+                if isinstance(cpm_var, NegBoolView):
+                    # special case, negative-bool-view: work directly on var inside the view
+                    revar = TEMPLATEpy.negate(self.solver_var(cpm_var._bv))
+                else:
+                    revar = TEMPLATEpy.NewBoolVar(str(cpm_var))
+            else:
+                revar = TEMPLATEpy.NewIntVar(cpm_var.lb, cpm_var.ub, str(cpm_var))
+            self._varmap[name] = revar
+            return revar
+
+        if is_int(cpm_var):  # shortcut, eases posting constraints
             return cpm_var
 
-        # [GUIDELINE] some solver interfaces explicitely create variables on a solver object
-        #       then use self.TPL_solver.NewBoolVar(...) instead of TEMPLATEpy.NewBoolVar(...)
-
-        # special case, negative-bool-view
-        # work directly on var inside the view
-        if isinstance(cpm_var, NegBoolView):
-            return TEMPLATEpy.negate(self.solver_var(cpm_var._bv))
-
-        # create if it does not exist
-        if cpm_var.name not in self._varmap:
-            if isinstance(cpm_var, _BoolVarImpl):
-                revar = TEMPLATEpy.NewBoolVar(str(cpm_var))
-            elif isinstance(cpm_var, _IntVarImpl):
-                revar = TEMPLATEpy.NewIntVar(cpm_var.lb, cpm_var.ub, str(cpm_var))
-            else:
-                raise NotImplementedError("Not a known var {}".format(cpm_var))
-            self._varmap[cpm_var.name] = revar
-
-        # return from cache
-        return self._varmap[cpm_var.name]
+        raise NotImplementedError("Not a known var {}".format(cpm_var))
 
 
     # [GUIDELINE] if TEMPLATE does not support objective functions, you can delete this function definition
--- a/cpmpy/solvers/TEMPLATE.py
+++ b/cpmpy/solvers/TEMPLATE.py
@@ -266,9 +266,7 @@ class CPM_template(SolverInterface):
             or returns a constant if the variable is a constant
         """
         if isinstance(cpm_var, _NumVarImpl):
-            # [GUIDELINE] some solver interfaces explicitely create variables on a solver object
-            #       then use self.TPL_solver.NewBoolVar(...) instead of TEMPLATEpy.NewBoolVar(...)
-
+            
             name = cpm_var.name
             revar = self._varmap.get(name)
             if revar is not None:
--- a/cpmpy/solvers/TEMPLATE.py
+++ b/cpmpy/solvers/TEMPLATE.py
@@ -278,9 +278,9 @@ class CPM_template(SolverInterface):
                     # special case, negative-bool-view: work directly on var inside the view
                     revar = TEMPLATEpy.negate(self.solver_var(cpm_var._bv))
                 else:
-                    revar = TEMPLATEpy.NewBoolVar(str(cpm_var))
+                    revar = TEMPLATEpy.NewBoolVar(name)
             else:
-                revar = TEMPLATEpy.NewIntVar(cpm_var.lb, cpm_var.ub, str(cpm_var))
+                revar = TEMPLATEpy.NewIntVar(cpm_var.lb, cpm_var.ub, name)
             self._varmap[name] = revar
             return revar
 
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -91,11 +91,6 @@ class CPM_choco(SolverInterface):
                                     "min", "max", "div", "mod", "pow", "abs", "mul", "count", "element", "nvalue", "among"})
     supported_reified_global_constraints = supported_global_constraints  # choco supports everything reified
 
-    # min and max bounds for integer variables
-    # https://github.com/chocoteam/choco-solver/blob/master/solver/src/main/java/org/chocosolver/solver/variables/IntVar.java#L43
-    MIN_INTEGER_BOUND = -2e31 // 100
-    MAX_INTEGER_BOUND = 2e31 // 100
-
     @staticmethod
     def supported():
         # try to import the package
@@ -333,16 +328,13 @@ class CPM_choco(SolverInterface):
                     revar = self.chc_model.bool_not_view(self.solver_var(cpm_var._bv))
                 else:
                     revar = self.chc_model.boolvar(name=name)
-            elif cpm_var.lb >= self.MIN_INTEGER_BOUND and cpm_var.ub <= self.MAX_INTEGER_BOUND:
-                revar = self.chc_model.intvar(cpm_var.lb, cpm_var.ub, name=name)
             else:
-                raise ChocoBoundsException(f"Choco does not accept variables with bounds outside of range ({self.MIN_INTEGER_BOUND}..{self.MAX_INTEGER_BOUND})")
+                revar = self.chc_model.intvar(cpm_var.lb, cpm_var.ub, name=name)
+           
             self._varmap[name] = revar
             return revar
 
         if is_int(cpm_var):  # shortcut, eases posting constraints
-            if cpm_var < self.MIN_INTEGER_BOUND or cpm_var > self.MAX_INTEGER_BOUND:
-                raise ChocoBoundsException(f"Choco does not accept integer literals with bounds outside of range ({self.MIN_INTEGER_BOUND}..{self.MAX_INTEGER_BOUND})")
             return int(cpm_var)
 
         raise NotImplementedError("Not a known var {}".format(cpm_var))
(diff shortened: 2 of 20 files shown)


tias:
also on my machine it went from 4s to 3.2s, which is really surprising.

so then I realized that NegBoolViews were not in the varmap... but the varmap is a private object right, we can put anything we want in it?

So I did that and cleaned it up a bit further (we were using str(var) as well as var.name? only use one that we precompute?)

went from 3.2s to 3s.

IgnaceBleukx:
Great, I've ported the new pattern to the other solvers too, which makes this a larger PR than I originally anticipated...

Some surprises/concerns I found along the way:
- For pindakaas, we had a recursive call to solver_vars when we encode an integer variable. I have now also saved the integer variables in the varmap, which saves a lot of calls.
- For pysat, we don't actually have a varmap, so there I could not do the same trick
- There was all kind of weirdness going on in the CPO solver_var function, which made it such that a negbool was translated to `(bv == 1) == 0)` which is correct but yeah.... It now uses the proper `logical_not` construct.
- Some cleanups in the choco interface.

tias:
right... I approve the ortools one?

I guess there is no step-by-step, so... lets see how I will proceed on that

cpmpy/solvers/choco.py
```diff
@@ -313,34 +318,34 @@ def solver_var(self, cpm_var):
...
+                    # special case, negative-bool-view: work directly on var inside the view
+                    revar = self.chc_model.bool_not_view(self.solver_var(cpm_var._bv))
+                else:
+                    revar = self.chc_model.boolvar(name=name)
+            elif cpm_var.lb >= self.MIN_INTEGER_BOUND and cpm_var.ub <= self.MAX_INTEGER_BOUND:
+                revar = self.chc_model.intvar(cpm_var.lb, cpm_var.ub, name=name)
             else:
-                raise NotImplementedError("Not a known var {}".format(cpm_var))
-            self._varmap[cpm_var.name] = revar
+                raise ChocoBoundsException(f"Choco does not accept variables with bounds outside of range ({self.MIN_INTEGER_BOUND}..{self.MAX_INTEGER_BOUND})")
```
tias: doesn't choco check and raise this themselves when you try to create the variable?

cpmpy/solvers/choco.py
```diff
@@ -313,34 +318,34 @@ def solver_var(self, cpm_var):
...
+            self._varmap[name] = revar
+            return revar
+
+        if is_int(cpm_var):  # shortcut, eases posting constraints
+            if cpm_var < self.MIN_INTEGER_BOUND or cpm_var > self.MAX_INTEGER_BOUND:
+                raise ChocoBoundsException(f"Choco does not accept integer literals with bounds outside of range ({self.MIN_INTEGER_BOUND}..{self.MAX_INTEGER_BOUND})")
+            return int(cpm_var)
```
tias, line 338: this will also 'int' the bools, is that intended?

cpmpy/solvers/cpo.py
```diff
@@ -342,31 +342,35 @@ def solver_var(self, cpm_var):
...
+            # not yet created, make a new solver var
+            docp = self.get_docp()
+            if cpm_var.is_bool():
+                if isinstance(cpm_var, NegBoolView):
+                    # special case, negative-bool-view: work directly on var inside the view
+                    revar = docp.modeler.logical_not(self.solver_var(cpm_var._bv))
+                else:
+                    # note that a binary var is an integer var with domain (0,1), you cannot do boolean operations on it.
+                    # we should add == 1 to turn it into a boolean expression
+                    revar = docp.expression.binary_var(name) == 1
```
tias: this is weird...? if we do a wsum with booleans, we would be posting ==1's in it?

I think instead we return the binary var; then, if we are in a logical constraint, only then do we do this trick? Its still weird...

cpmpy/solvers/cpo.py
```diff
@@ -342,31 +342,35 @@ def solver_var(self, cpm_var):
...
             else:
-                raise NotImplementedError("Not a known var {}".format(cpm_var))
-            self._varmap[cpm_var.name] = revar
-            self.cpo_model.add(revar >= cpm_var.lb) # ensure the model also has the variable
+                revar = docp.expression.integer_var(min=cpm_var.lb, max=cpm_var.ub, name=name)
+            self._varmap[name] = revar
+            self.cpo_model.add(revar >= cpm_var.lb)  # ensure the model also has the variable
```
tias: whot? that's not how we do it... we are careful not to remove constraints, but we don't do this?

I know you did not introduce it, but since we are touching it...

cpmpy/solvers/exact.py
```diff
@@ -393,33 +393,34 @@ def solver_var(self, cpm_var):
...
+            # not yet created, make a new solver var
+            if cpm_var.is_bool():
+                # special case, negative-bool-view. Should be eliminated in linearize
+                if isinstance(cpm_var, NegBoolView):
+                    raise NotSupportedError("Negative literals should not be left as part of any equation. Please report.")
+                self.xct_solver.addVariable(name)
+            else:
+                lb, ub = cpm_var.get_bounds()
+                if self.encoding is None:
+                    encoding = "order" if ub-lb < 8 else "log"  # heuristic bound
```
tias, line 413: I think exact has its own rules? no need to set one ourself if we don't ahve to...

cpmpy/solvers/gcs.py
```diff
@@ -371,29 +371,32 @@ def solver_var(self, cpm_var):
...
+            revar = self._varmap.get(name)
+            if revar is not None:
+                return revar
+
+            # not yet created, make a new solver var
+            if cpm_var.is_bool():
+                if isinstance(cpm_var, NegBoolView):
+                    # special case, negative-bool-view: work directly on var inside the view
+                    # gcs only works with integer variables, so not(x) = -x + 1
+                    revar = self.gcs.add_constant(self.gcs.negate(self.solver_var(cpm_var._bv)), 1)
```
tias, line 387: other solvers have the same, but we have a transformation for that? or is that only after linearisation? could we still use it so that at least the linears dont have this (or maybe we do that already)?

cpmpy/solvers/hexaly.py
```diff
@@ -248,29 +248,35 @@ def solver_var(self, cpm_var):
...
-            return ~self.solver_var(cpm_var._bv)
-
-        # create if it does not exist
-        if cpm_var.name not in self._varmap:
-            if isinstance(cpm_var, _BoolVarImpl):
-                revar = self.hex_model.bool()
-            elif isinstance(cpm_var, _IntVarImpl):
-                revar = self.hex_model.int(cpm_var.lb, cpm_var.ub)
+        if isinstance(cpm_var, _NumVarImpl):
+            # special case, negative-bool-view
```
tias: this line should be 9 lines lower?

cpmpy/solvers/hexaly.py
```diff
@@ -248,29 +248,35 @@ def solver_var(self, cpm_var):
...
+                if isinstance(cpm_var, NegBoolView):
+                    revar = ~self.solver_var(cpm_var._bv)
+                else:
+                    revar = self.hex_model.bool()
             else:
-                raise NotImplementedError("Not a known var {}".format(cpm_var))
+                revar = self.hex_model.int(cpm_var.lb, cpm_var.ub)
+            
             # set name of variable
             revar.set_name(str(cpm_var))
```
tias: use 'name'? thats what we do elsewhere

tias: well, we should not do this line when revar=... in NegBoolView, so this needs change imho

cpmpy/solvers/minizinc.py
```diff
@@ -519,48 +519,48 @@ async def _solveAll(self, display=None, time_limit:Optional[float]=None, solutio
...
 
-        if cpm_var.name not in self._varmap:
-            # clean the varname
-            varname = cpm_var.name
-            mzn_var = varname.replace(',', '_').replace('.', '_').replace(' ', '_').replace('[', '_').replace(']', '')
+            # not yet created, make a new solver var
+            mzn_var = name.replace(',', '_').replace('.', '_').replace(' ', '_').replace('[', '_').replace(']', '')
 
             # test if the name is a valid minizinc identifier
             if not self.mzn_name_pattern.search(mzn_var):
```
tias, line 543: ah, now I understand the test error, this is also checked for a NegBoolView whose name is '~Bv'

cpmpy/solvers/pindakaas.py
```diff
@@ -215,23 +215,40 @@ def solve(self, time_limit: Optional[float] = None, assumptions: Optional[Iterab
...
-                enc = self.ivarmap[cpm_var.name]
-            return self.solver_vars(enc.vars())
-        else:
-            raise TypeError(f"Unexpected type: {cpm_var}")
+                if name not in self.ivarmap:
+                    enc, cons = _encode_int_var(self.ivarmap, cpm_var, _decide_encoding(cpm_var, None, encoding=self.encoding))
+                    self.add(cons)
+                else:
+                    enc = self.ivarmap[name]
+                revar = self.solver_vars(enc.vars())
```
tias, line 243: that is weird and untypable...

it might need a bigger refactor, where solver_vars is only used in constraints;
I think the other place is in add, that we do solver_vars(v) to make sure they exist/the solvers knows them, but there we should special-treat the ints and check the ivarmap

cpmpy/solvers/pysat.py
```diff
@@ -313,32 +313,37 @@ def _process_solution(self, sol):
...
 
             Transforms cpm_var into CNF literal using ``self.pysat_vpool``
             (positive or negative integer).
 
             So vpool is the varmap (we don't use _varmap here).
         """
-
-        # special case, negative-bool-view
-        # work directly on var inside the view
         if isinstance(cpm_var, BoolVal):
```
tias: bit surprising that we would for pysat not put the most common _NumVarImpl case/check first?

cpmpy/solvers/pysat.py
```diff
@@ -313,32 +313,37 @@ def _process_solution(self, sol):
...
-        elif isinstance(cpm_var, NegBoolView):
-            # just a view, get actual var identifier, return -id
-            return -self.pysat_vpool.id(cpm_var._bv.name)
-        elif isinstance(cpm_var, _BoolVarImpl):
-            return self.pysat_vpool.id(cpm_var.name)
-        elif isinstance(cpm_var, _IntVarImpl):  # intvar
-            if cpm_var.name not in self.ivarmap:
+
+        if isinstance(cpm_var, _NumVarImpl):
+            if cpm_var.is_bool():
```
tias, line 326: declare name= before this line?

cpmpy/solvers/pysat.py
```diff
@@ -313,32 +313,37 @@ def _process_solution(self, sol):
...
+
+            # intvar
+            name = cpm_var.name
+            if name not in self.ivarmap:
                 enc, cons = _encode_int_var(self.ivarmap, cpm_var, _decide_encoding(cpm_var, None, encoding=self.encoding))
                 self.add(cons)
             else:
-                enc = self.ivarmap[cpm_var.name]
+                enc = self.ivarmap[name]
             return self.solver_vars(enc.vars())
```
tias, line 338: equally weird/untypable as pindakaas

cpmpy/solvers/TEMPLATE.py
```diff
@@ -263,30 +263,33 @@ def solver_var(self, cpm_var):
...
-            return TEMPLATEpy.negate(self.solver_var(cpm_var._bv))
-
-        # create if it does not exist
-        if cpm_var.name not in self._varmap:
-            if isinstance(cpm_var, _BoolVarImpl):
-                revar = TEMPLATEpy.NewBoolVar(str(cpm_var))
-            elif isinstance(cpm_var, _IntVarImpl):
-                revar = TEMPLATEpy.NewIntVar(cpm_var.lb, cpm_var.ub, str(cpm_var))
+        if isinstance(cpm_var, _NumVarImpl):
+            # [GUIDELINE] some solver interfaces explicitely create variables on a solver object
```
tias: I found this confusing... surely a solver developer will know whether to call it on the library or the solver... can be cut (its weirdly placed here now)

cpmpy/solvers/TEMPLATE.py
```diff
@@ -263,30 +263,33 @@ def solver_var(self, cpm_var):
...
+            if revar is not None:
+                return revar
+
+            # not yet created, make a new solver var
+            if cpm_var.is_bool():
+                if isinstance(cpm_var, NegBoolView):
+                    # special case, negative-bool-view: work directly on var inside the view
+                    revar = TEMPLATEpy.negate(self.solver_var(cpm_var._bv))
+                else:
+                    revar = TEMPLATEpy.NewBoolVar(str(cpm_var))
```
tias: use 'name' here and 2 lines lower

tias (changes requested):
OK, I did get to the end.

But some might go a bit further... not sure you want to keep this in one PR or split some off.

cpmpy/solvers/choco.py
IgnaceBleukx: It seems that in newer versions pychoco indeed raises an appropriate `Overflow` error. It used to just crash the java backend.

cpmpy/solvers/choco.py
IgnaceBleukx: yes, choco is quite strict on typing, and we don't ever want constant bools when they are not intended as ints (the should already be removed in simplify_bool)

cpmpy/solvers/cpo.py
IgnaceBleukx: Yes excellent point... I've now changed this, where _cpo_expr takes an extra argument "boolexpr" to indidate whether the result should be bool or not. Based on that, we can add the `== 1` or `== 0` part. solver_var now just returns the binary variable.

cpmpy/solvers/cpo.py
IgnaceBleukx: Yes, so its tricky with cpo... You create the variable, but the model does not necessarily know it yet, unless it is used in a constraint. So if we simplify a user variable away, just calling `def solver_var` on it is not enough.

cpmpy/solvers/exact.py
IgnaceBleukx: It seems to just be "log" by default, no heuristic on when to change the encoding.

cpmpy/solvers/gcs.py
IgnaceBleukx: Hmm, "only_positive_bv" expects linear expressions as input, we can fix this in a future PR, but out of scope for this one imo

cpmpy/solvers/pindakaas.py
IgnaceBleukx: Yes I agree this should change... but its out of scope for this PR

IgnaceBleukx:
Processed the review, should be ready for another round.

cpmpy/solvers/cpo.py
tias: I see, its documented so maybe its OK

cpmpy/solvers/pindakaas.py
tias: ok

tias (approved):
nice and consistent

except pindakaas/pysat returning a list of vars; can you open an issue that references here so we know this is to-fix?

PR #996, opened 2026-05-22
canonicalize inequalities

OrestisLomis:
to all have the form BV[x >= k], this will allow for more CSE. Also eliminate trivial bounds

### DIFF
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -1,7 +1,7 @@
 import copy
 from typing import Optional, Any, cast, overload, Literal
 from ..expressions.core import Expression, Comparison
-from ..expressions.utils import is_int
+from ..expressions.utils import is_int, is_bool
 from ..expressions.variables import NegBoolView, boolvar, intvar, _IntVarImpl, _BoolVarImpl
 
 
@@ -99,29 +99,33 @@ class CSEMap:
 
 
     def _canonicalize_boolexpr(self, expr: Expression) -> tuple[Expression, bool]:
-        """canonicalize a Boolean expression, results in more hits in the flat_map"""
+        """
+        Canonicalize any comparison between an expression `expr` and a constant `const`, results in more hits in the flat_map.
+        
+        """
 
         if isinstance(expr, Comparison):
             lhs, rhs = expr.args
-            if expr.name == "!=" and is_int(rhs):
-                # b <-> (expr != val) :: (~b) <-> (expr == val)
-                new_expr = Comparison("==", lhs, rhs)
-                new_expr._has_subexpr = expr._has_subexpr
-                return new_expr, True
-            elif isinstance(lhs, _IntVarImpl) and not lhs.is_bool() and expr.name == ">" and is_int(rhs):
-                # b <-> (expr > val) :: b <-> (expr >= val + 1)
-                new_expr = Comparison(">=", lhs, rhs + 1)
-                new_expr._has_subexpr = expr._has_subexpr
-                return new_expr, False
-            elif isinstance(lhs, _IntVarImpl) and not lhs.is_bool() and expr.name == "<=" and is_int(rhs):
-                # b <-> (expr <= val) :: (~b) <-> (expr >= val + 1)
-                new_expr = Comparison(">=", lhs, rhs + 1)
-                new_expr._has_subexpr = expr._has_subexpr
-                return new_expr, True
-            elif isinstance(lhs, _IntVarImpl) and not lhs.is_bool() and expr.name == "<" and is_int(rhs):
-                # b <-> (expr < val) :: (~b) <-> (expr >= val)
-                new_expr = Comparison(">=", lhs, rhs)
-                new_expr._has_subexpr = expr._has_subexpr
-                return new_expr, True
+            if is_int(rhs):
+                if expr.name == "!=":
+                    # b <-> (expr != val) :: (~b) <-> (expr == val)
+                    new_expr = Comparison("==", lhs, rhs)
+                    new_expr._has_subexpr = expr._has_subexpr
+                    return new_expr, True
+                elif expr.name == ">":
+                    # b <-> (expr > val) :: b <-> (expr >= val + 1)
+                    new_expr = Comparison(">=", lhs, rhs + 1)
+                    new_expr._has_subexpr = expr._has_subexpr
+                    return new_expr, False
+                elif expr.name == "<=":
+                    # b <-> (expr <= val) :: (~b) <-> (expr >= val + 1)
+                    new_expr = Comparison(">=", lhs, rhs + 1)
+                    new_expr._has_subexpr = expr._has_subexpr
+                    return new_expr, True
+                elif expr.name == "<":
+                    # b <-> (expr < val) :: (~b) <-> (expr >= val)
+                    new_expr = Comparison(">=", lhs, rhs)
+                    new_expr._has_subexpr = expr._has_subexpr
+                    return new_expr, True
 
         return expr, False
--- a/cpmpy/transformations/cse.py
+++ b/cpmpy/transformations/cse.py
@@ -100,8 +100,17 @@ class CSEMap:
 
     def _canonicalize_boolexpr(self, expr: Expression) -> tuple[Expression, bool]:
         """
-        Canonicalize any comparison between an expression `expr` and a constant `const`, results in more hits in the flat_map.
-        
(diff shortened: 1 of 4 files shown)


tias:
you seem to be doing two things in this PR... I expect this one to only be about canonalizing the inequalities? its something we can just put in master; it does not have to be xcsp3-only

OrestisLomis:
Yeah I should've mentioned this PR depends on PR #994 so I tacked the changes for this one onto that branch. Unless you're talking about the trivial bounds elimination? I can isolate that if needed.

OrestisLomis:
Aah wait actually, I'm wrong.. I can indeed isolate this one as well. Just have to change the testcase a bit

cpmpy/transformations/cse.py
IgnaceBleukx: Please update this doc with smt like:

```
Currently implemented:
- Convert any comparison between expr and constant to `expr = v` or `expr >= v`
```

cpmpy/transformations/cse.py
```diff
@@ -108,5 +108,20 @@ def _canonicalize_boolexpr(self, expr: Expression) -> tuple[Expression, bool]:
                 new_expr = Comparison("==", lhs, rhs)
                 new_expr._has_subexpr = expr._has_subexpr
                 return new_expr, True
-            
+            elif isinstance(lhs, _IntVarImpl) and not lhs.is_bool() and expr.name == ">" and is_int(rhs):
```
IgnaceBleukx: Why do you have the extra checks here? Do you even need them? I think lhs can also just be an expression, no?
Also not sure why the `is_bool` check needs to be here? b <-> expr > val :: b <-> expr >= val + 1 is still valid if expr is boolean? Note that in theory this case will never happen, as all comparions between Booleans and constants should be eliminated by `cpmpy.transformations.normalize.simplify_boolean`

tests/test_flatten.py
```diff
@@ -120,15 +120,15 @@ def test_get_or_make_var__bool(self):
         assert str(v) == "BV8"
         assert {str(c) for c in cons} == {
             "((BV7) == (BV0)) == (BV8)",
-            "(IV0 > 10) == (BV7)",
+            "(IV0 >= 11) == (BV7)",
```
IgnaceBleukx: I would prefer to just update the input expression instead of the output. We're not testing the normalization here (or at least thats not the point imo)

IgnaceBleukx: Same with the other tests in `test_flatten.py`

cpmpy/transformations/cse.py
IgnaceBleukx: Also, these blocks can be re-arranged to save some isinstance checks:

```python
if is_int(rhs):
   if expr.name == "!=" :  ...
   if expr.name == ">" : ...
   if expr.name == "<=": ...
   if expr.name == ">=": ...

```

IgnaceBleukx (changes requested):
Great, this should indeed result in more hits in the csemap (at least in theory), let's see how it plays out in practice.

For the tests: I would prefer a set of dedicated tests in `test_cse.py` that explicitely test the normalization functionality. Now they are hidden in `test_flatten.py`.

cpmpy/transformations/cse.py
OrestisLomis: Good points

OrestisLomis:
Alright I went through the comments, feel free to have another look

cpmpy/transformations/cse.py
```diff
@@ -99,14 +99,33 @@ def get_or_make_var(self, expr: Expression) -> tuple[_IntVarImpl, Optional[Expre
...
+                    new_expr = Comparison("==", lhs, rhs)
+                    new_expr._has_subexpr = expr._has_subexpr
+                    return new_expr, True
+                elif expr.name == ">":
+                    # b <-> (expr > val) :: b <-> (expr >= val + 1)
+                    new_expr = Comparison(">=", lhs, rhs + 1)
+                    new_expr._has_subexpr = expr._has_subexpr
+                    return new_expr, False
+                elif expr.name == "<=":
+                    # b <-> (expr <= val) :: (~b) <-> (expr >= val + 1)
```
IgnaceBleukx: add an inbetween step here to make it more clear?
```
# b <-> (expr <= val) :: (~b) <-> (expr > val) :: (~b) <-> (expr >= val + 1)
```

IgnaceBleukx: In that sense, it would also be better to move this case to the last one, as it is the most complex

tests/test_cse.py
```diff
@@ -43,22 +43,49 @@ def test_flatten(self):
...
         flat_cons = flatten_constraint(nested_cons2, csemap=csemap)
 
         assert len(flat_cons) == 1
         assert str(flat_cons[0]) == "BV0 xor a"
 
+    def test_canonicalize_inequalities(self):
+        x,y,z = cp.intvar(0,10, shape=3, name=tuple("xyz"))
+        csemap = CSEMap()
+
+        flat_cons = flatten_constraint((x > 5) | (y + z > 7), csemap=csemap)
```
IgnaceBleukx: you should check all of them.... not just `>`

tests/test_trans_linearize.py
```diff
@@ -642,28 +642,21 @@ def test_linearize_reified_variables_over_multiple_constraints(self):
...
-        assert str(out) == "[(BV[a >= 1]) or (BV[a >= 2]), sum([1, -1] * (BV[a >= 2], BV[a >= 1])) <= 0, sum([1, -1] * (BV[a >= 3], BV[a >= 2])) <= 0]"
+        self.csemap = CSEMap()
+        out = linearize_reified_variables(self.linearize((a < 2) | (a <= 2) | (a < 3)), min_values=2, csemap=self.csemap, ivarmap=self.ivarmap)
+        assert str(out) == "[or(~BV2, ~BV3, ~BV3), (a >= 2) == (BV2), (a >= 3) == (BV3)]"
 
+    # The following tests are marked with `xfail` because they are expected to fail, because they are not yet implemented; to see the current output compared with the desired output in the test, run with `pytest --runxfail`
     @pytest.mark.xfail(reason="aspirational")
-    def test_linearize_reified_inequalities_variations(self):
-        """Use order encoding on inequalities and replace other types of inequality expressions"""
+    def test_linearize_reified_disequalities(self):
```
IgnaceBleukx, line 645: I don't understand this: it is an aspirational test so why did you change it?
Either it is no longer aspirational, and the xfail should be removed, or it is still aspirational and then it should not be touched?

tests/test_flatten.py
```diff
@@ -302,7 +302,7 @@ def test_constraint(self):
 
         assert  str(flatten_constraint( a + b == c )) == "[((IV0) + (IV1)) == (IV2)]"
         #self.assertEqual( str(flatten_constraint( c != a + b )), "[((IV0) + (IV1)) != (IV2)]" ) # TODO, make it do the swap (again)
-        assert  str(flatten_constraint( ((a > 5) == (b < 3)) )) == "[(IV0 > 5) == (BV8), (IV1 < 3) == (BV8)]"
+        assert  str(flatten_constraint( ((a > 5) == ~(b >= 3)) )) == "[(IV0 > 5) == (~BV8), (IV1 >= 3) == (BV8)]"
```
IgnaceBleukx: Why change the input here this way? Just `b >= 3` should be fine, no?

IgnaceBleukx (changes requested):
Updated a bit of the code myself, and reverted some of the flatten tests to be more like on master (just changed `>` to `>=` instead of both `>=` and the rhs).

Last open questions are all test-related.

tests/test_flatten.py
OrestisLomis: Yeah that's true. I meant to keep the original meaning of the test, but I guess that's not really necessary.

tests/test_trans_linearize.py
OrestisLomis: Euhm I think it looks weird in the diff because I took the second aspirational test and moved it up to the non-aspirational tests. I did not touch the other aspirational tests in this PR.

OrestisLomis: I put the test back down in between the aspirational tests, because I might cause merge conflicts like this later..

IgnaceBleukx (approved):
The docstring felt way too AI generated... It was overly verbose and confusing.

I've updated it to somehting that reflects the actual purpose of the function.
Also refactored how the tests work (not altered any of the tests themselves).

If the tests pass, feel free to merge into master!

PR #998, opened 2026-05-27
use pyscipopt cumulative if it exists

tias:
to use with the pyscipopt pr https://github.com/scipopt/PySCIPOpt/pull/1223/changes

can we (easily) test it on an cumulative-rich benchmark, with solution checking (requires a local install of the above PR)?


IgnaceBleukx:
Cumulative is in main pyscipopt now, so should be fine to merge after tests succeed

PR #1000, opened 2026-05-28
Selective channeling for Exact

OrestisLomis:
When an integer variable is encoded (for now only directly, but with #994 this can also be orderly) a solver without an ivarmap (so any non-SAT solver) will post both defining ```sum(bvs) == 1``` as well as channeling ```wsum(ws, bvs) == iv``` to the solver. However, the channeling constraints are not necessary when the ```iv``` is not actually used in any constraint or the objective. This PR adds an ivarmap to Exact and allows for selective posting of channeling constraints, i.e. only when actually necessary.

### DIFF
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -148,6 +148,7 @@ class CPM_exact(SolverInterface):
         self.encoding = None
         self.ivarmap = dict()
         self._channeled_ivars = set()
+        self.reified_channeling_threshold = 1000
 
         # for solving with assumption variables,
         self.assumption_dict = None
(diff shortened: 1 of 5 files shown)


tias:
why is it so special?

pindakaas (pb interface) uses ivarmap and all is fine, if we didn't do that in exact yet, can't you just... do it exactly the same way?

OrestisLomis:
I agree the helper functions were kind of unnecessary and I removed them, but otherwise there is not much special going on.

IgnaceBleukx:
So as discussed offline, we'll first do some experimentation before this is merged into master; I've removed myself from the "awaiting review" list, but please feel free to renominate me when ready for review/discussion : )

PR #1001, opened 2026-05-29
set stale values to None after infeasibility (Exact and pindakaas)

OrestisLomis:
Exact and pindakaas both had a bug where values where not always set to None after an infeasible solve. Both bugs are slightly different, but makes sense to group them together in this PR.

Has to be noted that the bug related to Exact might be upstream bug, at least it didn't seem like expected behaviour. @JoD ?

### DIFF
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -162,7 +162,7 @@ class CPM_exact(SolverInterface):
         return self.xct_solver
 
     def _fillVars(self):
-        if not self.xct_solver.hasSolution():
+        if not self._solve_return(self.cpm_status):
             self.objective_value_ = None
             for cpm_var in self.user_vars:
                 cpm_var._value = None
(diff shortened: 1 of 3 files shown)


JoD:
Could Exact's behavior be related to my comment on this? https://gitlab.com/nonfiction-software/exact/-/work_items/26

I've never really considered how hasSolution() behaves under repeated calls (e.g., with different assumptions and constraints). Happy to hear your thoughts!

OrestisLomis:
The bug we had in CPMpy is definitely related to the linked issue. It was indeed hasSolution() returning True even though I wouldn't really expect this after a constraint is added and the solver is called again. Now if this is truly unexpected behaviour or not is a design choice I guess, but personally I would say it is. Perhaps the last solution can be efficiently verified with the new assumptions/constraints? And then if it is not a verified solution you indeed make hasSolution() return False.

PR #1002, opened 2026-06-01
fix tests and test docs

tias:
using `requires_solver()` in tests ALSO requires listing a 'solver' parameter in the function definition, otherwise the test is always run


OrestisLomis (approved):
Made some small changes, because it seemed to me like a test was misplaced + I added in the native mus cleanup here. Good to merge for me :)

OrestisLomis:
aah wait forgot that not all solvers can do MUS with assumptions..

OrestisLomis:
Okay, so to be clear I changed some more things but all is still scoped to this PR. It now works fine, but I had to set Exact as the solver for the standard MUS tests because #961 does not work (and also isn't in master yet).

PR #1003, opened 2026-06-01
Fix special case for not in decompose

IgnaceBleukx:
When refactoring the `decompose_in_tree` transformation in #929 we missed the special case where a global constraint occurs in a "not" expression.

In this PR, we again push down negation in the decomposition.


IgnaceBleukx:
Superseded by #916

PR #1006, opened 2026-06-02
Positive globals

ThomSerg:
Collection of positive decompositions of globals.

Table
- [x] decomp
- [x] decomp_linear -> MDD
- [x] decomp_pos -> NonReifiedTable

ShortTable
- [x] decomp
- [X] decomp_pos -> RowSelectingShortTable (issue with solution count)

NegTable
- [x] decomp

NegShortTable
- [x] decomp
- [ ] decomp_pos

Circuit
- [x] decomp
- [x] decomp_pos
- [x] decomp_lin -> has a branch

MDD
- [x] decomp
- [x] decomp_pos  

Regular
- [x] decomp -> MDD

InDomain
- [x] decomp_lin: `any(var == val)` instead of var != val


IgnaceBleukx:
Slightly misused this PR, and added a linear decomposition for InDomain.
I'm assuming this will improve quite a bit for linear solvers, as we avoid all of the != constraints like this.

tias:
To revise: which one to propose for master (rest, revert/remove from this PR)

cpmpy/expressions/globalconstraints.py
```diff
@@ -463,6 +463,34 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+        """
+        Linear decomposition of the Circuit global constraint, inspired by Miller-Tucker-Zemlin formulation for TSPs.
+        This linear decomposition is only valid in positive context.
+        """
+
+        succ = self.args
+        n = len(succ)
+        order = cp.intvar(0, n - 1, shape=n)
+
+        constraints = [x >= 0 for x in succ] + [x < n for x in succ]  # bounds on successors
```
tias: we now do `constraints.extend([list])`, also not += [item] but constraints.append(item), it saves needles list creations.

cpmpy/expressions/globalconstraints.py
```diff
@@ -855,39 +920,139 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
+
+        """
+
+        arr, transitions, start, accepting = self.args
+
+        if len(accepting) == 0:
+            return [cp.BoolVal(False)], [] # no accepting states, cannot be satisfied
+
+        # Collect all nodes and all transition values in the DFA
+        nodes = sorted({src for src, _, _ in transitions} |
```
tias: were walking the transitions list 3 times, can we do better?

cpmpy/tools/xcsp3/parser_callbacks.py
```diff
@@ -265,18 +265,18 @@ def strwildcard(x):
...
             if positive:
-                self.cpm_model += xglobals.NonReifiedTable(cpm_vars, tuples)
+                self.cpm_model += cp.Table(cpm_vars, tuples)
             else:
-                self.cpm_model += cp.NegativeTable(cpm_vars, tuples)
+                self.cpm_model += ~cp.Table(cpm_vars, tuples)
 
     def ctr_regular(self, scope: list[Variable], transitions: list, start_state: str, final_states: list[str]):
-        self.cpm_model += xglobals.Regular(self.get_cpm_vars(scope), transitions, start_state, final_states)
+        self.cpm_model += cp.Regular(self.get_cpm_vars(scope), transitions, start_state, final_states)
```
tias, line 279: does this mean xglobals.regular could be deleted too?

tias (commented):
great, minor feedback

cpmpy/expressions/globalconstraints.py
```diff
@@ -588,6 +617,24 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
         arr, tab = self.args
         return [cp.any([cp.all([ai == ri for ai, ri in zip(arr, row)]) for row in tab])], []
 
+    def decompose_positive(self) -> tuple[list[Expression], list[Expression]]:
+        """
+        Positive decomposition of the Table global constraint.
+
+        Returns:
+            tuple[list[Expression], list[Expression]]: A tuple containing the constraints representing the constraint value and the defining constraints
+        """
```
IgnaceBleukx, line 631: The `row -> cp.all(...)` should be a defining constraint? That helps if we are decomposing the table on `->` to keep the contents of the implication small.
You can also drop the special case for len(tab) == 1, with the following trick:
```python
arr, tab = self.args
row_selected = cp.boolvar(shape=(len(tab),)) # notice the `,` here to cover the len(tab) == 1 case
```

cpmpy/expressions/globalconstraints.py
```diff
@@ -722,6 +769,25 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
...
+        arr, tab = self.args
+
+        row_selected = boolvar(shape=(len(tab),))
+        cons = [cp.sum(row_selected) == 1]
+        for i, row in enumerate(tab):
+            subexpr = cp.all([ai == ri for ai, ri in zip(arr, row) if ri != STAR])
+            cons.append(row_selected[i].implies(subexpr))  # implication-only decomposition
+        return cons, []
```
IgnaceBleukx: same comments as for the regular table cons

cpmpy/expressions/globalconstraints.py
```diff
@@ -855,39 +921,144 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
-        transitions += [[sink, v, sink] for v in range(lb, ub + 1)]
+        if complete:
+            # add a sink node for transitions that are not defined. When the Regular constraint is in positive context, this is not needed
+            sink = len(self.nodes)
+            transitions.extend([[self.node_map[n], v, sink] for n in self.nodes for v in range(lb, ub + 1) if (n, v) not in self.trans_dict])
+            transitions.extend([[sink, v, sink] for v in range(lb, ub + 1)])
 
         # keep track of current state when traversing the array
-        state_vars = intvar(0, sink, shape=len(arr))
+        state_vars = intvar(0, len(self.nodes) if complete else len(self.nodes)-1, shape=len(arr))
```
IgnaceBleukx: Why not just add `sink` to self.nodes? Then you don't need this extra case?

IgnaceBleukx: I suppose you can also just take the domain 1..len(self.nodes)?

cpmpy/expressions/globalconstraints.py
```diff
@@ -855,39 +921,144 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
+            for v in range(len(values)):
+                flow_out[snk].append((v, snk))
+                flow_in[snk].append((v, snk))
+
+        defining = []
+        constraining = []
+
+        S = node_idx[start]
+        E = [node_idx[a] for a in accepting]
+        s = cp.boolvar(shape=(len(arr), len(values), len(nodes)))
```
IgnaceBleukx, line 1035: Some comment about what `s` is might be useful

cpmpy/expressions/globalconstraints.py
```diff
@@ -855,39 +921,144 @@ def __init__(self, array: ListLike[Expression], transitions: ListLike[tuple[int|
...
+                defining.append(cp.sum(s[0, j, q] for j in range(len(values))) == 0)
+
+        # Enforce flow constraints: flow in = flow out
+        for i in range(1, len(arr)):
+            for q in range(len(nodes)):
+                defining.append(cp.sum(s[i - 1, j, q_] for (j, q_) in flow_in[q]) == cp.sum(s[i, j, q] for (j, _) in flow_out[q]))
+                defining.append(cp.sum(s[i - 1, j, q_] for (j, q_) in flow_in[q]) <= 1) # redundant constraint
+                defining.append(cp.sum(s[i, j, q] for (j, _) in flow_out[q]) <= 1) # redundant constraint
+
+        # Accepting end nodes have one unit of flow in total
```
IgnaceBleukx, line 1052: Does this allow for DFAs with multiple ending nodes? I'm not entirely sure. Might be good to add a test for it in `test_globalconstraints` anyway

cpmpy/tools/xcsp3/parser_callbacks.py
```diff
@@ -265,18 +265,18 @@ def strwildcard(x):
...
 
     def ctr_mdd(self, scope: list[Variable], transitions: list):
         self.cpm_model += xglobals.MDD(self.get_cpm_vars(scope), transitions)
```
IgnaceBleukx: Use `cp.MDD` here? 
MDD can be removed from xglobals as well

cpmpy/transformations/decompose_global.py
```diff
@@ -41,7 +41,9 @@ def decompose_in_tree(lst_of_expr: list[Expression],
                       supported_reified: Optional[AbstractSet[str]] = None,
                       _toplevel=None, nested=False,
                       csemap: Optional[CSEMap] = None,
-                      decompose_custom: Optional[Dict[str, CustomDecomp]] = None) -> list[Expression]:
+                      decompose_custom: Optional[Dict[str, CustomDecomp]] = None,
```
IgnaceBleukx, line 46: Update docstring with new arguments

cpmpy/transformations/decompose_global.py
IgnaceBleukx, line 70: There should also be something about the order of custom decomposes that we use. E.g.,

```
To decompose a global constraint in positive context:
1. Check `decompose_positive_custom`
2. Check `decompose_custom`
3. Call `global.decompose_positive()`

The assumption is that `decompose_custom` generally works better compared to the standard global decomposition, even if the custom decomposition is not specialized for postive-only use.
```

or smt

IgnaceBleukx: The checks from before with the custom decomposes also need to be repeated here

cpmpy/transformations/linearize.py
```diff
@@ -627,22 +628,38 @@ def decompose_linear_objective(obj: Expression,
...
     return dict(
         alldifferent=AllDifferent.decompose_linear,
         element=Element.decompose_linear,
-        table=Table.decompose_linear
+        table=Table.decompose_linear,
+        short_table=ShortTable.decompose,
+        InDomain=InDomain.decompose_linear,
+        regular=Regular.decompose_linear,
     )
     # Should we add Gleb's table decomposition? or is it not non-reifiable?
```
IgnaceBleukx: Remove comment

tests/test_globalconstraints.py
```diff
@@ -545,7 +545,7 @@ def test_shorttable(self):
         # unconstrained
         true_cons = cp.ShortTable(iv, [[1,2,3],[STAR, STAR, STAR]])
         assert cp.Model(true_cons).solve(solver=solver)
-        assert cp.Model(true_cons).solveAll(solver=solver) == 17 ** 3
+        #assert cp.Model(true_cons).solveAll(solver=solver) == 17 ** 3
```
IgnaceBleukx, line 548: Hmm why is this not valid anymore?

tests/test_transf_decompose.py
```diff
@@ -233,9 +233,9 @@ def decompose_positive(self):
         assert set(map(str, decompose_in_tree([cons.implies(bv)]))) == {"(sum(a, b, c) == 1) -> (bv)","a == 1"}  # decompose standard
         assert set(map(str, decompose_in_tree([bv == (cons)]))) == {"(bv) == (sum(a, b, c) == 1)", "a == 1"} # decompose standard
 
-        # custom decompose has precedence over positive decompose
-        decompose_custom = {"mycustomglobal": lambda x : ([cp.sum(x.args) == 5], [])}
-        assert set(map(str, decompose_in_tree([cons], decompose_custom=decompose_custom))) == {"sum(a, b, c) == 5"} # custom decompose
+        # custom positive decompose has precedence over positive decompose
```
IgnaceBleukx: Also test the other cases of the custom and positive stuff here.

I.e., test whether we indeed first check decompose_custom_positive, then decompose_custom and then go to default

IgnaceBleukx: Actually, might be better to add a entire different test case for that?

IgnaceBleukx (changes requested):
Great, some simplifications are possible in the decompositions and the xglobals file : )

cpmpy/expressions/globalconstraints.py
WoutPiessens: Might make more sense to start with 0, but indeed I'll add the sink node in the complete case

tests/test_globalconstraints.py
WoutPiessens: Was still a leftover from the flow-based decomposition of ShortTable

tias (approved):
looks ready, great to see the full global treatment land!

cpmpy/transformations/decompose_global.py
```diff
@@ -52,7 +54,17 @@ def decompose_in_tree(lst_of_expr: list[Expression],
     :param supported_reified: a set of names of supported reified global constraints (those with Boolean return type only).
     :param _toplevel: DEPRECATED
     :param nested: DEPRECATED
-    :param csemap: a dictionary of 'expr: expr' mappings, for Common Subexpression Elimination
+    :param csemap: a dictionary of 'expr: expr' mappings, for Common Subexpression Elimination.
```
IgnaceBleukx: the csemap is a CSEMap object? Not a dict

smt like `:param csemap: CSEMap object used to avoid decomposing the same global constraint twice`

tests/test_globalconstraints.py
IgnaceBleukx: Ok so its disabled after all to make the CI happy; I'm ok with that as indeed locally it succeeds... Can you open an issue so we keep track of this? Not a blocking issue to merge this PR

tests/test_globalconstraints.py
```diff
@@ -639,6 +639,31 @@ def test_regular(self):
         assert num_true + num_false == 2**7
         assert len(true_sols & false_sols) == 0# no solutions can be in both
 
+    def test_regular_multiple_accepting_nodes(self):
```
IgnaceBleukx: I think my original comment was about the linear decomposition? That is not tested here I believe?
Ideally add a `solver` argument to the test, and use the solver arg in solveAll as well, then it will run for all solvers on the CI

IgnaceBleukx (approved):
Small comments, good to merge after they are processed : )

PR #1007, opened 2026-06-02
Improved GCC decomposition

IgnaceBleukx:
Adds a redundant constraint on the sum of counts for GCC.

Should allow for faster detection of failure in a branch during solving


tias (approved):
I don't mind.

Is there a paper documenting this, or from where does the idea come?

(maybe add a short comment that these are redundant? and if you edit it, there is a double space before 'for')

IgnaceBleukx:
Its propably described in some paper, but I got the idea while working on the half-reified paper : )
MiniZinc has the same formulation: https://github.com/MiniZinc/libminizinc/blob/master/share/minizinc/std/fzn_global_cardinality.mzn

PR #1009, opened 2026-06-03
Operator constructor cleaning

tias:
No more flatlist, and a bit cleaner paths.

Minor efficiency improvements.

Typing is still near-impossible because wsum requires 2 ListLikes and the other take 1 ListLike


tias:
Unmergable; `Operator` is probably the last constructor we will type, so redo later...

PR #1010, opened 2026-06-03
Save lhs and rhs of reification to csemap

IgnaceBleukx:
Improvement to cse in flattening:

If we encounter a constraint `lexpr == rexpr`, flatten will make a new variable `BV0` for `rexpr` and add a constraint: `rexpr == BV0`, and save `rexpr : BV0` to the csemap.

In this PR, I extend this to also add `lexpr : BV0` to the csemap. 
This allows to re-use the already introduced variable `BV0` again if we encounter `lexpr` as a subexpression.
Also increases the likelihood of an intvar being encoded with a direct encoding using `linearize_reified_varvals`, as the csemap can now contain more cases of `var == val : BV`. (e.g., if the user wrote `(x == 3) == b` -- see tests)
 
Also includes a small change to builtins.sum, returning `expr` when calling `cp.sum([expr])`; this improves the linearization for MDD, as we avoid the cases where we call `sum(edge_vars) == (var == val)` where `edge_vars` has a length of 1.

### DIFF
--- a/cpmpy/transformations/flatten_model.py
+++ b/cpmpy/transformations/flatten_model.py
@@ -259,15 +259,17 @@ def flatten_constraint(expr, csemap=None):
             # Reification (double implication): Boolexpr == Var
             # normalize the lhs (does not have to be a var, hence we call normalize instead of get_or_make_var
             if exprname == '==' and lexpr.is_bool():
+                lhs = None
                 if rvar.is_bool():
                     if csemap is not None:
-                        lexpr_norm, negate = csemap._canonicalize_boolexpr(lexpr) # canonicalize the lexpr
-                        if negate:
-                            csemap.flat_map[lexpr_norm] = ~rvar # save this reificsation in the csemap
+                        other_expr = csemap.get(lexpr)
+                        if other_expr is not None:
+                            lhs = other_expr # already have a constraint lexpr == bvar, use bvar instead
+                            lcons = []
                         else:
-                            csemap.flat_map[lexpr_norm] = rvar # save this reificsation in the csemap
-                    # this is a reification
-                    (lhs, lcons) = normalized_boolexpr(lexpr, csemap=csemap)
+                            csemap.put(lexpr, rvar)
+                    if lhs is None:
+                        (lhs, lcons) = normalized_boolexpr(lexpr, csemap=csemap)
                 else:
                     # integer comparison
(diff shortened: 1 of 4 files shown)


cpmpy/transformations/flatten_model.py
```diff
@@ -260,6 +260,12 @@ def flatten_constraint(expr, csemap=None):
             # normalize the lhs (does not have to be a var, hence we call normalize instead of get_or_make_var
             if exprname == '==' and lexpr.is_bool():
                 if rvar.is_bool():
+                    if csemap is not None:
+                        lexpr_norm, negate = csemap._canonicalize_boolexpr(lexpr) # canonicalize the lexpr
+                        if negate:
+                            csemap.flat_map[lexpr_norm] = ~rvar # save this reificsation in the csemap
```
tias: typo in comment

cpmpy/transformations/flatten_model.py
```diff
@@ -260,6 +260,12 @@ def flatten_constraint(expr, csemap=None):
             # normalize the lhs (does not have to be a var, hence we call normalize instead of get_or_make_var
             if exprname == '==' and lexpr.is_bool():
                 if rvar.is_bool():
+                    if csemap is not None:
+                        lexpr_norm, negate = csemap._canonicalize_boolexpr(lexpr) # canonicalize the lexpr
```
tias: is this the best way we can do this? we dont have some csemap.store(lexr, rvar)? because you have to remember to call canonicalize idd, otherwise it wont match the csemap.. but so its a bit unusual to be calling a _canon semi-privatley named function

but if thats the right way then fine

tests/test_cse.py
```diff
@@ -135,12 +135,54 @@ def test_flat_reification_is_not_cse_canonicalized(self):
         b = cp.boolvar(name="b")
 
         flat_cons = flatten_constraint((x > 5) == b, csemap=self.csemap)
+        print(flat_cons)
```
tias: stray print

tias (commented):
ah yes, nice.

Some minor comments

cpmpy/transformations/flatten_model.py
IgnaceBleukx: I've added a `.put` to csemap to fix this.

PR #1012, opened 2026-06-09
Bounds for pow and mod

### LINKED ISSUE #1011: Minimizing squared integer expression returns non-optimal solution (0.10.0)
The following minimal model seems to return a non-optimal solution with CPMpy 0.10.0 and the default solver.

```python
from cpmpy import *
import importlib.metadata as md

x = intvar(-10, 10, name="x")
model = Model(minimize=x**2)

print(md.version("cpmpy"), md.version("ortools"))
print(model.solve())
print(x.value(), (x**2).value(), model.objective_value())
```

Observed output:

```text
0.10.0 9.14.6206
True
-10 100 100
```

Expected optimum is `x = 0` with objective `0`.

A constrained variant also shows the same issue:

```python
from cpmpy import *
import importlib.metadata as md

x = intvar(-10, 10, name="x")
model = Model(x != 0, minimize=x**2)

print(md.version("cpmpy"), md.version("ortools"))
print(model.solve())
print(x.value(), (x**2).value(), model.objective_value())
```

Observed output:

```text
0.10.0 9.14.6206
True
-10 100 100
```

Expected objective is `1`, with `x = -1` or `x = 1`.

**For reference, I reproduced the constrained variant with CPMpy 0.9.28 + OR-Tools 9.14.6206, where it correctly returns the expected objective `1`.**

IgnaceBleukx:
Fix bound implementation for `pow`, also includes more precise bound computation for Modulo when the domain is small.
Fixes #1011

### DIFF
--- a/cpmpy/expressions/globalfunctions.py
+++ b/cpmpy/expressions/globalfunctions.py
@@ -637,14 +637,22 @@ class Modulo(GlobalFunction):
         if lb2 == ub2 == 0:
             raise ZeroDivisionError("Domain of {} only contains 0".format(self.args[1]))
 
-        # the (abs of) the maximum value of the remainder is always one smaller than the absolute value of the divisor
-        lb = lb2 + (lb2 <= 0) - (lb2 >= 0)
-        ub = ub2 + (ub2 <= 0) - (ub2 >= 0)
-        if lb1 >= 0:  # result will be positive if first argument is positive
-            return 0, max(-lb, ub, 0)  # lb = 0
-        elif ub1 <= 0:  # result will be negative if first argument is negative
-            return min(-ub, lb, 0), 0  # ub = 0
-        return min(-ub, lb, 0), max(-lb, ub, 0)  # 0 should always be in the domain
+        # For small domains, compute exact bounds by enumeration.
+        if (ub1 - lb1 + 1) * (ub2 - lb2 + 1) <= 10 ** 6:
+            xs = np.arange(lb1, ub1 + 1)
+            ys = np.arange(lb2, ub2 + 1)
+            ys = ys[ys != 0]  # divisor is never 0 (partial function)
+            
+            rem = np.fmod(xs[:, None], ys[None, :]) # np.fmod implements truncated-division remainder
+            return int(rem.min()), int(rem.max())
+
+        # Otherwise fall back to an analytical over-approximation.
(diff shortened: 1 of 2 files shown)


cpmpy/expressions/globalfunctions.py
```diff
@@ -703,9 +711,13 @@ def get_bounds(self):
             tuple[int, int]: A tuple of (lower bound, upper bound) for the power
         """
         base, exp = self.args
+        # exp is guaranteed to be a constant greater than 0
         lb_base, ub_base = get_bounds(base)
-
+    
         bounds = [lb_base ** exp, ub_base ** exp]
+        if lb_base < 0 < ub_base:
```
OrestisLomis, line 721: What if the exponent is actually 0? I feel like that such a case should not reach this part of the code (i.e. it should already be simplified before that), but you could also replace the if by `if lb_base < 0 < ub_base and exp != 0:`. I saw that `__pow__` already simplifies expressions with exponent 1 and we could easily add a case for exponent 0 here. But then a user might still use `cp.Power(x,0)`, which does not resolve the edgecase.

tests/test_globalconstraints.py
```diff
@@ -1480,14 +1480,26 @@ def test_bounds_pow(self):
...
+        assert lb == 0
+        assert ub == 8 ** 4
+        
+        x = cp.intvar(-5, 8)
+        op = cp.Power(x,3)
+        lb, ub = op.get_bounds()
+        assert lb == -5 ** 3
+        assert ub == 8 ** 3
+
+        op = cp.Power(x, 4)
```
OrestisLomis, line 1497: repeated?

OrestisLomis (changes requested):
I had also looked at it before I saw you already made a PR. I think there is one edgecase worth looking into when the exponent is 0.

tests/test_globalconstraints.py
IgnaceBleukx: x has changed bounds above (line 1491)

cpmpy/expressions/globalfunctions.py
IgnaceBleukx: Yes you're right, we should also handle that case, although I would prefer to just return (1,1) directly

OrestisLomis (approved):
Alright nice and simple fix

Issue #827, opened 2026-01-16
to_cnf 0 + 1 == 1


ThomSerg:
Most likely related to #778 

Reproducable model:

```python
import cpmpy as cp
import numpy as np

def model():
    p = cp.boolvar(name='p')
    a = cp.boolvar(name='a')
    b = cp.boolvar(name='b')
    c = cp.boolvar(name='c')

    constraints = [
        cp.Xor([~((p).implies((cp.Xor([a, b, c]) <= cp.BoolVal(True)))), (p).implies((cp.Xor([a, b, c]) <= cp.BoolVal(True)))]),
    ]
    return cp.Model(constraints)


if __name__ == '__main__':
    m = model()
    from cpmpy.transformations.to_cnf import to_cnf
    to_cnf(m.constraints)
```

Output:
```console
Traceback (most recent call last):
  File "/home/thomserg/Documents/programming/fuzz-tester/bug.py", line 24, in <module>
    to_cnf(m.constraints)
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/transformations/to_cnf.py", line 44, in to_cnf
    slv += constraints
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/solvers/pindakaas.py", line 268, in add
    for cpm_expr in self.transform(cpm_expr_orig):
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/solvers/pindakaas.py", line 251, in transform
    cpm_cons = linearize_constraint(
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/transformations/linearize.py", line 211, in linearize_constraint
    [cpm_expr] = canonical_comparison([cpm_expr])  # just transforms the constraint, not introducing new ones
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomserg/Documents/programming/fuzz-tester/libraries/cpmpy/cpmpy/transformations/linearize.py", line 515, in canonical_comparison
    assert not is_num(lhs), f"lhs cannot be an integer at this point! expr: {cpm_expr}"
           ^^^^^^^^^^^^^^^
[...]

PR #1014, opened 2026-06-11
Fix edge case in canonical_comparison

### LINKED ISSUE #620: Boolean constants not handled in Global constraints

IgnaceBleukx:
In some weird cases, the left hand side of a sum can contain only constants after previous transformations, which was not handled properly.

Fixes #620 and #827

### DIFF
--- a/cpmpy/solvers/pumpkin.py
+++ b/cpmpy/solvers/pumpkin.py
@@ -512,6 +512,8 @@ class CPM_pumpkin(SolverInterface):
         elif isinstance(cpm_expr, Operator):
             if cpm_expr.name == "or": 
                 return [constraints.Clause(self.solver_vars(cpm_expr.args), constraint_tag=tag)]
+            if cpm_expr.name == "and":
+                return [clause for arg in cpm_expr.args for clause in self._get_constraint(arg, tag=tag)]
 
             raise NotImplementedError("Pumpkin: operator not (yet) supported", cpm_expr)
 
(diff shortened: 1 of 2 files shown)


IgnaceBleukx:
Also small change to canonical comparison: it now expects a list of Expressions and returns one too. This saves us a shotgun call to toplevel_list. Needed to make some small tweaks to the callers, and pumpkin

tias (approved):
great, also to have this function now typed

PR #1015, opened 2026-06-12
Add CSPLib examples from DCP-Bench-Open

kostis-init:


examples/csplib/prob003_quasigroup_existence.py
```diff
@@ -0,0 +1,65 @@
...
+for them. We define two new relations, \\*321 and \\*312 by $a \\*321 b = c$ iff $c\\*b=a$ and
+$a \\*312 b = c$ iff $b\\*c=a$.
+
+QG3.m problems are order m quasigroups for which $(a\\*b)\\*(b\\*a) = a$.
+We only consider the QG3.m problem for this task.
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_003_quasigroup_existence/csplib_003_quasigroup_existence.cpmpy.py)
+"""
+
+from cpmpy import *
```
IgnaceBleukx: We should use import cpmpy as cp

examples/csplib/prob016_traffic_lights.py
```diff
@@ -0,0 +1,62 @@
...
+Pedestrian light states: 0=red, 1=green
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_016_traffic_lights/csplib_016_traffic_lights.cpmpy.py)
+"""
+
+from cpmpy import *
+
+
+def traffic_lights():
+    V1 = intvar(0, 3, name="V1")
```
IgnaceBleukx: You can make this vectorized, not requird tho : )

```python
vehicle_lights = cp.intvar(0,3,shape=4, name=tuple("V{i} for i (1,2,3,4))

```

examples/csplib/prob018_water_bucket.py
```diff
@@ -0,0 +1,110 @@
...
+
+        model += is_goal_t.implies((sequence[t + 1, :] == padding_value).all())
+        model += is_padded_t.implies((sequence[t + 1, :] == padding_value).all())
+
+        must_transfer = Table(np.hstack([sequence[t, :], sequence[t + 1, :]]), transitions) & \
+                        any(sequence[t + 1, :] != sequence[t, :])
+        model += (~is_goal_t & ~is_padded_t).implies(must_transfer)
+
+    # Goal must be reached
+    model += any([(sequence[t, :] == goal_state).all() for t in range(max_steps)])
```
IgnaceBleukx: These ones should really become "cp.any" instead

examples/csplib/prob022_bus_driver_scheduling.py
```diff
@@ -0,0 +1,65 @@
...
+            [0, 4, 8],
+            [0, 5, 10],
+            [1, 6, 11],
+            [2, 7, 9],
+            [3, 6, 8]
+        ]
+
+    model = Model()
+
+    x = boolvar(shape=num_shifts, name="x")
```
IgnaceBleukx: Ah so the goal is just to select the shifts, not necessarily asign them to worker too?

examples/csplib/prob022_bus_driver_scheduling.py
```diff
@@ -0,0 +1,65 @@
...
+
+The goal is to cover all tasks while minimizing the total number of shifts used.
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_022_bus_driver_scheduling/csplib_022_bus_driver_scheduling.cpmpy.py)
+"""
+
+from cpmpy import *
+
+
+def bus_driver_scheduling(num_work=12, num_shifts=14, shifts=None):
```
IgnaceBleukx: A docstring would be useful here, also num_work should be renamed to "num_drivers" I think?

examples/csplib/prob023_magic_hexagon.py
```diff
@@ -0,0 +1,76 @@
...
+def magic_hexagon(num_cells=19, magic_sum=38):
+    LD = intvar(1, num_cells, shape=num_cells, name="LD")
+    a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s = LD
+
+    model = Model()
+
+    # All numbers from 1 to 19 used exactly once
+    model += AllDifferent(LD)
+
+    # Rows (horizontal)
```
IgnaceBleukx: This hard code stuff does not work anymore when num_cells is not exactly 19, right?
Maybe it might be more useful to have the size of the edge of the hexagon as input, and derive everything from that? I.e., for the sample input the edge length would be 3.
And then you can just figure out which variables are on which row (keeping a flat list is probably the most convenient as you do now : ))

Smt like:
```
i = 0
length = n
while i < len(vars):
    model += sum(vars[i:i+length]) == magic_sum
    i += length
    ... some logic to deterine if we're going up or down with the length (always +1 / -1)

```

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,175 @@
...
+not even diagonally. The digits along the right side of and below the grid
+indicate the number of grid squares in the corresponding rows and columns that
+are occupied by vessels.
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_014_solitaire_battleships/csplib_014_solitaire_battleships.cpmpy.py)
+"""
+
+from cpmpy import *
+
+WATER = 0
```
IgnaceBleukx: Maybe add a small comment here on what these constants mean?

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,175 @@
...
+RIGHT = 4
+TOP = 5
+BOTTOM = 6
+MIDDLE = 7
+
+
+def solitaire_battleships(rows=10, cols=10, rowsum=None, colsum=None,
+                          fleet_counts=None, hints=None):
+    if rowsum is None:
+        rowsum = [0, 2, 3, 1, 2, 4, 2, 1, 2, 3]
```
IgnaceBleukx: assert that rows=10 and cols=10. Otherwise this is not valid

IgnaceBleukx: You can just raise a ValueError if the user did not provide rowsum, but did provide rows which are different from 10

examples/csplib/prob034_warehouse_location.py
```diff
@@ -0,0 +1,76 @@
...
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_034_warehouse_location/csplib_034_warehouse_location.cpmpy.py)
+"""
+
+from cpmpy import *
+
+
+def warehouse_location(n_suppliers=5, n_stores=10, building_cost=30,
+                       capacity=None, cost_matrix=None):
+    if capacity is None:
```
IgnaceBleukx: Also here this does not work if n_supppliers is not 5

IgnaceBleukx (changes requested):
Great! Thanks for the extra examples.

Most of them look very clean, but we should make sure they follow our own guidelines.
I.e., use `import cpmpy as cp` and `cp.sum` etc instead of `from cpmpy import *`.

Other comments are mainly related to default arguments, and what happens if we change them.

examples/csplib/prob021_crossfigures.py
```diff
@@ -0,0 +1,278 @@
...
+    Row -= 1
+    Col -= 1
+    tmp = intvar(0, 9999, shape=Len)
+    constraints = [to_num(tmp, Down, 10)]
+    for i in range(Len):
+        constraints += [Matrix[Row + i, Col] == tmp[i]]
+    return constraints
+
+
+def crossfigures():
```
IgnaceBleukx, line 111: I think this one can be improved massively?
I'm missing a lot of intuition about what the helper functions mean? 
Also not sure if we actually need all of the `tmp` helper variables? We typically try to avoid that (also the SAT solvers don't like the huge bounds on them ; ) )

examples/csplib/prob022_bus_driver_scheduling.py
kostis-init: Yes, it seems like a minimal version, in which each worker is assigned to a shift trivially.

kostis-init:
Thanks for the feedback, let me know if you have any further comments!

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -15,19 +15,41 @@
 Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_014_solitaire_battleships/csplib_014_solitaire_battleships.cpmpy.py)
 """
 
-from cpmpy import *
-
+import cpmpy as cp
+
+# Cell types in the grid. Each cell is either water or part of a ship.
+# Ship cells indicate their position within the vessel:
+#   CIRCLE  - submarine (1-cell ship, occupies a single cell)
```
IgnaceBleukx: Put these next to the constants? Seems a bit werid now

examples/csplib/prob016_traffic_lights.py
```diff
@@ -15,31 +15,23 @@
...
     allowed_tuples = [[0, 0, 2, 1], [1, 0, 3, 0], [2, 1, 0, 0], [3, 0, 1, 0]]
 
-    model = Model()
+    model = cp.Model()
 
     for i in range(4):
-        model += Table([vehicle_lights[i], pedestrian_lights[i],
-                        vehicle_lights[(i + 1) % 4], pedestrian_lights[(i + 1) % 4]],
+        model += cp.Table(
+            [vehicle_lights[i], pedestrian_lights[i], vehicle_lights[(i + 1) % 4], pedestrian_lights[(i + 1) % 4]],
```
IgnaceBleukx: Hmm this is some weird formatting. Maybe just make new variable `lights_i` or smt to put in the table constraint later? not sure

examples/csplib/prob021_crossfigures.py
```diff
@@ -89,36 +87,33 @@ def is_prime(n):
...
-    for i in range(Len):
-        constraints += [Matrix[Row + i, Col] == tmp[i]]
-    return constraints
+    digits = [Matrix[Row + i, Col] for i in range(Len)]
+    return [to_num(digits, Down, 10)]
 
 
 def crossfigures():
     n = 9
     D = 9999
```
IgnaceBleukx: How is this upper bound chosen?

examples/csplib/prob003_quasigroup_existence.py
```diff
@@ -0,0 +1,65 @@
...
+import cpmpy as cp
+
+def quasigroup_existence(m=8):
+    quasigroup = cp.intvar(0, m - 1, shape=(m, m), name="quasigroup")
+
+    model = cp.Model()
+
+    # Each element occurs once in every row
+    for i in range(m):
+        model += cp.AllDifferent(quasigroup[i, :])
```
IgnaceBleukx: In the Sudoku examples we use `model += [cp.AllDifferent(row) for row in quasigroup]` and similarly `model += [cp.AllDifferent(col) for col in quasigroup.T]

examples/csplib/prob003_quasigroup_existence.py
```diff
@@ -0,0 +1,65 @@
...
+QG3.m problems are order m quasigroups for which $(a\\*b)\\*(b\\*a) = a$.
+We only consider the QG3.m problem for this task.
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_003_quasigroup_existence/csplib_003_quasigroup_existence.cpmpy.py)
+"""
+
+import cpmpy as cp
+
+def quasigroup_existence(m=8):
+    quasigroup = cp.intvar(0, m - 1, shape=(m, m), name="quasigroup")
```
IgnaceBleukx: These domains don't match the example you give above? I think it should be domains 1..m instead

IgnaceBleukx: Ah but you need it for the Element constraints below? Hmm I would put the offset of 1 there then.

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,197 @@
...
+LEFT = 2
+RIGHT = 3
+TOP = 4
+BOTTOM = 5
+MIDDLE = 6
+
+
+def solitaire_battleships(rows=10, cols=10, rowsum=None, colsum=None,
+                          fleet_counts=None, hints=None):
+    if rows != 10 or cols != 10:
```
IgnaceBleukx: Actually, this stuff can be simplified I'm now realizing.

Just make the default arguments some constant above (after the circle/left/right stuff), and put them as default arguments to the function.
You just need some asserts that `len(rowsum) == rows` etc, which you should have anyway?

IgnaceBleukx: Or even better: just derive `rows` and `cols` from the length of rowsum and colsum?

examples/csplib/prob016_traffic_lights.py
```diff
@@ -0,0 +1,54 @@
...
+    allowed_tuples = [[0, 0, 2, 1], [1, 0, 3, 0], [2, 1, 0, 0], [3, 0, 1, 0]]
+
+    model = cp.Model()
+
+    for i in range(4):
+        model += cp.Table(
+            [vehicle_lights[i], pedestrian_lights[i], vehicle_lights[(i + 1) % 4], pedestrian_lights[(i + 1) % 4]],
+                       allowed_tuples)
+
+    return model, (lights,)
```
IgnaceBleukx: Why not just return `model, (vehicle_lights, pedestrian_lights)`?
You never use the `lighs` variable as such?

examples/csplib/prob018_water_bucket.py
```diff
@@ -0,0 +1,104 @@
...
+Find the minimum number of transfers of water between buckets to reach the goal.
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_018_water_bucket/csplib_018_water_bucket.cpmpy.py)
+"""
+
+import itertools
+import cpmpy as cp
+import numpy as np
+
+def water_bucket(capacities=[8, 5, 3], initial_state=[8, 0, 0], goal_state=[4, 4, 0], max_steps=20):
```
IgnaceBleukx, line 21: I revamped this one myself, there were quite some optimizations possible : )

examples/csplib/prob021_crossfigures.py
```diff
@@ -0,0 +1,273 @@
...
+
+    model = cp.Model()
+
+    for i in range(n):
+        for j in range(n):
+            if Valid[i][j] == B:
+                model += (M[i, j] == 0)
+
+    # Across variables
+    A1 = cp.intvar(0, D, name="A1")
```
IgnaceBleukx: Hmm, using that huge upper bound for all variables is fine for the CP solvers but not really for others :/ 
Ideally, you just use the length of the block to fill? E.g., if its length 2, then ub=99?

examples/csplib/prob021_crossfigures.py
```diff
@@ -0,0 +1,273 @@
...
+             [Z, Z, Z, Z, B, Z, Z, Z, Z],
+             [Z, B, Z, Z, B, Z, Z, B, Z],
+             [Z, Z, B, Z, Z, Z, B, Z, Z],
+             [Z, Z, Z, Z, B, Z, Z, Z, Z]]
+
+    M = cp.intvar(0, 9, shape=(n, n), name="M")
+
+    model = cp.Model()
+
+    for i in range(n):
```
IgnaceBleukx: In sudoku, we vectorize this (assuming Valid is a np array):

```
model += M[Valid == B] == 0
```

examples/csplib/prob021_crossfigures.py
```diff
@@ -0,0 +1,273 @@
...
+
+def crossfigures():
+    n = 9
+    D = 9999
+
+    primes = [i for i in range(2, D + 1) if is_prime(i)]
+    squares = [k * k for k in range(1, math.isqrt(D) + 1)]
+
+    Z = -1
+    B = -2
```
IgnaceBleukx: Maybe re-use the signs you used in the docstring? I.e., `_` and `X`?

examples/csplib/prob021_crossfigures.py
```diff
@@ -0,0 +1,273 @@
...
+    Valid = [[Z, Z, Z, Z, B, Z, Z, Z, Z],
```
IgnaceBleukx: lowercase?

examples/csplib/prob023_magic_hexagon.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # Build lookup: coordinate -> variable
+    var_at = {c: v for c, v in zip(coords, LD)}
+
+    model = cp.Model()
+    model += cp.AllDifferent(LD)
+
+    # Horizontal rows: group by r coordinate
+    for r in range(-(n - 1), n):
+        row = [var_at[(q, r)] for q in range(-(n - 1), n)
```
IgnaceBleukx: Euhm so here you assume the center is (0,0), right? May be good to put that explicitey somwhere
Also the formatting here is a bit weird with the if on the newline

examples/csplib/prob034_warehouse_location.py
```diff
@@ -0,0 +1,90 @@
...
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_034_warehouse_location/csplib_034_warehouse_location.cpmpy.py)
+"""
+
+import cpmpy as cp
+
+
+def warehouse_location(n_suppliers=5, n_stores=10, building_cost=30,
+                       capacity=None, cost_matrix=None):
+    # Default data is for 5 suppliers and 10 stores.
+    if n_suppliers != 5 or n_stores != 10:
```
IgnaceBleukx: Yeah so here as well, just put everthing as default args, and do asserts on the dimensions of all arguments to make sure they match

examples/csplib/prob041_n_fractions.py
```diff
@@ -0,0 +1,48 @@
...
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_041_n_fractions/csplib_041_n_fractions.cpmpy.py)
+"""
+import cpmpy as cp
+
+def n_fractions(n=9):
+    x = cp.intvar(1, n, shape=9, name="x")
+    A, B, C, D, E, F, G, H, I = x
+
+    D1 = cp.intvar(1, n * n, name="D1")
```
IgnaceBleukx: You don't actually need these variables?
Just do `D1 = 10 * B + C` instead of adding it as a constraint?

examples/csplib/prob057_killer_sudoku.py
```diff
@@ -0,0 +1,79 @@
...
+                   [14, [[7, 8], [7, 9], [8, 8], [8, 9]]], [8, [[8, 2], [9, 2]]],
+                   [16, [[8, 3], [9, 3]]], [15, [[8, 6], [8, 7]]],
+                   [13, [[9, 5], [9, 6], [9, 7]]], [17, [[9, 8], [9, 9]]]]
+
+    x = cp.intvar(1, n, shape=(n, n), name="x")
+
+    model = cp.Model()
+
+    model += [cp.AllDifferent(row) for row in x]
+    model += [cp.AllDifferent(col) for col in x.transpose()]
```
IgnaceBleukx: x.T is also fine

examples/csplib/prob067_quasigroup_completion.py
```diff
@@ -0,0 +1,65 @@
...
+    # We iterate through the start matrix and add a constraint for each non-zero cell.
+    start_np = np.array(start)
+    for i in range(N):
+        for j in range(N):
+            if start_np[i, j] != 0:
+                model += puzzle[i, j] == start_np[i, j]
+
+    # 2. Each row must contain unique values from 1 to N.
+    for i in range(N):
+        model += cp.AllDifferent(puzzle[i, :])
```
IgnaceBleukx: Same comments as on the normal quasigroup : )

IgnaceBleukx (changes requested):
Great improvement already. 
I've had some fun with prob18 as there were lots of simplifications possible, let me know if you don't like some stuff ; )

Some more comments, and a general theme: how to handle the default case?
I would just put them as default arguments, and add asserts to ensure the dimensions of all arguments provided to the function are compatible

kostis-init:
Thanks for the detailed comments, I think most of them are addressed, let me know if anything else remains!

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,196 @@
...
+
+    model = cp.Model()
+
+    # Hints constraint
+    for r, c, v in hints:
+        model += grid[r, c] == v
+
+    # Row and column sums
+    for i in range(rows):
+        model += cp.sum(grid[i, :] > WATER) == rowsum[i]
```
IgnaceBleukx, line 74: != to WATER would make more sense here? Its equivalent of course, but just because WATER happens to be 0 : )

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,196 @@
...
+            diag_is_water = []
+            if r > 0 and c > 0:
+                diag_is_water.append(grid[r - 1, c - 1] == WATER)
+            if r > 0 and c < cols - 1:
+                diag_is_water.append(grid[r - 1, c + 1] == WATER)
+            if r < rows - 1 and c > 0:
+                diag_is_water.append(grid[r + 1, c - 1] == WATER)
+            if r < rows - 1 and c < cols - 1:
+                diag_is_water.append(grid[r + 1, c + 1] == WATER)
+            model += (grid[r, c] > WATER).implies(cp.all(diag_is_water))
```
IgnaceBleukx, line 91: also here != WATER

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,196 @@
...
+                ortho_is_water.append(grid[r + 1, c] == WATER)
+            if c > 0:
+                ortho_is_water.append(grid[r, c - 1] == WATER)
+            if c < cols - 1:
+                ortho_is_water.append(grid[r, c + 1] == WATER)
+
+            # A CIRCLE must be entirely surrounded by water
+            model += (grid[r, c] == CIRCLE).implies(cp.all(ortho_is_water))
+
+            # LEFT piece
```
IgnaceBleukx, line 106: So here we are defining the ship-layouts, right? Maybe a small comment?

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,196 @@
...
+            model += (grid[r, c] == LEFT).implies(
+                ((grid[r, c + 1] == MIDDLE) | (grid[r, c + 1] == RIGHT) if c < cols - 1 else False) &
```
IgnaceBleukx, line 108: Could also use an "InDomain" global constraint here, but fine like this as well

examples/csplib/prob018_water_bucket.py
```diff
@@ -0,0 +1,104 @@
...
+    # Sequence must start with initial state
+    model += cp.all(sequence[0] == initial_state)
+
+    # Transition logic
+    goal_states = []
+    for t in range(max_steps - 1):
+        
+        is_goal = cp.all(sequence[t] == goal_state)
+        goal_states.append(is_goal)
+        model += is_goal.implies(cp.all(sequence[t+1] == sequence[t])) # no more pouring once the goal is reached
```
IgnaceBleukx, line 47: alternative would be to only allow transitions from goal to goal in transition table, but this is more explicit

examples/csplib/prob018_water_bucket.py
```diff
@@ -0,0 +1,104 @@
...
+    import argparse
+
+    parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument("-max_steps", type=int, default=20, help="Maximum number of steps")
+
+    max_steps = parser.parse_args().max_steps
+
+    model, (sequence,) = water_bucket(max_steps=max_steps)
+
+    # print(model)
```
IgnaceBleukx, line 91: stray comment

examples/csplib/prob021_crossfigures.py
```diff
@@ -0,0 +1,276 @@
...
+    for i in range(3, 1 + int(math.sqrt(n)), 2):
+        if n % i == 0:
+            return False
+    return True
+
+
+def to_num(a, n, base):
+    """Constrain digit list a to represent number n in the given base."""
+    tlen = len(a)
+    return n == cp.sum([(base ** (tlen - i - 1)) * a[i] for i in range(tlen)])
```
IgnaceBleukx, line 92: maybe make base a global constant? Then you don't need to pass it around

examples/csplib/prob021_crossfigures.py
```diff
@@ -0,0 +1,276 @@
...
+
+def across(Matrix, Across, Len, Row, Col):
+    """Link an across clue variable to its digit cells in the grid."""
+    Row -= 1
+    Col -= 1
+    digits = [Matrix[Row, Col + i] for i in range(Len)]
+    return [to_num(digits, Across, 10)]
+
+
+def down(Matrix, Down, Len, Row, Col):
```
IgnaceBleukx, line 103: we typically use lower-case arguments for functions

examples/csplib/prob023_magic_hexagon.py
```diff
@@ -0,0 +1,114 @@
...
+    """Build a magic hexagon model for edge length n.
+
+    Args:
+        n: Edge length of the hexagon (default 3).
+
+    Returns:
+        (model, vars) where vars is a flat list of decision variables.
+    """
+    num_cells = 3 * n * n - 3 * n + 1
+    magic_sum = num_cells * (num_cells + 1) // (2 * (2 * n - 1))
```
IgnaceBleukx, line 53: I'm not sure this is correct? Isn't the goal to find the magic sum? 
I.e., that you have `cp.AllEqual(list_sums_that_must_sum_to_the_same_magic_number)`?

IgnaceBleukx: for anything but 3 it is currently unsat

examples/csplib/prob039_rehearsal.py
```diff
@@ -0,0 +1,82 @@
...
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_039_rehearsal/csplib_039_rehearsal.cpmpy.py)
+"""
+
+import cpmpy as cp
+
+
+def rehearsal(num_pieces=9, num_players=5, duration=None, rehearsal_matrix=None):
+    if duration is None:
+        duration = [2, 4, 1, 3, 3, 2, 5, 7, 6]
+    if rehearsal_matrix is None:
```
IgnaceBleukx, line 23: here default arg is not following the other example patterns

examples/csplib/prob039_rehearsal.py
```diff
@@ -0,0 +1,82 @@
...
+    departure = cp.intvar(0, num_pieces - 1, shape=num_players, name="departure")
+
+    model += cp.AllDifferent(rehearsal_order)
+
+    # Link arrival and departure times to the rehearsal schedule.
+    # A player must be present for all pieces they play in.
+    for p in range(num_players):
+        for i in range(num_pieces):
+            # is_playing is an expression that is true if player p plays in the piece at slot i.
+            is_playing = (rehearsal[p, rehearsal_order[i]] == 1)
```
IgnaceBleukx, line 49: If you make an is_playing[p,i] array upfront, you can re-use it as is_not_playing below

examples/csplib/prob003_quasigroup_existence.py
```diff
@@ -0,0 +1,64 @@
+"""
+Quasigroup existence problem (QG3) in cpmpy.
+
+Problem 003 on CSPlib
+https://www.csplib.org/Problems/prob003/
+
+An order m quasigroup is a Latin square of size m. That is, a $m \\times m$ multiplication table in which each element
```
IgnaceBleukx, line 7: we never render latex? So docstrings should also not contain it

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,196 @@
...
+https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_014_solitaire_battleships/csplib_014_solitaire_battleships.cpmpy.py
+"""
+
+import cpmpy as cp
+
+
+# Cell types in the grid. Each cell is either water or part of a ship.
+# Ship cells indicate their position within the vessel.
+WATER = 0    # water
+CIRCLE = 1   # submarine, a 1-cell ship
```
IgnaceBleukx, line 25: Call it submarine then? : )

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,196 @@
...
+                             (grid[r - 1, c] == WATER if r > 0 else True) &
+                             (grid[r + 1, c] == WATER if r < rows - 1 else True)) if 0 < c < cols - 1 else False
+            is_ver_middle = (((grid[r - 1, c] == TOP) | (grid[r - 1, c] == MIDDLE)) &
+                             ((grid[r + 1, c] == BOTTOM) | (grid[r + 1, c] == MIDDLE)) &
+                             (grid[r, c - 1] == WATER if c > 0 else True) &
+                             (grid[r, c + 1] == WATER if c < cols - 1 else True)) if 0 < r < rows - 1 else False
+            model += (grid[r, c] == MIDDLE).implies(is_hor_middle | is_ver_middle)
+
+    # Fleet composition
+    model += cp.sum(grid == CIRCLE) == fleet_counts[1]
```
IgnaceBleukx, line 150: fleet_counts index should match `CIRCLE` constant

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,196 @@
...
+LEFT = 2     # leftmost cell of a horizontal ship
+RIGHT = 3    # rightmost cell of a horizontal ship
+TOP = 4      # topmost cell of a vertical ship
+BOTTOM = 5   # bottommost cell of a vertical ship
+MIDDLE = 6   # interior cell of a ship longer than 2
+
+
+DEFAULT_ROWSUM = (0, 2, 3, 1, 2, 4, 2, 1, 2, 3)
+DEFAULT_COLSUM = (1, 3, 3, 1, 5, 1, 2, 4, 0, 0)
+DEFAULT_FLEET_COUNTS = ((4, 1), (3, 2), (2, 3), (1, 4))
```
IgnaceBleukx, line 35: I would already make it a dict here, instead of in the model func

examples/csplib/prob014_solitaire_battleships.py
```diff
@@ -0,0 +1,196 @@
...
+            print("\n--- Solution Board ---")
+            symbols = {
+                WATER: '~', CIRCLE: 'O', LEFT: '<',
+                RIGHT: '>', TOP: '^', BOTTOM: 'v', MIDDLE: '#'
+            }
+            for r in range(args.rows):
+                print(" ".join(symbols[grid[r, c].value()] for c in range(args.cols)))
+        pretty_print(grid)
+        # get all solutions by restricting the grid to be different from previous solutions
+        for _s in range(10):  # limit to 10 solutions
```
IgnaceBleukx, line 188: Why get all sols here? This is a bit inconsistent with the other examples

examples/csplib/prob016_traffic_lights.py
```diff
@@ -0,0 +1,77 @@
...
+import cpmpy as cp
+
+
+def traffic_lights():
+    vehicle_lights = cp.intvar(0, 3, shape=4, name=tuple(f"V{i}" for i in range(1, 5)))
+    pedestrian_lights = cp.intvar(0, 1, shape=4, name=tuple(f"P{i}" for i in range(1, 5)))
+
+    # Allowed combinations for (V_i, P_i, V_{i+1}, P_{i+1})
+    allowed_tuples = [
+        [0, 0, 2, 1],
```
IgnaceBleukx, line 27: Use `RED` and `GREEN` constants instead?

examples/csplib/prob016_traffic_lights.py
```diff
@@ -0,0 +1,77 @@
...
+Problem 016 on CSPlib
+https://www.csplib.org/Problems/prob016/
+
+Imagine a four-way traffic junction with eight traffic lights. Four lights
+(V1 to V4) are for vehicles, and four (P1 to P4) are for pedestrians. The lights
+have different states (e.g., red, green). The problem parameters describe
+constraints on which combinations of light states are safe.
+
+Vehicle light states: 0=red, 1=red-yellow, 2=green, 3=yellow
+Pedestrian light states: 0=red, 1=green
```
IgnaceBleukx, line 13: May be going a bit far, but an ascii-layout of the junction showing could be useful? Not sure if necessary

examples/csplib/prob022_bus_driver_scheduling.py
```diff
@@ -0,0 +1,69 @@
...
+import cpmpy as cp
+
+
+def bus_driver_scheduling(num_tasks=12, num_shifts=14, shifts=None):
+    if shifts is None:
+        # shifts: a list of available shifts; each inner list contains the
+        # indices of the pieces of work (tasks) that that shift covers. Tasks
+        # are numbered 0..(num_tasks-1). The model will choose a subset of these
+        # shifts so that every task is covered exactly once.
+        shifts = [
```
IgnaceBleukx, line 26: make default arg constant like in the other examples

examples/csplib/prob022_bus_driver_scheduling.py
```diff
@@ -0,0 +1,69 @@
...
+            [0, 4, 8],
+            [0, 5, 10],
+            [1, 6, 11],
+            [2, 7, 9],
+            [3, 6, 8]
+        ]
+
+    model = cp.Model()
+
+    x = cp.boolvar(shape=num_shifts, name="x")
```
IgnaceBleukx, line 45: Its an edge-case, but this will fail if num_shifts is 1; to fix it you put `shape=(num_shifts,)` instead

IgnaceBleukx: also maybe a more intersting name than "x" would be nice? : )
Smth like "select_shift" or smth

examples/csplib/prob023_magic_hexagon.py
```diff
@@ -0,0 +1,114 @@
...
+
+    coords = _hex_coords(n)
+    assert len(coords) == num_cells
+
+    LD = cp.intvar(1, num_cells, shape=num_cells, name="LD")
```
IgnaceBleukx, line 58: What does "LD" stand for? Why not just "cells"?

examples/csplib/prob023_magic_hexagon.py
```diff
@@ -0,0 +1,114 @@
...
+        n: Edge length of the hexagon (default 3).
+
+    Returns:
+        (model, vars) where vars is a flat list of decision variables.
+    """
+    num_cells = 3 * n * n - 3 * n + 1
+    magic_sum = num_cells * (num_cells + 1) // (2 * (2 * n - 1))
+
+    coords = _hex_coords(n)
+    assert len(coords) == num_cells
```
IgnaceBleukx, line 56: You define this function yourself, so no need for the assert?

examples/csplib/prob023_magic_hexagon.py
```diff
@@ -0,0 +1,114 @@
...
+
+    LD = cp.intvar(1, num_cells, shape=num_cells, name="LD")
+
+    # Build lookup: coordinate -> variable
+    var_at = {c: v for c, v in zip(coords, LD)}
```
IgnaceBleukx, line 61: You don't need the `coords` variable, right?
Instead you can just do `var_at = {_hex_coords(c): v}`?

examples/csplib/prob034_warehouse_location.py
```diff
@@ -0,0 +1,97 @@
...
+)
+
+def warehouse_location(
+    n_suppliers=5,
+    n_stores=10,
+    building_cost=30,
+    capacity=DEFAULT_CAPACITY,
+    cost_matrix=DEFAULT_COST_MATRIX,
+):
+    """Build a warehouse location model.
```
IgnaceBleukx, line 44: not all examples have this style of docstring.
I'm not sure if its really required to have either; but it would be nice if its consistent among the csplib examples : )

examples/csplib/prob034_warehouse_location.py
```diff
@@ -0,0 +1,97 @@
...
+        capacity: Capacity of each warehouse, length n_suppliers.
+        cost_matrix: Supply cost matrix of shape n_stores x n_suppliers.
+
+    Returns:
+        (model, vars) where vars contains supplier assignments and open warehouses.
+    """
+    assert len(capacity) == n_suppliers
+    assert len(cost_matrix) == n_stores
+    assert all(len(row) == n_suppliers for row in cost_matrix)
+    assert sum(capacity) >= n_stores
```
IgnaceBleukx, line 59: This is not required? It will be UNSAT yes, but that's ok. This bit should just be about checking dimensions of the input params

examples/csplib/prob056_sonet.py
```diff
@@ -0,0 +1,69 @@
...
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_056_sonet/csplib_056_sonet.cpmpy.py)
+"""
+
+import cpmpy as cp
+
+
+def sonet(r=4, n=10, demand=None, capacity_nodes=None):
+    if demand is None:
+        demand = [[0, 1, 1, 0, 0, 0, 1, 1, 0, 0],
```
IgnaceBleukx, line 22: default demand should follow pattern of other examples

examples/csplib/prob057_killer_sudoku.py
```diff
@@ -0,0 +1,79 @@
...
+Killer sudoku problem in cpmpy.
+
+Problem 057 on CSPlib
+https://www.csplib.org/Problems/prob057/
+
+Killer sudoku (also killer su doku, sumdoku, sum doku, addoku, or
+samunamupure) is a puzzle that combines elements of sudoku and kakuro.
+Despite the name, the simpler killer sudokus can be easier to solve
+than regular sudokus, depending on the solver's skill at mental arithmetic;
+the hardest ones, however, can take hours to crack.
```
IgnaceBleukx, line 11: for humans I suppose?

examples/csplib/prob057_killer_sudoku.py
```diff
@@ -0,0 +1,79 @@
...
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_057_killer_sudoku/csplib_057_killer_sudoku.cpmpy.py)
+"""
+
+import cpmpy as cp
+
+
+def killer_sudoku(n=9, problem=None):
+    if problem is None:
+        problem = [[3, [[1, 1], [1, 2]]], [15, [[1, 3], [1, 4], [1, 5]]],
```
IgnaceBleukx, line 31: set default arg as global variable

examples/csplib/prob057_killer_sudoku.py
```diff
@@ -0,0 +1,79 @@
...
+
+    model = cp.Model()
+
+    model += [cp.AllDifferent(row) for row in x]
+    model += [cp.AllDifferent(col) for col in x.T]
+
+    for i in range(2):
+        for j in range(2):
+            cell = [x[r, c] for r in range(i * 3, i * 3 + 3) for c in range(j * 3, j * 3 + 3)]
+            model += [cp.AllDifferent(cell)]
```
IgnaceBleukx, line 57: just use the normal sudoku block formulation?

```
bsize = int(n**0.5)
for i in range(0,n,bsize):
     for j in range(0,n,bsize):
        cp.AllDifferent(x[i:i+bsize,j:j+bsize])
```

examples/csplib/prob057_killer_sudoku.py
IgnaceBleukx: Also I don't really undertand how this data is formatted... some comment/docstring explaining it would be helpful : )

examples/csplib/prob067_quasigroup_completion.py
```diff
@@ -0,0 +1,60 @@
...
+    model = cp.Model()
+
+    # Constraints
+    # 1. Pre-fill the grid with the starting values.
+    # We iterate through the start matrix and add a constraint for each non-zero cell.
+    start_np = np.array(start)
+    for i in range(N):
+        for j in range(N):
+            if start_np[i, j] != 0:
+                model += puzzle[i, j] == start_np[i, j]
```
IgnaceBleukx, line 38: do this vectorized:
```python
model += puzzle[start_np != 0] == start_np[start_np != 0]
```

examples/csplib/prob067_quasigroup_completion.py
```diff
@@ -0,0 +1,60 @@
...
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_067_quasigroup_completion/csplib_067_quasigroup_completion.cpmpy.py)
+"""
+
+import cpmpy as cp
+import numpy as np
+
+
+def quasigroup_completion(N=5, start=None):
+    if start is None:
+        start = [[1, 0, 0, 0, 0],
```
IgnaceBleukx, line 21: default arg pattern

examples/csplib/prob067_quasigroup_completion.py
```diff
@@ -0,0 +1,60 @@
...
+                 [0, 2, 0, 0, 0],
+                 [0, 0, 3, 0, 0],
+                 [0, 0, 0, 4, 0],
+                 [0, 0, 0, 0, 5]]
+
+    puzzle = cp.intvar(1, n, shape=(n, n), name="puzzle")
```
IgnaceBleukx, line 27: this does not use the `N` parameter! (rename the arg in the function)

examples/csplib/prob074_maximum_clique.py
```diff
@@ -0,0 +1,59 @@
...
+
+Model from DCP-Bench-Open (https://github.com/DCP-Bench/DCP-Bench-Open/blob/main/dataset/csplib_074_maximum_clique/csplib_074_maximum_clique.cpmpy.py)
+"""
+
+import cpmpy as cp
+
+
+def maximum_clique(n=5, adj=None):
+    if adj is None:
+        adj = [[0, 1, 0, 1, 0],
```
IgnaceBleukx, line 20: default arg pattern

IgnaceBleukx (changes requested):
Some more feedback : )

PR #1017, opened 2026-06-15
Solver var consistent return type

### LINKED ISSUE #999: Inconsistent return type for solver vars
As discussed in #995 , the `solver_var` function for PySAT and Pindakaas has a different return type, depending on the type of variable it gets as input.

For Boolean variables, the function returns a solver-specific Boolean variable, for _IntVarImpl's it returns a list of Boolean varibales, used in the encoding.

Handling of integer variables should probably be refactored to a helper function instead.

Dimosts:
fixes #999


cpmpy/solvers/pindakaas.py
```diff
@@ -123,18 +123,20 @@ def native_model(self):
         return self.pdk_solver
 
     def _int2bool_user_vars(self):
```
IgnaceBleukx: Could you type this? Also a docstring would be useful : )
I'm also not entire sure about the function name, but its already there from before so thats fine

IgnaceBleukx (changes requested):
Great, nice cleanups.
I would like to solver_vars to then also be typed? I think that was the original point anyway ; )

tias:
This would be nice to still have in V20.

I did a light-pass review and I approve : )

So @IgnaceBleukx can you do final check and merge?

IgnaceBleukx (approved):

PR #1018, opened 2026-06-15
Native mdd in minizinc

### LINKED ISSUE #967: Utilize MiniZinc native support for MDD global constraint
The MDD global constraint was added to CPMpy (see #924). The documentation of MiniZinc shows that an MDD global constraint is natively supported (https://docs.minizinc.dev/en/stable/lib-globals-extensional.html#mdd). However, using this when calling MiniZinc from CPMpy is not trivial; the required arguments are different compared to our implementation, and the conversion of the expression should be returned as a string.

Dimosts:
Fixing #967

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -968,9 +968,8 @@ class MDD(GlobalConstraint):
         assert len(sink_nodes) == 1
         self.sink_node = sink_nodes[0]
 
-        # reduce the MDD if requested
-        if reduce:
-            self._reduce()
+        # store whether the MDD should be reduced during decomposition
+        self.reduce = reduce
 
     def _reduce(self):
         """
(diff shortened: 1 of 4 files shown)


WoutPiessens (approved):
It looks good, I added some documentation and made a small change to node_map.

One open question: as it stands, the MDD _reduce() method (which identifies equivalent nodes in the MDD and merges them) is called in the MDD constructor, and therefore this operation would also be performed for solvers that support MDD natively. I'm not sure if we want this or not? (could be double work for solvers that natively reduce the MDD)

tias (approved):
had to merge in master, lets let the tests finish

re @WoutPiessens no, we just want to pass it on as is. If that is what is happening, do merge as is (otherwise tell dimos or update it yourself, thx)

WoutPiessens:
@tias I have double checked it, and currently the reduce step was already performed before constructing the MiniZinc MDD object, so I fixed it with a change to the MDD global. I think the cleanest solution is to only call _reduce() in decompose(), instead of calling it in the constructor of the MDD global, as we currently do. This also required some changes to the tests. If you approve I will merge with master.

I think this also opens the question if we even want to have a reduce flag in the MDD constructor, given that this would then be abstracted away in decompose(). But that's maybe not for this PR.

tias (changes requested):
I think how you did it now is fine: its a flag that impacts the decomposition. I guess it just needs an update to the docs that clarifies that it only has impact on the decomposition?

WoutPiessens:
Doc update made

tias (approved):
great

PR #1020, opened 2026-06-18
Add incremental flag for faster one-shot Exact and SCIP solves

### LINKED ISSUE #968: Non-incremental solve function in solver interface
As posted by Jo in another issue, for some solvers (e.g., Exact), there are more efficient API calls to use when "just solving" a CSP.

> A separate issue to note: `self.xct_solver.toOptimum` in `def solve(...)` in exact.py is slower than `self.xct_solver.runFull(true)`. The latter is allowed to add bound constraints. You really want to run the latter for regular optimization runs (e.g., competitions etc.). 

 _Originally posted by @JoD in [#896](https://github.com/CPMpy/cpmpy/issues/896#issuecomment-4300137970)_

We currently use the slow version, which does not post bound-constraints, as we allow to the use to change the objective later on, but that of course is only useful in a handful of usecases.

Maybe we can add a flag "incremental" or something to the constructor of the exact solver interface, which determines the solver-API call to use.
When initializing the solver from a `cpmpy.Model` class, this will always be set to False (somewhat similar to what we do with solveAll and the `call_from_model` argument)

tias:
Closes #968.

For one-shot solves via `Model.solve()`, Exact and SCIP were doing extra work meant for reusing the solver object. Exact always went through `toOptimum()`; SCIP always called `freeTransform()` after `optimize()`. Both are useful when you keep the solver and solve again, but `Model.solve()` throws the solver away anyway.

This adds an `incremental` flag on both interfaces (default `True`). `Model.solve()` passes `incremental=False` for exact and scip. Exact then uses `runFull`; SCIP skips `freeTransform()`.

Left `Model.solveAll()` alone — SCIP's generic enumeration calls `solve()` in a loop and adds nogoods, so it still needs the incremental path.

Added a couple of tests that check `Model.solve()` gives the same result as `SolverLookup.get(..., incremental=True)`.

### DIFF
--- a/cpmpy/model.py
+++ b/cpmpy/model.py
@@ -187,11 +187,20 @@ class Model(object):
         if kwargs and solver is None:
             raise NotSupportedError("Specify the solver when using kwargs, since they are solver-specific!")
 
+        if isinstance(solver, str):
+            name = solver.split(":", 1)[0]
+        elif isinstance(solver, type) and issubclass(solver, SolverInterface):
+            name = solver.__name__.removeprefix("CPM_").lower()
+        else:
+            name = None
+
+        init_kwargs = {"incremental": False} if name in ("exact", "scip") else {}
+
         if isinstance(solver, SolverInterface):
             # for advanced use, call its constructor with this model
-            s = solver(self)
+            s = solver(self, **init_kwargs)
         else:
-            s = SolverLookup.get(solver, self)
+            s = SolverLookup.get(solver, self, **init_kwargs)
 
         # call solver
         ret = s.solve(time_limit=time_limit, **kwargs)
--- a/cpmpy/solvers/exact.py
+++ b/cpmpy/solvers/exact.py
@@ -116,7 +116,7 @@ class CPM_exact(SolverInterface):
             return None
 
 
-    def __init__(self, cpm_model=None, subsolver=None, **kwargs):
+    def __init__(self, cpm_model=None, subsolver=None, incremental=True, **kwargs):
         """
         Constructor of the native solver object
 
@@ -126,6 +126,8 @@ class CPM_exact(SolverInterface):
         Arguments:
             cpm_model: Model(), a CPMpy Model() (optional)
             subsolver: None
+            incremental: if True (default), use toOptimum in solve() so the solver state
+                can be reused (e.g. change objective). If False, use runFull (faster one-shot).
 
         Exact takes options at initialization instead of solving.
         The Exact solver parameters are defined by https://gitlab.com/nonfiction-software/exact/-/blob/main/src/Options.hpp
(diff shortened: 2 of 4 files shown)


cpmpy/solvers/exact.py
```diff
@@ -116,7 +116,7 @@ def version() -> Optional[str]:
             return None
 
 
-    def __init__(self, cpm_model=None, subsolver=None, **kwargs):
+    def __init__(self, cpm_model=None, subsolver=None, incremental=True, **kwargs):
```
IgnaceBleukx, line 119: I feel it's too specific; we have a `call_from_model` flag that we pass to solveAll, we could re-use something similar here? There it is used to determine whether we want to print a warning or not I believe, here we can use it so select the solve func of exact.

cpmpy/solvers/exact.py
```diff
@@ -227,7 +230,11 @@ def solve(self, time_limit:Optional[float]=None, assumptions:Optional[Iterable[_
             
         # call the solver, with parameters
         start = time.time()
-        my_status, obj_val = self.xct_solver.toOptimum(timeout=timeout)
+        obj_val = None
+        if self.incremental:
+            my_status, obj_val = self.xct_solver.toOptimum(timeout=timeout)
```
IgnaceBleukx, line 235: I would not trust the objective val computed by exact in this special path? in the previous version, we would always just recompute it from the expresssion object, which I think is the right way

cpmpy/solvers/scip.py
```diff
@@ -183,7 +184,8 @@ def solve(self, time_limit=None, solution_callback=None, **kwargs):
         # transformed space, and just leave the original model. All solutions are kept, 
         # and the user can now change the model as they will. The downside is that potentially 
         # useful information for speeding up the next optimisation call is thrown out.
-        self.scip_model.freeTransform()
+        if self.incremental:
```
IgnaceBleukx, line 187: We could just always call freeTransform, no? If we do smt incremental then we should definetly, if its not incremental because we call from model, then it doesn't hurt either? Or can this call be expensive?

cpmpy/model.py
```diff
@@ -187,11 +187,20 @@ def solve(self, solver:Optional[str]=None, time_limit:Optional[int|float]=None,
         if kwargs and solver is None:
             raise NotSupportedError("Specify the solver when using kwargs, since they are solver-specific!")
 
+        if isinstance(solver, str):
```
IgnaceBleukx, line 190: If we decide to go the call_from_model flag-way then this is not needed

IgnaceBleukx (changes requested):
Seeing this, I'm not sure if we should add it to the constructor args? Maybe better to add a flag to the .solve function instead?

PR #1021, opened 2026-06-18
Microoptimization: slots for _NumVarImpl

IgnaceBleukx:
I recently learned about using `__slots__` in python objects
https://stackoverflow.com/questions/472000/usage-of-slots

The main advantage is that your objects take less memory, which may be very useful for models with loads of variables/constraints (e.g., files we read in from DIMACS?)

I've implemented it for _NumVarImpl, no real time gains, but memory footprint for a variable now goes from ±450B to ±80B which is nice : )

The real gains are probably when typing Operator and Comparison, or some of the globals (e.g., Multiplication).

### DIFF
--- a/cpmpy/expressions/variables.py
+++ b/cpmpy/expressions/variables.py
@@ -320,6 +320,8 @@ class _NumVarImpl(Expression):
     including bounds checking and value management. It should not be instantiated
     directly, but rather through the helper functions :func:`~cpmpy.expressions.variables.intvar` and :func:`~cpmpy.expressions.variables.boolvar`.
     """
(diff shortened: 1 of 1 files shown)


tias:
oh nice, this is some of the benefits I was thinking we could only get with mypyc, so if we can get it pure python, great.

In that SO post, it says parents also need slots, so strange this already works. But also, I wouldn't mind if 'name' becomes part of the slot of Expression, then there could be an even bigger win? or would that require using slots in all expressions?

IgnaceBleukx:
Yes good point, we might as well have `name` as a slot for Expression : )

PR #1023, opened 2026-06-20
Type nested constraints and cpm_array input

### LINKED ISSUE #1022: cpm_array type should be recursive/n-dimensional
it now pretends to only accept a ListLike[ExprLike], but e.g. the future 'multid_element' test in #926 makes a 2D array, perfectly validly

tias:
Fixes #1022.

Adds `NestedBoolExprLike` and uses it for `Model.add`, solver `add`/`transform`, and `normalize`. Types `cpm_array` as `ListLike[ExprLike | Any]` so nested/multi-dimensional python lists type-check.


cpmpy/expressions/core.py
```diff
@@ -105,6 +105,7 @@
 ListLike: TypeAlias = Union[Sequence[T], np.ndarray]  # similar to is_any_list() check  (Sequence a bit more general than list/tuple)
 ExprLike: TypeAlias = Union["Expression", int, np.integer, np.bool_]  # expression or int (incl np variants, e.g. user facing)
 BoolExprLike: TypeAlias = Union["Expression", bool, np.bool_]  # subtype of ExprLike (bool subtype int)
+NestedBoolExprLike: TypeAlias = Union[BoolExprLike, Sequence["NestedBoolExprLike"], np.ndarray]  # nested version of ListLike[BoolExprLike]
```
IgnaceBleukx: Do you need the ndarray? I wouth think the Sequence part already covers that

cpmpy/solvers/cplex.py
```diff
@@ -380,7 +380,7 @@ def _make_numexpr(self, cpm_expr):
         raise NotImplementedError("CPLEX: Not a known supported numexpr {}".format(cpm_expr))
 
 
-    def transform(self, cpm_expr):
+    def transform(self, cpm_expr: NestedBoolExprLike) -> list[Expression]:
```
IgnaceBleukx, line 384: More precisely: `list[BoolExpr]` as return type, but we don't have that type (yet?)

cpmpy/solvers/cplex.py
```diff
@@ -427,30 +427,30 @@ def add(self, cpm_expr_orig):
...
 
         :return: self
       """
       # add new user vars to the set
-      get_variables(cpm_expr_orig, collect=self.user_vars)
+      get_variables(cpm_expr, collect=self.user_vars)
 
       # transform and post the constraints
-      for cpm_expr in self.transform(cpm_expr_orig):
+      for con in self.transform(cpm_expr):
```
IgnaceBleukx, line 441: bit nitpicky, but I prefer `cons`

cpmpy/solvers/choco.py
```diff
@@ -435,10 +435,10 @@ def add(self, cpm_expr):
             the user knows and cares about (and will be populated with a value after solve). All other variables
             are auxiliary variables created by transformations.
 
-        :param cpm_expr: CPMpy expression, or list thereof
-        :type cpm_expr: Expression or list of Expression
+            :param cpm_expr: CPMpy expression, or list thereof
+            :type cpm_expr: NestedBoolExprLike
```
IgnaceBleukx: This docstring is different from other places? We normally use:
```
Arguments:
- cpm_expr (NestedBoolExpr): ...
```
I don't mind either, maybe @ThomSerg knows whats better for readthedocs rendering?

cpmpy/solvers/choco.py
```diff
@@ -422,7 +422,7 @@ def transform(self, cpm_expr):
 
         return cpm_cons
 
-    def add(self, cpm_expr):
+    def add(self, cpm_expr: NestedBoolExprLike) -> "CPM_choco":
```
IgnaceBleukx, line 427: There is also the `Self` type, which we might want to use instead?
https://docs.python.org/3/library/typing.html#typing.Self

IgnaceBleukx (commented):
Nice, a couple of open points, mainly (doc)style related

cpmpy/expressions/core.py
tias: not Sequence because ndarray is N-Dimensional, not 1Dim.
But with a ListLike instead of Sequence it does work; changing

cpmpy/solvers/cplex.py
tias: we do not/can not have a type-way to identify `Expression & Expr.is_bool()` because the latter is at runtime; a side-effect of a very early (now unfortunate) design decision of mine

We do have BoolExprLike because it accepts both Expression (iimplicitly only with is_bool()) as well as bool constants

cpmpy/solvers/cplex.py
tias: cons sounds plural, so I like that less. But even ortools uses `for con in self.transform(cpm_expr)` so I do think the current patch this is more consistent...

cpmpy/solvers/choco.py
tias: Yes, but its from 3.11 only and we still accept 3.10, not worth the bump imho.

IgnaceBleukx: Hmm I see, personally I wouldn't mind the bump as 3.10 will be end of life from october onwards, but ok to keep for now I suppose
https://devguide.python.org/versions/

IgnaceBleukx (approved):
Great, I've updated the docstring styles for the other functions touched by this PR to be consistent.
There is a slight mismatch between the docstring of `.add` for model and solver interfaces: in model we say `return self for method chaining` but I find that a bit confusing. Should we make it consistent or just keep as is?

Otherwise good to go (after highs test is fixed...)

PR #1024, opened 2026-06-23
Mark test as flaky

IgnaceBleukx:
Using the official pytest-rerunfailures plugin, we can allow flaky tests to be rerun automatically when they fail.
Enabled for the objective_numexprs test which fails quite often on the CI...


tias:
Ah yes, I had that 2 times this week too.

Do we know why it sometimes fails actually? (but also very happy with this fix)

IgnaceBleukx:
Sometimes when the servers on the CI are very busy, the solver does not find a feasible solution, or the objective value is not lower/higher than the lb/ub of the expression.

I've made the test sligthly more robust now, by explicitely checking for the result of the solver, but it can still fail because of a suboptimal sol, so kept flaky nevertheless

ThomSerg (approved):

PR #1026, opened 2026-06-23
Improvements to toplevel safening

IgnaceBleukx:
As shortly discussed in #926 there is some weirdness going on with safening of toplevel partial functions, which are addressed here:

1) When a partial function is used toplevel, but it is not supported, we can skip safening it as we know the decomposition will be total, and hence "safen" the function that way.
2) Actually, our previous implementation never did anything with safen_toplevel, as everything was assumed to be nested. This is now fixed here as well. Basically: everything is toplevel, until we pass a SECOND boolean expression. The first one is just the constraint, so the neirest Boolean context will be toplevel. Added some tests to clarify what I mean with this : )

Overall the fixes should result in fewer aux variables when solving a model with partial functions at the toplevel of the model.

To be clear, at some point, this did work fine but #875 introduced the faulty behaviour : )


cpmpy/solvers/cplex.py
```diff
@@ -397,7 +397,7 @@ def transform(self, cpm_expr):
         # apply transformations, then post internally
         # expressions have to be linearized to fit in MIP model. See /transformations/linearize
         cpm_cons = toplevel_list(cpm_expr)
-        cpm_cons = no_partial_functions(cpm_cons, safen_toplevel={"mod", "div", "element", "nd_element"})  # linearize and decompose expect safe exprs
+        cpm_cons = no_partial_functions(cpm_cons, safen_toplevel=None)
```
tias: default is None, so we can simply remove the argument? That would be least confusing (e.g. now it might look as if you have to 'deactivate' it)

cpmpy/solvers/pumpkin.py
```diff
@@ -379,7 +379,7 @@ def transform(self, cpm_expr):
         # apply transformations
         cpm_cons = toplevel_list(cpm_expr)
 
-        cpm_cons = no_partial_functions(cpm_cons, safen_toplevel={"element", "nd_element", "div", "mod"}) # safen toplevel elements, assume total decomposition for partial functions
+        cpm_cons = no_partial_functions(cpm_cons, safen_toplevel={"div", "element"})
```
tias: did you test this? CP solvers tend to assume element is closed...

cpmpy/solvers/TEMPLATE.py
```diff
@@ -406,7 +406,7 @@ def transform(self, cpm_expr):
         # apply transformations
         # XXX chose the transformations your solver needs, see cpmpy/transformations/
         cpm_cons = toplevel_list(cpm_expr)
-        cpm_cons = no_partial_functions(cpm_cons)  # to also safen at toplevel, add: `, safen_toplevel={"element", "nd_element", "div", "mod"})`
+        cpm_cons = no_partial_functions(cpm_cons, safen_toplevel=frozenset({"div"})) # global functions that are decomposed later don't have to be safened toplevel!
```
tias: I would not add anything to it in the default template.. more like before but with a clearer message:
"cpm_cons = no_partial_functions(cpm_cons)  # if the solver requires supported partial function globals to be safened, add them like this: `, safen_toplevel={"element", "nd_element", "div", "mod"})`"

tias: but then with frozenset I guess

cpmpy/transformations/safening.py
```diff
@@ -77,27 +77,34 @@ def no_partial_functions(lst_of_expr: list[Expression],
...
+            if args_changed:
+                changed = True
+                if len(toplevel) > 0:
+                    todolist.extend(toplevel)
+                if len(nbc) > 0:
+                    todolist.extend(nbc)
+                
+                expr = copy(expr)
+                expr.update_args(newargs)
+                newlist.append(expr)
```
tias: 96-100 can simply be a single `newlist.append(expr)` at the indent level of 99

tias (approved):
Great. So this clarifies: safen_toplevel is in case your solver supports a global that does not automatically safen its input.

Good to have it more extensively tested and to do that more robust check.

Minor tweak comments below

IgnaceBleukx:
Processed review, also added some documentation that explains the decomposition of a partial function should be total.

cpmpy/solvers/pumpkin.py
IgnaceBleukx: It used to not work, but now it does : )

PR #1031, opened 2026-06-28
improve docs for testing a (new) solver

tias:
Not sure if the explanation is now complete, but at least its now more up-to-date then it was.

Orestis, you have the most recent new solver PR, could you check the doc and see if you want to add or reword something?

### DIFF
--- a/docs/adding_solver.md
+++ b/docs/adding_solver.md
@@ -101,8 +101,14 @@ The CPMpy package provides a large testsuite on which newly added solvers can be
 Note that for this testsuite to work, you need to add your solver to the `SolverLookup` utility.
 This is done by adding an import statement in `/solvers/__init__.py` and adding an entry in the list of solvers in  `/solvers/utils.py`.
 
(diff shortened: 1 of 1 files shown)


tias:
Highs did a new release 11 hours ago, we probably have to upper-bound its version and then update it.

OrestisLomis (changes requested):
Obviously the updated doc is good. I'm just a bit confused as to why we still need the part about excluding certain globals/operators. We only have one global in there for minizinc, which seems to be resolved now: https://github.com/MiniZinc/libminizinc/issues/813. And the rest of the exclusions are simply related to the outdated transformation pipeline of PySDD, but most of the effort for this is already done through pindakaas' pipeline and the to_cnf transformation. 
All this is to say that in my opinion the current phrasing is a bit weak and may suggest that it is the rule that certain solvers do not accept certain globals/operators, but it is really the exception. In most cases just using the correct transformations (which are already fairly complete) should make a global/operator compatible.

OrestisLomis:
Updated the whole file a bit more extensively

tias:
great; and then I changed it some more based on highs' merge commit; please check

OrestisLomis (approved):
looks good to me!

Issue #1033, opened 2026-06-30
NDVarArray * numpy array silently ignores NumPy-style broadcasting


kostis-init:
Multiplying a CPMpy NDVarArray with a NumPy array silently creates the wrong expression instead of either broadcasting correctly or raising an error.

Minimal example:

```
import cpmpy as cp
import numpy as np

x = cp.intvar(0, 10, shape=(3, 4), name="x")
w = np.array([1, 2, 3, 4])

expr = x * w
print(expr)
```

Expected behavior:

Either:

behave like NumPy broadcasting, i.e. multiply each column by the corresponding weight:
```
[
  [x[0,0]*1, x[0,1]*2, x[0,2]*3, x[0,3]*4],
  [x[1,0]*1, x[1,1]*2, x[1,2]*3, x[1,3]*4],
  [x[2,0]*1, x[2,1]*2, x[2,2]*3, x[2,3]*4],
]
```

or raise an error saying that this kind of broadcasting is not supported.

Actual behavior:

The expression appears to be constructed by zipping over the first axis, so it behaves like:

```
[
  x[0, :] * 1,
  x[1, :] * 2,
  x[2, :] * 3,
]
```

and the final weight 4 is silently ignored.

A workaround seems to be using NumPy explicitly: `expr = np.multiply(x, w)` which appears to apply broadcasting correctly.

tias:
Oh, good catch. Seems to be a deeper issue in our vectorize, see PR

PR #1034, opened 2026-06-30
Fix NDVarArray vectorized ops to use NumPy broadcasting

tias:
- Fix `_vectorized` on `NDVarArray` to broadcast operands with `np.broadcast_to` and apply operators element-wise over `.flat`, fixing `x * w` for mismatched-but-broadcastable shapes (#1033).
- Reuse `_vectorized` for `implies`; update `LexLess`/`LexLessEq` decompositions to compare `bvar[:-1]` explicitly; add broadcast multiplication tests.


tias:
I also abuse this PR to update the DFL example so it no longer times out nor tries to show/wait for a plot

cpmpy/expressions/globalconstraints.py
```diff
@@ -2405,7 +2405,7 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
         # Constraint ensuring that each element in X is less than or equal to the corresponding element in Y,
         # until a strict inequality is encountered.
         defining = []
-        defining.extend(bvar == ((X <= Y) & ((X < Y) | bvar[1:])))  # vectorized expression, treat as list
+        defining.extend(bvar[:-1] == (X <= Y) & ((X < Y) | bvar[1:]))  # vectorized for all but the last
```
IgnaceBleukx, line 2453: This is actually one of the globals that we still need to convert to not use auxiliaries... but fix is ok for this PR

cpmpy/expressions/variables.py
```diff
@@ -660,23 +660,33 @@ def get_bounds(self) -> tuple[np.ndarray, np.ndarray]:
...
             NDVarArray: The vectorized result.
         """
-        if not isinstance(other, Iterable):
-            other = [other]*len(self)
-        # this is a bit cryptic, but it calls 'attr' on s with o as arg
+        if not isinstance(other, np.ndarray):
+            other = np.asarray(other)
+        try:
+            other = np.broadcast_to(other, self.shape)
+        except ValueError as e:
```
IgnaceBleukx, line 679: isn't there a more reliable way to check if they can be broadcasted? A ValueError can come from different things too, no?

cpmpy/expressions/variables.py
```diff
@@ -660,23 +660,33 @@ def get_bounds(self) -> tuple[np.ndarray, np.ndarray]:
...
+            raise ValueError(
+                f"operands could not be broadcast together with shapes {self.shape} {other.shape}"
+            ) from e
         # s.__eq__(o) <-> getattr(s, '__eq__')(o)
-        return cpm_array([getattr(s,attr)(o) for s,o in zip(self, other)])
+        # unwrap numpy scalars so e.g. int <= np.int64 does not return NotImplemented
+        return cpm_array([
+            getattr(s, attr)(o if not isinstance(o, np.generic) else o.item(), **kwargs)
```
IgnaceBleukx: add a small comment about the np.generic?

tests/test_expressions.py
```diff
@@ -253,6 +253,29 @@ def test_nullarg_mul(self):
...
+    def test_numpy_array_mul(self):
+        x = intvar(0, 10, shape=(3, 4), name="x")
+        w = np.array([1, 2, 3, 4])
+        expr = x * w
+        ref = np.multiply(x, w)
+        assert expr.shape == ref.shape
+        for idx in np.ndindex(expr.shape):
+            assert str(expr[idx]) == str(ref[idx])
+
+    def test_numpy_scalar_mul(self):
```
IgnaceBleukx, line 267: also add a test where intstead of `2` we multiply with a single expr?

IgnaceBleukx (commented):
Great! Some doc-related comments, and one thing I was wondering:

There is still some weird behaviour, not sure if it can be fixed easily in this PR or we have to go ufunc all the way...

```python
>>> x = cp.intvar(0,10,shape=3, name=tuple("abc"))
>>> y = cp.intvar(0,10,name="y")
>>> x * y
NDVarArray([(a) * (y), (b) * (y), (c) * (y)], dtype=object)
>>> y * x
(y) * ([a b c])
>>> 2 * x
NDVarArray([2 * (a), 2 * (b), 2 * (c)], dtype=object)
>>> x * 2
NDVarArray([2 * (a), 2 * (b), 2 * (c)], dtype=object)
```

cpmpy/expressions/variables.py
tias: yes but then we do two calls... see commit

IgnaceBleukx:
Had a bit of a fight with mypy on the CI and my local machine... latest numpy version annotates reshape to return a ndvararray, but that version is not compatible with Python 3.10; so the CI installed an older version, which does not have that annotation, which causes mypy to complain about an unused type ignore...

Anyway, all that to say: I updated the Python version to 3.11 on the linting workflow of the CI : )

Issue #341, opened 2023-06-02
Ufunc of NDVarArray


IgnaceBleukx:
Helene exposed an issue with our implementation of the NDVarArray overridden functions (__mul__, __add__....) in pull request #337 

To reproduce:
```python
import cpmpy as cp

w = cp.cpm_array([10,20,30])
x = cp.intvar(0,10, shape=3)

w *x 
>> NDVarArray(NotImplemented,NotImplemented,NotImplemented)
```

What happens is the `__mul__` method of NDVarArray `w` tries to call `__mul__` on each of the elements in that array. I.e., `10.__mul__(x[0])`, `20.__mul__(x[1])`... but the built-in type `int` has no idea how to multiply with our custom objects.
Instead of calling `__rmul__` on our objects as it does when expressing `10 * x[0]`, a `NotImplementedError` is returned.

Did some research on how to properly subclass a `np.ndarray` and it seems we don't really need the `_vectorized` thingy...
https://numpy.org/doc/stable/reference/arrays.classes.html

363734:
This is the code that triggered the discovery: tentative of refactoring the decompose method of the cumulative cst: 
```      
for t in range(lb,ub+1):
            demand_at_t = 0
            for job in range(len(start)):
                if is_num(demand):
                    demand_at_t += demand * ((start[job] <= t) & (t < end[job]))
                else:
                    demand_at_t += demand[job] * ((start[job] <= t) & (t < end[job]))
            cons += [capacity >= demand_at_t]
```
to
```      
for t in range(lb, ub+1):
            cons += [capacity >= sum(demand * ((start <= t) & (t < end)))]
```

Wout4:
this issue also mentions ufunc of ndvararray, so might also be solved? https://github.com/CPMpy/cpmpy/issues/124

hbierlee:
closed because seems to be working.

PR #1035, opened 2026-07-01
Fix 'Expr OP NDArr' by doing 'NDArr ROP Expr'

tias:
Fix 'Expr OP NDArr' by doing 'NDArr ROP Expr'

Fixes:
```
>>> x * y
NDVarArray([(a) * (y), (b) * (y), (c) * (y)], dtype=object)
>>> y * x
(y) * ([a b c])
```
from #1034

Curiously its actually a typing gap... that you can also get an
ndarray as second argument.

So with this now fixed, we might as well just go ahead and type all
these operators...?

### DIFF
--- a/tests/test_expressions.py
+++ b/tests/test_expressions.py
@@ -266,6 +266,15 @@ class TestMul:
         assert yx.shape == (3,)
         assert all(a is y or b is y for a, b in (e.args for e in yx))
 
+    def test_scalar_expr_with_numpy_array(self):
+        e = sum(cp.boolvar(3))
+        y = np.array([1, 2, 3])
+        ey, ye = e * y, y * e
+        assert isinstance(ey, NDVarArray)
+        assert ey.shape == ye.shape == (3,)
+        for a, b in zip(ey, ye):
+            assert str(a) == str(b)
+
     def test_nullarg_mul(self):
         x = intvar(0,5,shape=3, name="x")
         a = np.array([0,1,1], dtype=bool)
--- a/tests/test_expressions.py
+++ b/tests/test_expressions.py
@@ -243,6 +243,18 @@ class TestMul:
         assert expr.is_lhs_num is False
         assert expr.args[0] is x and expr.args[1] is y
 
+    def test_nullarg_mul(self):
+        x = intvar(0,5,shape=3, name="x")
+        a = np.array([0,1,1], dtype=bool)
+
+        prod = x * a
+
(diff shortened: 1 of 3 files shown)


tias:
all of this is really pointing to that we should switch to ufuncs #341 ... but that seems a bit rushed to do now?

tests/test_expressions.py
```diff
@@ -243,6 +243,38 @@ def test_mul_is_lhs_num(self):
         assert expr.is_lhs_num is False
         assert expr.args[0] is x and expr.args[1] is y
 
+    def test_scalar_expr_with_ndarray(self):
```
IgnaceBleukx, line 258: Move these to `TestArrayExpression` class instead? Now they are in the tests for multiplication... 
Ideally also test for all of the other operators, not just mul.

IgnaceBleukx (approved):
Yeah absolutely we should have another look at ufuncs... But it will require lots of refactoring which is out of scope for the next release.

Current appraoch is a bit hacky of course, but fine as a patch for now.

Minor comment on tests, otherwise good to go in.

IgnaceBleukx:
This actually also closes one of our oldest issues: #124

IgnaceBleukx:
There is an edge case apparently where a np array can have 0 dimensions: e.g., `np.array(3)` is a valid Numpy array, which causes our `_vectorized` helper function to fail (calls `len` on the array).
Simple fix would be to add 

```python
if isinstance(other, np.ndarray) and other.ndim == 0:
    other = other.item() if other.dtype != object else other[()]
if self.ndim == 0:
    s = self.item() if self.dtype == object else self[()]
    return getattr(s, attr)(other)
```

but it introduced another isinstance check... How shall we continue?

tias:
all this shows is that we are increasingly creeping up to re-implementing what ufunc would all handle...

also in the typing, there are more edge cases: we should also allow plain python lists for consistency, and probably smth else, and smth else...

So I've removed the type annotations again (they are incomplete, as is the code handling).


I'm OK to put this PR in, I'm also OK for you to cover the scalar case. But we have to draw a line for this release somewhere closeby.

And the next phase is just proper ufunc.

IgnaceBleukx:
I'm ok leaving the edge case for now, I cannot think of an actual use case where you would end up with such an "array", definitely not in common modeling. So indeed let's just draw the line here for this release, and v0.21 can have proper ufuncs

PR #1039, opened 2026-07-06
Decompose, `Regular` and `MDD` bugs from fuzztesting

WoutPiessens:
Fuzztesting revealed the following bug: in decompose_in_tree it is possible for an expression resulting from decomposition to not be a conjunction. Therefore the original assertion was invalid and the if-statement had to be rewritten.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -962,7 +962,7 @@ class Regular(GlobalConstraint):
         # define the rest of the automaton using transition table
         cons.extend(Table([state_vars[i - 1], arr[i], state_vars[i]], transitions) for i in range(1, len(arr)))
         # last state must be accepting
-        value = [InDomain(state_vars[-1], [self.node_map[e] for e in accepting])]
+        value : list[Expression] = [InDomain(state_vars[-1], [self.node_map[e] for e in accepting])]
         if complete:
             # constraint is satisfied iff last state is accepting
             return value, cons
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -958,12 +958,15 @@ class Regular(GlobalConstraint):
         state_vars = intvar(0, len(self.nodes)-1, shape=len(arr))
         id_start = self.node_map[start]
         # optimization: we know the entry node of the automaton, results in smaller table
-        defining: list[Expression] = [Table([arr[0], state_vars[0]], [[v, e] for s, v, e in transitions if s == id_start])]
+        cons: list[Expression] = [Table([arr[0], state_vars[0]], [[v, e] for s, v, e in transitions if s == id_start])]
         # define the rest of the automaton using transition table
-        defining.extend(Table([state_vars[i - 1], arr[i], state_vars[i]], transitions) for i in range(1, len(arr)))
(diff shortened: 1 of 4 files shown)


IgnaceBleukx:
Oh wow the alldiff except n bug is extremely tricky... Its a combination of many edge cases:
1. The test uses boolexprs as arguments (quite unlikely but supported so ok)
2. Another argument is `False` (again quite unlikely but supported by the type system)
3. The decomposition of alldiff does a `x == y` on every argument, and if you do `Boolexpr == False` we simplify this to `~Boolexpr` so thats where the not comes from, even if its not introduced explicitely...

Not sure how to fix this cleanly to be honest...

WoutPiessens:
The latest commit addresses the MDD bug that was found during fuzztesting: the invalid edges are now also built layer by layer, instead of directly going to the sink node (+ the defining/constraining constraints for the positively reified case are corrected, same as for Regular)

cpmpy/expressions/globalconstraints.py
```diff
@@ -1185,28 +1188,55 @@ def _reduce(self):
...
+
+        n = len(arr)
+        levels = self.levels.copy()
+
+        all_nodes = set(levels.keys())
+
+        dummy_nodes : list[str|int] = []
+        i = 0
+        while len(dummy_nodes) < n - 1:
+            dummy_nodes += [name] if (name := f"__dummy_{i}__") not in all_nodes else []
```
IgnaceBleukx: woah what is this beast??

I think this is way to overcomplicated...

You need a dummy node per level, so why not just create them on the fly in line 1220-1233? 
If you encounter a value in the domain for which you don't have an edge yet, you check if there is a dummy node for this level. If there is no node yet, make it and add an edge to the sink node?

Then you indeed still need the consistency edges to make a chain of dummy nodes, so thats fine to keep.

I suppose the issue you are trying to solve is how to name the new dummy node? The current approach does not guarantee that in any way either?
Maybe the first thing the decomposition should do is to normalize the node labels to just be 0...n, then you can make the label for the new dummy node n+1?

cpmpy/expressions/globalconstraints.py
```diff
@@ -1185,28 +1188,52 @@ def _reduce(self):
...
+                if v not in extended_mapping[id1]:
+                    extended_mapping[id1][v] = dummy_nodes[level]
+                    if level not in necessary_levels:
+                        necessary_levels.append(level)
                     invalid_edges.add((id1, v))
 
-        return extended_mapping, invalid_edges
+        if necessary_levels:
+            first_needed = min(necessary_levels)
+            for level in range(first_needed + 1, n):
```
IgnaceBleukx, line 1229: I'm a bit surprised that you don't iterate over necessary levels? But I suppose you actually need a dummy node for all layers after the first one that needs a dummy...

If this is indeed the case, please add a comment about it here

IgnaceBleukx (approved):
Great, some pretty critical bugs fixed in this PR!
Small comment about documentation, otherwise ready to be merged.

cpmpy/expressions/globalconstraints.py
WoutPiessens: Yes, this is indeed the case, the layer at which you 'diverge' from the positive edges you have to construct the negative edges layer by layer (in order to always admit a solution to the flow problem, otherwise flow constraints cannot be defining). I'll add a comment.

PR #1040, opened 2026-07-06
Flatten wsum remove inplace edits

ThomSerg:
The flattening of wsum was making in-place update to the original wsum argument expressions, which we promise our transformations will never do (always copy). Causes bugs when other expressions re-used wsum's arguments.


IgnaceBleukx (approved):

IgnaceBleukx:
This probably existed in the code base for a very long time already, but we are now much more careful when to copy an expression in all of our transformations; so indeed some original expression objects may reach flatten : )

PR #1041, opened 2026-07-07
Trivial decompose

OrestisLomis:
In some edgecases a global constraints would decompose into 0 constraints (e.g. AllDifferent with 0/1 elements). These 0 constraints would then be conjoined, which would lead to a crash. This fixes that.


IgnaceBleukx (approved):
Yes, I suppose ideally this would be handled in the global constraint itself (e.g., AllDiff(x)) should just raise a ValueError...

But this patch is fine to go in

PR #1042, opened 2026-07-07
Fix sat boolval clause

OrestisLomis:
Eliminate BoolVal creation for linearize_constraint which is used after simplify_boolean

### DIFF
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -215,13 +215,13 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum","->"}, reified=Fal
             lhs, rhs = cpm_expr.args
 
             if lhs.name == "sum" and len(lhs.args) == 1 and isinstance(lhs.args[0], _BoolVarImpl) and "or" in supported:
-                # very special case, avoid writing as sum of 1 argument
+                # very special case, avoid writing as sum of 1 argument, write as disjunction of 1 arg instead
                 new_expr = simplify_boolean([eval_comparison(cpm_expr.name,lhs.args[0], rhs)])
                 assert len(new_expr) == 1
                 if isinstance(new_expr[0], BoolVal):
                     if new_expr[0].value() is True:
                         continue # skip or([BoolVal(True)])
-                    newlist.append(new_expr[0])  # post BoolVal(False) directly
+                    newlist += linearize_constraint([BoolVal(False)], supported=supported, csemap=csemap)  # don't wrap in or(BoolVal(False))
                     continue
                 newlist.append(Operator("or", new_expr))
                 continue
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -221,7 +221,7 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum","->"}, reified=Fal
                 if isinstance(new_expr[0], BoolVal):
                     if new_expr[0].value() is True:
                         continue # skip or([BoolVal(True)])
-                    newlist.append(new_expr[0])  # post BoolVal(False) directly, not or([BoolVal(False)]); parent -> handler folds (~bv -> false)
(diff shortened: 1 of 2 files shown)


cpmpy/transformations/linearize.py
```diff
@@ -218,8 +218,11 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum","->"}, reified=Fal
                 # very special case, avoid writing as sum of 1 argument
                 new_expr = simplify_boolean([eval_comparison(cpm_expr.name,lhs.args[0], rhs)])
                 assert len(new_expr) == 1
-                if isinstance(new_expr[0], BoolVal) and  new_expr[0].value() is True:
-                    continue # skip or([BoolVal(True)])
+                if isinstance(new_expr[0], BoolVal):
+                    if new_expr[0].value() is True:
+                        continue # skip or([BoolVal(True)])
+                    newlist.append(new_expr[0])  # post BoolVal(False) directly
```
IgnaceBleukx: Have a look at what we do just below on line 239: better to post the linear version of BoolVal(False), we might come up with a better way to handle False in the future, and this is such an uncommon case that it will not hurt performance.

tests/test_trans_linearize.py
```diff
@@ -100,6 +100,21 @@ def test_single_boolvar(self):
...
 
+    def test_trivially_false_bool_comparison_not_wrapped_in_or(self):
+        b0 = cp.boolvar(name="b0")
+        b1 = cp.boolvar(name="b1")  # reifies ((~b0) + -7 <= -6)
+        cons = only_implies(only_bv_reifies([((~b0) + -7 <= -6) == b1]))
+        lin = linearize_constraint(cons, supported={"or", "->", "sum", "wsum", "and"})
+        # comparison is trivially true for any b0, so only b1 is forced true; b0 is not posted
+        assert str(lin) == "[or(b1)]"
+
+    @pytest.mark.requires_solver("pindakaas", "pysat")
```
IgnaceBleukx: The test don't require pindkaas or pysat to run? Its just that the bug only affects them; remove annotation

cpmpy/transformations/linearize.py
```diff
@@ -218,8 +218,11 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum","->"}, reified=Fal
                 # very special case, avoid writing as sum of 1 argument
```
IgnaceBleukx: add ", write as disjunction of 1 arg instead"

cpmpy/transformations/linearize.py
IgnaceBleukx: also add comment "don't wrap in or(BoolVal(False))" instead? Using "post" can be confusing here as we might be handling nested stuff

IgnaceBleukx (commented):
Some small proposals, mainly doc related

tests/test_trans_linearize.py
```diff
@@ -100,6 +100,20 @@ def test_single_boolvar(self):
...
+        cons = only_implies(only_bv_reifies([((~b0) + -7 <= -6) == b1]))
+        lin = linearize_constraint(cons, supported={"or", "->", "sum", "wsum", "and"})
+        # comparison is trivially true for any b0, so only b1 is forced true; b0 is not posted
+        assert str(lin) == "[or(b1)]"
+
+    def test_reified_trivially_false_bool_sum(self, solver):
+        b0, b1 = cp.boolvar(name="b0"), cp.boolvar(name="b1")
+        model = cp.Model(((~b0) + -7 <= -6) == b1)
+        assert model.solve(solver=solver)
+        assert b1.value() is True
```
IgnaceBleukx, line 116: You should check that also b0 has a value here, it should be known to the solver.
Indeed, the transformations will remove it, as your comment in the above test indicates, but the solver should always know all user variables and give them a value after solve.

IgnaceBleukx (approved):
Need one more assert in the test, otherwise good to merge.

PR #1043, opened 2026-07-07
Fix implication with single bool in lhs

OrestisLomis:
An implication of the form `BV0 -> or(BV1)` would crash for HiGHS, this fixes that. Also made the cases a bit more explicit.

### DIFF
--- a/cpmpy/transformations/linearize.py
+++ b/cpmpy/transformations/linearize.py
@@ -151,6 +151,9 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum","->"}, reified=Fal
                             assert isinstance(lin, Comparison), f"Expected a comparison as rhs of implication constraint, got {lin}"
                             lin_lhs, lin_rhs = lin.args
                             if lin_lhs.name not in {"sum", "wsum"}:
+                                if isinstance(lin_lhs, _BoolVarImpl):
+                                    indicator_constraints.append(cond.implies(lin))
+                                    continue
                                 assert lin_lhs.name in supported, f"Unexpected lhs of rhs of implication: {cpm_expr}, it is not supported ({supported})"
                                 indicator_constraints.append(cond.implies(lin))
                                 continue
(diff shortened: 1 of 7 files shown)


IgnaceBleukx:
I don't think this is the right fix for the problem.
The key issue is that the recursive call of linearize does not return a sum of a single argument, and instead just returns BV >= 1.

A better fix is to change line 121 to explicitely make a sum operator (and also line 117 for 'ands' actually)
This ensures that the output of linearize is indeed a linear inequality constraint with a sum/wsum on the lhs.

OrestisLomis:
Yes, I was looking into this again and I agree. Working on a better fix (and clearer tests).

cpmpy/transformations/linearize.py
```diff
@@ -114,11 +114,11 @@ def linearize_constraint(lst_of_expr, supported={"sum","wsum","->"}, reified=Fal
         elif isinstance(cpm_expr, Operator) and cpm_expr.is_bool():
             # conjunction
             if cpm_expr.name == "and" and cpm_expr.name not in supported:
-                newlist.append(sum(cpm_expr.args) >= len(cpm_expr.args))
+                newlist += linearize_constraint([sum(cpm_expr.args) >= len(cpm_expr.args)], supported=supported, reified=reified, csemap=csemap)
```
IgnaceBleukx: I'm ok with duplicating the `Operator("<name>")` logic here, you don't really need th recursive call to linearize?

tests/test_constraints.py
```diff
@@ -128,13 +128,15 @@ def bool_exprs(solver):
...
+            arg_counts = [1, len(BOOL_ARGS)] # singleton and full
 
-        yield Operator(name, operator_args)
-        # Negated boolean values
-        yield Operator(name, [~ arg for arg in operator_args])
+        for n in arg_counts:
+            operator_args = BOOL_ARGS[:n]
+            yield Operator(name, operator_args)
+            # Negated boolean values
+            yield Operator(name, [~ arg for arg in operator_args])
```
IgnaceBleukx, line 139: Ah yes thats also an option, I was thinking of making them dedicated tests, but this better as it actually tests the transformation stack and process

IgnaceBleukx (approved):

PR #1044, opened 2026-07-08
Fix normalized boolexpr with const int rhs

OrestisLomis:
For cases where we had a global expression in reified constant in a disequality with a constant there was a bug because the current code expected the rhs to be a var. In the code comments the assumption is that after get_or_make_var the rhs is always a var, but this is not necessarily true, a constant stays a constant. I believe the get_or_make_var behaviour is correct so I patched the code in flatten_constraint to handle such cases with a constant.


IgnaceBleukx:
So simplify_bool actually already takes care of this problem, as it transforms `a -> boolexpr != 1` to `a -> ~boolexpr`.
`normalized_boolexpr` is a helper function of flatten, which first runs simplify bool on the input already.

So this issue cannot occur in the wild : )

PR #1045, opened 2026-07-14
Remove mul from supported for CPLEX

### LINKED ISSUE #764: Linearisation of integer multiplication

OrestisLomis:
Multiplication was not supported for CPLEX so it needs to be decomposed, fixes #764 (which I think was temporarily fixed, but broken again when mul became a global).

### DIFF
--- a/cpmpy/solvers/cplex.py
+++ b/cpmpy/solvers/cplex.py
@@ -467,7 +467,8 @@ class CPM_cplex(SolverInterface):
                         self.cplex_model.add_constraint(self.cplex_model.max(self.solver_vars(lhs.args)) == cplexrhs)
                     elif lhs.name == 'abs':
                         self.cplex_model.add_constraint(self.cplex_model.abs(self.solver_var(lhs.args[0])) == cplexrhs)
-                    elif lhs.name == 'mul':
+                    elif lhs.name == 'mul': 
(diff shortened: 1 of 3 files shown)


cpmpy/solvers/cplex.py
IgnaceBleukx, line 458: I think we should just add mul in this set, and keep it supported.
We know CPLEX does not support all multiplications, but thats ok imho, it would be a waste of decomposing all multiplications if only a few of them don't work

OrestisLomis: but mul is not an operator so we do need to keep it with the globals

IgnaceBleukx:
Ok  I think decomposing mul is the right way after all... It turns out only veeeeery few cases are actually working with cplex, so best to just decompose it all together.
Raised an error if we now get a multiplication, should have been linearized already

Also fixed a test that was failing for gurobi/CPLEX. Good to merge for me

Issue #336, opened 2023-06-01
.value() fonction of constraints (globalconstraint.py)


363734:
Some of the constraints return None for this if one of its args is evaluated as None (such as Maximum), some of them check partially if there is a value None (such as Cumulative which checks if one is None, but forgets to flatten the arguments that are list and thus may miss some None elements in it) and some don't do any checks at all (such as Xor) and would simply return an error if this arises. 

Should we uniform this and add full None checks for each of the expression .values() functions?

363734:
Also, I just notice, by looking at the .value() method of GlobalCardinalityCount constraint, that we could, for all the constraints where the decomposition function is available, do just as the GGC is doing (which calling .value() on the decomposition). If the None is well handled on operators and comparisons, I think for the constraint with a decomposition (not Min, Max, element, count) could just call the evaluation on the decomposition to get the value, since the decomposition use in fine only operators and comparison. This could then be set in the GlobalConstraint class and override for the one without a decomposition

JoD:
What do we mean with `None`? Since we only have total functions, `value()` will always be computable if all variables (leafs in the expression tree) have a value. So `None` covers the case where not all variables may have a value? In that case, it should be properly implemented in all `value()` calls.

Agree with the proposal to have GlobalConstrain provide a default `value()` implementation that works with the decomposition of the constraint. More efficient implementations can then be overridden in the subclass.

363734:
value() seems to be ok if all the variables are assigned. But if some are not, then there are inconsistencies. 

I just notice that the xor passed the work to a sum operator which indeed checks for the None

but Cumulative does not catch it (because there is no flattening of the lists mapped using argval). And if one of the start is None, then it will try to do None + something. So for me, it is not properly computed as of now. 

And I think refactoring it by evaluating the decomposition will first remove lots of redundant code, and secondly allow to have only basic operations where the None is checked beforehand.

IgnaceBleukx:
Calling the value on the decomposition will not work when that decomposition introduces auxiliary variables.
The new variables will not have a value (as they are not passed through any solve-call) so the `.value()` of the global constraint will always be `None`.
Moreover, creating new expressions in the decomposition, only to check for their value is really inefficient. For example, for the Sudoku Assistant, we first had the `.value()` of the `AllDifferent` constraint to check the pairwise-inequalities for their value. It turned out that during the computation of our explanations, more than 60% of time was spent in the `.value()` function of `AllDifferent`! 
[...]

PR #1047, opened 2026-07-14
globalconstraint base value function

OrestisLomis:
#336 mentions a throwing a notimplementederror for the base global constraint class, this PR does that.

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -170,6 +170,19 @@ class GlobalConstraint(Expression):
         """
         return True
 
+    def value(self) -> Optional[bool]:
+        """
+        Returns whether the global constraint is satisfied under the current variable assignment.
+
+        Returns:
+            Optional[bool]: True or False when all variables in the constraint are assigned;
+            None if any variable is unassigned.
+
+        Raises:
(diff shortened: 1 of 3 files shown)


cpmpy/expressions/globalconstraints.py
```diff
@@ -170,6 +170,19 @@ def is_bool(self) -> bool:
         """
         return True
 
+    def value(self) -> Optional[bool]:
+        """
+        Returns whether the global constraint is satisfied under the current variable assignment.
+
+        Returns:
+            Optional[bool]: True or False when all variables in the constraint are assigned;
+            None if any variable is unassigned.
```
IgnaceBleukx: Any variable within its scope is more precise

cpmpy/expressions/globalconstraints.py
```diff
@@ -170,6 +170,19 @@ def is_bool(self) -> bool:
...
+
+        Raises:
+            NotImplementedError: if the subclass does not implement ``value()``.
```
IgnaceBleukx: Remove this? The exception is obvious :)

IgnaceBleukx (commented):
Yes that makes sense to add. 
Please also add it for global functions?

Issue #1048, opened 2026-07-15
Remove deprecated code after breaking change


ThomSerg:
With the introduction of `__all__` and the other changes to `__init__` in #905, a change that has not been accounted for / has not been documented is the removal of the old lower-case globals (alldifferent, allequal, circuit) and other old names (BoolVar, IntVar, cparray) from CPMpy's namespace. This means that they can no longer be imported, i.e. the following no longer works:
```python
model += cp.alldifferent(...)
```
This is a breaking change and should be correctly documented as such. The code in question was already labeled as "deprecated", but with #905 it (unintentionally) is now a breaking change.

I think we should now take the opportunity to remove this old deprecated code from the codebase for the 1.0 release. Old existing user code is already broken due to the namespace change, so why still keep this code around?

IgnaceBleukx:
Can be closed now?

PR #1050, opened 2026-07-15
Remove old deprecated code

ThomSerg:
Removing old code that has been deprecated for a while. Some are still left in place due to
- being deprecated just now for the 1.0 release
- being deprecated for longer, but not correctly being labeled as such. Here we still need to give users time to make the switch.

Partially solves #1048. The removed code will be documented in the changelog of v1.0.0 and in the docs #1038

### DIFF
--- a/cpmpy/transformations/decompose_global.py
+++ b/cpmpy/transformations/decompose_global.py
@@ -42,6 +42,7 @@ class CustomDecomp(Protocol):
 def decompose_in_tree(lst_of_expr: list[Expression],
                       supported: Optional[AbstractSet[str]] = None,
                       supported_reified: Optional[AbstractSet[str]] = None,
+                      _toplevel=None, nested=False,
                       csemap: Optional[CSEMap] = None,
                       decompose_custom: Optional[Dict[str, CustomDecomp]] = None,
                       decompose_custom_positive: Optional[Dict[str, CustomDecomp]] = None,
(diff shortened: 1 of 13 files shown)


cpmpy/transformations/decompose_global.py
```diff
@@ -73,8 +70,6 @@ def decompose_in_tree(lst_of_expr: list[Expression],
     :func:`cpmpy.transformations.reification.reify_rewrite`
     E.g. ``bv -> NumExpr <comp> Var/Const`` will then be rewritten as  ``[bv -> IV0 <comp> Var/Const, NumExpr == IV0]``.
     """
-    assert _toplevel is None, "decompose_in_tree: argument '_toplevel' is deprecated, do not use/modify it"
```
IgnaceBleukx, line 76: For this one I'm not sure, we have not had a release yet where it is deprecated, so maybe wait one more?

cpmpy/transformations/safening.py
```diff
@@ -13,8 +13,6 @@
 
 
 def no_partial_functions(lst_of_expr: list[Expression],
-                        _toplevel: Optional[Any]=None,
```
IgnaceBleukx, line 16: Same here, we haven't had a release yet where this had a warning

IgnaceBleukx (commented):
Nice cleanup, I would keep the deprecated args to the transformation functions for a next release, we haven't had a release yet where the user was warned about them with a proper error message?

cpmpy/transformations/decompose_global.py
ThomSerg: I added the arguments back, but with a warning instead of an assert (since the latter would be an equal sudden breaking change to removing the argument)

OrestisLomis:
Are many of the examples not broken now? I should resume my clean up of those..

IgnaceBleukx:
I can confirm all examples run, so should be good to merge? (also very nicely documented in the changelog btw : ) )

PR #1052, opened 2026-07-17
Remove xfailing test for Z3

IgnaceBleukx:
Z3 had a new release, about 5h after I merged #1049 ...
So this PR removes the xfail from the test.

Not sure if we should add the min version in the supported check of the z3 interface, its not that we really rely on features in the new z3 version, its just that the result might be wrong due to internal bugs...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ with open("README.md", "r", encoding="utf8") as readme_file:
 
 solver_dependencies = {
     "ortools": ["ortools>=9.3.10497,!=9.9.*,!=9.10.*,!=9.11.*"], # exclusion due to bug #191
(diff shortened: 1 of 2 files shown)


ThomSerg:
I would say to not include a version check in supported, to still allow users to use a non-supported version at their own risk (to not hard-lock it). But indeed maybe we should adopt giving warnings to prompt users to update their solvers if below lower limit to make them aware they might encounter issues. Could be useful for those who update CPMpy but not their optional solver deps (pip auto-updates hard deps like OR-Tools based on version compatibility checks, but not the optional deps of the other solvers)

ThomSerg (approved):

PR #1056, opened 2026-07-17
XCSP3 ne parser fix

### LINKED ISSUE #1027: Parser error in xcsp3
This instance raises an error when the trying to parse with the current xcsp3 parser.
`CSP25/RotatingRostering-008-2-3_c25.xml.lzma`

Not sure which constraint is causing it

ThomSerg:
Closes #1027 

'ne' is an n-ary operator

### DIFF
--- a/cpmpy/tools/xcsp3/parser_callbacks.py
+++ b/cpmpy/tools/xcsp3/parser_callbacks.py
@@ -87,7 +87,9 @@ class CallbacksCPMPy(Callbacks):
         "le": (2, lambda x, y: x <= y),
         "ge": (2, lambda x, y: x >= y),
         "gt": (2, lambda x, y: x > y),
-        "ne": (2, lambda x, y: x != y),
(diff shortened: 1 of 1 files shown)


cpmpy/tools/xcsp3/parser_callbacks.py
```diff
@@ -87,7 +87,9 @@ def ctr_intension(self, scope: list[Variable], tree: Node):
         "le": (2, lambda x, y: x <= y),
         "ge": (2, lambda x, y: x >= y),
         "gt": (2, lambda x, y: x > y),
-        "ne": (2, lambda x, y: x != y),
+        # ne is n-ary in XCSP3 (like eq): ne(x0,...,xn) is the logical negation of eq, i.e. "not all equal"
+        # (not pairwise AllDifferent). Binary case stays a plain disequality.
```
IgnaceBleukx: You mean "not allequal"? Maybe we should just post is as that as well? `~cp.AllEqual(x)`

PR #1057, opened 2026-07-17
Flatten table args for pumpkin

ThomSerg:
Fix originating from cecbbc1c5. Flatten nested expressions from `InDomain` before posting to Pumpkin's table.


IgnaceBleukx:
But... `InDomain(sum(x,y,z), [1,5,7])` is not flat? Such constraints should never reach the Pumpkin interface..

ThomSerg:
So the issue is here then? 

https://github.com/CPMpy/cpmpy/blob/7e6a56dadc05f81a17c7d1ddcd393b53bb85418a/cpmpy/expressions/globalconstraints.py#L1409

should `has_subexpr` be true as soon as `expr` is not a variable?

IgnaceBleukx:
Yes indeed, we fixed it for Tables here #1029 but apparently I forgot to check InDomain

PR #1058, opened 2026-07-17
XCSP3 small parser fixes

ThomSerg:
Two small fixes to the xcsp3 parser coming from bd9b6a723 and 7e72cc896

### DIFF
--- a/cpmpy/tools/xcsp3/parser_callbacks.py
+++ b/cpmpy/tools/xcsp3/parser_callbacks.py
@@ -805,12 +805,7 @@ class CallbacksCPMPy(Callbacks):
         return cpm_exprs
 
     def get_cpm_var(self, x):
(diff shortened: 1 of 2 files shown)


cpmpy/tools/xcsp3/parser_callbacks.py
```diff
@@ -805,7 +805,9 @@ def exprs_from_node(self, node):
         return cpm_exprs
 
     def get_cpm_var(self, x):
-        if isinstance(x, XVar):
+        # also map non-XVar keys
```
IgnaceBleukx: So this function is now a fancy "self.cpm_variables.get(x,x)"?

IgnaceBleukx (approved):
Small cleanup possible, otherwise fine

PR #1059, opened 2026-07-17
Dynamic dispatch `decompose_linear`

ThomSerg:
`decompose_linear` should use dynamic dispatch so that globals inheriting from other globals can still overwrite their linear decomposition.

Fix coming from 57349a829


IgnaceBleukx:
Good, but ideally just adopt this pattern for all globals?

PR #1063, opened 2026-07-19
Add example for packing solver logos

IgnaceBleukx:
A fun example showing a 2d packing problem for tightly packing the solver logos in CPMpy


examples/packing_logos.ipynb
```diff
@@ -0,0 +1,274 @@
...
+        "    m = cp.Model()\n",
+        "\n",
+        "    # No-overlap: every logo fully above, below or next to every other logo\n",
+        "    for i, j in all_pairs(range(n)):\n",
+        "        m += ((pos_x[i] + widths[i] <= pos_x[j]) |\n",
+        "              (pos_x[j] + widths[j] <= pos_x[i]) |\n",
+        "              (pos_y[i] + heights[i] <= pos_y[j]) |\n",
+        "              (pos_y[j] + heights[j] <= pos_y[i]))\n",
+        "\n",
+        "    # Keep the packing roughly square (avoid long strips)\n",
```
OrestisLomis, line 70: These two constraints are a bit weird to me.. they don't really speed up solving and the solution without them is exactly the same (ignoring symmetric logos). I would remove them.

examples/packing_logos.ipynb
```diff
@@ -0,0 +1,274 @@
...
+        "LOGOS = [\n",
+        "    (\"CPMpy\", \"cpmpy.png\"),\n",
+        "    (\"OR-Tools\", \"ortools.png\"),\n",
+        "    (\"Z3\", \"z3.png\"),\n",
+        "    (\"MiniZinc\", \"minizinc.png\"),\n",
+        "    (\"GCS\", None),\n",
+        "    (\"Gurobi\", \"gurobi.png\"),\n",
+        "    (\"PySAT\", \"pysat.png\"),\n",
+        "    (\"PySDD\", None),\n",
+        "    (\"Exact\", None),\n",
```
OrestisLomis, line 157: Exact does have a logo, it's two cutting planes forming an X and spelling out exact. You can get a low quality version from the GitLab, though I couldn't find a high quality version anywhere myself so we could ask Jo. I guess my bigger concern is that this playful example might have some 'political' implications especially if we keep adding solvers as to why specifically this set of solvers was chosen to be displayed here. Not sure if the concern is valid tbh...

examples/packing_logos.ipynb
```diff
@@ -0,0 +1,274 @@
...
+        "    pos_y = cp.intvar(0, area_max_y, shape=n)\n",
+        "\n",
+        "    total_x = cp.max([x + w for x, w in zip(pos_x, widths)])\n",
+        "    total_y = cp.max([y + h for y, h in zip(pos_y, heights)])\n",
+        "\n",
+        "    m = cp.Model()\n",
+        "\n",
+        "    # No-overlap: every logo fully above, below or next to every other logo\n",
+        "    for i, j in all_pairs(range(n)):\n",
+        "        m += ((pos_x[i] + widths[i] <= pos_x[j]) |\n",
```
OrestisLomis, line 65: since you're using OR-Tools as a solver you could actually use the 2D NoOverlap natively

examples/packing_logos.ipynb
```diff
@@ -0,0 +1,274 @@
...
+        "    # Keep the packing roughly square (avoid long strips)\n",
+        "    m += (total_x <= 2 * total_y)\n",
+        "    m += total_x <= total_y\n",
+        "\n",
+        "    # Hard-center the CPMpy logo (allow 1 unit slack for odd sizes)\n",
+        "    m += cp.abs(2 * pos_x[center_idx] + widths[center_idx] - total_x) <= 1\n",
+        "    m += cp.abs(2 * pos_y[center_idx] + heights[center_idx] - total_y) <= 1\n",
+        "\n",
+        "    # Minimize half-perimeter\n",
+        "    m.minimize(total_x + total_y)\n",
```
OrestisLomis, line 79: to note for my own curiosity, you do get a different solution when optimising for area

examples/packing_logos.ipynb
```diff
@@ -0,0 +1,274 @@
...
+        "    (\"Pumpkin\", \"pumpkin.png\"),\n",
+        "    (\"IBM CPLEX / CPO\", \"ibm.png\"),\n",
+        "    (\"Pindakaas\", \"pindakaas.png\"),\n",
+        "    (\"HiGHS\", \"highs.png\"),\n",
+        "    (\"Hexaly\", \"hexaly.png\"),\n",
+        "    (\"RC2\", None),\n",
+        "    (\"SCIP\", \"scip.png\"),\n",
+        "]\n",
+        "\n",
+        "LOGO_SIZE = 100       # longer side for solver logos\n",
```
OrestisLomis, line 168: Similar 'political' comment as before. If you just scale them such that the longest side has size 100, then you're making square logos look proportionally larger.. not sure if that's desired

OrestisLomis (commented):
I think it's a fun example, but since we are using logos of different institutions/corporations I do have some concern over how they are being displayed.. Not sure if the criticisms are valid tho. Some other minor comments on the choice of how to model an optimal packing problem.

IgnaceBleukx:
I think I'm just going to abandon this one, its indeed too tricky to keep the right version for each solver logo etc.

PR #1065, opened 2026-07-20
paramita (max)sat modeling library

tias:
homepage: https://ulog.udl.cat/static/doc/paramita/

passes test-suite and code looks alright at first glance

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -20,6 +20,14 @@ jobs:
           - solver: pindakaas
           - solver: paramita
             install_commands: |
+              # Ubuntu's wget (1.21.x) gets a 404 from bitbucket.org's edge for the
+              # ipasir.h/ipamir.h raw-file fetch in the EvalMaxSAT2022 plugin's CMake
+              # patch step, while curl succeeds against the same URL. Shadow wget with
+              # a curl-based shim so that fetch works on the CI runner.
+              mkdir -p /tmp/wget-shim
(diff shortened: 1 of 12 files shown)


tias:
To the paramita devs:

I'm now caching the Bool() objects as we discussed.

I also use model to get the model at once, but here I notice that it returns the dimacs_lits, as does the core. So I need one way to get the 'Bool()' but also a way to get the 'dimacs_lit' for a CPMpy variable... can I get the dimacs_lit from the Bool()?

And related to that, when I set assumption literals, I need to set the dimacs literal it seems, not the Bool() object, is that correct?

LogicOptimizationGroup:
Hi! This is Paramita devs.

> To the paramita devs:
> 
> I'm now caching the Bool() objects as we discussed.
> 
> I also use model to get the model at once, but here I notice that it returns the dimacs_lits, as does the core. So I need one way to get the 'Bool()' but also a way to get the 'dimacs_lit' for a CPMpy variable... can I get the dimacs_lit from the Bool()?

Probably you should use our [DimacsVariablePool](https://ulog.udl.cat/static/doc/paramita/reference/api_paramita_modelling/#paramita.modelling.DimacsVariablePool) for that.

> And related to that, when I set assumption literals, I need to set the dimacs literal it seems, not the Bool() object, is that correct?

Yes, it should be the literal, so you can recover that using the same DimacsVariablePool.

On Monday, we will release a new version of Paramita with some updates in the interfaces. The most important thing is that we added the benefit and cost methods to MaxSatSolver and NonIncrMaxSatSolver interfaces.

So it might be better to wait until we release this to merge the PR, but the changes that you mentioned should be OK for that new release too.

Thank you!

PR #1067, opened 2026-07-23
Sudoku dataset

IgnaceBleukx:
Implemeting the Scaled Sudoku dataset as a CPMpy dataset


cpmpy/tools/datasets/scaledsudoku.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        return parse_sudoku(instance, open=cls.open)
+
+    def categories(self) -> dict[str, Any]:
+        return {
+            "kind": self.kind,
+            "size": self.size,
+        }
+
+    def collect_instance_metadata(self, file: pathlib.Path) -> dict[str, Any]:
```
ThomSerg, line 99: Might it not also be useful to have access to the "solution_grid" in the metadata? Or maybe it should be returned by the parser? So returning a tuple of (puzzle, solution, ?walk?). 

It just seems like the dataset has a lot more data available on the instances with currently not a good way for the user to retrieve it.

ThomSerg: Like it feels like the sidecar file of the original dataset contains more information that our dataset object should also give access to.

ThomSerg (commented):
Should the sudoku's solution also be parsed / included as metadata? And what with the "walks"?

IgnaceBleukx:
Yeah it indeed makes sense to just pass the entire sidecar to the user in the metadata, modified it now.

I also noticed that if I want to set the root to "~/..." it does not get expanded correctly, that is fixed as well now by adding the root dir as an argument to the constructor. I had to do a bit of a hack for the MSE dataset to supply the subdirs, could you have another look at that?

Sudoku-bit is ready to go I think : )

ThomSerg:
Checked the `dataset_dir` part, it was a small oversight from me leaving it in the constructor like that, different from all the other datasets. But thanks to your great rewrite bringing some of the logic to the root class, all still works as intended; the user can still pass a more custom install location.

ThomSerg (approved):

PR #1068, opened 2026-07-24
Xcsp3 fixes

tias:
see individual commits

most also apply when using plain ortools; some from when trying to benchmark the paramita PR


ThomSerg (approved):

ThomSerg:
I applied some of the fixes to the parser / runner that currently are only present in branch `xcsp3_26`.

PR #1070, opened 2026-07-24
Hermax MaxSAT modeling library

IgnaceBleukx:
Interface to the HerMax MaxSAT library.

Supports CP-level constraints, so does not require linearize/int2bool.

Two open issues:
1. `__eq__` between two variables with disjoint domains raises a ValueError (see failing test). @josalhor I think it might be more natural to either return an empty clause instead, or return a specialized "trivial unsat" error or something. Then we can catch it at the CPMpy level and convert it to `BoolVal(False)` instead. I don' dare to catch ValueErrors as they can mean lots of things : ) 
2. Time limits are not supported as far as I can tell; I can of course interrupt the process with a Python Timer (as we do in PySAT), but HerMax does not provide a feasible solution in that case

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -19,6 +19,7 @@ jobs:
               sudo snap install minizinc --classic
           - solver: pindakaas
           - solver: pumpkin
+          - solver: hermax
           - solver: scip
           - solver: highs
     uses: ./.github/workflows/test-solver.yml
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ CPMpy can translate to a wide variety of constraint solving paradigms, including
 * **GO Solvers**: Hexaly (license required)
 * **SMT Solvers**: Z3
 * **PB Solvers**: Exact
+* **MaxSAT Solvers**: RC2 (via PySAT), Hermax + solvers
 * **SAT Encoders and Solvers**: PySAT+solvers, Pindakaas
 * **Decision Diagrams**: PySDD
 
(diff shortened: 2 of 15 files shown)


josalhor:
Hello :D

> __eq__ between two variables with disjoint domains raises a ValueError (see failing test).

I think you mean between a variable and a constant, is that correct?
Also, I see no failing test (?)

> Time limits are not supported as far as I can tell; I can of course interrupt the process with a Python Timer (as we do in PySAT), but HerMax does not provide a feasible solution in that case

After talking with other people in the community I think I know how to support this.
Would you be interested only in optimal solutions with time limit setting or also suboptimals?

I will do a review of the integration soon

josalhor:
> I think you mean between a variable and a constant, is that correct?

This has been fixed as of Hermax 1.2.4 uploaded today.

> Time limits are not supported as far as I can tell

Time limits are now supported as of 1.2.4. Also, I share more tails with you through email regard this matter.

tias:
Hi Jose,

The equality seems gone yes.

For the time limit, there seem to be two special cases:

* Hermax's live incremental SAT backend cannot interrupt. The current code falls-back to non-incremental SAT, which does support time-limit it seems. Could you also support it for the incremental one?
* One-shot timed MaxSAT (PortfolioSolver + RC2) can return "error" when the worker is killed at the limit without a model; we catch that and treat as UNKNOWN then.

josalhor:
Please try 1.2.5: https://hermax.readthedocs.io/en/latest/changelog.html
I think I fixed both things.

Regards,

PR #1072, opened 2026-07-31
Fix quickxplain hard unsat

OrestisLomis:
The implementation of quickxplain (and the original paper) assumes that the given hard constraints are satisfiable by themselves. I'm not sure if this is a fair assumption so with this change the algorithm also works without the assumption.


cpmpy/tools/explain/mus.py
```diff
@@ -127,6 +127,9 @@ def do_recursion(soft, hard, delta):
 
     # optimization: find max index of solver core
     solver_core = frozenset(s.get_core())
+    if not solver_core:
```
IgnaceBleukx, line 130: I like to be more explicit here: `if len(solver_core) == 0`

cpmpy/tools/explain/mus.py
```diff
@@ -275,7 +278,7 @@ def quickxplain_naive(soft, hard=[], solver="ortools"):
     def do_recursion(soft, hard, delta):
 
         m = cp.Model(hard)
-        if len(delta) != 0 and m.solve(solver) is False:
+        if m.solve(solver) is False:
```
IgnaceBleukx, line 281: I don't think you can delete this one? delta2 can be empty in the call on line 295
In the non-naive version the check for the lenght of solver core is outside the recursive call.

IgnaceBleukx (changes requested):

PR #1073, opened 2026-08-04
Rewrite Boolean comparisons to Boolean operators

IgnaceBleukx:
Improvement to flattening to avoid making some auxiliary variables.

I'm not conviced yet it should live in flatten, might be better to make it part of simplify_bool instead?

(Some int2bool tests still failing, will investigate)

### DIFF
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -471,6 +471,8 @@ class CPM_choco(SolverInterface):
                 return self.chc_model.and_(self.solver_vars(cpm_expr.args))
             elif cpm_expr.name == 'or':
                 return self.chc_model.or_(self.solver_vars(cpm_expr.args))
+            elif cpm_expr.name == 'not':
+                return self.chc_model.not_(self._get_constraint(cpm_expr.args[0]))
 
             elif cpm_expr.name == "->":
                 cond, subexpr = cpm_expr.args
--- a/cpmpy/solvers/choco.py
+++ b/cpmpy/solvers/choco.py
@@ -63,7 +63,7 @@ from ..transformations.flatten_model import flatten_constraint, get_or_make_var
 from ..transformations.comparison import only_numexpr_equality
 from ..transformations.linearize import canonical_comparison
 from ..transformations.safening import no_partial_functions, safen_objective
-from ..transformations.negation import push_down_negation
+from ..transformations.negation import push_down_negation, push_down_negation_objective
 from ..transformations.reification import reify_rewrite
 from ..exceptions import ChocoBoundsException, NotSupportedError
 
@@ -365,7 +365,7 @@ class CPM_choco(SolverInterface):
                                                supported=self.supported_global_constraints,
                                                supported_reified=self.supported_reified_global_constraints,
                                                csemap=self._csemap)
-
+        obj = push_down_negation_objective(obj)
         # make objective function non-nested
         obj_var, obj_cons = get_or_make_var(obj) # do not pass csemap here, we will still transform obj_var == obj...
 
--- a/cpmpy/solvers/pumpkin.py
+++ b/cpmpy/solvers/pumpkin.py
@@ -56,7 +56,7 @@ from ..transformations.normalize import toplevel_list
 from ..transformations.decompose_global import decompose_in_tree, decompose_objective
 from ..transformations.flatten_model import flatten_constraint, get_or_make_var
 from ..transformations.comparison import only_numexpr_equality
-from ..transformations.negation import push_down_negation
+from ..transformations.negation import push_down_negation, push_down_negation_objective
 from ..transformations.reification import reify_rewrite, only_bv_reifies, only_implies
 from ..transformations.safening import no_partial_functions, safen_objective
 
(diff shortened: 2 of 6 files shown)


IgnaceBleukx:
Also updated a few other things related to negation which is different in the journal (and updated the grammar at the top of the flatten file. 
Should be good to review now

IgnaceBleukx:
Ok so this PR seems to be blocked by a bizarre performance bug in GCS.
I can reproduce it on the current pip-installable version locally too; but installing from source seems to work fine (commit 8b2cc112 on gcs repo).
 
The model where the tests currently stop is:
```python
import cpmpy as cp

p = cp.boolvar(name="p")
start = cp.intvar(0, 10, shape=3, name="start")

cons = (p <= cp.Cumulative(start, [1, 4, 3], demand=[4, 5, 7], capacity=10)) == p
assert cp.Model(cons).solve(solver="gcs", time_limit=15) # timeout
```
where the nested cumulative is decomposed using the time-decomposition.

Maybe @ciaranm or @mmcilree might know what has changed since the last release of gcs that could have caused/fixed this?
In any case, the CPMpy CI will probably keep timing out until the next release of gcs, unless we start installing from source there too : )

ciaranm:
Can confirm we can reproduce this, and that it's not Python-related. I've got Claude narrowing it down, but it looks like the cause might be that we fixed a propagator bug in reified comparisons that's then leading to a much larger search tree due to heuristic things.

ciaranm:
(Ciaran's Claude here — he set me looking at this in detail. Everything below is
measured on his machine; he's read it before posting.)

We can reproduce this, and there are two separate things going on — one of them
is ours, but fixing it won't be enough on its own.

**First: the model in the comment above isn't the one that hangs.** It drops the
`end` array, and that's the load-bearing part — as written it solves in 25 nodes
on the released gcspy. The case your suite actually stops on is the one
`test_constraints.py` generates with explicit ends:

```python
p = cp.boolvar(name="p")
s = cp.intvar(0, 10, shape=3, name="start")
e = cp.intvar(0, 10, shape=3, name="end")
cons = (p <= cp.Cumulative(s, [1, 4, 3], e, [4, 5, 7], 10)) == p
```

Running your branch's suite here against `pip install gcspy` stalls at the same
96% mark the CI job does, and that's the test. It isn't a hang — it's an honest
search, which is why `time_limit` doesn't rescue it: the solve simply doesn't
finish.

**Second: there is a real gcs bug in 0.1.9, and it is exactly what your PR
started tripping over.** Our reified comparison propagator only enforced one
direction when the reification is *false*: `b <-> (x >= 1)` with `b` false left
`x` at [0..10] instead of pinning `x = 0`. Your old flattening emitted
`b <-> (start <= t)` and asserted it positively; the new one emits
`b <-> (start >= t+1)` and asserts `~b`, which is precisely the broken
direction. So every one of the ~66 time-point Booleans stopped propagating back
to `start`/`end`, and the search enumerates the auxiliaries instead. On the
model above that's 48.7M nodes versus 25.

That's fixed on our main (commit 4cca9fc6, back in March) — which is why a
source install works for you. We'll get a release out; @mmcilree owns the PyPI
side.

**But a release won't make this test fast, and that's the part worth talking
about.** On current gcs main, *both* flattenings take ~7.08M nodes on this
model. The old one only looked fine because it won a coin toss: every auxiliary
Boolean here has domain size 2, so our default `dom_then_deg` picks the root
variable purely on degree, and `p` used to edge out the `and` reification
variable. We changed how degree is counted in July and the tie flipped — same
propagation either side, 25 nodes versus 7.08M.

Numbers on current main, same model, varying only the branching:

| branching | nodes | time |
|---|---|---|
| default (`dom_then_deg` over all 135 variables) | 7,085,707 | 5.6s |
| `dom_wdeg` over all 135 variables | 93,212 | 89ms |
| decision variables (`p`, `start`, `end`) first | **6** | 0ms |

So the honest diagnosis is that this model is hypersensitive to variable
ordering, and right now the ordering is left entirely to chance: gcspy's
`solve()` calls into the solver with no search specification at all, so every
CPMpy model gets `dom_then_deg` + smallest-value over *every* variable —
including the ~130 auxiliaries your flattening introduces. CPMpy knows which
variables are the user's and which are flattening artefacts; the solver doesn't.

The durable fix is for us to expose branching in gcspy and for the CPMpy
backend to pass the model's decision variables. On our side that's
`branch_sequence(dom_then_deg(decision_vars), dom_then_deg(everything))` — the
fallback keeps it complete if any auxiliary is left unfixed — and it's the
6-node row above. I'll open an issue for the binding; happy to do the CPMpy side
too if you'd rather not.

In the meantime, `--solver gcs` on this test will pass once we release, just
slowly.

One aside: your `minizinc` lane is also sitting at 6h on this run. I haven't
checked whether it's the same test, but if it is, the sensitivity isn't
gcs-specific and specifying a search is worth doing regardless.

IgnaceBleukx:
Hi Ciaran,

Thanks for the extensive report!
So indeed, we'll wait for a new gcs release to merge this PR; regarding search-orders: this is a bigger point in CPMpy in general, as we currently don't have "CPMpy native" way of setting the search order by the user.
Although for GCS, we could indeed hard-code some search strategy in the interface (e.g., by first branching over the decision vars, and then aux)

PR #1075, opened 2026-08-08
typing abs and mult

tias:


cpmpy/expressions/globalfunctions.py
```diff
@@ -210,7 +210,7 @@ def __init__(self, arg_list: ListLike[ExprLike]):
     
     @property
     def args(self) -> tuple[int|Expression, ...]:
-        """ READ-ONLY, well-typed argument of this global function, no numpy ints"""
+        """ READ-ONLY, well-typed argument of this global function"""
```
IgnaceBleukx, line 213: I actually like the "no np ints" bit in the original docstring, or do we want to streamline it for all operators/globals? Then it indeed makes sense to just put this

cpmpy/expressions/globalfunctions.py
```diff
@@ -388,64 +393,82 @@ def __init__(self, x: ExprLike, y: ExprLike):
...
-            (x, y) = (y, x)
-            is_lhs_num = True
-        elif isinstance(x, (float, np.floating)):
-            raise TypeError(f"Multiplication does not support float constants, got: {x}")
-        elif isinstance(y, (float, np.floating)):
-            raise TypeError(f"Multiplication does not support float constants, got: {y}")
-
-        super().__init__("mul", (x, y))
+            if not isinstance(x, int):
+                x = int(x)  # type: ignore  # it can't see we're removing the np.integers
```
IgnaceBleukx: Would annotating "is_int" with a typeguard solve this?

cpmpy/expressions/globalfunctions.py
```diff
@@ -388,64 +393,82 @@ def __init__(self, x: ExprLike, y: ExprLike):
...
+        else:
+            if not isinstance(x, Expression):
+                raise TypeError(f"Multiplication does not support float constants, got: {type(x)}")
```
IgnaceBleukx: Not sure about mentioning floats here: it can be anything at this point...
Alternative: `f"Multiplication only supports integers or Expressions, but got {x} of type {type(x)}"`

Or if you really want the float-relatred stuff in there:
```
f"Multiplication only supports integers or Expressions, but got {x} of type {type(x)}, to use floats in a weighted sum, use the `FloatSum` object. (<point to docs>)"
```

cpmpy/expressions/globalfunctions.py
```diff
@@ -388,64 +393,82 @@ def __init__(self, x: ExprLike, y: ExprLike):
...
+        super().__init__("mul", args)
         self.is_lhs_num = is_lhs_num
+    
+    @property
+    def args(self) -> tuple[int|Expression, Expression]:
+        """ READ-ONLY, well-typed argument of this global function"""
+        return self._args
 
-    def update_args(self, args):
+    def update_args(self, args: Iterable[int|Expression], has_subexpr: Optional[bool] = None):
```
IgnaceBleukx: Its a bit surprising that update_args has stricter types then the constrctor? Although we only call it during our transforms so maybe its fine

cpmpy/expressions/globalfunctions.py
```diff
@@ -388,64 +393,82 @@ def __init__(self, x: ExprLike, y: ExprLike):
...
+    def __repr__(self) -> str:
         x, y = self.args
 
         if self.is_lhs_num:
             return "{} * ({})".format(x, y)
 
         return "({}) * ({})".format(x, y)
 
-    def __neg__(self):
+    def __neg__(self) -> "Expression":
```
IgnaceBleukx: this does not match superclass type (untyped), is that an issue?

IgnaceBleukx (commented):
Mostly small comments, there is some room for improvements on the decomposition for sure, but out of scope for this PR : )

cpmpy/expressions/globalfunctions.py
tias: indeed, the default will be no numpy ints, and it is what the type signature below says, no need to overspecify the comment

cpmpy/expressions/globalfunctions.py
tias: I'll try

cpmpy/expressions/globalfunctions.py
tias: right, it should have been generic, will fix

cpmpy/expressions/globalfunctions.py
tias: you are right... subclass args should only be more general... hmm, will investigate

cpmpy/expressions/globalfunctions.py
tias: untyped is ignored, I will remove this from this PR, its for after we type the superclass, good catch

cpmpy/expressions/globalfunctions.py
```diff
@@ -416,7 +416,7 @@ def args(self) -> tuple[int|Expression, Expression]:
         """ READ-ONLY, well-typed argument of this global function"""
         return self._args
 
-    def update_args(self, args: Iterable[int|Expression], has_subexpr: Optional[bool] = None):
+    def update_args(self, args: Iterable[Any], has_subexpr: Optional[bool] = None):
```
IgnaceBleukx, line 419: At some point, we will probably want a helper function that processes the args, which is called from both the constructor and the update_args() function; now we have to duplicate the code (which is ok for now as we only have  a couple of classes that have a custom update_args func

IgnaceBleukx (approved):
Great, ready to go in

cpmpy/expressions/globalfunctions.py
tias: Yes, a bigger refactor would be to have an explicit copy constructor (a mypyc compatible one) instead of the generic update args, which probably requires such a helper function.

Issue #1077, opened 2026-08-10
Variadic constraints (non) trivial arity


ThomSerg:
After merging master into the CVC5 branch to bring it up-to-date, new test suddenly started failing. Somewhere in the transformations things changed that caused 1-ary "And"s and "Or"'s to reach the solver, which CVC5 doesn't support (expects arity of at least 2, since the 1 cases are trivial). Looking into this, some other stuff came up, all related to variadic globals (globals that accept a variable amount of arguments like `AllDifferent`). Some cases that we don't properly define / check / handle.

- [x] Globals that are ill defined for 0-arity but still allow for 0 aguments: `AllDifferent`, `AllDifferentExceptN`,  `AllEqual`, `AllEqualExceptN`, `Increasing`, `Decreasing`, `IncreasingStrict`, `DecreasingStrict`, `Xor` (missing check in constructor)
- [x] Globals that do support 0-arity but aren't handled: `And`, `Or` : per definition, the truth value in 0-arity should be the identity element (and->True, or->False)
- [ ] 1-arity operator simplification: currently handled in the solver interface (`And`,  `Or`, see Z3) -> why not part of `simplify_boolean`?  And(x) -> x, Or(x) -> x. (I have code ready for this)
- [ ] Globals that don't constrain for 1-arity: `AllDifferent`, `AllDifferentExceptN`,  `AllEqual`, `AllEqualExceptN`, `Increasing`, `Decreasing`, `IncreasingStrict`, `DecreasingStrict`, `Xor` -> call decompose? not post? enforce arity >= 2? what with impact on solution count?

This so far covers the `Operators` and the `GlobalConstraints` (what I found so far), there will be similar things for `GlobalFunctions` (min, max, nvalue, ... a lot of constructors with hidden arity assumptions). To be expanded upon.

So a discussion starter on how to handle these cases and a place to track progress.

PRs:
- [ ] Arity checks for globals and operators #1078 
- [ ] Simplify trivial arity And/Or #1080

IgnaceBleukx:
Hmm I see, my take would be:

- Ill-defined globals for 0 args: indeed we just need to add a check in the constructor that raises a ValueError, we probably have something similar for Operator("sum", []) (didn't check it)
- I suppose you mean operators? For me they can also just error out, it is an indication something is going wrong... For user-interface functions like cp.all and cp.any we already handle this properly I believe (though there could be an inconsistency betweeen False/BoolVal(False))
[...]

PR #1078, opened 2026-08-10
Arity checks

ThomSerg:
Part of #1077 

Adds missing arity checks to some global functions and constraints and some cleanup for existing checks in operators.

Whilst technically / logically some constraints allow for 0-arity, this is most likely an indication of user error. We thus chose to raise an error to catch these mistakes instead of supporting these edge-cases (which don't add any modeling expressiveness). Some user-facing apis do still support them, like `cp.all([])`, but the model-facing `cp.Operator("And", [])` is not.

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -729,8 +729,8 @@ class Operator(Expression):
     it is stored first (as expr[0]), this eases weighted sum detection
 
     N-ary operators (``and``, ``or``, ``sum``) require at least one argument (arity >= 1).
-    Empty argument lists are rejected: while logically plausible, they usually indicate a
-    construction / user mistake (e.g. unexpectedly empty input).
+    ``wsum`` likewise requires at least one term. Empty argument lists are rejected: while
+    logically plausible, they usually indicate a construction / user mistake (e.g. unexpectedly empty input).
     """
     allowed: Final[dict[str, tuple[int, bool]]] = {
         #name: (arity, is_bool)       arity 0 = n-ary, min 1
@@ -794,7 +794,12 @@ class Operator(Expression):
                     weights.append(int(a)) # bool or int, simplifies things later on
                 else:
                     weights.append(a) # can be float
-            arg_list = (weights, arg_list[1])
+            exprs = arg_list[1]
+            if len(weights) != len(exprs):
+                raise ValueError(f"Operator 'wsum' expects equal lengths, got {len(weights)} weights and {len(exprs)} expressions")
+            if len(weights) == 0:
+                raise ValueError("Operator 'wsum' must be given at least one term")
+            arg_list = (weights, exprs)
 
         # small cleanup: nested n-ary operators are merged into the toplevel
         # (this is actually against our design principle of creating
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -727,9 +727,13 @@ class Operator(Expression):
 
     Convention for 2-ary operators: if one of the two is a constant,
     it is stored first (as expr[0]), this eases weighted sum detection
+
+    N-ary operators (``and``, ``or``, ``sum``) require at least one argument (arity >= 1).
+    Empty argument lists are rejected: while logically plausible, they usually indicate a
+    construction / user mistake (e.g. unexpectedly empty input).
     """
     allowed: Final[dict[str, tuple[int, bool]]] = {
-        #name: (arity, is_bool)       arity 0 = n-ary, min 2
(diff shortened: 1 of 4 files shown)


ThomSerg:
"and" or "or" with 0-arity can still occur in `simplify_boolean`, so we should return respectively `BoolVal(False)` and `BoolVal(True)` there instead of constructing an empty operator. Will be part of another PR.

ThomSerg:
Added some more checks, enforcing more arities in globals. But some of these might come a bit closer to becoming breaking changes, eg `LexLess` did correctly handle the 0-arity case in its decompose. But in general I don't think we can rely on the solver to support this edge case. Also it might be again an indicator of modelling error, so best to throw exception?

E.g. for `Regular`, the minizinc backend and the decompose handle 0-arity, but OR-Tools and Choco (probably more) don't allow it in their api.

Haven't added tests verifying constructor validation, seems better to do this properly in #870.

tias:
"Some user-facing apis do still support them, like cp.all([]), but the model-facing cp.Operator("And", []) is not."

that is the right design choice: user facing API's can accept more sometimes, but the objects we create should be meaningful. (it slightly contradicts with creating objects as specified by the user, but there are also other trivialities that we rewrite, so its fine)

tias (approved):
Great!

Added 2 more commits:
- make element check consistent
- table check is on variables, docstring was on nr rows (do you also want to check nr rows?)

auto-review also highlighted that Count/Among/GCC still allow [], which might be intentional?

ThomSerg:
For the constraints you list, skipping them wasn't intentional.

I would argue that they also need checks. e.g. a `Count(arr,val)` with an empty arr is just 0, same with empty val. Both of them seem more like user mistakes than an intentionally modeled expression. Same with `Among`, why create the expression if you're not going to pass it variables?

Will add those missing ones.

ThomSerg:
So apparently there are a bunch more edge cases. For example, we currently support `InDomain` with an empty domain, which is trivially False. Do we keep supporting this, or do we report this as a user modeling error and throw an exception?

1) One approach would be to just look at what OR-Tools accepts and replicate that. 
2) We could look at more than one solver, and see what they overall support and agree on.
3) A different approach would be to try to detect all possible user modeling errors and thus be more strict with the allowed expression arguments. e.g. OR-Tools accepts `AllDifferent([])`, but this most likely indicates a mistake.

I currently went for option 3, we can always roll back some of the commits.

An overview (A=Accept, R=Reject):
| Global | OR-Tools | Choco | MiniZinc | CPMpy | Suggested if OR-Tools leads |
|---|---|---|---|---|---|
| AllDifferent | A-true | R | A | **R** | A-true |
| AllDifferentExceptN/0 | — | R | A | **R** | R (Choco) |
| AllEqual / ExceptN | — | R | A | **R** | R (Choco) |
| Xor | A-unsat | — | — | **R** | A-unsat *or* R |
| Circuit | R | R | A | **R** | R |
| Inverse | R | R | A | **R** | R |
| Table vars | R | R | — | **R** | R |
| Table rows | A-unsat | A | type err | **R** | A-unsat |
| NegativeTable vars | R | R | — | **R** | R |
| NegativeTable rows | A-true | A | — | **R** | A-true |
| Regular array | R | — | — | **R** | R |
| Regular transitions | R | crash | — | **R** | R |
| Regular accepting | R | — | — | **R** | R |
| MDD array/trans | — | needs MDD | — | **R** | R |
| Cumulative | A-true | R | A | **R** | A-true |
| CumulativeOptional | (same family) | — | — | **R** | A-true |
| NoOverlap | A-true | — | A | **R** | A-true |
| NoOverlapOptional | (same) | — | — | **R** | A-true |
| Increasing* / Decreasing* | — | R | A | **R** | R |
| LexLess(Eq) | — | R | A | **R** | R |
| LexChain* | — | R | — | **R** | R |
| Precedence vars | — | R | A | **R** | R |
| Precedence values | — | A | — | **R** | A *or* R |
| GCC vars/vals | — | A | A | **R** | A-true / A-0 |
| InDomain arr | A-unsat | A | A | **R** | A-unsat |
| Minimum / Maximum | A-unsat | A | A | **R** | A-unsat *or* R |
| Element / NDElement | R | R | — | **R** | R |
| Count | — | A | A | **R** | A-0 |
| Among arr/vals | — | A | A | **R** | A-0 |
| NValue / Except | — | R | A | **R** | R |
| IfThenElse / Abs / Mul / Div / Mod / Pow | n/a | n/a | n/a | n/a | n/a |
| `and` / `or` Operator | A-true / A-unsat | R | — | **R** | see builtins |
| `sum` / `wsum` | 0 / — | R | — | **R** | R for Operator; builtin sum → 0 |

PR #1080, opened 2026-08-10
Simplify trivial arity And/Or

ThomSerg:
Related to #1077 

Adds cases for 0 and 1 arity versions of `And` and `Or` to `simplify_boolean`.
Returning either a trivial value or the single argument on its own

| Before | After |
| - | - |
| Or([]) | BoolVal(False) |
| And([]) | BoolVal(True) |
| Or([b]) | b |
| And([b]) | b |

### DIFF
--- a/cpmpy/transformations/normalize.py
+++ b/cpmpy/transformations/normalize.py
@@ -125,7 +125,7 @@ def _simplify_boolean_expr(expr: Expression, num_context=False) -> tuple[bool, E
                     newargs.append(a)
 
             if newargs is None:
-                newargs = args
+                newargs = list(args)
 
             if len(newargs) == 0: # empty disjunction
                 return True, BoolVal(False)
@@ -150,7 +150,7 @@ def _simplify_boolean_expr(expr: Expression, num_context=False) -> tuple[bool, E
                     newargs.append(a)
 
             if newargs is None: # when args where simplified recursively above
-                newargs = args
+                newargs = list(args)
 
             if len(newargs) == 0: # empty conjunction
                 return True, BoolVal(True)
--- a/cpmpy/transformations/normalize.py
+++ b/cpmpy/transformations/normalize.py
@@ -128,7 +128,7 @@ def _simplify_boolean_expr(expr: Expression, num_context=False) -> tuple[bool, E
                 newargs = list(args)
 
             if len(newargs) == 0: # empty disjunction
-                return True, BoolVal(False)
+                return True, (0 if num_context else BoolVal(False))
             elif len(newargs) == 1: # single argument, drop the operator
                 return True, newargs[0]
             elif changed:
@@ -153,7 +153,7 @@ def _simplify_boolean_expr(expr: Expression, num_context=False) -> tuple[bool, E
                 newargs = list(args)
 
             if len(newargs) == 0: # empty conjunction
-                return True, BoolVal(True)
+                return True, (1 if num_context else BoolVal(True))
             elif len(newargs) == 1: # single argument, drop the operator
                 return True, newargs[0]
             elif changed:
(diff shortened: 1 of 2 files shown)


tests/test_trans_simplify.py
```diff
@@ -37,6 +37,27 @@ def test_bool_ops(self):
...
+        assert str(simplify_boolean([Operator("or", [False, False])])) == "[boolval(False)]"
+        assert str(simplify_boolean([Operator("and", [True, True])])) == "[boolval(True)]"
+
+        # also in a nested context
+        assert str(self.transform(Operator("or", [bv[0]]) == bv[1])) == "[(bv[0]) == (bv[1])]"
+        assert str(self.transform(Operator("and", [bv[0]]).implies(bv[1]))) == "[(bv[0]) -> (bv[1])]"
+        assert str(self.transform(Operator("or", [Operator("and", [bv[0]]), bv[1]]))) == "[(bv[0]) or (bv[1])]"
+
+        # more than one argument is left untouched
+        assert str(simplify_boolean([Operator("or", [bv[0], bv[1]])])) == "[(bv[0]) or (bv[1])]"
```
IgnaceBleukx: This is a really typical agent-thing to do: apply the patch and verify the previous case is unchanged.
While useful in development, I don't think these sanity check tests should remain part of the testsuite, they feel out of place

tests/test_trans_simplify.py
```diff
@@ -37,6 +37,27 @@ def test_bool_ops(self):
...
 
+    def test_degenerate_bool_ops(self):
+        bv = self.bvs
+
+        # a single argument does not need the operator
+        assert str(simplify_boolean([Operator("or", [bv[0]])])) == "[bv[0]]"
+        assert str(simplify_boolean([Operator("and", [bv[0]])])) == "[bv[0]]"
+        # also when the other arguments are constants that get removed
+        assert str(simplify_boolean([Operator("or", [bv[0], False])])) == "[bv[0]]"
+        assert str(simplify_boolean([Operator("and", [bv[0], True])])) == "[bv[0]]"
```
IgnaceBleukx, line 80: Also check with BoolVal(False)/BoolVal(True)

cpmpy/transformations/normalize.py
```diff
@@ -122,9 +124,14 @@ def _simplify_boolean_expr(expr: Expression, num_context=False) -> tuple[bool, E
...
                     newargs.append(a)
 
-            if changed:
-                if newargs is None: # when args where simplified recursively above
-                    newargs = list(args)
+            if newargs is None:
+                newargs = list(args)
+
+            if len(newargs) == 0: # empty disjunction
+                return True, BoolVal(False)
```
IgnaceBleukx: This should check whether we are in a numercial context or not: a + OR([]) should become a + 0, not a +  False

cpmpy/transformations/normalize.py
```diff
@@ -142,9 +149,14 @@ def _simplify_boolean_expr(expr: Expression, num_context=False) -> tuple[bool, E
...
 
-            if changed:
-                if newargs is None: # when args where simplified recursively above
-                    newargs = list(args)
+            if newargs is None: # when args where simplified recursively above
+                newargs = list(args)
+
+            if len(newargs) == 0: # empty conjunction
+                return True, BoolVal(True)
+            elif len(newargs) == 1: # single argument, drop the operator
```
IgnaceBleukx, line 157: Same here

cpmpy/transformations/normalize.py
IgnaceBleukx: Ideally also add a test for this case

IgnaceBleukx (changes requested):
One missing check for num/bool contexts, and some comments on the tests : )

ThomSerg:
Still blocked on your request for changes

IgnaceBleukx (approved):

PR #1082, opened 2026-08-11
typing of comparison (and fall-out)

tias:
as became apparent in the previous type PR, time for Comparison.

Two notes:
- normalize makes the arg[0] always an Expression, but this is not explicit in the typing, hence some added asserts (could also be casts... casts are lighter than asserts I think... I'm fine with this)
- there is an open TODO on IncompleteFunctionError, to be discussed separately


tias:
I should have checked the old PR #879 before...

Some points from that discussion still valid
- due to how Python creates comparisons, the arg[0] will actually always be an Expression... so maybe we can/should type it that way (though in 879, Ignace called it a side-effect) though the tests do generate Comparisons with a constant arg0
- there is a comment in a test I don't understand
- "value() on comparison with undefined functions... what is the right behaviour? (catch the Incomplete or let it propagate up?)" is what we should discuss in a separate PR or issue; is there any documentation of the current intention?

cpmpy/expressions/core.py
```diff
@@ -692,7 +692,23 @@ def __init__(self, name: str, left: ExprLike, right: ExprLike) -> None:
         We expect at least one of the two to be an :class:`Expression`.
         """
         assert (name in Comparison.allowed), f"Symbol {name} not allowed"
-        super().__init__(name, (left, right))
+        # convert numpy bool/integers to Python natives (keep bool as bool, unlike npint2int)
```
IgnaceBleukx, line 695: Hmm I'm not sure; I would be fine just converting it to ints actually.
In `simplify_boolean` we'll handle this anyway, and its not too hard for a user to see that True changed to 1 when printing the comparison.

IgnaceBleukx (commented):
> due to how Python creates comparisons, the arg[0] will actually always be an Expression... so maybe we can/should type it that way (though in 879, Ignace called it a side-effect) though the tests do generate Comparisons with a constant arg0

The way I see it now I would just be strict and indeed enforce the left-hand side to always be an expression. Fine for me to updated the tests.

cpmpy/expressions/core.py
```diff
@@ -709,7 +725,7 @@ def __bool__(self) -> bool:
     # return the value of the expression
     # optional, default: None
     def value(self) -> Optional[bool]:
-        arg_vals = argvals(self.args)
+        arg_vals = argvals(self.args)  # TODO: argvals_intexpr, but argvals handles IncompleteFunctionError...
```
IgnaceBleukx, line 728: Yes, icomplete func errors to be handled in a different PR, as discussed offline

cpmpy/solvers/cplex.py
```diff
@@ -476,7 +477,7 @@ def add(self, cpm_expr: NestedBoolExprLike) -> "CPM_cplex":
                         "Not a known supported cplex comparison '{}' {}".format(lhs.name, con))
             else:
                 raise NotImplementedError(
-                "Not a known supported cplex comparison '{}' {}".format(lhs.name, con))
+                "Not a known supported cplex comparison '{}' {}".format(con.name, con))
```
IgnaceBleukx, line 480: If we're changing this line anyway: I would prefer to keep the Error 1 line : )

IgnaceBleukx (commented):
was a bit too quick pressing "submit" in my prev review

cpmpy/expressions/core.py
IgnaceBleukx: @tias I think this is the only open point of dicsussion in this PR : )

Issue #966, opened 2026-04-30
Map missing global constraints to MiniZinc interface


WoutPiessens:
The MDD global constraint was added to CPMpy (see #924), and some solvers (e.g. choco) support it natively. However, more solvers (e.g. ortools) support the Regular global constraint natively, and since an MDD is a restricted form of a finite automaton (more specifically, an MDD is a finite automaton without loops, and with one root node and one sink node), the native support for the Regular global constraint could be used directly.

IgnaceBleukx:
Lets use this more like a tracker issue of globals we still need to map to MinZinc native?

- [x] Regular (currently in PR #952 )
- [x] MDD (this issue; after #924 is merged)
- [ ] NoOverlapOptional (addressed in #1083)
- [ ] CumulativeOptional (addressed in #1083)

PR #1083, opened 2026-08-13
NoOverlapOptional and CumulativeOptional mapping to MiniZinc

WoutPiessens:
This PR addresses the two global constraints left in #966: NoOverlapOptional and CumulativeOptional, which were not yet mapped to the MiniZinc interface. Similar to the mappings for NoOverlap and Cumulative respectively, except for the necessary changes for optional tasks.

### DIFF
--- a/cpmpy/solvers/minizinc.py
+++ b/cpmpy/solvers/minizinc.py
@@ -776,16 +776,16 @@ class CPM_minizinc(SolverInterface):
             demand, demand_cons = get_nonneg_args(demand, is_present)
             extra_cons += demand_cons
 
-            # absent tasks consume no resources
-            demand_str = "[{}]".format(",".join(
-                "if {} then {} else 0 endif".format(self._convert_expression(p), self._convert_expression(d))
-                for d, p in zip(demand, is_present)))
+            # absent tasks do not have to be scheduled: expressed with an optional start time
+            opt_start = ["if {} then {} else <> endif".format(self._convert_expression(p),
+                                                              self._convert_expression(s))
+                         for s, p in zip(start, is_present)]
 
             format_str = "forall(" + self._convert_expression(extra_cons) + " ++ [" + global_str + "])"
 
-            return format_str.format(self._convert_expression(start),
+            return format_str.format("[{}]".format(",".join(opt_start)),
                                      self._convert_expression(dur),
-                                     demand_str,
+                                     self._convert_expression(demand),
                                      self._convert_expression(capacity))
 
         elif expr.name == "no_overlap":
@@ -818,7 +818,7 @@ class CPM_minizinc(SolverInterface):
             dur, dur_cons = get_nonneg_args(dur, is_present)
             extra_cons += dur_cons
 
-            # absent tasks do not have to be scheduled
+            # absent tasks do not have to be scheduled: expressed with an optional start time
             opt_start = ["if {} then {} else <> endif".format(self._convert_expression(p),
                                                               self._convert_expression(s))
                          for s, p in zip(start, is_present)]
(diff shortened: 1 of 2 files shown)


tests/test_globalconstraints.py
```diff
@@ -1256,63 +1256,93 @@ def test_cumulative_negative_dur(self):
         assert cp.Model(bv == expr).solve()
         assert not bv.value()
 
-    def test_optional_cumulative(self):
+    @pytest.mark.usefixtures("solver")
+    def test_optional_cumulative(self, solver):
+        if solver in ("pysat", "pysdd", "pindakaas", "rc2"):
+            pytest.skip(f"{solver} does not support integer variables")
```
IgnaceBleukx, line 1282: pysat, pindakaas and rc2 now do! So just pysdd should be skipped
Ah and also rc2 because it only supports optmization

cpmpy/solvers/minizinc.py
```diff
@@ -777,6 +805,28 @@ def zero_based(array):
...
+                start, dur, end, is_present = expr.args
+                extra_cons += [implies(p, s + d == e) for s, d, e, p in zip(start, dur, end, is_present)]
+
+            global_str = "disjunctive({},{})"
+            # ensure duration is non-negative (only enforced for the tasks that are present)
+            dur, dur_cons = get_nonneg_args(dur, is_present)
+            extra_cons += dur_cons
+
+            # absent tasks do not have to be scheduled
+            opt_start = ["if {} then {} else <> endif".format(self._convert_expression(p),
```
IgnaceBleukx, line 822: Hmm I see this might also be fine actually, although not very minzinc-y ; )

cpmpy/solvers/minizinc.py
```diff
@@ -760,6 +760,34 @@ def zero_based(array):
...
+            # ensure duration is non-negative (only enforced for the tasks that are present)
+            dur, dur_cons = get_nonneg_args(dur, is_present)
+            extra_cons += dur_cons
+            # ensure demand is non-negative (only enforced for the tasks that are present)
+            demand, demand_cons = get_nonneg_args(demand, is_present)
+            extra_cons += demand_cons
+
+            # absent tasks consume no resources
+            demand_str = "[{}]".format(",".join(
+                "if {} then {} else 0 endif".format(self._convert_expression(p), self._convert_expression(d))
```
IgnaceBleukx: This is not actually required, minizinc handles this themself.
The only difference between the non-optional cumulative and the optinonal one should be the array of `start` vars (it does get some getting used to)
https://docs.minizinc.dev/en/2.9.3/lib-globals-scheduling.html#mzn-ref-globals-scheduling-cumulative

The starts become optional integers, which you can do just using an `if` statement 
```
[s | i, s in enumerate(_convert_expression(start)) if is_present[i]]
```
I forgot what the exact minizinc-equivalent is of `enumerate(start)` but you get the point : )

IgnaceBleukx (changes requested):
You can use a bunch more minizinc features actually, so you don't have to manually handle all of the opt-stuff : )

cpmpy/solvers/minizinc.py
WoutPiessens: The reason for using this syntax is that when using something like this:

``
[s | i, s in enumerate(_convert_expression(start)) if is_present[i]]
``

if is_present has a fixed value, the expression is evaluated at compile time, so the start array is not equal in length to the others.

cpmpy/solvers/minizinc.py
WoutPiessens: Yes that's better, also more similar to NoOverlapOptional that way. I'll also use the same syntax as NoOverlapOptional, for the same reason as discussed in that comment

IgnaceBleukx:
Lets first determine what we do with #1103, we might need to change to disjunctive_strict here as well

PR #1087, opened 2026-08-16
Fix name of InDomain global constraint

IgnaceBleukx:
For some historical reason, the `InDomain` global constraint was the only on with a PascalCased name, which has been a source of bugs in the past (due to using the wrong string in the supported set in solver interfaces).

This PR brings it in line with the other globals


Dimosts (approved):
Simple to review, fine to go for me

PR #1091, opened 2026-08-26
Split up value/defining constraints in all transformations

OrestisLomis:
This PR splits up value and defining constraints throughout the whole transformation pipeline. On master it is only done in some parts, e.g. decomposition of globals. This PR allows the split to be used elsewhere as well, specifically it will enable a fix for a bug noticed by @IgnaceBleukx in #986. The diff is quite large and while the core idea is simple it is important to verify that the split is correct for all transformations.

### DIFF
--- a/cpmpy/solvers/TEMPLATE.py
+++ b/cpmpy/solvers/TEMPLATE.py
@@ -61,7 +61,7 @@ from ..transformations.get_variables import get_variables
 from ..transformations.normalize import toplevel_list
 from ..transformations.safening import no_partial_functions
 from ..transformations.decompose_global import decompose_in_tree, decompose_objective
-from ..transformations.flatten_model import flatten_constraint, flatten_objective
+from ..transformations.flatten_model import flatten_constraint, flatten_objective, apply_transform
 from ..transformations.comparison import only_numexpr_equality
 from ..transformations.reification import reify_rewrite, only_bv_reifies
 from ..transformations.safening import safen_objective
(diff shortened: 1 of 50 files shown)


IgnaceBleukx:
Hmm it feels very hacky and a big workaround for a small feature (the native IIS for Gurobi).
I'll have a think about what could be a better solution. From the top of my head an idea could be to force the transformations to append all toplevel constraints at the back instead; but then we get into trouble when splitting top-level AND constraints...

Maybe to fix the immediate issue, just turn off cse for Gurobi when doing the IIS extraction? That should fix it for now

OrestisLomis:
Definitely agree that it is worth discussing a bit further, since it is quite a big change. Though I would disagree that it's hacky (even if the specific implementation might need a second look), it is just a more principled split which already exist in some parts of the code. It's just that we generally don't need the split so the two list are often concatenated before the return. And while the motivation is  the native MUS/IIS algorithms, this has some use in general. E.g. for proofs using exact you would want to have a map from user constraints -> solver constraints; and since the defining constraint are not necessarily related to the specific constraint that was transformed we would like to treat them differently. 

As for the specific native MUS/IIS bug, we can indeed turn of CSE for now.

IgnaceBleukx:
I see, so if we decide to go trough with it, I would like to take a bit more of a structured approach to this.
Its probably best to work transformation by transformation to make the split of value and defining already explicit within the function. For now, we can then just end the function with `return value +  defining` and when all transformations are ported to this new patter, we simply replace the return with `return value, defining`,  which is a super simple change to review.

Also toplevel_list is a bit of a weird one to let it split up value and defining... its by definition all value/defining as everything is put in a toplevel list : )

I'd say don't start on it yet; lets discuss on the next dev meeting first if/how/when we go through with this : )

PR #1095, opened 2026-09-04
Fix MUS tests

IgnaceBleukx:
Now runs each mus test really on every solver:
- Previously it required exact to be installed for some reason
- For solvers that do not support assumptions, there is an explicit pytest.skip now
- Native MUS solvers are still parametrized manually (we could change that but will result in lots of skips)

### DIFF
--- a/cpmpy/tools/explain/marco.py
+++ b/cpmpy/tools/explain/marco.py
@@ -3,6 +3,7 @@
 """
 
 import cpmpy as cp
+from cpmpy.exceptions import NotSupportedError
 from cpmpy.transformations.get_variables import get_variables
 
 from .utils import make_assump_model
@@ -40,12 +41,14 @@ def marco(soft, hard=[], solver="ortools", map_solver="ortools", return_mus=True
 
     # map solver for computing hitting sets
     map_solver = cp.SolverLookup.get(map_solver)
-    do_solution_hint = do_solution_hint and hasattr(map_solver, 'solution_hint')  # solver may not support solution hinting...
-
     map_solver += cp.any(assump)
+    
     if do_solution_hint:
         hint = [1]*len(assump)
-        map_solver.solution_hint(assump, hint) # we want large subsets, more likely to be a MUS
+        try:
+            map_solver.solution_hint(assump, hint) # we want large subsets, more likely to be a MUS
+        except NotSupportedError: # not all solvers support solution hinting
+            do_solution_hint = False
 
     deletion_order = {a : -len(get_variables(dmap[a])) for a in assump} # avoid recomputing
 
--- a/tests/test_tools_mus.py
+++ b/tests/test_tools_mus.py
@@ -1,31 +1,61 @@
+import inspect
 import pytest
 
 import cpmpy as cp
 from cpmpy.tools import mss_opt, marco, OCUSException
 from cpmpy.tools.explain import mus, mus_naive, quickxplain, quickxplain_naive, optimal_mus, optimal_mus_naive, mss, mcs, ocus, ocus_naive, mus_native
 
-@pytest.mark.requires_solver("exact")
+
+# Each shared case runs twice: assumption-based (or native) MUS, and the naive variant.
+# Unsupported solvers skip only the non-naive case; naive still runs.
+mus_or_naive = pytest.mark.parametrize("naive", [False, True], ids=["mus", "naive"])
+
+
 class TestMus:
     def setup_method(self):
         self.mus_func = mus
         self.naive_func = mus_naive
 
-    def test_circular(self, solver):
+    def _supported_solver(self, solver):
+        s = cp.SolverLookup.get(solver)
+        solve_arguments = inspect.signature(s.solve).parameters
+        return solve_arguments.get("assumptions") is not None
+
+    def _unsupported_reason(self, solver):
+        return f"Solver {solver} does not support assumption-based MUS"
+
+    def _test_mus(self, cons, hard, solver, verify_func, naive=False, **kwargs):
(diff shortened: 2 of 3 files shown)


tests/test_tools_mus.py
```diff
@@ -1,31 +1,61 @@
...
 
 import cpmpy as cp
 from cpmpy.tools import mss_opt, marco, OCUSException
 from cpmpy.tools.explain import mus, mus_naive, quickxplain, quickxplain_naive, optimal_mus, optimal_mus_naive, mss, mcs, ocus, ocus_naive, mus_native
 
-@pytest.mark.requires_solver("exact")
+
+# Each shared case runs twice: assumption-based (or native) MUS, and the naive variant.
+# Unsupported solvers skip only the non-naive case; naive still runs.
+mus_or_naive = pytest.mark.parametrize("naive", [False, True], ids=["mus", "naive"])
```
OrestisLomis: To be honest, I did not really understand how this was dispatched until some further digging and the doc didn't really help me so much. Not sure how important it is for someone to understand how the pytest 'collecting' works when going through this...

tests/test_tools_mus.py
```diff
@@ -1,31 +1,61 @@
...
 class TestMus:
     def setup_method(self):
         self.mus_func = mus
         self.naive_func = mus_naive
 
-    def test_circular(self, solver):
+    def _supported_solver(self, solver):
+        s = cp.SolverLookup.get(solver)
+        solve_arguments = inspect.signature(s.solve).parameters
+        return solve_arguments.get("assumptions") is not None
```
OrestisLomis, line 21: We had a discussion on this before and my opinion was that it is not really necessary to test mus functionality for all solvers. The solvers themselves are already tested elsewhere, for mus extraction they are just an oracle. The tests should be about the algorithms themselves and so there is no need to do duplicate work.

cpmpy/tools/explain/marco.py
```diff
@@ -40,12 +41,14 @@ def marco(soft, hard=[], solver="ortools", map_solver="ortools", return_mus=True
...
     map_solver = cp.SolverLookup.get(map_solver)
-    do_solution_hint = do_solution_hint and hasattr(map_solver, 'solution_hint')  # solver may not support solution hinting...
-
     map_solver += cp.any(assump)
+    
     if do_solution_hint:
         hint = [1]*len(assump)
-        map_solver.solution_hint(assump, hint) # we want large subsets, more likely to be a MUS
+        try:
+            map_solver.solution_hint(assump, hint) # we want large subsets, more likely to be a MUS
```
OrestisLomis, line 49: this should best be in a separate PR I think

OrestisLomis (commented):
To be honest, I don't really see the point of the PR. I think you can just change the default for the tests back to ortools (during testing it was easier to have it be a solver that supported native muses), but other than that it's unnecessary imo.

The Marco fix can be in a separate PR as well.

tests/test_tools_mus.py
IgnaceBleukx: I don't agree. Bugs like for example the solution hinting stuff are ONLY exposed by running all solvers in the MUS tool. Indeed, the solvers are used as an orcacle, but there are clearly untested code paths.

tests/test_tools_mus.py
IgnaceBleukx: Yeah doc can be improved for sure. We do use the parametrization elsewhere too, it basically duplicates the test by assigning the given variable each value in the list (in this case False and True for the `naive` parameter)

IgnaceBleukx: Yeah doc can be improved for sure. We do use the parametrization elsewhere too, it basically duplicates the test by assigning the given variable each value in the list (in this case False and True for the `naive` parameter)

tests/test_tools_mus.py
OrestisLomis: But there are separate tests for solution hinting, so if those tests don't catch the bugs I would be more inclined to argue that those tests need to be more complete. Though I suppose marco application can still be a nice integrated test.

tests/test_tools_mus.py
IgnaceBleukx: Can you check this is more clear?

IgnaceBleukx:
So after some offline discussion we agreed that the integration tests with the full MUS algorithm are quite useful to have as testcases. The infrasstructure also allows to much easier add mus testcase in the future.

@OrestisLomis if you can check the doc update, I believe this can be merged and #987 can be rebased from this one

tests/test_tools_mus.py
```diff
@@ -1,31 +1,60 @@
+import inspect
 import pytest
 
 import cpmpy as cp
 from cpmpy.tools import mss_opt, marco, OCUSException
 from cpmpy.tools.explain import mus, mus_naive, quickxplain, quickxplain_naive, optimal_mus, optimal_mus_naive, mss, mcs, ocus, ocus_naive, mus_native
 
-@pytest.mark.requires_solver("exact")
+
+# annotation to run both MUS and naive varaint for the test
```
OrestisLomis: I think the terminology is still a bit confusing. Now we call the two variants MUS and naive, but really both are MUS variants and one can be called naive while the other can be called incremental or something like that? Also "variants" has a typo.

OrestisLomis:
As said offline I believe we should treat 'MUS' more as the concept and not the specific incremental deletion-based algorithm. For this PR that comment mostly means that I think the wording should be more clear, but in general perhaps the code-structure can also be changed. To be discussed for other PRs.

OrestisLomis (approved):
Nice!

PR #1096, opened 2026-09-04
Large neighborhood search tool

IgnaceBleukx:
Generic tool to do LNS-style local search.

Some design desisions to discuss:
- The destroy operator is now designed to be a function with side effects in order to be as generic as possible
- We currently require a full solution as input. I'm not sure if this is what we want: alternatively, we can also allow non-initialized variables and just start the improving loop after the first solve call; Both have advantages and disadvantages: if we start from a given solution, a user can construct it heuristically or manually even, but if the inital solution was found using a CPMpy solver, we might miss out on potentiall incrementality.

### DIFF
--- a/cpmpy/tools/local_search.py
+++ b/cpmpy/tools/local_search.py
@@ -98,6 +98,7 @@ def large_neighborhood_search(model: cp.Model,
 
     best_sol = [var.value() for var in model_vars]
     best_obj = model.objective_value()
+    assert best_obj is not None, "Not all variables are assigned, LNS requires a full solution to start from."
     sol = [var.value() for var in model_vars]
     if verbose >= 1:
         print(f"Starting LNS ({len(model_vars)} vars, solver={solver})")
@@ -140,6 +141,7 @@ def large_neighborhood_search(model: cp.Model,
 
         sol = [var.value() for var in model_vars]
         obj = model.objective_value()
+        assert obj is not None # to make mypy happy
         if model.objective_is_min:
             improved = obj < best_obj
         else:
--- a/cpmpy/tools/local_search.py
+++ b/cpmpy/tools/local_search.py
@@ -1,0 +1,161 @@
+"""
+Local search methods built on top of CPMpy solvers.
+
+Implements Large Neighborhood Search (LNS): iteratively destroy part of a
+solution and re-optimize the resulting neighborhood with a CP solver.
+
+=================
+List of functions
+=================
+
+.. autosummary::
+    :nosignatures:
+
+    large_neighborhood_search
+"""
+from typing import Optional, Callable
+import inspect
+import time
+
+import cpmpy as cp
+from cpmpy.expressions.variables import _NumVarImpl
+from cpmpy.transformations.get_variables import get_variables_model
+from cpmpy.transformations.normalize import toplevel_list
+
+
+def large_neighborhood_search(model: cp.Model, 
+                              destroy: Callable,
+                              max_iterations: Optional[int] = None,
+                              total_time_limit: Optional[float] = None,
+                              incremental = True,
+                              verbose = 1,
+                              solver = "ortools",
+                              **kwargs):
+    """
+    Improve a solution of an optimization model using Large Neighborhood Search (LNS).
+
+    Each iteration calls ``destroy`` to unassign part of the current solution, then re-solves
+    the resulting neighborhood. The best assignment found is stored in the model's variable values.
+
+    All decision variables in the model must already be assigned, e.g. from a previous
+    :meth:`solve() <cpmpy.model.Model.solve>` call or a manually constructed solution.
+
+    At least one of ``max_iterations`` or ``total_time_limit`` should be set, otherwise the
+    search runs indefinitely.
+
+    Arguments:
+        - model (:class:`~cpmpy.model.Model`): CPMpy optimization model
+        - destroy (Callable): callback invoked each iteration; should unassign part of the current solution
+            by calling `_NumVarImpl.clear()` on the variables to unassign. Invoked after each solver call.
+        - max_iterations (int, optional): maximum number of destroy-and-repair iterations
+        - total_time_limit (float, optional): overall time budget in seconds
+        - incremental (bool, optional): whether to use assumption variables to fix the values of non-destroyed variables.
+            Avoids re-initializing the solver between iterations, but can introduce overhead.
+        - verbose (int, optional): how much information to print (0=none, default: 1)
+        - solver (str): name of a solver to use, defaults to "ortools"
+        - **kwargs: extra keyword arguments passed to the solver (e.g., ``time_limit`` to set the time limit per iteration)
+
+    Example:
+        .. code-block:: python
+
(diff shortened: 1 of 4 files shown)


cpmpy/tools/local_search.py
```diff
@@ -0,0 +1,163 @@
...
+
+            x = cp.boolvar(shape=len(weights))
+            model = cp.Model(cp.sum(weights * x) <= capacity, 
+                             maximize=cp.sum(vals * x))
+
+            # greedy solution, assign variables with highest values until capacity is reached
+            sorted_vars = sorted(enumerate(x), key=lambda v: vals[v[0]] / weights[v[0]], reverse=True)
+            for i, var in sorted_vars:
+                if used_capacity + weights[i] <= capacity:
+                    var._value = 1
```
ThomSerg, line 74: maybe we want a proper setter instead of accessing a private field. with something like `set_value()` we could then also check for whether the value lies within the domain bounds

cpmpy/tools/local_search.py
```diff
@@ -0,0 +1,163 @@
...
+                    var._value = 0
+
+            def destroy(): # clear first half of the variables
+                for i in range(len(x)//2):
+                    x[i].clear()
+
+            large_neighborhood_search(model, destroy, max_iterations=10, time_limit=1)
+    """
+
+    if model.objective_ is None:
```
ThomSerg, line 86: if not model.has_objective()

cpmpy/tools/local_search.py
```diff
@@ -0,0 +1,163 @@
...
+    s = cp.SolverLookup.get(solver, model)
+    incremental = incremental and "assumptions" in inspect.signature(s.solve).parameters
+
+    # keep track of best solution and objective value
+    start = time.time()
+
+    best_sol = [var.value() for var in model_vars]
+    best_obj = model.objective_value()
+    assert best_obj is not None, "Not all variables are assigned, LNS requires a full solution to start from."
+    sol = [var.value() for var in model_vars]
```
ThomSerg, line 102: sol = best_sol.copy() (or drop the copy if no edit needed, if can be same list object)

cpmpy/tools/local_search.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Arguments:
+        - model (:class:`~cpmpy.model.Model`): CPMpy optimization model
+        - destroy (Callable): callback invoked each iteration; should unassign part of the current solution
+            by calling `_NumVarImpl.clear()` on the variables to unassign. Invoked after each solver call.
+        - max_iterations (int, optional): maximum number of destroy-and-repair iterations
+        - total_time_limit (float, optional): overall time budget in seconds
+        - incremental (bool, optional): whether to use assumption variables to fix the values of non-destroyed variables.
+            Avoids re-initializing the solver between iterations, but can introduce overhead.
+        - verbose (int, optional): how much information to print (0=none, default: 1)
```
ThomSerg, line 54: document verbosity levels, or does this follow a standard?

cpmpy/tools/local_search.py
```diff
@@ -0,0 +1,163 @@
...
+        # destroy (part of) the solution
+        destroy()
+        if verbose >= 2:
+            n_destroyed = sum(1 for var in model_vars if var.value() is None)
+            print(f"Iteration {num_iter}: destroyed {n_destroyed}/{len(model_vars)} variables")
+
+        # fix the variables that are not destroyed
+        if incremental:
+            bv = cp.boolvar()
+            s += bv.implies(cp.all(var == sol[i] for i, var in enumerate(model_vars) if var.value() is not None))
```
ThomSerg, line 131: could it be worth it to have some reuse? With LNS I expect that iterations will often assign the same values to variables.

cpmpy/tools/local_search.py
```diff
@@ -0,0 +1,163 @@
...
+    Each iteration calls ``destroy`` to unassign part of the current solution, then re-solves
+    the resulting neighborhood. The best assignment found is stored in the model's variable values.
+
+    All decision variables in the model must already be assigned, e.g. from a previous
+    :meth:`solve() <cpmpy.model.Model.solve>` call or a manually constructed solution.
+
+    At least one of ``max_iterations`` or ``total_time_limit`` should be set, otherwise the
+    search runs indefinitely.
+
+    Arguments:
```
ThomSerg, line 46: In the future we might want a solution callback? Or do we document that they can (ab)use the `destroy` one for this purpose. Or a new one that is only called upon an improving solution.

examples/advanced/vrp_lns.py
```diff
@@ -0,0 +1,105 @@
...
+    load = cp.intvar(0, capacity, shape=n_nodes, name="load")
+    dist_cpm = cp.cpm_array(dist_ext)
+    demand_cpm = cp.cpm_array(demand)
+
+    model = cp.Model(cp.Circuit(succ))
+    model += [load[n + k] == 0 for k in range(n_vehicle)]
+    for i in range(n_nodes):
+        model += (succ[i] < n).implies(load[succ[i]] == load[i] + demand_cpm[succ[i]])
+
+    # multi-objective: minimize traveled distance and number of vehicles used
```
ThomSerg, line 39: not really multi-objective. Maybe very indirectly due to minimising distance also minimises repeated visits to the depot. But there could still be cases where an unnecessary additional stop to the depot is made.

examples/advanced/vrp_lns.py
```diff
@@ -0,0 +1,105 @@
...
+        removed = [v]
+        while len(removed) < num_to_remove:
+            v = removed[int(np.random.randint(len(removed)))]
+            inplan.sort(key=lambda j: dist[v, j])  # closest = most related
+            chosen = inplan.pop(min(int(len(inplan) * np.random.random() ** D), len(inplan) - 1))
+            removed.append(chosen)
+
+        sv = [succ[i].value() for i in range(n_nodes)]
+        for c in removed:
+            pred = next(i for i in range(n_nodes) if sv[i] == c)
```
ThomSerg, line 69: nice trick, will remember this one!

tests/test_tools_lns.py
```diff
@@ -0,0 +1,95 @@
...
+            x[0].clear()
+            x[1].clear()
+
+        large_neighborhood_search(model, destroy, max_iterations=5, incremental=False)
+
+        assert all(var.value() is not None for var in x)
+        assert (2*x[0] + 3*x[1] + 4*x[2] + 5*x[3] <= 5).value() is True
+        assert model.objective_value() >= 0
+
+    def test_time_limit(self):
```
ThomSerg, line 65: also test for max iterations?

ThomSerg (commented):
Some comments. Also do we want a version that enforces an objective lower bound? (does require some bigger changes tho, maybe this already works good enough)

cpmpy/tools/local_search.py
IgnaceBleukx: You're right, adding a `display` argument would be useful! Although we need to be careful that they don't clash with passing `display` to the solver call : )

PR #1097, opened 2026-09-07
Fix typos in docstrings, comments, and error messages

tias:
## Summary
- Fixed a handful of typos found by manual review and a `codespell` scan of the `cpmpy/` package (docstrings, comments, and error messages).
- Purely textual changes — no logic touched.

## Test plan
- [x] `import cpmpy` succeeds
- [x] All edited files still parse (`ast.parse`)

### DIFF
--- a/cpmpy/solvers/solver_interface.py
+++ b/cpmpy/solvers/solver_interface.py
@@ -65,7 +65,7 @@ class SolverInterface(object):
 
     def __init__(self, name="dummy", cpm_model=None, subsolver=None):
         """
-            Initalize solver interface
+            Initialize solver interface
 
             - name: str: name of this solver
             - subsolver: string: not used/allowed here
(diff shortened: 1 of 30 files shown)


cpmpy/solvers/solver_interface.py
```diff
@@ -65,7 +65,7 @@ def version(cls) -> Optional[str]:
 
     def __init__(self, name="dummy", cpm_model=None, subsolver=None):
         """
-            Initalize solver interface
+            Initialize solver interface
```
OrestisLomis, line 68: this is US English spelling, it would be initialise in UK English

OrestisLomis (approved):
These are correct indeed and can easily be accepted. One thing to note is that we do mix UK/US English. E.g. optimisation vs optimization, modelling vs modeling, etc. Personally for the docs I don't think these are really worth keeping consistent, but I did think it was worth noting.

PR #1098, opened 2026-09-08
Description context manager 

IgnaceBleukx:
Context manager to make it more natural for a user to add a description to a set of constraints added to the model.

Before, we had to wrap the epxressions into an `and`, set the description on that conjunction and then add it to the model.
This PR simplifies it in a more Pythononic way.

One thing thats currently unclear for me is how/whether we want to support nested descriptions.
Its a bit tricky as for Operators, we merge nested ands at construction time (which is something we probably want to keep doing). So if we would have something like this:

```python
with model.description("foo"):
    model += x + y <= 1
    with model.description("bar"):
        model += x == 4
        model += y != 1
```
we would merge the args of bar into foo: and(x + y <= 1, x == 4,  y != 1)

One solution could be to only allow adding expressions at the lowest level of the description manager, the nested structure would then just be glorified "ands" with a description attached to it.

@WoutPiessens probably has some interesting thoughts on this : )

### DIFF
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -792,7 +792,7 @@ class Operator(Expression):
         # small cleanup: nested n-ary operators are merged into the toplevel
         # (this is actually against our design principle of creating
         #  expressions the way the user wrote them)
-        if arity == 0:
+        if arity == 0 and name != 'and': # special handling for nested description groups
             arg_list = list(arg_list)  # make sure its a writable list
             i = 0 # length can change
             while i < len(arg_list):
--- a/cpmpy/expressions/core.py
+++ b/cpmpy/expressions/core.py
@@ -792,7 +792,7 @@ class Operator(Expression):
         # small cleanup: nested n-ary operators are merged into the toplevel
         # (this is actually against our design principle of creating
         #  expressions the way the user wrote them)
-        if arity == 0 and name != 'and': # special handling for nested description groups
+        if arity == 0:
             arg_list = list(arg_list)  # make sure its a writable list
             i = 0 # length can change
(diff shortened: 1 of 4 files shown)


WoutPiessens:
I think that only allowing to add expressions at the lowest level makes sense, this is what I do in the context of the work on interactive hierarchical explanations. Alternatively, expressions at a different level could be pushed to the lowest level, by adding a dummy description so that
```
with model.description("foo"):
    model += x + y <= 1
    with model.description("bar"):
        model += x == 4
        model += y != 1
```

is equivalent to something like
```
with model.description("foo"):
    with model.description(""):
        model += x + y <= 1
    with model.description("bar"):
        model += x == 4
        model += y != 1
```

But I have not really encountered any need for adding expressions on a higher level.

Then I think it makes sense that the different-level descriptions could be stacked. These would then represent the different-level meanings of the constraints, e.g.

```
with model.description("Person unavailable"):
   with model.description("Monday"):
      ...
   with model.description("Tuesday"):
     ....
```

could result in a description such as ```str(cons) = ("Person unavailable", "Monday")``` Also, that way a hierarchy based on the conceptual meaning of the constraints can be derived without having to store a tree object.

IgnaceBleukx:
I very much like the idea of stacking the descriptions, but typing-wise its a bit tricky... str() always returns a string of course, so we cannot return a tuple all of a sudden.

For now, I think its ok to only allow adding descriptions at the lowest level, and to just print the top-level description if there are nested ones descriptions. Its the users responsibility to just walk the expression if they want to do so.
Actually... we could add a tool to print the hierarchical constraint tree? Just like the `tree` program on linux

PR #1100, opened 2026-09-10
Skills infrastructure

ThomSerg:
All the infrastructure needed to add Agent skills to the repo. For now includes one reference skill to expand upon.

Skills are located under /skills. This is inline with skill.sh's expectation (the somewhat defacto cli for installing agent skills).
After merging into master, you can run:
```console
npx skills add CPMpy/cpmpy
``` 
Right now you need:
```console
npx skills add https://github.com/CPMpy/cpmpy/tree/skills
```
It gets all the skill markdown files directly from the github repo. After enough usage, our skill should appear in their [skills directory](https://www.skills.sh/)

Python also has the upcoming [library-skills](https://library-skills.io/), which starts to get well-accepted within the python community (FastAPI, Streamlit, Typr, ...). It is a standard for bundling skill files with python library wheels. The library-skills library with its cli can than discover them and add to your local project's .agents or .claude. But contrary to skills.sh, it expects the files to be located under cpmpy/.agents. So to prevent duplicating files, I've included some build scripts that auto run on pip install that copy /skills over to that location. Also works for editable installs from source.

So the proposed flow:
A) Use skill.sh
```console
npx skills add CPMpy/cpmpy
``` 
B) Install from cpmpy wheel
1) run command to setup skills in local project
```console
cpmpy skill
```
-> will complain about missing deps:
```console
pip install cpmpy[skills]
```
(this installs library-skills)
C) Users already using library-skills
```console
library-skills
```
-> will auto-discover cpmpy skills from python environment's installed wheels

Skills can have reference files, like excerpts from the user docs. Builds on #1099 for agent-friendly markdown version of the docs.

Once merged in master, we can add a shield to the GitHub readme with a link to the skills. 

Useful links:
| | |
| - | - |
| Python's `library-skills` | https://library-skills.io/ |
| Skills.sh | https://www.skills.sh/ |
| FastAPI's skills | https://github.com/fastapi/fastapi/tree/master/fastapi/.agents/skills/fastapi |
| Nvidia skills for cuOpt | https://github.com/NVIDIA/skills/blob/main/skills/cuopt-numerical-optimization-api/SKILL.md |
| Agent Skills standard spec | https://agentskills.io/specification |
| blogposts from person at Nvidia | https://jacobtomlinson.dev/posts/2026/documenting-your-software-libraries-for-agents/ |

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -48,7 +48,8 @@ extensions = [
     'sphinx_automodapi.smart_resolver',
     'sphinx.ext.napoleon',
     'sphinx.ext.todo',
-    'sphinx.ext.autosectionlabel'
+    'sphinx.ext.autosectionlabel',
+    'sphinx_llm.txt'
 ]
 
 myst_enable_extensions = [
@@ -118,3 +119,8 @@ html_static_path = ['_static']
 
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'cpmpy'
+
+# Config options for sphinx_llm
+llms_txt_exclude = [
+    'docs_todo',
+]
--- a/skills/cpmpy-model/SKILL.md
+++ b/skills/cpmpy-model/SKILL.md
@@ -1,21 +1,12 @@
 ---
 name: cpmpy-model
-description: Write and solve problems in CPMpy, a constraint programming and modelling library in Python, based on numpy, with direct solver access. 
-version: 0.0.1
+description: Write and solve problems in CPMpy, a constraint programming and modelling library in Python, based on numpy, with direct solver access.
 license: Apache-2.0
+compatibility: Requires Python 3.10+
 metadata:
   author: CPMpy team
   domain: constraint-programming
   library: cpmpy
-  tags:
-    - constraint-programming
-    - combinatorial-optimization
-    - constraint-modeling
-    - decision-variables
-    - global-constraints
-    - python
-    - or-tools
-    - optimization
 ---
 
 # CPMpy modeling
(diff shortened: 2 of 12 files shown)


docs/conf.py
```diff
@@ -48,7 +48,8 @@
     'sphinx_automodapi.smart_resolver',
     'sphinx.ext.napoleon',
     'sphinx.ext.todo',
-    'sphinx.ext.autosectionlabel'
+    'sphinx.ext.autosectionlabel',
+    'sphinx_llm.txt'
```
kostis-init, line 52: Should this also be added to docs/requirements.txt?

skills/cpmpy-model/SKILL.md
```diff
@@ -0,0 +1,21 @@
...
+
+# CPMpy modeling
+
+Model and solve constrained (optimisation) problems using CPMpy with one of its many solver backends across SMT, CP, ILP, PB, (max)SAT. Model once in a numpy-style high-level language, using **decision variables** plus **constraints** over them, optionally with an **objective**, then solve with any backend.
+
+For an overview of the available backends and their abilities: [index](https://cpmpy.readthedocs.io/en/latest/index.html)
+
+For a compact, complete API listing: [references/api-cheatsheet.generated.md](references/api-cheatsheet.generated.md)
+
+For the full narrative documentation (global constraints catalog, solver selection, incremental solving, I/O, debugging), see [documentation](https://cpmpy.readthedocs.io/en/latest/llm.txt).
```
kostis-init, line 20: Should this be "..../llms-full.txt"? The sphinx extension seems to generate llms.txt and llms-full.txt, but this link points to llm.txt instead.

kostis-init (commented):
Seems good as a first version of a cpmpy skill. I just added 2 minor comments.

PR #1101, opened 2026-09-10
Fix linearize when trivially unsat comparison

Dimosts:
When facing a trivially unsat comparison, linearize just posts boolval(false) and breaks. I do not think this is intended behaviour. First, it just continues in other cases of trivially unsat (e.g. lines 224-225). Second, although in a single solve this will have the same result, in other cases, like explanations etc, it will just ignore several constraints of the problem.

This PR just changes the "break" to "continue", and adds a test.


IgnaceBleukx (approved):
Hmmm I see, indeed if you have a conflict outside of just the BoolVal(False), then some explanation techniques (notably the native ones) won't find all MUSes for example...

For the CPMpy-implemented techniques this shouldn't really be an issue, right? Because then we just have `assump_var -> BoolVal(False)`?

I any case fine to go in : )

Issue #1084, opened 2026-08-13
Inconsistency between NoOverlap and Cumulative


WoutPiessens:
Cumulative should be equivalent to NoOverlap when demand and capacity are equal to 1.
However, this is not the case for the following simple example:

```text

NoOverlap(start=[0,1], duration=[2,0]) -> value() = False

Cumulative(start=[0,1], duration=[2,0], demand=1, capacity=1) -> value() = True

```
It seems to me that the problem lies with the current NoOverlap decomposition. In my opinion, a task with duration 0 should not constrain the NoOverlap constraint any further, it cannot overlap with any other task by definition, even though it's starting time is in the middle of another task.

IgnaceBleukx:
I'm not sure, solvers also use the current semantics; which is inconsistent with the global constraint catalog.
I would be inclined to keep the current semantics and follow the solvers; we'll need to update it of course (and add an explicit test so we check it if future solvers are added)

hbierlee:
Note that MiniZinc has both disjunctive (nooverlap) and disjunctive_strict, so both semantics exist. Not sure if this is something that CPMpy should adopt as well, but if so, you can consider a keyword argument e.g. `strict=False`

WoutPiessens:
I think it's a good idea to have this ``strict`` keyword argument, so that both semantics can be supported.

However, when checking these two values:
```
NoOverlap(start=[0,1], duration=[2,0]).value()

Cumulative(start=[0,1], duration=[2,0], demand=1, capacity=1).value()
```
for several different solvers with native support for these globals (and also Gurobi, which does not), I found inconsistencies between solvers. Here they are summarized in a table:

| Solver | `NoOverlap` | `Cumulative` |
|---|---|---|
| or-tools | `False` | `True` |
| gurobi (time-based decomposition) | `False` | `True` |
| gurobi (task-based decomposition) | `False` | `False` |
| choco | `True` | `True` |
| pumpkin | `True` | `True` |
| minizinc | `True` | `True` |


e.g. for Minizinc, this is caused by the fact that we map CPMpy's NoOverlap to Minizinc's disjunctive, which should be disjunctive_strict if we want to maintain the current semantics. So after making a decision, we'll have to adjust some of these mappings.

IgnaceBleukx:
Hmm ok so I agree that it must be consistent among solvers of course, but I don't think we need the strict/non-strict version.
[...]

PR #1102, opened 2026-09-10
Cumulative no duration demand fix

Dimosts:
In Cumulative, we currently have an inconsistency.

When the duration of a task is 0, native cumulative (I tested ortools, choco, minizinc) does not take into account its demand, even if it is alone larger than the capacity. In our implementation, we also have this assumption in .value() and in time decomposition.

However, in task decomposition we actually did not follow this, getting different results.

In this PR, I changed task decomposition (also in the optional), to follow this assumption. Also added a test for it.

### DIFF
--- a/tests/test_globalconstraints.py
+++ b/tests/test_globalconstraints.py
@@ -1134,6 +1134,24 @@ class TestGlobal:
         # also test decomposition
         assert not cp.Model(cons.decompose()).solve()# capacity was not taken into account and this failed
 
+    def test_cumulative_zero_duration(self):
+        # Zero-duration tasks occupy no resource
+        from cpmpy.expressions.utils import argval
+
+        s = cp.intvar(0, 0, name="s")
+        d = cp.intvar(0, 0, name="d")
+        cons = cp.Cumulative([s], [d], demand=[2], capacity=1)
+
+        assert cp.Model([s == 0, d == 0]).solve()
+        assert cons.value() is True
+
+        task, _ = cons.decompose(how="task")
+        time, _ = cons.decompose(how="time")
+        assert all(argval(q) for q in task)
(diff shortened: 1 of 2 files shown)


OrestisLomis:
There is some more discussion on this here: #1084

IgnaceBleukx (approved):
Yes indeed, so my current take (and also preparing a PR with that) is to not allow overlap 0-duration tasks in the NoOverlap constraint (following OR-Tools I believe), but to allow it in Cumulative, as your PR already does for the task-decomposition of Cumulative

Maybe do add something about 0-duration tasks in the docstring of the global constraint?

IgnaceBleukx:
Updated the tests to run for all solvers, and added a similar test for Cumulative optional.
This uncovered one [inconsistency in Pumpkin](https://github.com/ConSol-Lab/Pumpkin/issues/577), as well as a small oversight in the translation of multiplication for Minizinc.

Should be ready to go in after tests complete

tests/test_globalconstraints.py
```diff
@@ -1134,6 +1134,58 @@ def test_cumulative_decomposition_capacity(self):
         # also test decomposition
         assert not cp.Model(cons.decompose()).solve()# capacity was not taken into account and this failed
 
+    def test_cumulative_zero_duration(self, solver):
+        # Zero-duration tasks occupy no resource
+        if solver == "pumpkin":
+            pytest.skip("Bug in Pumpkin Cumulative, see test below")
```
ImkoMarijnissen, line 1140: The bug should have now been fixed in Pumpkin!

IgnaceBleukx: Amazing! We'll keep the xfail test in there until the next Pumpkin release : ) Then the CI will tell us we can remove it : )

PR #1103, opened 2026-09-10
NoOverlap zero duration

IgnaceBleukx:
Following #1102 and the discussion in #1084 I changed the semantics of NoOverlap to follow the strict version of disjunctive in MiniZinc.

Concretely, this means 0-duration tasks cannot overlap with other tasks. They can be scheduled (i.e., be non-optional) the start has to be before/after other tasks.

I checked the MiniZinc parsers for pumpkin and choco, they both post disjunctive_strict using the decomposition. In this PR, I slightly changed this so that both the Cumulative and decomposition are posted. 

Also fixes an existing bug in CPO and implemented a faster checker for NoOverlap (which is how I got to this issue in the first place...)

### DIFF
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -1959,6 +1959,9 @@ class NoOverlap(GlobalConstraint):
     Enforces that a set of tasks are scheduled without overlapping, and enforces:
         - duration >= 0
         - start + duration == end
+
+    Tasks with zero duration are considered to overlap, similar to MiniZinc's `disjunctive_strict` constraint.
+    If you need to enforce that tasks with zero duration are allowed to overlap, use `cp.Cumulative(start, duration, end, demand=1, capacity=1)` instead.
     """
 
     def __init__(self, start: ListLike[ExprLike], duration: ListLike[ExprLike], end: Optional[ListLike[ExprLike]] = None):
@@ -2020,17 +2023,17 @@ class NoOverlap(GlobalConstraint):
             end = [s + d for s,d in zip(start, duration)]
         else:
             start, duration,end = argvals(self.args)
-            if any(a is None for a in [start, duration,end]):
+            if any(a is None for a in start + duration + end):
                 return None
         
-        tasks = sorted(zip(start, duration,end), key=lambda x: x[0])
+        tasks = sorted(zip(start, duration, end), key=lambda x: (x[0], x[1] != 0)) # break ties, sort 0-duration tasks first
 
         for i, (s,d,e) in enumerate(tasks):
             if d < 0:
                 return False
             if s + d != e:
                 return False
-            if i < len(tasks) - 1 and e < tasks[i+1][0]: # overlaps with next
+            if i < len(tasks) - 1 and e > tasks[i+1][0]: # overlaps with next
                 return False
                 
         return True
@@ -2115,13 +2118,14 @@ class NoOverlapOptional(GlobalConstraint):
             if any(a is None for a in start + duration + end + is_present):
                 return None
 
-        tasks = sorted(((s,d,e) for s,d,e,p in zip(start, duration,end, is_present) if p), key=lambda x: x[0])
+        tasks = sorted(((s,d,e) for s,d,e,p in zip(start, duration,end, is_present) if p),
+                       key=lambda x: (x[0], x[1] != 0)) # break ties, sort 0-duration tasks first
         for i, (s,d,e) in enumerate(tasks):
             if d < 0:
                 return False
             if s + d != e:
                 return False
-            if i < len(tasks) - 1 and e < tasks[i+1][0]: # overlaps with next
+            if i < len(tasks) - 1 and e > tasks[i+1][0]: # overlaps with next
                 return False
     
         return True
--- a/cpmpy/expressions/globalconstraints.py
+++ b/cpmpy/expressions/globalconstraints.py
@@ -2022,14 +2022,17 @@ class NoOverlap(GlobalConstraint):
             start, duration,end = argvals(self.args)
             if any(a is None for a in [start, duration,end]):
                 return None
-       
-        if any(d < 0 for d in duration):
-            return False
-        if any(s + d != e for s,d,e in zip(start, duration,end)):
-            return False
-        for (s1,d1), (s2,d2) in all_pairs(zip(start,duration)):
-            if s1 + d1 > s2 and s2 + d2 > s1:
+        
+        tasks = sorted(zip(start, duration,end), key=lambda x: x[0])
+
+        for i, (s,d,e) in enumerate(tasks):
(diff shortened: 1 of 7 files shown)


cpmpy/expressions/globalconstraints.py
Dimosts, line 2051: Also in this docstring?

cpmpy/expressions/globalconstraints.py
```diff
@@ -1959,6 +1959,9 @@ class NoOverlap(GlobalConstraint):
     Enforces that a set of tasks are scheduled without overlapping, and enforces:
         - duration >= 0
         - start + duration == end
+
+    Tasks with zero duration are considered to overlap, similar to MiniZinc's `disjunctive_strict` constraint.
```
Dimosts, line 1963: I think similar docstring changes should also be at Cumulative and CumulativeOptional (now they state that they are equivalent to nooverlap in case of demand = capacity = 1

cpmpy/solvers/choco.py
```diff
@@ -90,7 +90,7 @@ class CPM_choco(SolverInterface):
                                     "increasing", "decreasing", "strictly_increasing", "strictly_decreasing",
                                     "lex_lesseq", "lex_less", "mdd",
                                     "min", "max", "div", "mod", "pow", "abs", "mul", "count", "element", "nvalue", "among"})
-    supported_reified_global_constraints = supported_global_constraints  # choco supports everything reified
+    supported_reified_global_constraints = supported_global_constraints - {"no_overlap"}  # choco supports everything reified
```
Dimosts, line 93: something in the comment on then why we remove nooverlap?

cpmpy/solvers/choco.py
```diff
@@ -679,10 +679,17 @@ def _get_constraint(self, cpm_expr):
...
                 else:
                     start, dur, end = cpm_expr.args
-                    return self._get_constraint(Cumulative(start, dur, end, demand=1, capacity=1))
+                    chc_cons = self._get_constraint(Cumulative(start, dur, end, demand=1, capacity=1))
+                    
+                if any(lb <= 0 for lb in get_bounds(dur)[0]): # zero-duration tasks cannot overlap, not enforced by Cumulative
+                    assert "no_overlap" not in self.supported_reified_global_constraints # otherwise cannot post decomposition toplevel
+                    self.add(cpm_expr.decompose())
+
+                return chc_cons
```
Dimosts, line 691: So, we also post the cumulative, together with the decomposition right? But then, we do not need the full decomp but only the constraints about the constraints related to 0 dur tasks?

WoutPiessens:
Looks good, but I think there is still one inconsistency remaining. In the table in #1084, the task-based decomposition of Cumulative still uses the strict semantics for zero-duration tasks. Since we want to use the non-strict semantics for Cumulative consistently, this should be updated as well.

cpmpy/solvers/choco.py
```diff
@@ -679,10 +679,17 @@ def _get_constraint(self, cpm_expr):
...
                     start, dur = cpm_expr.args
-                    return self._get_constraint(Cumulative(start, dur, demand=1, capacity=1))
+                    chc_cons = self._get_constraint(Cumulative(start, dur, demand=1, capacity=1))
                 else:
                     start, dur, end = cpm_expr.args
-                    return self._get_constraint(Cumulative(start, dur, end, demand=1, capacity=1))
+                    chc_cons = self._get_constraint(Cumulative(start, dur, end, demand=1, capacity=1))
+                    
+                if any(lb <= 0 for lb in get_bounds(dur)[0]): # zero-duration tasks cannot overlap, not enforced by Cumulative
+                    assert "no_overlap" not in self.supported_reified_global_constraints # otherwise cannot post decomposition toplevel
```
Dimosts, line 688: We could avoid dropping it from reified right? I mean, we could only post top level the defining constraints and return a conjunction of the value constraints and chc_cons? Then we could keep it in reified for positive duration only cases

cpmpy/solvers/pumpkin.py
```diff
@@ -582,10 +582,16 @@ def _get_constraint(self, cpm_expr, tag=None):
...
                     start, dur = cpm_expr.args
-                    return self._get_constraint(Cumulative(start, dur, demand=1, capacity=1), tag=tag)
+                    pum_cons = self._get_constraint(Cumulative(start, dur, demand=1, capacity=1), tag=tag)
                 else:
                     start, dur, end = cpm_expr.args
-                    return self._get_constraint(Cumulative(start, dur, end, demand=1, capacity=1), tag=tag)
+                    pum_cons = self._get_constraint(Cumulative(start, dur, end, demand=1, capacity=1), tag=tag)
+                
+                if any(lb <= 0 for lb in get_bounds(dur)[0]): # zero-duration tasks cannot overlap, not enforced by Cumulative
+                    assert f"no_overlap" not in self.supported_reified_global_constraints # otherwise cannot post decomposition toplevel
```
Dimosts, line 591: check my comments on choco

cpmpy/solvers/scip.py
```diff
@@ -478,6 +478,9 @@ def _add_transformed_constraint(self, cpm_expr):
                 else:
                     start, dur, end = cpm_expr.args
                     self._add_transformed_constraint(Cumulative(start, dur, end, demand=1, capacity=1))
+                
+                if any(lb <= 0 for lb in get_bounds(dur)[0]): # zero-duration tasks cannot overlap, not enforced by Cumulative
+                    self.add(cpm_expr.decompose())
```
Dimosts, line 483: check my comments on choco

tests/test_globalconstraints.py
```diff
@@ -1543,6 +1543,43 @@ def check_val():
 
         cp.Model(~cons).solveAll(display=check_val)
 
+    def test_no_overlap_zero_duration(self, solver):
```
Dimosts, line 1546: Maybe also a test where dur is var (with 0 in its domain).

Dimosts (commented):
In general I agree that this should be the intended behavior in nooverlap for 0 duration tasks.

My comments mainly have to do with:
- Making docstrings/comments consistent with the change
- Some discussion/questions on how this should be implemented in solvers that it is enforced through cumulative. (e.g. avoiding the removal from the reified globals in choco)

Other than that I agree.

I think for merging both this and #1102 we should confirm the intended behavior next cpmpy meeting

Dimosts:
@WoutPiessens this will be fixed at #1102

IgnaceBleukx:
We decided offline that it would make the most sense to follow the global constraint catalog and go for the non-strict version of NoOverlap. https://sofdem.github.io/gccat/gccat/Cdisjunctive.html

For OR-Tools, this means we will post NoOverlap constraints as Cumulatives when 0 is in the domain of the duration of any task. Closing this PR for now, I will open another one soon with the updated semantics.

PR #1105, opened 2026-09-22
NoOverlap non-strict semantics for zero duration tasks

WoutPiessens:
This PR addresses #1084 and fixes the semantics of NoOverlap for zero duration tasks. In order to respect the global constraint catalog, we decided to use the non-strict semantics in all cases, meaning that zero-duration tasks cannot overlap with any other task. 

The necessary changes are made in converting NoOverlap to native solver globals: in particular, for OR-Tools Cumulative with unit demand and capacity is used, since NoOverlap always uses the strict semantics.


cpmpy/expressions/globalconstraints.py
```diff
@@ -2020,15 +2022,15 @@ def value(self) -> Optional[bool]:
...
+            if any(a is None for a in start + duration + end):
                 return None
        
         if any(d < 0 for d in duration):
             return False
         if any(s + d != e for s,d,e in zip(start, duration,end)):
             return False
         for (s1,d1), (s2,d2) in all_pairs(zip(start,duration)):
-            if s1 + d1 > s2 and s2 + d2 > s1:
+            if d1 > 0 and d2 > 0 and s1 + d1 > s2 and s2 + d2 > s1:
```
IgnaceBleukx: This is a good opportunity to fix the value compuation while we're at it:
- sort tasks by start time
- filter out 0 duration tasks
- linear scan through list of tasks, ensure start of task i+1 is after end of task i

cpmpy/expressions/globalconstraints.py
```diff
@@ -2091,9 +2095,9 @@ def decompose(self) -> tuple[list[Expression], list[Expression]]:
         else:
             start, duration, end, is_present = self.args
             cons += [implies(is_present[i], start[i] + duration[i] == end[i]) for i in range(len(start))]
-        
-        for (s1, e1, p1), (s2, e2, p2) in all_pairs(zip(start, end, is_present)):
-            cons += [implies(p1 & p2, (e1 <= s2) | (e2 <= s1))]
+
+        for (s1, d1, e1, p1), (s2, d2, e2, p2) in all_pairs(zip(start, duration, end, is_present)):
+            cons += [implies(p1 & p2, cp.any([d1 == 0, d2 == 0, e1 <= s2, e2 <= s1]))]
```
IgnaceBleukx, line 2101: Actually: we should probably just make this a big disjunction then?
So we get:
```
~p1 | ~p2 | d1 == 0 | d2 == 0 | e1 <= s2 | e2 <= s1
```

cpmpy/solvers/cpo.py
```diff
@@ -613,8 +613,7 @@ def _cpo_expr(self, cpm_con, boolexpr=False):
                     end = None
                 else:
                     start, dur, end = cpm_con.args
-                tasks, cons = self._make_tasks(start, dur, end, None)
-                return cons + [dom.no_overlap(tasks)]
+                return self._make_no_overlap(start, dur, end, None)
```
IgnaceBleukx: Hmm I'm not sure the helper function is needed.
I would prefer it to be very explicit:
```python
if any(lb <= 0 <= ub for lb,ub in zip(*get_bounds(dur)):
    return self._cpo_expr(cp.Cumulative(start, dur..))
else: # what we had before
```

cpmpy/solvers/ortools.py
```diff
@@ -677,7 +677,10 @@ def _post_constraint(self, cpm_expr, reifiable=False):
                 # make interval variables
                 tasks, task_cons = self._get_ort_intervals(start, dur, end)
                 self.add(task_cons)
-                return self.ort_model.AddNoOverlap(tasks)
+                # posted as cumulative with unit demand and capacity:
```
IgnaceBleukx: I would only do this if 0 is in the domain of a task duration, not always

IgnaceBleukx (changes requested):
Logic is mostly good, some cleanups possible.

Main comment is about the update of the solver interfaces: we should only post the Cumulative if 0 is in the domain of some task, otherwise just post the no-overlap as usual.

cpmpy/expressions/globalconstraints.py
WoutPiessens: Done

cpmpy/expressions/globalconstraints.py
WoutPiessens: This causes a problem when p1 or p2 is a python bool. ~True flips to -2. This problem has already been discussed in a previous issue (#152), but it hasn't been fixed in all cases (because of a limitation of operator overloading in Python).

So for now I have kept it as is. Also, I like that it separates the active/non-active logic from the no overlap requirement, might make it a bit easier to read.

cpmpy/solvers/cpo.py
WoutPiessens: Done

cpmpy/solvers/ortools.py
WoutPiessens: Done

WoutPiessens:
One additional comment: for CP Optimizer, the semantics for Cumulative were also strict (seems to be something on our side, since cumulative is not natively supported). Therefore I added code to set all zero duration tasks as inactive. But I'm not sure if it's the cleanest way, and it may also be out of scope of this PR.

cpmpy/expressions/globalconstraints.py
IgnaceBleukx: I see, ok good to keep like this then indeed

cpmpy/solvers/cpo.py
```diff
@@ -613,6 +613,12 @@ def _cpo_expr(self, cpm_con, boolexpr=False):
                     end = None
                 else:
                     start, dur, end = cpm_con.args
+
+                if any(lb <= 0 <= ub for lb, ub in zip(*get_bounds(dur))):
+                    # CP Optimizer's NoOverlap has strict semantics for zero duration tasks: they cannot start when another task is planned.
```
IgnaceBleukx: the bit after the colon can go, in this comment
Can be a bit more compact in general:
```
CPO has strict semantics for NoOverlap, post as Cumulative instead
```

cpmpy/solvers/ortools.py
```diff
@@ -677,6 +677,11 @@ def _post_constraint(self, cpm_expr, reifiable=False):
                 # make interval variables
                 tasks, task_cons = self._get_ort_intervals(start, dur, end)
                 self.add(task_cons)
+
+                if any(lb <= 0 <= ub for lb, ub in zip(*get_bounds(dur))):
+                    # OR-Tools' NoOverlap has strict semantics for zero duration tasks: they cannot start when another task is planned.
+                    # In order to use the non-strict semantics, we rewrite NoOverlap as Cumulative.
```
IgnaceBleukx: same here as in cpo

cpmpy/solvers/cpo.py
```diff
@@ -598,11 +598,11 @@ def _cpo_expr(self, cpm_con, boolexpr=False):
                         continue
                     else:
                         task_demand = dom.pulse(task, get_bounds(h))
+                        # a zero-duration task has an empty pulse, so it occupies no resource and its demand should not be channeled
```
IgnaceBleukx: comment is a bit confusing... rather: `a zero-duration task occupies no resource, set as non-active`

tests/test_globalconstraints.py
```diff
@@ -1543,6 +1543,71 @@ def check_val():
...
+        dur = [0,5,3]
+        cons = cp.NoOverlap(start, dur)
+
+        assert cp.Model(cons, start == start_val).solve(solver=solver)
+        assert cons.value() is True
+        assert cp.Model(cons.decompose(), start == start_val).solve(solver=solver)
+        # NoOverlap has the same semantics as Cumulative with demand and capacity equal to 1
+        assert cp.Model(cp.Cumulative(start, dur, demand=1, capacity=1), start == start_val).solve(solver=solver)
+
+        # tasks with a non-zero duration must still not overlap
```
IgnaceBleukx: isn't this a bit redundant? Ok to keep but feels like its toom much

tests/test_globalconstraints.py
```diff
@@ -1543,6 +1543,71 @@ def check_val():
...
+        # same, but now 0 is only in the domain of the duration variables
+        start = cp.intvar(0,5, shape=3, name="start")
+        dur = cp.intvar(0,5, shape=3, name="dur")
+        cons = cp.NoOverlap(start, dur)
+
+        assert cp.Model(cons, start == [1,0,5], dur == [0,5,3]).solve(solver=solver)
+        assert cons.value() is True
+        assert cp.Model(cons, start == [1,0,2], dur == [0,5,3]).solve(solver=solver) is False
+
+    def test_no_overlap_optional_zero_duration(self, solver):
```
IgnaceBleukx, line 1633: also add the _var variant for no-overlap optional

IgnaceBleukx (approved):
Some small comments on the docs and tests, otherwise looks ready!

